PR #62, opened 2014-06-16
Feature/initializer list

awulkiew:
This adds support for std::initializer_list for some of the models (see the test for example usage).
model::point's constructor is no longer explicit to allow using C++11 unified initialization syntax.


awulkiew:
I believe the support for initializer lists is ready for merging. To allow using Boost.Assign and therefore ensure backward-compatibility I commented-out the assignment operators taking `std::initializer_list<>`. I tested it on msvc (10, 12), gcc (4.8, 4.9) and clang (3.4).

Currently when an initializer list is assigned to one of the geometries, first a geometry object should be created using a ctor taking `std::initializer_list<>` and then it should be moved using implicitly defined move assignment operator.

If Boost.Assign was somehow fixed (there are 2 tickets for this https://svn.boost.org/trac/boost/ticket/5419 4yo and https://svn.boost.org/trac/boost/ticket/7364 2yo) or the support for Assign was dropped on C++11 compilers (since Assign doesn't work with STL containers anyway), assignment operators taking `std::initializer_list<>` could be restored.

awulkiew:
I've shielded the support with macro and enabled it for one test, testing together `std::initializer_list<>`, uniform initialization syntax and Boost.Assign's `list_of()`, `pair_list_of()` and `tuple_list_of()`. This way we would be able to check if everything works before enabling the support for the users.

PR #190, opened 2015-01-05
Fix/centroid

awulkiew:
Various improvements for centroid():
- more deterministic result for invalid geometries, this "fixes" the issue introduced by https://github.com/boostorg/geometry/pull/155, the modification (translation) of a result even if it wasn't correctly calculated,
- check for `count == 0` in average strategy and returning `false` in this case, this also makes it CentroidStrategyConcept-conformant,
- replacement of `int` with `std::size_t` for counters (should really be `range_size<>::type` but still safer than `int`),
- use of default template parameters to improve readability.


barendgehrels:
Thanks Adam. I'm OK with merging it.

PR #194, opened 2015-01-14
Moving from boost.timer old to new API

raffienficiaud:
Please find here the PR for fixing the issues you have in geometry, due to the changes in boost.test.
http://article.gmane.org/gmane.comp.lib.boost.devel/256800

### DIFF
--- a/test/algorithms/buffer/multi_point_buffer.cpp
+++ b/test/algorithms/buffer/multi_point_buffer.cpp
@@ -174,7 +174,7 @@ void test_growth(int n, int distance_count)
 {
     srand(int(time(NULL)));
     //std::cout << typeid(bg::coordinate_type<P>::type).name() << std::endl;
-    boost::timer t;
+    boost::timer::cpu_timer t;
 
     namespace buf = bg::strategy::buffer;
     typedef bg::model::polygon<P> polygon;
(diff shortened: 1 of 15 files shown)


barendgehrels:
Thanks for the PR.

My (our) main objection is that Boost.Test was usable as a header-only library. I assume, and really hope, that it will stay to be usable as a header-only library. I understand that it now depends on Boost.Timer v2 and Boost.Timer v2 is not header-only. But that should be solved somehow.

This PR advances Boost.Timer v1 to Boost.Timer v2 (with the side-effect that it fixes Boost.Test issues) but that is not enough. If we apply this it would require linking with Boost.Timer, which is unwished (and not part of this PR).

Besides that, the robustness tests are NOT unit-tests and (AFAIR) not dependant on Boost.Test. They can for now stay as they are.

I will send a message to the list about my opinion how to solve this.

raffienficiaud:
I agree with your remarks. I separated the commits on these parts, in case you want to do a selective merge/cherry pick. 
There are several options on boost.test side as you should know. The header only variant is definitely something we want to keep (although I do not know how it works with boost.system).

barendgehrels:
Good to hear you want to keep it. Will react later.

raffienficiaud:
I rolled back the changes in boost.test. Should work now. Let me know if you experience any issue with this.

barendgehrels:
Thanks - I assume you noted on the list that Beman will create a header-only option for Boost.Timer v2. So that will make all your changes OK! One or two weeks.
https://svn.boost.org/trac/boost/ticket/10935

PR #195, opened 2015-01-19
Fix/cart intersect

awulkiew:
This PR fixes the assertion failure for specific cases when a collinear segments' endpoints are not detected as intersecting in `cart_intersect -> direction policy`, but should be detected as such if e.g. `equals()` function was used. It's because only the ratios are compared. This may lead to some inconsistencies, e.g. the generation of only one intersection point for a segment detected as collinear (should be detected as collinear-touch) and an assertion failure in `get_turn_info()` algorithm analysing the `intersection_result`.


barendgehrels:
That is quick! Thanks. I looked at the changes and I'm OK with merging.

PR #197, opened 2015-01-22
Feature/support for one point geometries

mkaravel:
In this PR I modify the `range_segment_iterator` class to support ranges with less than two elements.
With this modification of the `range_segment_iterator` class, the `segment_iterator` can support (invalid) geometries that contain ranges with less than two points.

A consequence of this PR is that `bg::distance` can now compute the distance for (invalid) geometries that contain ranges of less than two points (e.g., one-point linestrings, non-empty closed rings with less than four points, non-empty open rings with less than three points, non-empty multilinestrings that contain one-point linestrings (or even empty linestrings), one-point polygons, etc.)

The unit tests in this PR rely on the fix proposed in PR #196, so PR #196 should be merged first.

### DIFF
--- a/include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
+++ b/include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
@@ -104,7 +104,7 @@ struct range_segment_iterator
 
     // default constructor
     range_segment_iterator()
-        : m_it(), m_has_less_than_two_elements()
+        : m_it(), m_has_less_than_two_elements(false)
     {}
 
     // for begin
--- a/include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
+++ b/include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
@@ -91,7 +91,7 @@ struct range_iterator_end<Range, open>
 
 
 template <typename Range, typename Value, typename Reference = Value>
-class range_segment_iterator
+struct range_segment_iterator
     : public boost::iterator_facade
         <
             range_segment_iterator<Range, Value, Reference>,
@@ -100,16 +100,6 @@ class range_segment_iterator
             Reference
         >
 {
-private:
-    static bool has_less_than_two_elements(Range const& r)
-    {
-        iterator_type first = range_iterator_begin<Range>::apply(r);
-        iterator_type last = range_iterator_end<Range>::apply(r);
-
-        return first == last || ++first == last;
-    }
-
-public:
     typedef typename range_iterator_type<Range>::type iterator_type;
 
     // default constructor
@@ -120,15 +110,15 @@ public:
     // for begin
     range_segment_iterator(Range& r)
         : m_it(range_iterator_begin<Range>::apply(r))
-        , m_has_less_than_two_elements(has_less_than_two_elements(r))
+        , m_has_less_than_two_elements(boost::size(r) < 2u)
     {}
 
     // for end
     range_segment_iterator(Range& r, bool)
         : m_it(range_iterator_end<Range>::apply(r))
-        , m_has_less_than_two_elements(has_less_than_two_elements(r))
+        , m_has_less_than_two_elements(boost::size(r) < 2u)
     {
-        if (!m_has_less_than_two_elements)
+        if (! m_has_less_than_two_elements)
         {
             // the range consists of at least two items
             --m_it;
--- a/include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
+++ b/include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
@@ -100,23 +100,40 @@ class range_segment_iterator
             Reference
         >
 {
+private:
+    static bool has_less_than_two_elements(Range const& r)
+    {
+        iterator_type first = range_iterator_begin<Range>::apply(r);
+        iterator_type last = range_iterator_end<Range>::apply(r);
+
+        return first == last || ++first == last;
+    }
+
 public:
     typedef typename range_iterator_type<Range>::type iterator_type;
 
     // default constructor
     range_segment_iterator()
-        : m_it()
+        : m_it(), m_has_less_than_two_elements()
     {}
 
     // for begin
     range_segment_iterator(Range& r)
         : m_it(range_iterator_begin<Range>::apply(r))
+        , m_has_less_than_two_elements(has_less_than_two_elements(r))
     {}
 
     // for end
     range_segment_iterator(Range& r, bool)
-        : m_it(--range_iterator_end<Range>::apply(r))
-    {}
+        : m_it(range_iterator_end<Range>::apply(r))
+        , m_has_less_than_two_elements(has_less_than_two_elements(r))
+    {
+        if (!m_has_less_than_two_elements)
+        {
+            // the range consists of at least two items
+            --m_it;
+        }
+    }
 
     template
     <
@@ -151,6 +168,11 @@ private:
 
     inline Reference dereference() const
     {
+        if (m_has_less_than_two_elements)
+        {
+            return Reference(*m_it, *m_it);
+        }
+
         iterator_type next(m_it);
         ++next;
         return Reference(*m_it, *next);
@@ -184,6 +206,7 @@ private:
 
 private:
     iterator_type m_it;
+    bool m_has_less_than_two_elements;
 };
 
 
--- a/test/algorithms/distance/distance_linear_areal.cpp
+++ b/test/algorithms/distance/distance_linear_areal.cpp
@@ -65,6 +65,10 @@ void test_distance_segment_polygon(Strategy const& strategy)
     tester::apply("segment(-1 20,-1 -20)",
                   "polygon((-10 -10,10 -10,10 10,-10 10,-10 -10))",
                   0, 0, strategy);
+
+    tester::apply("segment(0 0,1 1)",
+                  "polygon((2 2))",
+                  sqrt(2.0), 2, strategy);
 }
 
 //===========================================================================
@@ -93,6 +97,14 @@ void test_distance_linestring_polygon(Strategy const& strategy)
     tester::apply("linestring(-1 20,1 20,1 -20)",
                   "polygon((-10 -10,10 -10,10 10,-10 10,-10 -10))",
                   0, 0, strategy, true);
+
+    tester::apply("linestring(-2 1)",
+                  "polygon((0 0,10 0,10 10,0 10,0 0))",
+                  2, 4, strategy, true);
+
+    tester::apply("linestring(-5 1,-2 1)",
+                  "polygon((0 0))",
+                  sqrt(5.0), 5, strategy, true);
 }
 
 //===========================================================================
@@ -139,6 +151,14 @@ void test_distance_multilinestring_polygon(Strategy const& strategy)
     tester::apply("multilinestring((-1 20,1 20,1 30),(-1 20,1 20,1 -20))",
                   "polygon((-10 -10,10 -10,10 10,-10 10,-10 -10))",
                   0, 0, strategy, true);
+
+    tester::apply("multilinestring((-100 -100,-90 -90),(1 20))",
+                  "polygon((-10 -10,10 -10,10 10,-10 10,-10 -10))",
+                  10, 100, strategy, true);
+
+    tester::apply("multilinestring((-100 -100,-90 -90),(-1 20,1 20,1 30))",
+                  "polygon((-110 -110))",
+                  sqrt(200.0), 200, strategy, true);
 }
 
 //===========================================================================
--- a/test/algorithms/distance/distance_linear_linear.cpp
+++ b/test/algorithms/distance/distance_linear_linear.cpp
@@ -79,6 +79,10 @@ void test_distance_segment_linestring(Strategy const& strategy)
     tester::apply("segment(1 1,2 2)",
                   "linestring(3 3,3 3)",
                   sqrt(2.0), 2, strategy);
+
+    tester::apply("segment(1 1,2 2)",
+                  "linestring(3 3)",
+                  sqrt(2.0), 2, strategy);
 }
 
 //===========================================================================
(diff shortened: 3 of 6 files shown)


awulkiew:
This version of segment_iterator is maybe less "clear" but in the same time in some cases (se below) it seems to be more robust. For a 1-point range a degenerated segment can be dereferenced instead an assertion failure. And if someone wanted to handle the degenerated ranges in other way he can still do it. Besides AFAIU it's always possible to refer to a degenerated segment anyway. Or are such segments normally filtered by the algorithm?

I wrote "in some cases" because there is a special case when this implementation could silently fail, i.e. if mutable segments was represented. Then suppose someone would like to modify the points through the refering segment. In such scenario the same iterator would be dereferenced for both points. So when someone modified the first point, the second one would also be modifed. This could be unexpected. I can't see why someone could want to modify the points through the segment iterator, still maybe it could be safer to dissallow it, so always return non-mutable Reference and write a comment about it. Or was this done already in other part of the code? I remember we had similar conversation some time ago.

Btw, what if a MultiLinestring contained an empty Linestring? Is such Linestring just ignored? By concatenating/flattening iterator or somehow else? Can default-created/singular/NULL iterators be compared or is the emptiness of a range checked explicitly somehow (`boost::empty()` or `boost::size()`)? I'm guessing that the latter is true, but I'm asking just in case. In C++03 the use of singular iterators is undefined behavior.

Regarding distance, personally I'd probably consider implementing the distance "the usual way" so "manually" wrapping an algorithm for single-geometry as a policy of an algorithm for multi-geometry in dispatch namespace.

mkaravel:
@awulkiew Please see below. I have tried to provide answers to your questions:
* Yes, you can always dereference degenerated segments through the `range_segment_iterator` or the `segment_iterator`. Whether the algorithm using the segment iterator filters degenerated segments is the responsibility of this algorithm.
* Correct: if the segment iterator it mutable, then you can have the problem you mentioned. The current implementation of the segment iterator is non-mutable, so this problem does not exist.
* Multilinestrings with empty linestrings are handled: the empty linestring is ignored. This is done through the flatten iterator automatically. See the unit test for the `segment_iterator` iterator class.
* The segment iterator is a bidirectional iterator, and its traversal tag is also the bidirectional traversal tag. Its unit test checks for the corresponding concepts from Boost.Concept and Boost.Range. Its emptiness is checked the usual way: `begin == end`. I have not checked `boost::empty()`, but I do expect that it works. Using `boost::size()` should be possible but a bad idea: it would take linear time to compute the size. See also the unit test for the segment iterator where all sorts of (invalid) geometries are tested. I do not test singular iterators at all.
* Regarding distance: I do not see the point of doing it the manual way. Suppose we have two multilinestrings. The current algorithm takes the segments of the smallest multilinestring (smallest in terms of number of segments), inserts its segments in an r-tree and then the r-tree is queried for nearest neighbors using the segments of the other multilinestring. The minimum distance among nearest neighbor pairs is returned as the end. Let's consider the alternative, manual way. First of all I would avoid reducing the problem to finding the minimum distance among all pairs of linestrings in the multilinestrings, as this would result in a quadratic algorithm. So what I would possibly do is to consider each multipolygon, manually convert 1-point linestrings to segments, and insert all segments inside an r-tree. But this is basically what is done now using the patched `range_segment_iterator` proposed in this PR. If I understand it correctly, your suggestion to wrap an algorithm for a single geometry as a policy for the multigeometry would easily result in quadratic behavior (think of two multilinestrings where each linestring consists of a single point or a single segment).

mkaravel:
Would it be possible to make a decision about this PR?
Ideally I would like it to be in 1.58.

include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
```diff
@@ -100,23 +100,40 @@ class range_segment_iterator
             Reference
         >
 {
+private:
+    static bool has_less_than_two_elements(Range const& r)
+    {
+        iterator_type first = range_iterator_begin<Range>::apply(r);
+        iterator_type last = range_iterator_end<Range>::apply(r);
+
+        return first == last || ++first == last;
```
barendgehrels: is std::distance not working?

barendgehrels:
See my question.
For the rest, Adam did have a close to this. If Adam is OK, I'm OK with merging

awulkiew:
Yes, I'm ok with merging. I'll wait for a response for the question about the std::distance().

include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
mkaravel: If the range is a random access range, calling `std::distance` is perfectly okay.
If the range is not a random access range, calling `std::distance` takes linear time in the size of the range.

It is true that in our context, the range passed here is either a ring or a linestring, and our corresponding concepts require that they are of random access. So it is okay to use the following return statement:
```
return std::distance(first, last) < 3u;
```
Let me know if this preferable, and I will make the change.

mkaravel: The return statement in my previous message should have been:
```
return std::distance(first, last) < 2u;
```

barendgehrels: If that is the case. Then I think boost::size(r) < 2u will also do at the caller side?
We don't need this function then at all. That looks better to me.
All (or at least most of) our functionality is based on random access ranges, accessing them with segment-id's and so on, I don't think we have to provide complexity to support non random access ranges here.

include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
```diff
@@ -100,23 +100,40 @@ class range_segment_iterator
...
     {}
 
     // for end
     range_segment_iterator(Range& r, bool)
-        : m_it(--range_iterator_end<Range>::apply(r))
-    {}
+        : m_it(range_iterator_end<Range>::apply(r))
+        , m_has_less_than_two_elements(has_less_than_two_elements(r))
+    {
+        if (!m_has_less_than_two_elements)
```
barendgehrels: We normally put a space after !

include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
```diff
@@ -100,23 +100,40 @@ class range_segment_iterator
...
+        return first == last || ++first == last;
+    }
+
 public:
     typedef typename range_iterator_type<Range>::type iterator_type;
 
     // default constructor
     range_segment_iterator()
-        : m_it()
+        : m_it(), m_has_less_than_two_elements()
```
barendgehrels: OK I've never seen this usage for bool but I assume it initializes with false

include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
mkaravel: All right. I will use `boost::size` then. This way I can avoid the additional `#include`.

include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
mkaravel: Will do. Thanks for pointing this out.

include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
mkaravel: Using the default constructed iterator is undefined behavior, which is why I think it is okay to leave it as it is.

include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
barendgehrels: OK, thanks. I'm OK with merging after the update

mkaravel:
This PR should be ready for merging now.

include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
awulkiew: Why not explicitly initialize it with false?
Consider the initialization of a pointer with NULL.
I see it in various places in BG code, so it's a more general problem...

barendgehrels: I agree with initializing it with false.

I don't agree with the statement "I see it in various places, it is a general problem": those are assignments of setting a numeric type T to 0 (T()), which is perfectly fine and allowed. (as long as udt do support also this construct)

awulkiew: As long as we're in C++11 or a constructor of T in C++98 initializes T to 0. Otherwise, this is not true. This is why someone was forced to add a default ctor to ttmath types in the past.

mkaravel: Initialization with `false` done.

barendgehrels: T x = T() for a built-in type is well defined zero-initialization in C++03
ttmath is not a built-in type so I was the one adding that ctor to support this construct
Here it was used differently so good Menelaos changed it

awulkiew: @mkaravel Thanks!

@barendgehrels Indeed scalar types should be zero-initialized assuming the compiler is standard-conformant. However Geometry should work also with non-scalar coordinate types and we shouldn't assume that a default ctor should initialize an object to an equivalent of 0. Is there a reason why do you think that value-initialization is better than explicit initialization with 0? This is one more, unnecessary requirement for coordinate type. The library already requires that it should be possible to construct an object with some int or double value, e.g. in distance_cross_track:

    return_type const quarter(0.25);

which also has its drawbacks but it's another story. Why not do this also for 0 and leave the default ctor out of this?

barendgehrels: @awulkiew That is a different thing.You mentioned it is a general problem in BG. I reacted on that: it is not a general problem in BG.

UDT are a different thing - if you cannot assume they have a default constructor, you can also not assume the presence of a constructor T(0). Therefore we selected 7 years ago the construction of T(): it is well defined.

Eehm, as far as I know I never wrote the construct as you mention for quarter - this should IMO *not* be done. It should read like this: return_type const quarter = 0.25; We should fix that.

barendgehrels:
Thanks for the PR and for merging

include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
awulkiew: @barendgehrels Why it can't be assumed that a class may not have a default ctor? It's a common thing! It may have any number of non-default ctors taking arbitrary types. I don't understand this reasoning.

Furthermore we shouldn't assume that a default-initialization is zero-initialization.

What is more, we already require that T must be initialized explicitly in other places. In your fix above (assignment instead of an explicit ctor call) it makes no difference because a ctor will be called (at some point on a path of implicit conversions). E.g. here:

    return_type const zero = 0; // implicit conversion <=> ctor

The library must and do require that there must be some path of implicit conversions from double/int to T. Effectively it's requiring ctor taking double or some other type constructible from numeric constant. So why add another requirement (zero-initializing default ctor)?

And why value-initialization is better than explicit construction? It's worse, because with explicit construction you see immediately what's passed, exactly like with `false` or `NULL`.

Or maybe I'm missing something?

brunolalande: As far as I know, you choose the T() initialization when you want to say something like "if it's an object it's supposed to know how to properly initialize itself (and I'm not supposed to know) and if it's a fundamental type then I know I want it to be 0 (which the C++ standard guarantees)". Iterators are a typical example: if it's a raw pointer you want 0, if it's an object you want it to do whatever is needed for it to be ready for iteration. So you write T() (you'll see it everywhere in the STL).

In the case you are manipulating something that's supposed to represent a scalar value, and you want it to "mathematically represent" X, then you are not really in the situation described above. You're saying "whatever you are, you are now equal to X". So you'll rather use T(X).

This is all general guidelines of course.

Bruno

include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
```diff
@@ -91,7 +91,7 @@ struct range_iterator_end<Range, open>
 
 
 template <typename Range, typename Value, typename Reference = Value>
-class range_segment_iterator
+struct range_segment_iterator
```
barendgehrels: NOTE: apprently this gives errors for clang because elsewhere you have declared this as a friend.

````
"clang++" -x c++ -Wno-unneeded-internal-declaration -std=c++11 -O0 -g -O0 -fno-inline -Wall -g -DBOOST_ALL_NO_LIB=1 -I".." -I"../boost/geometry/extensions/contrib/ttmath" -I"../libs/geometry/test" -c -o "/Volumes/Data/kbelco/boost_testing/results/boost/bin.v2/libs/geometry/test/iterators/segment_iterator.test/clang-darwin-4.2.1/debug/segment_iterator.o" "../libs/geometry/test/iterators/segment_iterator.cpp"

In file included from ../libs/geometry/test/iterators/segment_iterator.cpp:46:
In file included from ../boost/geometry/iterators/segment_iterator.hpp:22:
In file included from ../boost/geometry/iterators/detail/segment_iterator/iterator_type.hpp:27:
../boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp:157:12: warning: class template 'range_segment_iterator' was previously declared as a struct template [-Wmismatched-tags]
    friend class range_segment_iterator;
           ^
../boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp:94:8: note: previous use is here
struct range_segment_iterator
       ^
../boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp:157:12: note: did you mean struct here?
    friend class range_segment_iterator;
           ^~~~~
           struct
1 warning generated.
            
```

Please fix before we can merge this to master...

mkaravel: Fix done.
Also fixed a bug that was introduced by the usage of `boost::size` to initialize the `m_has_less_than_two_elements` member variable.
See PR #203.

include/boost/geometry/iterators/detail/segment_iterator/range_segment_iterator.hpp
awulkiew: Exactly, this is also my point, it's even semantically more clear to initialize coordinates explicitly passing a value. But currently in many places value-initialization is used, so my question is why it was choosen? For me it adds unnecessary requirement for a coordinate type and is less clear. But maybe I'm missing something?

In case my reasoning was correct, I propose to not use zero-initialization and initialize objects with constants explicitly when a representation of a specific scalar/numeric value is needed. This is valid not only for coordinate types.

PR #198, opened 2015-01-24
[test][algorithms][distance] replace use of internal Boost.Test tools

mkaravel:
by BOOST_CHECK and BOOST_CHECK_CLOSE


barendgehrels:
Thanks for your fast action!

I see the problem - I would like to have a BOOST_CHECK_CLOSE_MESSAGE too!

I will merge this right away.

PR #202, opened 2015-01-28
Fix/is simple closed linestrings

mkaravel:
`bg::is_simple` was handling incorrectly simple closed linestrings within multilinestrings. For example the geometry with WKT `MULTILINESTRING((0 0,10 0,10 10,0 10,0 0))` was failing.

The reason is that the linestring was accepted as a valid linestring, and then for verifying the validity of the multi-linestring, the self-turns of the multi-linestring were computed; the closing point of the linestring was reported as a turn which was wrongly considered as an unacceptable turn.

In this PR such turns are checked explicitly and reported as acceptable.


barendgehrels:
I'm OK with merging this

PR #204, opened 2015-01-29
Fix/is simple wrong turns detection

mkaravel:
This PR essentially deals with two issues:
* Optimizing the `bg::is_simple` computation for multilinestrings. In the older implementation self turns were computed for each linestring in the multilinestring individually, and then again if needed at the multilinestring level. In the new implementation the self turns of the linestrings are not computed, while the self turns of the entire multilinestring are used to determine/decide simplicity.
* The change in the above item conveniently deals with and fixes a bug in `bg::intersects`: `bg::intersects` cannot identify an open linestring whose one endpoint is the same as an interior point as self-intersecting (for example: `LINESTRING(4 1,10 8,4 6,4 1,10 5,10 3)`). This immediately fixes a bug in `is_simple`, which, previously to this PR, erroneously considers linestrings, like the ones described just above, as simple.

### DIFF
--- a/include/boost/geometry/algorithms/detail/is_simple/linear.hpp
+++ b/include/boost/geometry/algorithms/detail/is_simple/linear.hpp
@@ -92,7 +92,7 @@ public:
         return m_is_closed
             && turn.method == overlay::method_none
             && check_segment_indices(turn, boost::size(m_linestring) - 2)
-            && geometry::equals(range::front(m_linestring), turn.point);
+            && turn.operations[0].fraction.is_zero();
     }
 
 private:
@@ -132,7 +132,7 @@ private:
             &&
             geometry::equals(range::front(linestring), range::back(linestring))
             &&
-            geometry::equals(range::front(linestring), turn.point)
+            turn.operations[0].fraction.is_zero();
             ;
     }
 
--- a/include/boost/geometry/algorithms/detail/is_simple/linear.hpp
+++ b/include/boost/geometry/algorithms/detail/is_simple/linear.hpp
@@ -1,6 +1,6 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
-// Copyright (c) 2014, Oracle and/or its affiliates.
+// Copyright (c) 2014-2015, Oracle and/or its affiliates.
 
 // Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
 
@@ -19,6 +19,7 @@
 #include <boost/geometry/core/closure.hpp>
 #include <boost/geometry/core/coordinate_type.hpp>
 #include <boost/geometry/core/point_type.hpp>
+#include <boost/geometry/core/tag.hpp>
 #include <boost/geometry/core/tags.hpp>
 
 #include <boost/geometry/util/range.hpp>
@@ -29,8 +30,10 @@
 
 #include <boost/geometry/algorithms/equals.hpp>
 #include <boost/geometry/algorithms/intersects.hpp>
+#include <boost/geometry/algorithms/not_implemented.hpp>
 
 #include <boost/geometry/algorithms/detail/check_iterator_range.hpp>
+#include <boost/geometry/algorithms/detail/signed_index_type.hpp>
 
 #include <boost/geometry/algorithms/detail/disjoint/linear_linear.hpp>
 #include <boost/geometry/algorithms/detail/overlay/get_turn_info.hpp>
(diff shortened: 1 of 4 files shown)


include/boost/geometry/algorithms/detail/is_simple/linear.hpp
```diff
@@ -54,128 +57,210 @@ namespace detail { namespace is_simple
...
+        return
+            turn.method == overlay::method_none
+            &&
+            check_segment_indices(turn, boost::size(linestring) - 2)
+            &&
+            geometry::equals(range::front(linestring), range::back(linestring))
+            &&
+            geometry::equals(range::front(linestring), turn.point)
+            ;
+    }
```
barendgehrels: The whole fix looks more like a whole rewrite (no problem).

What I'm wondering about this method is: can't we make better use of the turn's properties? They have already the segment-index, they know how the intersection is made, there are fractions, arrival-information, etc, why is it still necessary to call geometry::equals here? That information should be known here... At least, for A/A those info is surely known

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
```diff
@@ -54,128 +57,210 @@ namespace detail { namespace is_simple
...
+                        Linestring, closed
+                    >::apply(linestring)
+            && ! (CheckSelfIntersections && has_self_intersections(linestring));
+    }
+};
 
 
+template <typename MultiLinestring>
+struct is_simple_multilinestring
+{
```
barendgehrels: This struct looks really good! :-)

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
```diff
@@ -54,128 +57,210 @@ namespace detail { namespace is_simple
...
-            && !detail::is_valid::has_spikes
-                    <
-                        Linestring, closed
-                    >::apply(linestring)
-            && !(CheckSelfIntersections && geometry::intersects(linestring));
+        BOOST_ASSERT(boost::size(m_linestring) > 1);
+        return m_is_closed
+            && turn.method == overlay::method_none
+            && check_segment_indices(turn, boost::size(m_linestring) - 2)
+            && geometry::equals(range::front(m_linestring), turn.point);
```
barendgehrels: Same question about turn-info here: can we reuse?

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
mkaravel: I assume you are referring to the last call to `goemetry::equals`. I will try to rewrite the test using segment ratios.

Note that we also need to fix the bug in bg::intersects that takes a single argument. It is a bit more complicated there as we need to account for the possibility to have duplicate points at the beginning and and of the linestring.

barendgehrels:
I've a few questions but basically I'm in favour of merging this

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
mkaravel: Same answer: I will try to replace the call to `geometry::equals` by means of the segment ratios.

mkaravel:
@barendgehrels Thanks for taking a look at it so quickly.
I will try to use segment ratios.

mkaravel:
There is general question related to `bg::is_simple`: recently I modified `bg::is_valid` to accept empty multi-geometries as valid. Should `bg::is_simple` be modified accordingly to accept empty multi-geometries as simple?

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
awulkiew: Ratios not necessarily must be compatible with points. Remember the fix I made some time ago in cart_intersect?

barendgehrels:
Good question - don't know (yet). What does the standard say here?

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
awulkiew: And here again equals(ratio, 0) not necessarily is the same as equals(point, front)

mkaravel:
I could not finding anything in the standard about this.
When asking the question, I was mostly thinking in terms of consistency of our own code: if we accept empty multi-geometries as valid, why not accept them as simple as well?

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
mkaravel: @awulkiew Do you have a specific example where there is a mismatch between points and rations. It would be really helpful to have it and use it as test case.

awulkiew: Yes, however have in mind that it was detected only for floating-point ratios (no rational) so when rescaling is disabled. See e.g. https://github.com/boostorg/geometry/commit/eeb33369ecfc2ed8104ab5207694027a30c67f04 for the test cases and https://github.com/boostorg/geometry/commit/38389101829ed993523730ecedcd097fbf1ec70b for the fix. E.g. see the "simpler case" test in line 309, points (-2.8823529411764710 -7.7647058823529411) and (-2.8823529411764706 -7.7647058823529411) are spatially equal but when ratios are checked they're not detected as equal. It's because both points and ratios are compared using machine EPS but in both cases the actual values being compared are not compatible. E.g. ratios values depends on the lengths of segments.

awulkiew: Btw, if you used the new get_turn_info_ll policy you shouldn't be forced to deal with this problem since for the first and the last point of non-degenerated segment a special flag (position) is set in operation member of turn_info.

mkaravel: Please correct me if I am wrong: if I am to use the `get_turn_info_ll` policy, then I would need to call turns by passing the same argument twice, which means that I will get a lot more turns than what I do now. I am just saying that the way I do it is a bit more efficient.

awulkiew: My point is, you could pass this new policy into the `detail::self_get_turn_points::get_turns<>::apply()`. Here in the implementation of 1-argument `intersects()` https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/intersects.hpp in the lines 57 and 65 the regular (A/A) turn_info and policy types are defined. But you could use the turn_info and policy specific for the pair of input parameters, like it's done in relate https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/relate/turns.hpp in the lines 46 and 41.

Another alternative would be to modify the regular (A/A) policy and just set those position flags if specific turn_info type was used, or always for Linear geometries. The required parameters are passed into all turn policies, in regular policy they're just ignored because Areal geometries doesn't have the front or back (see https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp line 910).

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
barendgehrels: ratio = 0 should be checked. Checking equals(point,front) can make the result inconsistent.

awulkiew: In this case indeed the first and the last point should be EXACTLY the same so yes, maybe you're right that it'd be sufficient to check the ratios for exactly 0 or 1.

mkaravel:
Ratios are now used for testing the equality between turn points and endpoints.

PR #207, opened 2015-02-03
Fix/make side by triangle permutation invariant

mkaravel:
Make the `side_by_triangle` strategy permutation invariant: the strategy takes three points as arguments, and its result is in theory invariant under cyclic permutations of these three arguments.
To make the strategy consistent with respect to cyclic permutations of its three arguments, we cyclically permute them so that the first argument is always the lexicographically smallest point.

### DIFF
--- a/include/boost/geometry/strategies/cartesian/side_by_triangle.hpp
+++ b/include/boost/geometry/strategies/cartesian/side_by_triangle.hpp
@@ -114,17 +114,31 @@ public :
         geometry::detail::relate::less less;
 
         promoted_type s;
-        if (less(p2, p1) && less(p2, p))
+        if (less(p, p1))
         {
-            s = side_value<coordinate_type, promoted_type>(p2, p, p1);
-        }
-        else if (less(p, p1) && less(p, p2))
-        {
-            s = side_value<coordinate_type, promoted_type>(p, p1, p2);
+            if (less(p, p2))
+            {
+                // p is the lexicographically smallest
+                s = side_value<coordinate_type, promoted_type>(p, p1, p2);
+            }
+            else
+            {
+                // p2 is the lexicographically smallest
+                s = side_value<coordinate_type, promoted_type>(p2, p, p1);
+            }
         }
         else
         {
-            s = side_value<coordinate_type, promoted_type>(p1, p2, p);
+            if (less(p1, p2))
+            {
+                // p1 is the lexicographically smallest
+                s = side_value<coordinate_type, promoted_type>(p1, p2, p);
+            }
+            else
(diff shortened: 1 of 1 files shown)


barendgehrels:
The side_value was (and is) twice the area of a triangle formed by p,p0,p1 (either positive or negative). That is still the case because you always keep the order p,p1,p2 the same, be it in a permutation of the same sequence. So the polygon formed by p,p0,p1 keeps the same orientation. Maybe you can add this in a comment, like you wrote just above.

I'm OK with merging, because of this and because by mail you said all unit tests pass with this change.

barendgehrels:
On further thought - the new implementation does 4 point-equal-comparisons and 2 less-comparisons more than the old, which was just a simple calculation.

Therefore I propose to make this a separate strategy, keeping the old one intact. We can still consider setting it as the default (or not), but I want to have the additional calculations optional, also because for integer it is not necessary at all...

mkaravel:
Sure, I will add a comment about the invariance.

mkaravel:
I would prefer to add a boolean template parameter that handles the option of performing these checks or not, instead of creating a new strategy.
Using this template parameter, I can also disable automatically the additional tests for fundamental integral types.
Is that solution okay, or do you really prefer another strategy?

We also need to think about documentation: do we prefer to have two different strategies documented or just one with an additional template parameter?

awulkiew:
We should prefer the correctness over the performance.

I think there is no need for additional parameters or strategies. We could use the old code for integral robust coordinates and the new one for floating-point (and probably user-defined) coordinates.

In case this solution wasn't 100% consistent with the intersection calculation IMHO we should use exactly the same method for sides and intersection (i.e. cramer's rule in cartesian CS) which could be even slower.

mkaravel:
I agree with Adam. I think it simply suffices to keep the same interface (towards the user) and internally optimize for integral coordinates.
I would treat user-defined types the same way as floating-point types.

barendgehrels:
Of course we should prefer correctness over performance, I therefore agreed with the solution. But it should be optimal, and that is easy to do here. An automatic solution  (specialization) to avoid the comparisons is OK for me

mkaravel:
@barendgehrels Please see the last commit: the optimization is implemented there.

mkaravel:
@awulkiew I do not see the problem. Before calling `relate::less` I have already checked for equality, which means that when `relate::less` is called the points are far enough away.

awulkiew:
@mkaravel Ah yes, indeed. Besides degenerated segments should be filtered out.

barendgehrels:
I'm OK with this.

PR #213, opened 2015-02-09
Fix: replace assertion by exception

mkaravel:
The assertion failure that is replaced indicates that something went wrong in the turns' computation, and more precisely that the turns computed are not consistent; such a case should better be handled by throwing an exception rather than an assertion.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/follow_linear_linear.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/follow_linear_linear.hpp
@@ -38,21 +38,16 @@ namespace boost { namespace geometry
 #if ! defined(BOOST_GEOMETRY_OVERLAY_NO_THROW)
 class inconsistent_turns_exception : public geometry::exception
 {
-    std::string message;
-
 public:
 
-    inline inconsistent_turns_exception()
-    {
-        message = "Boost.Geometry Inconsistent Turns exception";
-    }
+    inline inconsistent_turns_exception() {}
 
     virtual ~inconsistent_turns_exception() throw()
     {}
 
     virtual char const* what() const throw()
     {
-        return message.c_str();
+        return "Boost.Geometry Inconsistent Turns exception";
     }
 };
 #endif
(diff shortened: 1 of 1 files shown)


include/boost/geometry/algorithms/detail/overlay/follow_linear_linear.hpp
```diff
@@ -35,6 +35,28 @@
 namespace boost { namespace geometry
 {
 
+#if ! defined(BOOST_GEOMETRY_OVERLAY_NO_THROW)
+class inconsistent_turns_exception : public geometry::exception
+{
+    std::string message;
```
barendgehrels: m_message (btw it seems const, you can probably also just return it in what())

include/boost/geometry/algorithms/detail/overlay/follow_linear_linear.hpp
```diff
@@ -304,7 +326,14 @@ class follow_linestring_linear_linestring
...
 
-        BOOST_ASSERT( enter_count == 0 );
+#if ! defined(BOOST_GEOMETRY_OVERLAY_NO_THROW)
+        if (enter_count != 0)
+        {
+            throw inconsistent_turns_exception();
+        }
+#else
+        BOOST_ASSERT(enter_count == 0);
+#endif
```
barendgehrels: OK for me.
We might think of a macro which implements this behaviour in a sort of generic way, for example
```
BOOST_GEOMETRY_THROW_OR_ASSERT(enter_count == 0, 
  inconsistent_turns_exception, 
  "Enter count: " << enter_count);
```
so the third argument is a streamable thing part of the exception

include/boost/geometry/algorithms/detail/overlay/follow_linear_linear.hpp
mkaravel: Done as suggested.

include/boost/geometry/algorithms/detail/overlay/follow_linear_linear.hpp
mkaravel: I like the idea. Will work on it on another PR.

barendgehrels:
I'm OK with merge

PR #214, opened 2015-02-10
Fix/set ops ll add test cases and tolerance

mkaravel:
* Add support for checking equality of results with a specified tolerance.
* Add more test cases.


barendgehrels:
+1 for added test coverage. Thanks. I'm OK with merging

mkaravel:
@barendgehrels FYI, I replaced checking against an absolute tolerance by checking against a relative one (see last commit above b7ccd1f).

PR #218, opened 2015-02-11
Fix/warnings

awulkiew:
This PR fixes the MSVC compiler warning C4127: conditional expression is constant in all parts of the library.

For this purpose a macro BOOST_GEOMETRY_CONDITION was added. It should be used in cases when a conditional expression may be constant.
Note that the same expression may or may not be constant in various cases when template specialization is involved.


awulkiew:
This fixes the ticket: https://svn.boost.org/trac/boost/ticket/10666

include/boost/geometry/util/condition.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include <boost/config.hpp>
+
+
+#ifdef BOOST_MSVC
+
+//namespace boost { namespace geometry { namespace detail {
+//BOOST_FORCEINLINE bool condition(bool const b) { return b; }
+//}}} // boost::geometry::detail
+//#define BOOST_GEOMETRY_CONDITION(CONDITION) boost::geometry::detail::condition(CONDITION)
```
mkaravel: Shouldn't these comments go away?

include/boost/geometry/util/condition.hpp
```diff
@@ -0,0 +1,34 @@
...
+#define BOOST_GEOMETRY_CONDITION(CONDITION) ((void)0, (CONDITION))
+
+#else
+
+#define BOOST_GEOMETRY_CONDITION(CONDITION) (CONDITION)
+
+#endif
+
+
+#endif // BOOST_GEOMETRY_UTIL_STATIC_CONDITION_HPP
```
mkaravel: Macro name in comment does not match macro in `#ifndef` and `#define` directives at beginning of file.

include/boost/geometry/util/condition.hpp
awulkiew: I'd prefer to keep it in case if some other or future version of MSVC didn't work as expected or started to generate a warning in additional cases, e.g. for the workaround below.

include/boost/geometry/algorithms/convex_hull.hpp
```diff
@@ -123,7 +125,7 @@ struct convex_hull<Box, box_tag>
         boost::array<typename point_type<Box>::type, 4> range;
         geometry::detail::assign_box_corners_oriented<Reverse>(box, range);
         geometry::append(out, range);
-        if (Close)
+        if ( BOOST_GEOMETRY_CONDITION(Close) )
```
barendgehrels: Please remove the extra spacing you added - it is not according to our guidelines, and never used in this sourcefile (at least not in the original version).

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
```diff
@@ -892,7 +893,7 @@ inline void buffer_inserter(GeometryInput const& geometry_input, OutputIterator
 
     collection.get_turns();
     collection.classify_turns(linear);
-    if (areal)
+    if ( BOOST_GEOMETRY_CONDITION(areal) )
```
barendgehrels: Idem, and everywhere

include/boost/geometry/algorithms/detail/overlay/get_turn_info_la.hpp
```diff
@@ -261,7 +263,7 @@ struct get_turn_info_linear_areal
                         AssignPolicy::apply(tp, pi, qi, inters);
                         
                         // conditionally handle spikes
-                        if ( ! handle_spikes
+                        if ( ! BOOST_GEOMETRY_CONDITION(handle_spikes)
```
barendgehrels: I don't object to extra spaces in multi line expressions as this makes the layout quite clear here

include/boost/geometry/algorithms/detail/overlay/get_turn_info_la.hpp
```diff
@@ -529,7 +531,9 @@ struct get_turn_info_linear_areal
                                               bool is_p_last, bool /*is_q_last*/,
                                               OutIt out)
     {
-        bool is_p_spike = ( Version == append_touches ?
+        static const bool is_version_touches = (Version == append_touches);
```
barendgehrels: Hmm, I see why this is done, but this makes the review much more work.
I would have preferred if these had been a separate commit
Anyway, no need to change now

include/boost/geometry/util/condition.hpp
barendgehrels: OK - please add a comment then, in case anyone later thinks to remove old code

test/test_geometries/custom_lon_lat_point.hpp
```diff
@@ -72,14 +72,7 @@ struct access<rw_lon_lat_point<CoordinateType, CoordinateSystem>, Dimension>
...
     {
-        if (Dimension == 0)
-        {
-            p.longitude = value;
-        }
-        else
-        {
-            p.latitude = value;
-        }
+        ( Dimension == 0 ? p.longitude : p.latitude ) = value;
```
barendgehrels: Never seen this before. Does this work everywhere?
I prefer the old version, but than with BOOST_GEOMETRY_CONDITION as done everywhere else

barendgehrels:
Thanks for the PR. See my comments but basically I'm OK having this in our library

test/test_geometries/custom_lon_lat_point.hpp
awulkiew: If by "everywhere" you mean "C++ standard-conformant compiler" then yes :)
The result of conditional operator is an l-value if both the second and third operands are of the same type and both are l-values. In C this operator always returns r-value.

In C++ this operator allows to do many more interesting things. E.g. see the example here: http://en.cppreference.com/w/cpp/language/operator_other#Conditional_operator

I've implemented it this way because it's "compatible" with the implementation of `get()` method, the conditional operator is also used there. It doesn't require any additional utils/macros, the language allows us to do it. And it's only a test.

Though I agree that this is not a well-known feature and a reader may be surprised seing this. So if you strongly prefer the "longer" version with macro I'll change it.

barendgehrels: Thanks for the link.
OK, if this is standard, and supported by all compilers we support, we are allowed to use it. It is indeed much shorter.
So I changed opinion, no need to change, unless we later see in the matrix that this breaks some compilers which are for the rest working for us.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_la.hpp
awulkiew: Yes, this is why I started to use this convention in all cases, to be consistent.

awulkiew:
Ok, thanks for the review. I addressed all of the issues. I'm merging. In case of problems I'll act accordingly.

PR #219, opened 2015-02-11
Fix point_type traits for variant over MPL sequence

sdebionne:
When variant types are generated from MPL sequences the actual variant
type is `boost::variant<over_sequence<Sequence...`
and thus T0 is `over_sequence<>`. This fix use the first bounded type `boost::variant<>::types`
instead of T0.

### DIFF
--- a/include/boost/geometry/geometries/variant.hpp
+++ b/include/boost/geometry/geometries/variant.hpp
@@ -23,10 +23,10 @@ namespace boost { namespace geometry {
 
 
 template <BOOST_VARIANT_ENUM_PARAMS(typename T)>
-struct point_type<variant<BOOST_VARIANT_ENUM_PARAMS(T)> >
+struct point_type<boost::variant<BOOST_VARIANT_ENUM_PARAMS(T)> >
     : point_type<
-        typename mpl::front<
-            typename variant<BOOST_VARIANT_ENUM_PARAMS(T)>::types
+        typename boost::mpl::front<
+            typename boost::variant<BOOST_VARIANT_ENUM_PARAMS(T)>::types
         >::type
     >
 {};
--- a/include/boost/geometry/geometries/variant.hpp
+++ b/include/boost/geometry/geometries/variant.hpp
@@ -16,14 +16,19 @@
 
 
 #include <boost/variant/variant_fwd.hpp>
+#include <boost/mpl/front.hpp>
 
 
 namespace boost { namespace geometry {
 
 
 template <BOOST_VARIANT_ENUM_PARAMS(typename T)>
-struct point_type<boost::variant<BOOST_VARIANT_ENUM_PARAMS(T)> >
-    : point_type<T0>
+struct point_type<variant<BOOST_VARIANT_ENUM_PARAMS(T)> >
+    : point_type<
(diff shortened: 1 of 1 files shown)


include/boost/geometry/geometries/variant.hpp
```diff
@@ -16,14 +16,19 @@
...
+#include <boost/mpl/front.hpp>
 
 
 namespace boost { namespace geometry {
 
 
 template <BOOST_VARIANT_ENUM_PARAMS(typename T)>
-struct point_type<boost::variant<BOOST_VARIANT_ENUM_PARAMS(T)> >
-    : point_type<T0>
+struct point_type<variant<BOOST_VARIANT_ENUM_PARAMS(T)> >
```
mkaravel: Although we are not 100% consistent about this, external (to Boost.Geometry) libraries used in Boost.Geometry code are qualified with `boost::`. So here I would expect `boost::variant` (basically as it was before).

include/boost/geometry/geometries/variant.hpp
```diff
@@ -16,14 +16,19 @@
...
+    : point_type<
+        typename mpl::front<
```
mkaravel: Same here: `mpl` should be `boost::mpl`.

include/boost/geometry/geometries/variant.hpp
```diff
@@ -16,14 +16,19 @@
...
+            typename variant<BOOST_VARIANT_ENUM_PARAMS(T)>::types
```
mkaravel: Same here for `variant`

awulkiew:
Thanks for catching this! So AFAIU this version should work for any wrapper used internally in `boost::variant<>`.

Btw, does it make sense to check all point types to see if the dimensions and coordinate systems are the same or somehow compatible? Or to pick the one with the most precise coordinate type? Or something similar.

sdebionne:
@mkaravel The missing `boost::` namespaces are fixed.

@awulkiew Yes this version should work for any wrapper used internally in boost::variant<> since it uses the public interface `variant<>::types` to get the list of bounded types. I did not check if that fix should be applied somewhere else. I guess @brunolalande would know better.

As for your second question, AFAIU the (variant) `point_type` traits is mostly used to compute the return type of (variant) algorithms. See for instance the `length` algorithm. More complex algorithm, such as `distance`, use a `distance_result` metafunction that computes a variant return type according to the input variant bounded types. In the same vein, an algorithm like `length` could return a variant with say an mp float number or a double (`variant<gmp:float, double>`) depending on the input coordinate types. That would be the most flexible solution, but that may not be worth the trouble... Maybe, for `length`,  checking if all the coordinate types are convertible to double would be enough.

awulkiew:
@sdebionne Actually I had something else in mind. For instance we could store in a variant geometries using different coordinate types. E.g. the first could be a point defined by floats and the second one a polygon defined by doubles. And such variant would be passed into some algorithm internally using a point_type. In such case a better point_type would be the polygon's. Anyway it's just a thought. I don't think we should expect such cases.

I'm in favor of merging but we should probably wait for a while to give the others the time to comment.

PR #220, opened 2015-02-11
[test][strategies][douglas peucker] make the unit test for this strategy

mkaravel:
platform portable and platform independent

### DIFF
--- a/test/strategies/douglas_peucker.cpp
+++ b/test/strategies/douglas_peucker.cpp
@@ -26,6 +26,7 @@
 #include <boost/test/included/unit_test.hpp>
 
 #include <boost/assign/list_of.hpp>
+#include <boost/core/ignore_unused.hpp>
 #include <boost/type_traits/is_same.hpp>
 #include <boost/tuple/tuple.hpp>
 
@@ -39,8 +40,8 @@
 #include <boost/geometry/geometries/adapted/boost_tuple.hpp>
 #include <boost/geometry/geometries/register/multi_point.hpp>
 
-#include <boost/geometry/algorithms/equals.hpp>
 #include <boost/geometry/algorithms/comparable_distance.hpp>
+#include <boost/geometry/algorithms/equals.hpp>
 
 #include <boost/geometry/io/wkt/wkt.hpp>
 #include <boost/geometry/io/dsv/write.hpp>
@@ -188,6 +189,8 @@ struct test_one_case
 #endif
 
         strategy.apply(geometry, std::back_inserter(result), max_distance);
+
+        boost::ignore_unused(strategy);
 
 #ifdef BOOST_GEOMETRY_TEST_DEBUG
         print_point_range(std::cout, boost::begin(result), boost::end(result),
--- a/test/strategies/douglas_peucker.cpp
+++ b/test/strategies/douglas_peucker.cpp
@@ -40,6 +40,7 @@
 #include <boost/geometry/geometries/register/multi_point.hpp>
 
 #include <boost/geometry/algorithms/equals.hpp>
+#include <boost/geometry/algorithms/comparable_distance.hpp>
 
 #include <boost/geometry/io/wkt/wkt.hpp>
 #include <boost/geometry/io/dsv/write.hpp>
@@ -221,12 +222,15 @@ inline void test_with_strategy()
 
     typedef bg::model::point<CoordinateType, 2, bg::cs::cartesian> point_type;
     typedef bg::model::linestring<point_type> linestring_type;
+    typedef bg::model::segment<point_type> segment_type;
     typedef test_one_case<linestring_type> tester;
 
     linestring_type ls1, ls2;
     linestring_type res1, res2;
+    point_type const p1(-6,-13), p2(0,-15);
+    segment_type const s(point_type(12,-3), point_type(-12,5));
 
-    if (boost::is_same<CoordinateType, long double>::value)
+    if (bg::comparable_distance(p1, s) >= bg::comparable_distance(p2, s))
     {
         tester::apply("LINESTRING(12 -3, 4 8,-6 -13,-9 4,0 -15,-12 5)",
                       10,


test/strategies/douglas_peucker.cpp
```diff
@@ -221,12 +225,15 @@ inline void test_with_strategy()
...
     linestring_type ls1, ls2;
     linestring_type res1, res2;
+    point_type const p1(-6,-13), p2(0,-15);
+    segment_type const s(point_type(12,-3), point_type(-12,5));
 
-    if (boost::is_same<CoordinateType, long double>::value)
+    if (bg::comparable_distance(p1, s) >= bg::comparable_distance(p2, s))
     {
         tester::apply("LINESTRING(12 -3, 4 8,-6 -13,-9 4,0 -15,-12 5)",
                       10,
```
barendgehrels: I think we should slightly change one of the offending coordinates instead of this fix.

barendgehrels: I don't understand this case. In SQL Server this linestring is not simplified until the distance is about 13.6
If so, the output is LINESTRING (12 -3, -6 -13, -9 4, 0 -15, -12 5) so it is 4 8 which is removed here. So how can we have for our case both points removed, in one case, and not in another case?
Don't know the exact algorithm, but it is at least suspicious.

mkaravel: The algorithm that has been implemented is the one described in the following wikipedia page:
http://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm

What happens is the following. The end points `(12 -3)` and `(-12 5)` are fixed. The remaining points are considered, in order to find the one at maximum distance from the segment of the two endpoints. The points at maximum distance from this segment are `(-6 -13)` and `(0 -15)`; which one is really further away depends on the accuracy of the computations, so rounding errors play a role here (which is why in the unit test there is this check of comparable distances).

Assume for now that the point that is further away from the segment is `(0 -15)`. This point is marked as point to be kept as its distance from the segment is about 15.17 which is greater than the distance passed to the algorithm which is 10. The algorithm then continues recursively with the two linestrings `LINESTRING(12 -3,4 8,-6 -13,-9 4,0 -15)` and `LINESTRING(0 -15,-12 5)`. the second is not simplified as it has only two points. For the first one the segment `(12 -3,0 -15)` is considered and the point that is further away from the segment is `(-9 4)` at distance 19.8. Again this distance is above 10 so the point is kept and the algorithm applied recursively to the two linestrings `LINESTRING(12 -3,4 8,-6 -13,-9 4)` and `LINESTRING(-9 4,0 -15)`. The second has only two points, so it is not simplified. For the first, the distances of `(4 8)` and `(-6 -13)` from the segment `(12 -3,-9 4)` are computed, and `(-6 -13)` is found to be further away, at distance 15.17. So the point `(-6 -13)` is kept and the algorithm continues recursively with the linestrings `LINESTRING(12 -3,4 8,-6 -13)` and `LINESTRING(-6 -13,-9 4)`. The second is not simplified, while for the first the distance of `(4 8)` from the segment `(12 -3,-6 -13)` is computed. This distance is 13.5, which is again above 10, so the point `(4 8)` is kept. The last recursive calls are on the linestrings `LINESTRING(12 -3,4 8)` and `LINESTRING(4 8,-6 -13)`, which both have two points so they are not simplified.
Bottom line: all points are kept and the input linestring is not simplified at all.

Assume now that the point that is further away from the segment is `(-6 -13)`. This point is kept, and the algorithm recursively considers the linestrings `LINESTRING(12 -3,4 8,-6 -13)` and `LINESTRING(-6 -13,-9 4,0 -15,-12 5)`. For the first linestring there is no simplification (see explanation above in the previous case), while for the second the point that is further away from the segment `(-6 -13,-12 5)` is the point `(0 -15)` at distance 6.32. This distance is below 10, so the linestring `LINESTRING(-6 -13,-9 4,0 -15,-12 5)` is simplified to `LINESTRING(-6 -13,-12 5)`. So the end result is `LINESTRING(12, -3,4, 8,-6 -13,-12 5)`.

Finally, regarding PostGIS: without looking at the sources, it implements exactly the same algorithm, and it behaves in exactly the same way as our implementation does when the coordinates are doubles. At the first recursive step the point `(0 -15)` is found as further away from the segment connecting the points of the original linestring, at distance 15.17. Then essentially the linestring `LINESTRING(12 -3,4 8,-6 -13,-9 4,0 -15)` is considered, and the point further away from it is `(-9 4)` at distance 19.79. Then essentially the next linestring considered is `LINESTRING(12 -3,4 8,-6 -13,-9 4)` and the point further away from it is `(-6 -13)` at distance 15.17. The next linestring considered is  `LINESTRING(12 -3,4 8,-6 -13)` and the distance of `(4 8)` from the endpoints of this linestring is computed. This distance is 13.50087. For an input distance of 10 (like in the unit test), the point `(4 8)` is kept, and the entire original linestring is returned as is. For an input simplifcation distance of more than 13.50087 (like the distance 13.6 in your case), and of course less than 15.17 (because then we would have had simplifications at an earlier stage) the point `(4 8)` is simplified and the final outcome is `LINESTRING(12 -3,-6 -13,-9 4,0 -15,-12 5)`, which is precisely what PostGIS does.

As far as I am concerned, this is a perfectly valid, and IMHO very interesting, example, I do not see any problems whatsoever, and our results are consistent with what PostGIS does.

One final note: all the distances I mentioned above are not really computed by our algorithm, but rather their comparable counterparts, and each comparable distance computed is checked against the square of the simplification distance.

barendgehrels: With "I don't know the algorithm", I meant that I don't know which algorithm is used in SQL Server. Which was indeed not clear from my comment.

The Douglas Peucker algorithm I know very well, I implemented all this code, first together with a colluege back in 1995, and in 2007 again for Boost.Geometry...

For the rest, thanks for your explanations, did not yet read them all but will take some time for that tomorrow.

mkaravel: @barendgehrels From the PostGIS online manual:
http://www.postgis.org/docs/ST_Simplify.html

PR #222, opened 2015-02-13
Fix/re factor code for get rescale policy

mkaravel:
The purpose of this PR is to factor out common code related to `get_rescale_policy`.


barendgehrels:
This is very convenient, I planned to do this twice but each time something else got my priority.
Thanks a lot!

PR #223, opened 2015-02-13
[geometry] qualify mpl:: namespace by boost:: (namespace external to the

mkaravel:
Boost.Geometry library); fix long lines produced by the addition of "boost::"


barendgehrels:
I'm OK with merging

PR #224, opened 2015-02-13
Fix: replace access method for accessing the front element of rings

mkaravel:
Change access of a ring's front range element from using the `.front()` method by the `range::front` Boost.Geometry utility function.


awulkiew:
If there is a problem with added not being properly used as an object of a type adapted to Ring concept then why only front() was changed? AFAIU empty() and back() should also be replaced.

mkaravel:
@awulkiew I agree. See the last commit.

mkaravel:
@awulkiew I am referring to commit 2f5d497

PR #226, opened 2015-02-17
Fix/test is valid add case ids

mkaravel:
In this PR:
* String-based IDs are added to the test cases (for easier identification).
* Additional test cases (coming from buffer computations) have been added.

### DIFF
--- a/test/algorithms/is_valid.cpp
+++ b/test/algorithms/is_valid.cpp
@@ -1,7 +1,7 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 // Unit Test
 
-// Copyright (c) 2014, Oracle and/or its affiliates.
+// Copyright (c) 2014-2015, Oracle and/or its affiliates.
 
 // Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
 
@@ -18,6 +18,7 @@
 
 #include <from_wkt.hpp>
 #include "test_is_valid.hpp"
+#include <boost/geometry/algorithms/reverse.hpp>
 
 
 BOOST_AUTO_TEST_CASE( test_is_valid_point )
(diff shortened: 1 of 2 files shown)


awulkiew:
I'm of course ok with the change. However AFAIK in most cases the ID is the first parameter of test functions in Geometry. Then the geometries, then expected results, and additional parameters. Though I don't know if this order is required or if there is some guideline. Therefore I'll wait for Barend's comments before merging.

mkaravel:
I have not checked this, but I take your word for it. I usually do it the other way around in my unit tests: I indicate the case id as the last parameter passed to the tester (see for example set operations).

I had not realized we have a rule about that. Let's indeed wait for Barend's feedback on this.

barendgehrels:
Yes, I normally pass the case id first. It's not in the guidelines, it's more a habit. The order is as Adam describes.

barendgehrels:
But there are more differences: the statements
```
 test::apply(from_wkt<G>("MULTILINESTRING()"), true);
```
Look highly repetitive.
I usually call
```
 test<G>("mycase", "MULTILINESTRING()"), true);
```
instead

mkaravel:
Adopted both suggestions (see commit 4ab7a4f).
The tester now takes plain WKTs as arguments, and the case ID is the first argument.

I have kept the `apply()` static method, as my tester is a class rather than a function. I could fake a function like syntax by overloading the constructor, but I do not think this is necessary or that important.

barendgehrels:
Thanks for adapting. It looks good.
No, please don't create a constructor, that is indeed not the intention.

I have functions, not classes. But no need to change this - if you need a class, fine

mkaravel:
Okay, let's merge this PR then.

PR #229, opened 2015-02-20
New algorithm: is valid reason

mkaravel:
This PR proposes a new algorithm: `is_valid_reason`.
The free function `is_valid_reason` takes a geometry and returns a string indicating is the geometry is valid or a message indicating the source of invalidity.

Internally this has been implemented by introducing visitors in the `is_valid` algorithm: inside the `is_valid` algorithm, the visitor sets an appropriate failure type value, and possibly registers additional data related to the failure; it is then up to the visitor implementation to decide what to do with the data passed.

Two visitors have been implemented as policies (stored in `boost/geometry/policies/is_valid`):
* `is_valid_null_policy`: does nothing with the passed data (this policy is used by default in the `is_valid` algorithm)
* `failing_reason_policy`: keeps track of the failure value and possibly the additional data, and outputs a relevant message to an `std::ostringstream`; this policy is used by the `is_valid_reason` algorithm (the message in the string stream is returned by the algorithm)

Things to do and/or consider:
* Is the enum type `validity_failure_type` at the appropriate location in BG's file tree?
* The documentation for `is_valid_reason` is not there yet; will be added once the code is reviewed and merged.
* To what level or detail should the failure messages dive into? For example for a polygon (invalid actually) where on interior ring is touching from the outside the exterior ring, the current version of the algorithm reports this case as an "invalid self-intersection point"; should the message rather be that the interior ring is outside the exterior ring (needs additional code to implement this feature).

A final note on the policy design: the policy's visitor is designed to take the failure value as a template parameter rather than as an argument. This has been done on purpose in order to allow for the possibility to call the visitor's apply method that take data with different types of data (and whose types depends of course on the type of failure). So with this design, for example, if the failure value is `failure_self_intersections` all turns are passed, where as if the failure value is `failure_spikes` the apex of the spike is passed.


include/boost/geometry/policies/is_valid/failing_reason_policy.hpp
```diff
@@ -0,0 +1,150 @@
...
+    case failure_wrong_orientation:
+        return "Geometry has wrong orientation";
+    case failure_holes_outside:
+        return "Geometry has holes defined outside the outer boundary";
+    case failure_nested_holes:
+        return "Geometry has nested holes";
+    case failure_disconnected_interior:
+        return "Geometry has disconnected interior";
+    case failure_intersecting_interiors:
+        return "Multi-polygon has intersecting interiors";
```
barendgehrels: Should this not be "Geometry" ? Also a polygon can have intersection interiors

include/boost/geometry/policies/is_valid/failing_reason_policy.hpp
```diff
@@ -0,0 +1,150 @@
...
+    case failure_self_intersections:
+        return "Geometry has invalid self-intersections";
+    case failure_wrong_orientation:
+        return "Geometry has wrong orientation";
+    case failure_holes_outside:
+        return "Geometry has holes defined outside the outer boundary";
+    case failure_nested_holes:
+        return "Geometry has nested holes";
+    case failure_disconnected_interior:
+        return "Geometry has disconnected interior";
```
barendgehrels: What is the difference between holes outside boundary?

include/boost/geometry/policies/is_valid/failing_reason_policy.hpp
```diff
@@ -0,0 +1,150 @@
...
+    switch (failure)
+    {
+    case no_failure:
+        return "Geometry is valid";
+    case failure_few_points:
+        return "Geometry has too few points";
+    case failure_wrong_dimension:
+        return "Geometry has wrong dimension";
+    case failure_not_closed:
+        return "Geometry is defined as closed but is open";
```
barendgehrels: Do we have "... as open but is closed" too? Or is that no problem?

include/boost/geometry/policies/is_valid/failing_reason_policy.hpp
```diff
@@ -0,0 +1,150 @@
...
+        return "Geometry has invalid self-intersections";
+    case failure_wrong_orientation:
+        return "Geometry has wrong orientation";
+    case failure_holes_outside:
+        return "Geometry has holes defined outside the outer boundary";
+    case failure_nested_holes:
+        return "Geometry has nested holes";
+    case failure_disconnected_interior:
+        return "Geometry has disconnected interior";
+    case failure_intersecting_interiors:
```
barendgehrels: We should talk about either interiors (we have that everywhere) or holes (not in OGC spec) but not mixed.

include/boost/geometry/policies/is_valid/failing_reason_policy.hpp
mkaravel: If a polygon has holes that intersect, then we either have invalid self-intersection points or nested holes (see cases above).
This refers to multi-polygons whose individual polygons are valid, but are such that their interiors intersect (in which case the multi-polygon is invalid, but not the individual polygons).
This is why I restrict this to multi-polygons.

include/boost/geometry/policies/is_valid/failing_reason_policy.hpp
mkaravel: The holes-outside-boundary case refers to the case where an interior boundary is not inside the exterior boundary.
This refers to the case were all holes are inside the (closure of the) exterior boundary, but the interior of the polygon is disconnected. If you look at the example in the documentation of `is_valid`, it refers to exactly this case.

barendgehrels:
I like the design with visit_policy

include/boost/geometry/policies/is_valid/failing_reason_policy.hpp
mkaravel: This can never happen. An open geometry is by default closed (the closing point is automatically added). If you define a geometry as open, but the two endpoints coincide, you simply have a duplicate point at the end.

include/boost/geometry/policies/is_valid/failing_reason_policy.hpp
mkaravel: A hole for me is the interior of an inner ring. The interior above refers to the interior of the areal geometry (so what is outside holes).
This is what I had in mind at least. I will check again to make sure I am self-consistent.

include/boost/geometry/algorithms/detail/is_valid/box.hpp
```diff
@@ -30,15 +31,23 @@ namespace detail { namespace is_valid
...
+    template <typename VisitPolicy>
+    static inline bool apply(Box const& box, VisitPolicy& visitor)
     {
-        if ( geometry::get<geometry::max_corner, I-1>(box)
-             <=
-             geometry::get<geometry::min_corner, I-1>(box) )
+        if (math::equals(geometry::get<geometry::min_corner, I-1>(box),
+                         geometry::get<geometry::max_corner, I-1>(box)))
         {
+            visitor.template apply<failure_wrong_dimension>();
```
barendgehrels: So this is invalid (was already). I find the word "dimension" a little confusing - it is the topological dimension, no? Maybe just add "topological" here?

mkaravel:
Should we also have a version of the `is_valid()` and/or `is_valid_reason()` free functions that take a visitor as a second argument?

include/boost/geometry/algorithms/detail/is_valid/box.hpp
mkaravel: Yes, it is topological dimension.
Will add it in both the enum type value as well as the message.
Thanks for pointing this out.

include/boost/geometry/algorithms/detail/is_valid/pointlike.hpp
```diff
@@ -42,14 +45,26 @@ struct is_valid<Point, point_tag>
...
+            // we allow empty multi-geometries, so an empty multipoint
+            // is considered valid
+            visitor.template apply<no_failure>();
+            return true;
+        }
+        else
+        {
+            // we do not allow an empty multipoint
+            visitor.template apply<failure_few_points>();
+            return false;
```
barendgehrels: I don't know if it is a good idea, but just brainstorm: we have this pattern often, visitor.template ... no_failure, return true, or visitor.template failure, return false. Can we not just return visitor.template apply<....> ? So then the apply with no-failure always returns true and the rest false? Or does this have drawbacks?
Anyway, if good idea, you don't have to apply that now, we can postpone that action.

include/boost/geometry/algorithms/is_valid_reason.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+// Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
+
+// Licensed under the Boost Software License version 1.0.
+// http://www.boost.org/users/license.html
+
+#ifndef BOOST_GEOMETRY_ALGORITHMS_IS_VALID_REASON_HPP
+#define BOOST_GEOMETRY_ALGORITHMS_IS_VALID_REASON_HPP
+
+#include <string>
```
barendgehrels: #include <sstream>

include/boost/geometry/algorithms/is_valid_reason.hpp
```diff
@@ -0,0 +1,45 @@
...
+*/
+template <typename Geometry>
+inline std::string is_valid_reason(Geometry const& geometry)
+{
+    std::ostringstream sstr;
+    failing_reason_policy policy_visitor(sstr);
+    resolve_variant::is_valid<Geometry>::apply(geometry, policy_visitor);
+    return sstr.str();
+}
+
```
barendgehrels: Looks very good and very small, nice.
sstr -> I would call it just "stream" that avoids the abbreviation

include/boost/geometry/algorithms/detail/is_valid/pointlike.hpp
mkaravel: What I can do it is to have the visitor's apply method return the boolean. If that is what you mean it is easy to do. I already considered doing so, but I decided to avoid it: the drawback that I see is that it "complicates" the interface of the visitor. I would prefer the visitor to just take info from the algorithm passively, rather than actively participate in computing the result. This way we can avoid potential "malicious" policies that disregard the failure value and return whatever....

What I think (I have to double-check this, since I think I already tried it) I cannot do is to eliminate the `if (...)` statement. When I call the visitor I need the exact precise failure type at compile time.

test/algorithms/is_valid.cpp
```diff
@@ -238,7 +238,7 @@ BOOST_AUTO_TEST_CASE( test_is_valid_multilinestring )
 }
 
 template <typename Point, bool AllowDuplicates>
-void test_open_rings()
+inline void test_open_rings()
```
barendgehrels: The inline is not needed in a cpp file. Anyway, no problem further

include/boost/geometry/algorithms/is_valid_reason.hpp
mkaravel: Will add it. Thanks.

include/boost/geometry/algorithms/is_valid_reason.hpp
mkaravel: Sure. I will do this also.

test/algorithms/is_valid.cpp
mkaravel: In the past you had "complained" about it, so I decided to consistently use `inline` even in `.cpp` files.

barendgehrels:
I like it. It is a large change-list but most changes follow the same pattern. It is complete with tests, and you mention docs will follow later. Thanks a lot, I'm OK with merging.

test/algorithms/is_valid.cpp
barendgehrels: ;-) I hope I complained about it in a headerfile - there they belong...

include/boost/geometry/algorithms/detail/is_valid/pointlike.hpp
barendgehrels: OK, but it can be an advantage too, to have a specific visitor that copies normal behaviour but for one specific situation it reverses the result...
It is for me not necessary to avoid the if, I was just considering letting the visitor return the result indeed. If you considered this and still like it, OK for me.

barendgehrels:
You mean, as a sort of strategy? Good idea but I would postpone this a 
bit, let's first release this, get the visitor interface completely 
settled down, and then consider it again.

include/boost/geometry/algorithms/detail/is_valid/box.hpp
mkaravel: Done. See commits 9357d09, 667d5ea and 0987bc9.

include/boost/geometry/algorithms/is_valid_reason.hpp
mkaravel: Done. See commit c098f89.

include/boost/geometry/algorithms/is_valid_reason.hpp
mkaravel: Done. See commit c098f89.

barendgehrels:
I don't know if I should mention this now, but it occurred to me that is_valid_reason is actually a weird name. I suggested it myself, plus it is in PostGIS
http://postgis.net/docs/ST_IsValidReason.html
so we could comply with that.
However, it is not the reason for validity - it is the reason for invalidity... get_reason_for_invalidity... hmmm, I don't know. Maybe just leave it as it is

mkaravel:
About `is_valid()` and `is_valid_reason()` 2nd argument: okay, let's postpone it for now.
It is an easy fix in the code anyways (more work on the documentation side).

mkaravel:
Yes, it is a weird name, I agree with you.
When you suggested it, I looked up the PostGIS manual and found it. This was the reason I decided to name it this way.
BTW, I like the one-line description of the function in the PostGIS manual. I will most probably use that in out documentation as well.

include/boost/geometry/algorithms/detail/is_valid/pointlike.hpp
mkaravel: I have to admit I did not think of is as an advantage, but you are right.

Here is one specific advantage I can think of: right now we allow duplicate points, and this is done via a template parameter that is by default set to `true` (allow duplicate points). If we switch the default value to `false`, then different visitors can allow or disallow duplicate points, and this can be done at run-time instead of compile time.

So here is what I propose: change the interface of the visitor's `apply` methods to return the boolean success/failure value, which will then be used by `is_valid()` as the return value.
Also change the default value of the `AllowDuplicates` template parameter of `dispatch::is_valid` to `false`, and modify the two visitors we have now to accept duplicate points as valid. The drawback I see here is that `is_valid()` will now check for duplicate points, which means a small time overhead (personally, I am okay with this additional overhead).

What do you think?

include/boost/geometry/policies/is_valid/failing_reason_policy.hpp
mkaravel: I went through the values of the `validity_failure_type` enum. I prefer them as they are now: "holes" are the inner rings, and "interior" refers to the interior of the geometries.

I can replace `holes` by `interior_rings` is that is preferable for you.

mkaravel: Second thoughts: we never use "holes" in BG, but rather "interior rings".
So I will just simply go ahead and rename "holes" to "interior rings": this way it will be consistent with the rest of the library.

mkaravel: I just grepped the BG source code.
The word "holes" is used both in comments and code, but only internally.
Given that the validity failure enum values are visible to the user and error messages are part of the output, I prefer to change them as I mention in my previous message,

barendgehrels: Agreed, thanks.

barendgehrels:
OK, that description explains also the name a bit, indeed

include/boost/geometry/policies/is_valid/failing_reason_policy.hpp
mkaravel: Done. see commits: 5a47531, 004b5af, ef32b10 and 31ecf74.

mkaravel:
As far as I can tell, there is one more remaining issue: whether the policy's `apply` methods should return a boolean.
@barendgehrels Could you please comment on the suggestion I made?

include/boost/geometry/algorithms/detail/is_valid/pointlike.hpp
barendgehrels: I'm OK with this proposal. Yes it makes it more generic - probably you don't even need the AllowDuplicates template parameter anymore? Because it's then the visitors responsibility (it can still be compiletime then, I think)

mkaravel:
Changes implemented in today's commits:
* The handling of spikes, for linear geometries, is now also done by the visitor.
* The free function `is_valid_reason()` has been removed and in its place an overload of `is_valid()` that takes a geometry and a (reference to a) string has been introduced.
* The documentation for the new overload above has been added.
* Two more free function overloads have been added for `is_valid()`: one that takes a geometry and a (reference to a) visitor, and one that takes a geometry a (reference to a) validity failure type. Both these functions are still undocumented. They will be documented in the boost 1.58 time frame if time permits.
* The `is_simple()` has been modified to use an appropriate policy in order to check for simplicity failures related to spikes and duplicate points.

mkaravel:
In commit 1917641, the `failing_reason_policy` is modified so it prints more information about the invalid self-intersection points found.

barendgehrels:
Thanks for processing all so quickly. The new interface (overload with string) is a good idea.
I'm OK with merging

test/algorithms/is_valid_failure.cpp
```diff
@@ -0,0 +1,979 @@
...
+    case bg::failure_disconnected_interior:
+        return "failure_disconnected_interior";
+    case bg::failure_intersecting_interiors:
+        return "failure_intersecting_interiors";
+    case bg::failure_wrong_corner_order:
+        return "failure_wrong_corner_order";
+    default:
+        return ""; // to avoid warnings (-Wreturn-type)
+    }    
+}
```
awulkiew: Is `to_string()` test function really needed when there is identical `validity_failure_type_message()`? This is probably used to test the message but I think it's not really needed, especially if we consider that the message could be extended to e.g. contain some helper data like point coordinates. The `failure_type` would stay the same, so I think it would be sufficient to just compare the failure types and maybe to check if the failure type is consistent with the message, no need to compare the messages.

awulkiew:
Yes, this is a great addition! And I'm ok with merging. I saw one place where msvc could generate a "constant conditional expression" warning but we can test and fix any warnings after merging.

I also think that the name `is_valid_reason()` is not a good name, for me it'd be a function returning `true` if a reason was valid. I prefer 2 argument `is_valid()`, taking a reference to string. I'm even ok with not documenting the other overload at all, at least until it's explicitly requested by the users (btw we could ask about it on the list).

include/boost/geometry/algorithms/detail/is_valid/multipolygon.hpp
```diff
@@ -89,22 +94,29 @@ class is_valid_multipolygon
...
-        return !visitor.items_overlap;
+        if (item_visitor.items_overlap)
+        {
+            return visitor.template apply<failure_intersecting_interiors>();
+        }
+        else
+        {
+            return visitor.template apply<no_failure>();
+        }
     }
```
barendgehrels: Looks much better now, returning apply.

mkaravel:
Thank you both for the comments and for merging.
I will prepare another PR with the documentation for `validity_failure_type` and the version of `is_valid()` that takes a reference to this enum as second argument. It should be ready later on today.

PR #234, opened 2015-02-26
[strategies][cartesian][buffer][point_circle] fix internal point count

mkaravel:
for small input values (less than 3): when the input point count is less
than 3, set the internal point count to 3; this is important for generating
valid polygons;


barendgehrels:
Thanks

PR #235, opened 2015-02-27
Fix range back() function

sdebionne:
`*(--boost::end(rng));` is not valid for range with raw pointer as iterator (the compiler complains that `'--' needs l-value`. Since the range is supposed to be BidirectionalRange, use `*rbegin()` instead. To illustrate the issue, the following code doesn't not compile:
```
typedef boost::array<double, 2> point_t;
typedef boost::iterator_range<point_t*> linestring_t;

linestring_t l;

point_t back = *(--boost::end(l));
```
NOTE:  `*(--boost::end(rng));` is used in other places (e.g. point_in_geopetry.hpp:197) and should probably use geometry::back instead. If the fix is OK for you I could try to track and fix them.

### DIFF
--- a/include/boost/geometry/algorithms/detail/within/point_in_geometry.hpp
+++ b/include/boost/geometry/algorithms/detail/within/point_in_geometry.hpp
@@ -36,6 +36,7 @@
 #include <boost/geometry/strategies/within.hpp>
 #include <boost/geometry/strategies/covered_by.hpp>
 
+#include <boost/geometry/util/range.hpp>
 #include <boost/geometry/views/detail/normalized_view.hpp>
 
 namespace boost { namespace geometry {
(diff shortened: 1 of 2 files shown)


awulkiew:
Thanks for catching this!

To avoid the creation of an object of reverse_iterator we could manually implement it as:

    typename boost::range_iterator<BidirectionalRange>::type
        it = boost::end(rng);
    return *(--it);

though I'm not sure if this is needed, your version is shorter and elegant. Thoughts?

Of course, feel free to track and fix those places.

P.S. Please add a tag at the begin of the commit message indicating a part of a library. Such name may be a name of a directory or namespace. In the case of the above commit it could be [util] or [range]. In the case of point_in_geometry it'd be [algorithms], etc.

sdebionne:
It turns out that point_in_geometry.hpp was the only file that uses `*(--boost::end(rng));`. I have added the modified file to the PR.

I would expect the reverse_iterator construction to be completely inlined and optimized away by the compiler but would not put my fingers on it.

How shall I amend my previous commit to add the [range] prefix ? Is it OK for this time or should I "rewrite history" (since the commit has already published to github) ?

mkaravel:
Since the changes you have made are really simple, I would suggest (at least this is what I would do), to close this PR and create another one with the proper commit messages.
Also please use the `[algorithms]` tag in front of `[range]` or `[within]`. Something like:
```
[algorithms][range] Fix range back
*(--boost::end(rng)); is not valid for range with pointer as iterators.
since the range is supposed to be BidirectionalRange, use rbegin
instead.
```

sdebionne:
Allright, I'm closing this PR and will reopen a new one.

PR #236, opened 2015-03-02
Fix range back() function

sdebionne:
`*(--boost::end(rng));` is not valid for range with raw pointer as iterator (the compiler complains that `'--' needs l-value`. Since the range is supposed to be BidirectionalRange, use `*rbegin()` instead. See [PR#235](https://github.com/boostorg/geometry/pull/235) for the full description and discussion.


mkaravel:
I am okay with merging.
Thanks a lot for the fix.

PR #237, opened 2015-03-02
[strategies][cartesian][buffer][end_round] fix internal point count

mkaravel:
for small input values (less than 4): when the input point count is less
than 4, set the internal point count to 4; for values less than 4 the
generated ends do not look round;


barendgehrels:
Thnks

barendgehrels:
BTW it is OK for now, for next release I'm considering throwing exceptions if the number of points is not enough, or too much, for join strategies (and probably for miter factor or other relevant values)

PR #238, opened 2015-03-03
Fix number of generated points in join_round buffer strategy

mkaravel:
Make sure that at least one interior point is generated by the strategy.

### DIFF
--- a/include/boost/geometry/strategies/cartesian/buffer_join_round.hpp
+++ b/include/boost/geometry/strategies/cartesian/buffer_join_round.hpp
@@ -96,10 +96,8 @@ private :
         // visually correct also for a low number of points / circle
         // n is set to be at least 2 so that at least one interior
         // point is generated
-        int const n = static_cast<int>
-            (
-                 (std::max)(m_points_per_circle * dangle / two_pi, two)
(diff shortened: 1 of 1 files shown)


barendgehrels:
Thanks - I will merge this right away.
Can you in a follow-up PR create a unit-test for this change (you can put it in linestring_buffer.cpp), and maybe also for the changes related to min.number of points you did recently?
Actually these are more strategy-changes. We don't have unit tests for them. So either you can add them to linestring_buffer, or (if you have the time) you can create a separate unit test (but based on linestring_buffer, I want to use test_buffer.hpp also for strategies).

include/boost/geometry/strategies/cartesian/buffer_join_round.hpp
```diff
@@ -83,15 +90,16 @@ private :
         {
             angle2 -= two_pi;
         }
+        PromotedType const dangle = angle1 - angle2;
```
awulkiew: This name may be confusing. It's like if there was something dangling, like a spike below a polygon or a pointer.
I'd probably use `andle_diff`. Just saying, no need to change it.

PR #239, opened 2015-03-03
Fix point_iterator with raw pointer base

sdebionne:
Fix `point_iterator` (and adapt `reverse_point_iterator` to the fix) to support linestrings/rings implemented with raw pointer as iterator. The issue can be reproduced with:
```
boost::iterator_range<const point_t *> l;
points_begin(l);
```
The compiler complain that `const point_t *` is not a legal base class of `point_iterator`.

The fix uses `boost::iterator_adaptor` that implements base() a synonyme of base_ptr() that consequently has been removed. The operators are also redondant and removed.

### DIFF
--- a/include/boost/geometry/iterators/point_iterator.hpp
+++ b/include/boost/geometry/iterators/point_iterator.hpp
@@ -11,6 +11,7 @@
 #define BOOST_GEOMETRY_ITERATORS_POINT_ITERATOR_HPP
 
 #include <boost/assert.hpp>
+#include <boost/iterator/iterator_adaptor.hpp>
 #include <boost/mpl/assert.hpp>
 #include <boost/type_traits/is_convertible.hpp>
 #include <boost/range.hpp>
@@ -245,33 +246,26 @@ struct points_end<MultiPolygon, multi_polygon_tag>
 // MK:: need to add doc here
 template <typename Geometry>
 class point_iterator
-    : public detail::point_iterator::iterator_type<Geometry>::type
+    : public boost::iterator_adaptor <
+        point_iterator<Geometry>
+      , typename detail::point_iterator::iterator_type<Geometry>::type
+    >
 {
 private:
     typedef typename detail::point_iterator::iterator_type<Geometry>::type base;
-
-    inline base* base_ptr()
-    {
-        return this;
-    }
-
-    inline base const* base_ptr() const
-    {
-        return this;
-    }
 
     template <typename OtherGeometry> friend class point_iterator;
     template <typename G> friend inline point_iterator<G> points_begin(G&);
     template <typename G> friend inline point_iterator<G> points_end(G&);
 
-    inline point_iterator(base const& base_it) : base(base_it) {}
+    inline point_iterator(base const& base_it) : point_iterator::iterator_adaptor_(base_it) {}
 
 public:
     inline point_iterator() {}
 
     template <typename OtherGeometry>
     inline point_iterator(point_iterator<OtherGeometry> const& other)
-        : base(*other.base_ptr())
+        : point_iterator::iterator_adaptor_(other.base())
     {
         static const bool is_conv
             = boost::is_convertible<
@@ -288,32 +282,6 @@ public:
         BOOST_MPL_ASSERT_MSG((is_conv),
                              NOT_CONVERTIBLE,
                              (point_iterator<OtherGeometry>));
-    }
-
-    inline point_iterator& operator++() // prefix
-    {
-        base::operator++();
-        return *this;
-    }
-
-    inline point_iterator& operator--() // prefix
-    {
-        base::operator--();
-        return *this;
-    }
-
-    inline point_iterator operator++(int) // postfix
-    {
-        point_iterator copy(*this);
-        base::operator++();
-        return copy;
-    }
-
-    inline point_iterator operator--(int) // postfix
-    {
-        point_iterator copy(*this);
-        base::operator--();
-        return copy;
     }
 };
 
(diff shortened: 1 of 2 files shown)


include/boost/geometry/iterators/point_iterator.hpp
```diff
@@ -245,33 +246,26 @@ struct points_end<MultiPolygon, multi_polygon_tag>
 // MK:: need to add doc here
 template <typename Geometry>
 class point_iterator
-    : public detail::point_iterator::iterator_type<Geometry>::type
+    : public boost::iterator_adaptor <
+        point_iterator<Geometry>
+      , typename detail::point_iterator::iterator_type<Geometry>::type
+    >
```
mkaravel: In order to follow our coding guidelines please re-write the above as follows:
```
template <typename Geometry>
class point_iterator
    : public boost::iterator_adaptor
        <
            point_iterator<Geometry>,
            typename detail::point_iterator::iterator_type<Geometry>::type
        >
{
    // ...
}

File: include/boost/geometry/iterators/point_iterator.hpp [hunk 2]
```diff
@@ -245,33 +246,26 @@ struct points_end<MultiPolygon, multi_polygon_tag>
...
 {
 private:
     typedef typename detail::point_iterator::iterator_type<Geometry>::type base;
```
mkaravel: The type `base` is not really needed now that you are using `boost::iterator_adaptor<>`. It is defined in `boost::iterator_adaptor<>` as `base_type`.

include/boost/geometry/iterators/point_iterator.hpp
```diff
@@ -245,33 +246,26 @@ struct points_end<MultiPolygon, multi_polygon_tag>
...
-    {
-        return this;
-    }
-
     template <typename OtherGeometry> friend class point_iterator;
     template <typename G> friend inline point_iterator<G> points_begin(G&);
     template <typename G> friend inline point_iterator<G> points_end(G&);
 
-    inline point_iterator(base const& base_it) : base(base_it) {}
+    inline point_iterator(base const& base_it) : point_iterator::iterator_adaptor_(base_it) {}
```
mkaravel: This line seems too long (more than 80 characters). Please re-write it as:
```
inline point_iterator(typename point_iterator::base_type const& base_it)
    : point_iterator::iterator_adaptor_(base_it)
{}
```
Notice that I also used `base_type` from `boost::iterator_adaptor<>` that I have mentioned in another line comment.

mkaravel:
Samuel thanks a lot for the fix.
Please see the line notes I made.

include/boost/geometry/iterators/point_reverse_iterator.hpp
```diff
@@ -29,16 +29,6 @@ class point_reverse_iterator
 private:
     typedef std::reverse_iterator<point_iterator<Geometry> > base;
```
mkaravel: Both `g++` and `clang++` confuse the `base` type with the `base()` method called in the constructor.
Please rename `base` to `base_type`.

include/boost/geometry/iterators/point_reverse_iterator.hpp
```diff
@@ -54,7 +44,7 @@ class point_reverse_iterator
     template <typename OtherGeometry>
```
mkaravel: In the private constructor above please rename `base` to `base_type`.
To avoid a long line, please write it as:
```
inline point_reverse_iterator(base_type const& base_it)
    : base_type(base_it)
{}
```

include/boost/geometry/iterators/point_reverse_iterator.hpp
```diff
@@ -54,7 +44,7 @@ class point_reverse_iterator
...
     inline
     point_reverse_iterator(point_reverse_iterator<OtherGeometry> const& other)
-        : base(*other.base_ptr())
+        : base(other.base())
```
mkaravel: Please replace `base` by `base_type`.

mkaravel:
Just one more comment: the `base()` method you are using in `point_reverse_iterator<>` has nothing to do with the changes in `point_iterator<>`. It is coming from `std::reverse_iterator<>`.
I should have used it in the first place.

mkaravel:
Once again, thanks for the fix.

sdebionne:
And thank you for the thoroughfull review ! I have made the changes you mention in your line comments. And you are right about `point_reverse_iterator<>`, I was confused when I looked for any other class that used the `base_ptr()` function.

mkaravel:
Thanks for taking care of this. Please run the units tests if you have not done already.
I am okay with merging this.

barendgehrels:
Thanks for the contribution/review
I'm OK too, as soon as it is confirmed that all unit tests are run successfully (especially because the branch closes today)

sdebionne:
...patience...
...patience...
...updated 449 targets...

All tests passed on Windows 7 64 bit, MSVC 2013.

I had to clone the full boost repository again, and that takes ages. Side question : does git really scale well for boost ?

mkaravel:
@sdebionne Interestingly enough, some of my unit tests fail (both in the algorithms and in the iterators directories). The reason is a missing `const` in the private `point_iterator<>` constructor. I am curious as to what is the cause of the different behavior of `g++` and `clang++` with respect to MSVC 2013 on this issue.
No need to do anything on your side: I have filed PR #243 to fix the problem.

sdebionne:
@mkaravel I have just seen your PR, sorry about that. I had corrected the missing const on my branch as well but for some obscure reason this was not pushed to Github... Interestingly the computer where I ran the tests was a fresh checkout from Github so did NOT have the const, so I would like to investigate. Here is what I did (basically following your Contributing Tutorial):

* Clone boost modular (recursive)
* Generate the headers (b2 headers)
* Add a new remote for geometry fork
* Checkout my branch `sdebionne:fix/iterators`
* run b2 in geometry\tests

AFAICT all targets build successfully with *passed*. Is there any generated report that I could check ?

mkaravel:
@sdebionne This is what I do also: run `b2` and inspect the output. I am not aware of any other reporting mechanism.

PR #243, opened 2015-03-06
[iterators][point iterator] add missing const in constructor

mkaravel:


mkaravel:
After merging PR #239 the iterator unit test for `point_iterator<>` started to fail.
The reason is a missing `const` in the point iterator constructor.
This PR fixes this problem and iterator unit tests pass again.

PR #244, opened 2015-03-06
Fix: make partition work for forward ranges

mkaravel:
The current `partition<>` implementation, the one or two input collections passed are assumed internally to be random access containers (to be precise, it is required that the collection's `operator[]` is defined).
In fact, this is requirement is a bit too strict. `partition<>` can easily work even for forward ranges, and this is the focus of this PR.

In this PR, `partition<>` has been modified so as to work for forward ranges. This eliminates among other things the need for the range to define `operator[]`. The idea behind the re-implementation is as follows: instead of storing and keeping track of indices of collection items (these are the indices that are used as arguments in `operator[]` calls), iterators to collection items are stored and manipulated. The `index_vector_type` is consistently replaced by a vector of iterators (possibly of two different types if the two input collections are different). Accessing the collection items in this context simply amounts to de-referencing the iterators stored in the iterator vectors. The only operators used with respect to the iterators of the input collections are `operator++` and `operator*`.

On the practical side: this allows to use partition with ranges which are not random access. Two such cases are ranges coming from BG's `point_iterator<>` and `segment_iterator<>` which are bidirectional iterators.

All algorithm unit tests pass with the re-designed `partition<>`.

### DIFF
--- a/include/boost/geometry/algorithms/detail/partition.hpp
+++ b/include/boost/geometry/algorithms/detail/partition.hpp
@@ -1,6 +1,11 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
-// Copyright (c) 2011-2014 Barend Gehrels, Amsterdam, the Netherlands.
+// Copyright (c) 2011-2015 Barend Gehrels, Amsterdam, the Netherlands.
+
+// This file was modified by Oracle on 2015.
+// Modifications copyright (c) 2015 Oracle and/or its affiliates.
+
+// Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
 
 // Use, modification and distribution is subject to the Boost Software License,
 // Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
@@ -9,18 +14,19 @@
 #ifndef BOOST_GEOMETRY_ALGORITHMS_DETAIL_PARTITION_HPP
 #define BOOST_GEOMETRY_ALGORITHMS_DETAIL_PARTITION_HPP
 
+#include <cstddef>
 #include <vector>
-#include <boost/range/algorithm/copy.hpp>
-#include <boost/geometry/algorithms/assign.hpp>
+#include <boost/range.hpp>
+#include <boost/geometry/core/access.hpp>
 #include <boost/geometry/core/coordinate_type.hpp>
+#include <boost/geometry/algorithms/assign.hpp>
+
 
 namespace boost { namespace geometry
 {
 
 namespace detail { namespace partition
 {
-
-typedef std::vector<std::size_t> index_vector_type;
 
 template <int Dimension, typename Box>
 inline void divide_box(Box const& box, Box& lower_box, Box& upper_box)
@@ -38,31 +44,33 @@ inline void divide_box(Box const& box, Box& lower_box, Box& upper_box)
     geometry::set<min_corner, Dimension>(upper_box, mid);
 }
 
-// Divide collection into three subsets: lower, upper and oversized
+// Divide forward_range into three subsets: lower, upper and oversized
 // (not-fitting)
 // (lower == left or bottom, upper == right or top)
-template <typename OverlapsPolicy, typename InputCollection, typename Box>
+template
+<
+    typename OverlapsPolicy,
+    typename ForwardRange,
+    typename Box,
+    typename IteratorVector
+>
 inline void divide_into_subsets(Box const& lower_box,
         Box const& upper_box,
-        InputCollection const& collection,
-        index_vector_type const& input,
-        index_vector_type& lower,
-        index_vector_type& upper,
-        index_vector_type& exceeding)
+        ForwardRange const& /*forward_range*/,
+        IteratorVector const& input,
+        IteratorVector& lower,
+        IteratorVector& upper,
+        IteratorVector& exceeding)
 {
-    typedef boost::range_iterator
+    typedef typename boost::range_iterator
         <
-            index_vector_type const
-        >::type index_iterator_type;
+            IteratorVector const
+        >::type it_type;
 
-    for(index_iterator_type it = boost::begin(input);
-        it != boost::end(input);
-        ++it)
+    for(it_type it = boost::begin(input); it != boost::end(input); ++it)
     {
-        bool const lower_overlapping = OverlapsPolicy::apply(lower_box,
-                    collection[*it]);
-        bool const upper_overlapping = OverlapsPolicy::apply(upper_box,
-                    collection[*it]);
+        bool const lower_overlapping = OverlapsPolicy::apply(lower_box, **it);
+        bool const upper_overlapping = OverlapsPolicy::apply(upper_box, **it);
 
         if (lower_overlapping && upper_overlapping)
         {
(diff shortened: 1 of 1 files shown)


include/boost/geometry/algorithms/detail/partition.hpp
```diff
@@ -84,23 +92,34 @@ inline void divide_into_subsets(Box const& lower_box,
...
+<
+    typename ExpandPolicy,
+    typename Box,
+    typename ForwardRange,
+    typename IteratorVector
+>
 inline void expand_with_elements(Box& total,
-                InputCollection const& collection,
-                index_vector_type const& input)
+                ForwardRange const& /*forward_range*/,
```
barendgehrels: So we can remove this argument

include/boost/geometry/algorithms/detail/partition.hpp
```diff
@@ -84,23 +92,34 @@ inline void divide_into_subsets(Box const& lower_box,
...
-inline void handle_one(InputCollection const& collection,
-        index_vector_type const& input,
+// Match forward_range with itself
+template
+<
+    typename ForwardRange,
+    typename Policy,
+    typename IteratorVector
+>
+inline void handle_one(ForwardRange const& /*forward_range*/,
```
barendgehrels: Idem

include/boost/geometry/algorithms/detail/partition.hpp
```diff
@@ -589,53 +653,57 @@ public :
...
+            <
+                ForwardRange2 const
+            >::type iterator_type2;
+
+        if (std::size_t(boost::size(forward_range1)) > min_elements
+            && std::size_t(boost::size(forward_range2)) > min_elements)
         {
-            index_vector_type index_vector1, index_vector2;
+            std::vector<iterator_type1> iterator_vector1;
+            std::vector<iterator_type2> iterator_vector2;
```
barendgehrels: A vector of iterators. What is the size compared to integers? Can you do a performance comparison test too with old/new?

barendgehrels:
I wonder why you call this "Fix". It is obviously not a fix, but an improvement, making it more generi

barendgehrels:
Please postpone this until after 1.58

barendgehrels:
I also wonder why you call this "reimplement" and "redesign". Anyway, does not matter, just wondering - it is just changing a part of the implementation

include/boost/geometry/algorithms/detail/partition.hpp
mkaravel: Yes, we can. Will do it.

include/boost/geometry/algorithms/detail/partition.hpp
mkaravel: Same here. Will do it.

include/boost/geometry/algorithms/detail/partition.hpp
mkaravel: I will try to. More to come.

mkaravel:
* About "Fix": partly lack of creativity in the naming, and partly that it "fixes" the "shortcoming" of working for random access ranges only.
* About 1.58: no problem: my usage case is difference and intersection of pointlike/linear geometries, so it is not really needed for 1.58.
* About "reimplement" and "redesign": partly lack of creativity and imagination for naming what I did, and partly the fact that I use iterators instead of indices (which in my mind is a design choice). Honestly, I never intended to mean that this is my/new design or my/new implementation; it seemed more natural to call what I did a "redesign" or a "reimplementation" rather than a "change". But this is just my own opinion, and I do not have strong feelings about it. What is important for me here is the added functionality.

include/boost/geometry/algorithms/detail/partition.hpp
mkaravel: I remember now why I decided to keep it: in order to allow for the possibility to pass it, if we want at a later stage, to any of the policies visitors and used in the algorithm.
If you think this is too much I can remove it.

barendgehrels: OK, I see. For Expand I don't think it will be useful, so IMO you can remove it - we can always readd it if really needed

mkaravel: Okay. Will remove it.

barendgehrels:
OK, thanks for the explanations. Good to hear you want to use it for point like differences.

mkaravel:
It is for pointlike-linear differences. Pointlike-pointlike are already implemented in a different way (sorting and binary search; I did benchmarks then, and it was faster that way compared to using partition).

barendgehrels:
Sure, pointlike and linear. We still have to make partition so fast that it can compare to std::sort ....

include/boost/geometry/algorithms/detail/partition.hpp
mkaravel: Done. See commit ee1e0e0.

include/boost/geometry/algorithms/detail/partition.hpp
mkaravel: Regarding the size: it depends on the size of the iterator of course.

My understanding is that for the existing use cases we should not see a difference in the size of the vector: what was passed was a random access container/range and an integer index; now what I pass is an iterator to the random access container/range, and I expect the size of this iterator to be the same as the size of the index.

It is true that if you pass a possibly more complicated iterator (like `point_iterator<>` or `segment_iterator<>`) then the size of the vector may be larger; on the other hand such usages where not possible before, and in that respect I do not see an issue with respect to size.

I will investigate this in more detail as part of the performance comparison.

mkaravel: One more update on this: on my computer the size of `std::size_t` and the size of iterators of, for example, `bg::multi_point<>` and `bg::multi_linestring<>`, are both 8 bytes.
So my understanding is that for all existing usages, there is no overhead in using a vector of iterators instead of a vector of `std::size_t`'s.

mkaravel:
## **_Performance comparison_**

### **Setup**
I tried the following scenario: I considered random points in a `[0,1] x [0,1]` square, seen as multipoints. For these multipoints I called partition. The visitor passed is not doing something very meaningful, but still I wanted it to do something to avoid any possible optimizations by the compiler. What it does is that for the points in each bottom-most level box, it computes the minimum distance among them. I considered running the old and new implementation of partition either for a single multipoint (one-collection version of partition) or for two multipoints (two-collections version of partition).

See the following files (the partition implementations were given different names and put in different namespace so that I can use both of them in the same program):
* Comparison/timing program: https://github.com/mkaravel/geometry/blob/performance-partition/performance/algorithms/partition/multipoint_multipoint.cpp
* Old partition implementation: https://github.com/mkaravel/geometry/blob/performance-partition/performance/algorithms/partition/partition_old.hpp
* New partition implementation: https://github.com/mkaravel/geometry/blob/performance-partition/performance/algorithms/partition/partition_new.hpp

### **Compilation**
The comparison program was compiled with `clang++` version 3.4 with the flags `-O2` and `-DNDEBUG`.

### **Timings**
The following two tables show the elapsed times for the old and new partition implementation with respect to the size of the multipoint(s) passed.
Times are in seconds. `Boost.Timer` has been used for performing the timings. Smallest time is shown in bold-face.

| *Multipoint size* | *Old* | *New* |
|:-------------:|--------:|--------:|
| 10^3 | 0.000618 | **0.000585** |
| 10^4 | 0.002773 | **0.002450** |
| 10^5 | **0.035126** | 0.036622 |
| 10^6 | 0.431059 | **0.394527** |
| 10^7 | 5.861435 | **5.680503** |


| *Multipoint  # 1 size* | *Multipoint  # 2 size* | *Old* | *New* |
|:-------------:|:-------------:| --------:|--------:|
| 10^3 | 10^3 | 0.001106 | **0.000997** |
| 10^4 | 10^4 | 0.005468 | **0.005185** |
| 10^5 | 10^5 | 0.076411 | **0.065967** |
| 10^6 | 10^6 | 0.849864 | **0.813731** |
| 10^7 | 10^7 | 12.670377 | **12.326983** |

mkaravel:
My understanding is that there is no show-stopper for this PR.

Since we have decided to leave it for boost 1.59, please just comment that the PR is indeed okay, and we can merge once our develop branch is open for 1.59

barendgehrels:
The timings look convincing, thanks for this. We already earlier noticed that changing from index operator to iterators is measurably faster.
So yes, we can merge it for 1.59

PR #245, opened 2015-03-06
Remove misleading overloads of front() and back() in range and fix a missing const in sub_range

sdebionne:
Remove misleading overloads of `front()` and `back()` since the constness of a Range does not mean that the underlying iterator (and so value type) is const (e.g. you can have a const Range of mutable iterators and vice versa). Use `boost::range_reference<>`.

In the same vein, I wonder if `range::resize()`, `range::clear()` and `range::push_back()` should be moved in a new file "util/container.hpp" to make it more obvious that these fonctions require Containers ?

### DIFF
--- a/include/boost/geometry/algorithms/detail/sub_range.hpp
+++ b/include/boost/geometry/algorithms/detail/sub_range.hpp
@@ -14,7 +14,12 @@
 #ifndef BOOST_GEOMETRY_ALGORITHMS_DETAIL_SUB_RANGE_HPP
 #define BOOST_GEOMETRY_ALGORITHMS_DETAIL_SUB_RANGE_HPP
 
+#include <boost/geometry/algorithms/not_implemented.hpp>
+#include <boost/geometry/core/tags.hpp>
+#include <boost/geometry/core/ring_type.hpp>
 #include <boost/geometry/util/range.hpp>
+
+#include <boost/range/size.hpp>
 
 namespace boost { namespace geometry {
 
@@ -44,7 +49,7 @@ struct sub_range<Geometry, Tag, false>
 template <typename Geometry>
 struct sub_range<Geometry, polygon_tag, false>
 {
-    typedef typename geometry::ring_type<Geometry>::type & return_type;
+    typedef typename geometry::ring_type<Geometry>::type const& return_type;
 
     template <typename Id> static inline
     return_type apply(Geometry & geometry, Id const& id)
--- a/include/boost/geometry/util/range.hpp
+++ b/include/boost/geometry/util/range.hpp
@@ -107,20 +107,8 @@ at(RandomAccessRange & rng,
 \ingroup utility
 */
 template <typename Range>
-inline typename boost::range_value<Range>::type const&
+inline typename boost::range_reference<Range>::type
 front(Range const& rng)
-{
-    BOOST_ASSERT(!boost::empty(rng));
-    return *boost::begin(rng);
-}
-
-/*!
-\brief Short utility to conveniently return the front element of a Range.
-\ingroup utility
-*/
-template <typename Range>
-inline typename boost::range_value<Range>::type &
-front(Range & rng)
 {
     BOOST_ASSERT(!boost::empty(rng));
     return *boost::begin(rng);
@@ -133,27 +121,13 @@ front(Range & rng)
 \ingroup utility
 */
 template <typename BidirectionalRange>
-inline typename boost::range_value<BidirectionalRange>::type const&
+inline typename boost::range_reference<BidirectionalRange>::type
 back(BidirectionalRange const& rng)
 {
     BOOST_RANGE_CONCEPT_ASSERT(( boost::BidirectionalRangeConcept<BidirectionalRange const> ));
     BOOST_ASSERT(!boost::empty(rng));
     return *(boost::rbegin(rng));
 }
-
-/*!
-\brief Short utility to conveniently return the back element of a BidirectionalRange.
-\ingroup utility
-*/
-template <typename BidirectionalRange>
-inline typename boost::range_value<BidirectionalRange>::type &
-back(BidirectionalRange & rng)
-{
-    BOOST_RANGE_CONCEPT_ASSERT(( boost::BidirectionalRangeConcept<BidirectionalRange> ));
-    BOOST_ASSERT(!boost::empty(rng));
-    return *(boost::rbegin(rng));
-}
-
 
 /*!
 \brief Short utility to conveniently clear a mutable range.


sdebionne:
Now that I think of it, removing the overloads may break codes that use Containers as Ranges...

A possible solution to the problem of constness (Containers have deep constness, so iterators of const containers are const_iterators while Ranges behave more like pointers) could be to internally convert user inputs to Ranges so that anywhere in Boost.Geometry code Ranges could assumed.

There are other places where the issue may arise, see for instance: 
```
template <typename Polygon>
struct ring_return_type<polygon_tag, Polygon>
{
    typedef typename boost::remove_const<Polygon>::type nc_polygon_type;

    typedef typename mpl::if_
        <
            boost::is_const<Polygon>,
            typename traits::ring_const_type<nc_polygon_type>::type,
            typename traits::ring_mutable_type<nc_polygon_type>::type
        >::type type;
};
```
Here, the constness of the ring type is based on the constness of the Polygon template parameter (and so is assuming a Container). But a Polygon could be implemented as a Range of Rings (with two special accessors `outer()` and `inners()`).

Hoping it's not too confusing...

awulkiew:
Thanks for the PR!
Indeed range_reference<> should be used.

In Boost.Geometry the algorithms take const Geometries as input parameters and non-const as output. My understanding is that they are Geometries concepts which behave like Ranges, not the other way around. And the information about constness is propagated into the internals of algorithms. I'm not sure if messing with this machinery is a good idea. But if this is a problem we certainly should change or tweak it.

Let's consider 2 possibly problematic scenarios.

 - A const Range is passed into an algorithm as mutable Geometry. Such situation is only possible if this Range is an output Geometry. Output geometries are changed by the algorithm, that is an algorithms clears them resizes and adds points to them. So they must be Containers, not Ranges.

- A mutable Range is passed as input Geometry. The algorithm takes const reference so this range will be passed as const and this constness will be propagated into the lower levels.

Is the purpose of this PR to solve some real problematic issue?
Btw, you should probably add the tests for it as well.

include/boost/geometry/algorithms/detail/sub_range.hpp
```diff
@@ -44,7 +49,7 @@ struct sub_range<Geometry, Tag, false>
 template <typename Geometry>
 struct sub_range<Geometry, polygon_tag, false>
 {
-    typedef typename geometry::ring_type<Geometry>::type & return_type;
+    typedef typename geometry::ring_type<Geometry>::type const& return_type;
```
mkaravel: I am confused about this change. Why is the `const` required here?

include/boost/geometry/util/range.hpp
```diff
@@ -107,25 +107,13 @@ at(RandomAccessRange & rng,
...
-\ingroup utility
-*/
-template <typename Range>
-inline typename boost::range_value<Range>::type &
-front(Range & rng)
-{
-    BOOST_ASSERT(!boost::empty(rng));
-    return *boost::begin(rng);
-}
-
```
awulkiew: `front()` taking non-const reference shouldn't be removed. With your change it isn't possible to return a non-const reference if a non-const Container is passed because it'd be passed as const into the function. Consider:

    std::vector<int> v(10, 0);
    int & i = range::front(v);

include/boost/geometry/util/range.hpp
```diff
@@ -142,20 +130,6 @@ back(BidirectionalRange const& rng)
...
-\ingroup utility
-*/
-template <typename BidirectionalRange>
-inline typename boost::range_value<BidirectionalRange>::type &
-back(BidirectionalRange & rng)
-{
-    BOOST_RANGE_CONCEPT_ASSERT(( boost::BidirectionalRangeConcept<BidirectionalRange> ));
-    BOOST_ASSERT(!boost::empty(rng));
-    return *(boost::rbegin(rng));
-}
```
awulkiew: Same here, this should stay.

include/boost/geometry/util/range.hpp
```diff
@@ -107,25 +107,13 @@ at(RandomAccessRange & rng,
 \ingroup utility
 */
 template <typename Range>
-inline typename boost::range_value<Range>::type const&
+inline typename boost::range_reference<Range>::type
```
awulkiew: To work properly also with Containers the `Range const` should be passed into the `range_reference<>`.

include/boost/geometry/util/range.hpp
```diff
@@ -133,7 +121,7 @@ front(Range & rng)
 \ingroup utility
 */
 template <typename BidirectionalRange>
-inline typename boost::range_value<BidirectionalRange>::type const&
+inline typename boost::range_reference<BidirectionalRange>::type
```
awulkiew: Same here `BidirectionalRange const`.

include/boost/geometry/algorithms/detail/sub_range.hpp
awulkiew: Why is this a problem? Isn't `ring_type<>` returning a type of rings suitable for the passed Geometry? So const ring for const Geometry and non-const otherwise?

include/boost/geometry/util/range.hpp
sdebionne: Yes that what I mentionned in my second comment. But consider:
```
boost::iterator_range<const int*> r(10, 0);
int & i = bg::range::front(r);
```
That won't compile. I believe this is the reason why Boost.Range does not have this function.
`boost::iterator_range<const point*>` should be valid Linestring, right ?

sdebionne:
You are right, the use of `range_reference<>` fix the problem. For ranges the two overloads are equivalent, but the const version is indeed necessary for Container. Am I right to say that `range::at()` should be fixed as well ?
Is it OK to update the branch or should I make an other PR ?

include/boost/geometry/algorithms/detail/sub_range.hpp
sdebionne: Well in my code `ring_type<>::type` is mutable while range::at() returns a const. My ring type is `iterator_range<const point *>`. Let me investigate futher...

awulkiew:
Yes, `at()` should be fixed as well. I'd create a new PR to not push commits reverting the changes done in previous commits. You should probably also add some tests. I recently added tests for `boost::iterator_range` and `std::pair` so cases for const pointers could be added there (in test/util/range.cpp).

Does it fix all of the problems noticed by you?
Or are there some other problematic places?

sdebionne:
The problem with `sub_range` is still there, but I'm not so sure that it's actullay `sub_range` the problem but may be `ring_type<>` and the way it dispatchs to `ring_const_type<>` and `ring_mutable_type<>`...

I'm going to submit a PR for `range::at()`, `range::front()` and `range::back()`.

And I'll add tests for const pointers as its a pattern I use a lot to adapt (read-only) BLOB geometries to Boost.Geometry.

awulkiew:
Great, thanks!

As for the `sub_range<>`, indeed the problem may be located in `ring_type<>` or some other metafunction. It'd be helpful to create some minimal failing example, though I'm guessing that it could require more code. Maybe it'd be a good idea to move this discussion to the mailing list?

sdebionne:
My plan is to submits several smaller PRs, I'm closing this one.

PR #248, opened 2015-03-08
Simplify detail::not_

mkaravel:
Modify the implementation of `detail::not_` by moving the geometry template parameters inside the class.
This way it is simpler to use `detail::not_` as it is not required to specify the geometries.


barendgehrels:
OK for me

PR #249, opened 2015-03-08
Fix: rename local variable

mkaravel:
Rename local variable so that it is more descriptive.


barendgehrels:
I did already rename locally it but did not announce it. But OK, thanks for this, will merge it.

PR #250, opened 2015-03-08
[strategies][cartesian][buffer][join_round] optimize expression

mkaravel:
calculating number of interior points


barendgehrels:
Looks good - I assume it is unit tested? If so I agree with merging

barendgehrels:
Oh now I relook at it, the std::max should be protected as boost recommends - can you have that too? Otherwise I do it later this week. I'm off now

mkaravel:
`std::max` is now protected (see commit b3b95ae); sorry for that, it slipped.

Yes, unit tests were run, but besides unit tests, you can verify that the two expressions are mathematically equivalent.

barendgehrels:
That's fast! Thanks!

PR #252, opened 2015-03-09
[range] Use range_reference

sdebionne:
This PR fixes `range::at()`, `range::front()` and `range::back()` with Ranges (such as `iterator_range<>`) that does not handle constness like containers.
This PR is a followup of [PR#245](https://github.com/boostorg/geometry/pull/245).

### DIFF
--- a/include/boost/geometry/util/range.hpp
+++ b/include/boost/geometry/util/range.hpp
@@ -81,7 +81,7 @@ pos(RandomAccessRange & rng,
 \ingroup utility
 */
 template <typename RandomAccessRange>
-inline typename boost::range_value<RandomAccessRange const>::type const&
+inline typename boost::range_reference<RandomAccessRange const>::type
 at(RandomAccessRange const& rng,
    typename boost::range_size<RandomAccessRange const>::type i)
 {
@@ -94,7 +94,7 @@ at(RandomAccessRange const& rng,
 \ingroup utility
 */
 template <typename RandomAccessRange>
-inline typename boost::range_value<RandomAccessRange>::type &
+inline typename boost::range_reference<RandomAccessRange>::type
 at(RandomAccessRange & rng,
    typename boost::range_size<RandomAccessRange>::type i)
 {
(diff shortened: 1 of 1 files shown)


include/boost/geometry/util/range.hpp
```diff
@@ -116,10 +116,10 @@ front(Range const& rng)
 
 /*!
 \brief Short utility to conveniently return the front element of a Range.
-\ingroup utility
-*/
+ \ingroup utility
+ */
```
mkaravel: Please remove the leading spaces in the two lines above.

include/boost/geometry/util/range.hpp
```diff
@@ -142,19 +142,18 @@ back(BidirectionalRange const& rng)
 }
 
 /*!
-\brief Short utility to conveniently return the back element of a BidirectionalRange.
-\ingroup utility
-*/
+-\brief Short utility to conveniently return the back element of a BidirectionalRange.
+-\ingroup utility
+-*/
```
mkaravel: Please remove the leading dashes in the three line above.

include/boost/geometry/util/range.hpp
sdebionne: OK, I have one git question: shall I amend the commit (like I just did to adjust the comment) or make a new commit with the removed spaces ?

mkaravel: I still see only one commit in this PR.

As far as I understand `git commit --amend` works for the tip of the tree. So, since you have a single commit you can certainly use it in your local branch to fix things in the commit message.
To actually redo the commit, you probably need to do `git reset <some hash>` or `git reset <some reflog>`, do the modifications and redo the commit.
All these are okay with your local branch, but I have no experience as to how they play along with pull requests.

sdebionne: I have done a `push --force` so that the PR remains a single simple commit (not three commits with space removed). It's bad practice, but I don't think anyone has pulled my branch in the latest 5 minutes... so it should be armless.

sdebionne:
@mkaravel I have removed all the spaces that were added by mistake. Sorry about that.

awulkiew:
Thanks for this!

PR #255, opened 2015-03-10
Feature/disjoint: implement disjoint for multipoint/[multi]linestring

mkaravel:
In this PR I implement `bg::disjoint()` for the following two (still missing) geometry combinations:
* multipoint/linestring
* multipoint/multilinestring

The implemented algorithm uses `bg::partition()`: `bg::partition()` is called on two collections: the points of the multipoint and the segments of the linear geometry. To extract the segments of the linear geometry the `bg::segment_iterator<>` is used.

Unit tests have been added as part of this PR, and the documentation for `bg::disjoint()` is also updated (including release notes).

This PR is targeted for boost 1.59 and depends on PR #244 (the version of `bg::partition()` in PR #244 can handle a range of segments created via the `bg::segment_iterator<>`, whereas this is not possible in the version of `bg::partition()` in the current version of BG's develop branch).

As part of this PR, I have also slightly re-arranged and simplified the implementation of `bg::disjoint()` for the geometry combinations point/segment, point/linear.

This PR can be merged independently of PR #244; the unit tests however will be failing until PR #244 is also merged.


include/boost/geometry/algorithms/detail/disjoint/point_geometry.hpp
```diff
@@ -39,20 +39,25 @@ namespace detail { namespace disjoint
...
+    static inline bool apply(Point const& point, Segment const& segment)
     {
-        return !geometry::covered_by(pt, g);
+        typedef geometry::model::referring_segment<Point const> other_segment;
+
+        other_segment other(point, point);
+        return disjoint_segment
+            <
+                Segment, other_segment
+            >::apply(segment, other);
```
awulkiew: What was wrong with the previous version?
Wasn't it the same as `reverse_covered_by` used below for Pt/Linear?
Or is `covered_by()` not defined for Pt/Seg?

Isn't Pt/Seg algorithm faster than Seg/Seg. AFAIU in Seg/Seg points must be checked for equality and then Pt/Seg version called. So there is additional condition and the algorithm contains unnecessary code (Seg/Seg even if only Pt/Seg is needed) which can potentialy harm the instructions cache. Or am I missing something?

barendgehrels: I think this is a new combination - the diff seems a bit off here
!geometry::covered_by is still used

mkaravel: The point-segment disjointness test was simply moved from one place of the file to another. The actual implementation was not changed.

What I eliminated (but it is not visible easily in the diff) is `disjoint_point_linear`. Its implementation was the same as `reverse_covered_by`, so I simply use `reverse_covered_by` in the new version.

The changes in this file are all about code refactoring and polishing. There is nothing new.

awulkiew: What I had in mind is, that `reverse_covered_by` could also be used for Pt/Seg. AFAIU there is no need to have `disjoint_point_segment` becasue assuming that `covered_by()` works for Pt/Seg, `reverse_covered_by` should work out of the box. Or am I missing something?

mkaravel: It seems that `covered_by` is indeed implemented for the point-segment combination.
I will double-check and modify the code as per your suggestion if `covered_by` does indeed work.

mkaravel: I have replace `disjoint_point_segment` by `reverse_covered_by` as suggested.
see commit 7bd759d.

awulkiew:
Thanks!

PR #256, opened 2015-03-10
Unit tests for set operations of pointlike/pointlike geometries

mkaravel:
In this PR:
* polish tests of set operations for pointlike/pointlike geometries (make the case ID the first argument of the tester)
* modify the common set operations test code so that it can be applied to set operations of pointlike/linear geometries as well


mkaravel:
This PR may wait for boost 1.59.

PR #261, opened 2015-03-11
Add support for ForwardRange to Range

sdebionne:
Add support for ForwardRange to `pos()`, `at()`...
Operations on RandomAccessRange remain O(1) while other Range operations are O(N).
The ForwardRange support is tested with `sd::list<>`

In my problematic use case, the range of interior rings of an adapted Polygon is a ForwardRange. I'm not sure what to expect in term performance penalty (the Ring itself is RandomAccessRange of Point), any clue ?

All tests passed on Windows.

### DIFF
--- a/include/boost/geometry/util/range.hpp
+++ b/include/boost/geometry/util/range.hpp
@@ -32,74 +32,76 @@ namespace boost { namespace geometry { namespace range {
 
 namespace detail {
 
-// NOTE: For SinglePassRanges pos could iterate over all elements until the i-th element was met.
 
-template <typename RandomAccessRange>
+template <typename ForwardRange>
 struct pos
 {
-    typedef typename boost::range_iterator<RandomAccessRange>::type iterator;
-    typedef typename boost::range_size<RandomAccessRange>::type size_type;
-    typedef typename boost::range_difference<RandomAccessRange>::type difference_type;
+    typedef typename boost::range_iterator<ForwardRange>::type iterator;
+    typedef typename boost::range_size<ForwardRange>::type size_type;
+    typedef typename boost::range_difference<ForwardRange>::type difference_type;
 
-    static inline iterator apply(RandomAccessRange & rng, size_type i)
+    static inline iterator apply(ForwardRange & rng, size_type i)
     {
-        BOOST_RANGE_CONCEPT_ASSERT(( boost::RandomAccessRangeConcept<RandomAccessRange> ));
-        return boost::begin(rng) + static_cast<difference_type>(i);
+        BOOST_RANGE_CONCEPT_ASSERT((boost::ForwardRangeConcept<ForwardRange>));
+        iterator it = boost::begin(rng);
+        std::advance(it, static_cast<difference_type>(i));
+        return it;
     }
 };
+
 
 } // namespace detail
 
 /*!
-\brief Short utility to conveniently return an iterator of a RandomAccessRange.
+\brief Short utility to conveniently return an iterator of a ForwardRange.
 \ingroup utility
 */
-template <typename RandomAccessRange>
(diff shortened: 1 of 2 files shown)


awulkiew:
The PR looks good, thanks for proposal!

Though I'm not sure if we should allow using ForwardRanges exactly because of the performance degradation. I know that it was done like this in the STL (`distance`, `advance`) but a function silently changing the complexity for a different input is probably not a good idea. Esspecially if it's called `at()` which is a `vector`'s method. This brings some user's expectations. For such functions with variading complexity, instead of `at()` we should probably use something like `at_advanced()` name. AFAIR `boost::size()` also expects RandomAccessRanges. Probably because of the confusion It could cause, for this the user should use `distance`.

On the other hand in the case of `erase()` instead of O(1) for a list as you'd probably expect the complexity would be O(N) as for `vector`, though maybe it could be "optimized" (with SFINAE or some ADL trick).

Answering your question about the performance, it probably depends on the algorithm and the input data. I'm guessing that typically it shouldn't be a problem since Polygons hasn't many holes in general, probably. You should do some benchmarking to see how does it look like in your case.

Maybe it'd be possible to tweak some algorithms and instead of storing indexes or along with indexes store iterator(s). This would be interesting.

But the most important question is, couldn't you just use `deque` or `vector` instead? It seems surprising that you need `list` here. It'd in most cases be faster than `list`. Or maybe `boost::stable_vector` if the reference stability is what you need? Though I'm not sure how fast it is.

barendgehrels:
I'm basically OK with this if Adam's questions are answered.

awulkiew:
I'm not really ok with it :)

What I had in mind is that technically the PR is good, but still probably shouldn't be merged.
`at()`, just like `size()` should expect RandomAccessRange.

The more general question is, should we support ForwardRanges of SingleGeometries in Multi, Rings in Polygon and/or Points in Range? If the answer was yes, then we should probably support it "properly" in algorithms, i.e. use/store iterators instead of indexes, don't use `at()` or `size()`, etc. But I do realize that it's probably more work.

barendgehrels:
I agree with your objections, and, I think, the answer is (for the foreseeable future): no, we should not support Forward ranges in algorithms - we use those indexes everywhere. It is a lot of work and for the moment we have other objectives.
However, I did not oppose to this PR which looks good. But I understand your reasoning, this will give the impression that it is supported.

sdebionne:
@awulkiew I get your point, that the reason why I mentioned the complexity in my initial post. In the other hand, concept checking should trap ForwardRanges in places where it is really not acceptable (e.g. Points in Range).

My only use case for ForwardRange is for interior rings of polygons. A better solution would be to relax this part of the polygon concept:

> The Polygon Concept is defined as following:
> - there must be a specialization of traits::interior_type defining the type of the collection of its interior rings as type; **this collection itself must fulfill a Boost.Range Random Access Range Concept**

but that would require changes (e.g. use of iterator and `advance()`, `next()`, `prior()`) in every algorithms that manipulates the interior rings and I have not idea what kind of rework that imply.

awulkiew:
It's good you reminded that it's explicitly said in the concept definition.

The library musn't be changed in order to support your specific (non-concept-conforming) rings container. You could implement a RandomAccess Adaptor wrapping `std::list` or any other Container with ForwardAccess. This adaptor then could be used anywhere, in particular to store interior rings of a Polygon. The effect would be the same. And as it is now, the compiler would report an error if invalid (non-RandomAccess) container was used which could harm the performance.

But I'd still consider replacing `std::list` with some other container. Would it be possible to share your use case, most preferably on a list? Maybe we could think about some better workaround?

sdebionne:
>The library musn't be changed in order to support your specific (non-concept-conforming) rings container. 

Agreed.

PR #262, opened 2015-03-11
Fix: remove obsolete includes referring to the "multi"directory

mkaravel:


barendgehrels:
Can't we remove the files themselves too?

awulkiew:
They were left for backward compatibility.
Do you think we should remove them?
All of them or e.g. only details?

barendgehrels:
I know but nobody (or hardly anybody) includes them directly, and they are already empty for several releases IIRC.
But this is indeed a "major change" possibly having backward compatability issues, so let's do it after the 1.58 release and put it in the notes there.

mkaravel:
I am doing the `#include` removals as I see them when working on code.
I think we should do a pass and remove all of them in our code.
This is independent of removing the files, which I agree should be done after 1.58 and not under time pressure.
I volunteer to remove the obsolete includes all over the library.

barendgehrels:
@mkaravel that is OK, thanks for volunteering

PR #263, opened 2015-03-11
Fix: allow geometry test common to be included in all unit tests

mkaravel:
In this PR I guard the inclusion of Unit.Test related include by the macro `BOOST_TEST_MODULE`.
This macro is defined in many unit tests (distance, set-ops, validity, simplicity and possibly others), and is used for using Boost.Test is a different way than in traditionally written unit tests.

When `<geometry_test_common.hpp>` is included in those tests, the `test_main()` function is defined twice and gives a compilation error.

With the guarding proposed in this PR, `<geometry_test_common.hpp>` can be safely included in those unit tests as well, without any problems.

### DIFF
--- a/test/geometry_test_common.hpp
+++ b/test/geometry_test_common.hpp
@@ -72,9 +72,11 @@
 #endif
 
 # include <boost/test/floating_point_comparison.hpp>
+#ifndef BOOST_TEST_MODULE
 # include <boost/test/included/test_exec_monitor.hpp>
 //#  include <boost/test/included/prg_exec_monitor.hpp>
 # include <boost/test/impl/execution_monitor.ipp>
+#endif
 
 #ifdef __clang__
 # pragma clang diagnostic pop


barendgehrels:
I'm OK

sdebionne:
IMHO, it would have been better to keep the inclusion of Unit.Test headers centralized in `<geometry_test_common.hpp>`.

What about:
```cpp
#ifdef BOOST_TEST_MODULE
# include <boost/test/included/unit_test.hpp>
#else
# include <boost/test/included/test_exec_monitor.hpp>
//#  include <boost/test/included/prg_exec_monitor.hpp>
# include <boost/test/impl/execution_monitor.ipp>
#endif
``` 
BOOST_TEST_MODULE would mean "use the UTF", the default being "use the test execution monitor". If sometime in the future you want to switch to the linked version of UTF, it will be much easier (but I don't add fuel to the fire)...

mkaravel:
The contents of `geometry_test_common.hpp` are quite neutral and generic so, I am not against this idea.

On the other hand this enforces the inclusion of `geometry_test_common.hpp` in all unit tests, and I am not sure this is something we want to enforce.

Let's wait for other opinions about this. If we agree that this is the way we want to go, I can modify the tests that need to be changed, and include this header file.

sdebionne:
The big picture:
- there are few test suites that are not build with Boost.Test at all (point_concept, robustness);
- some tests (distance, iterators, strategies, math_sqrt) use UTF and include `unit_test.hpp` directly;
- very few test suites (util_variant, io_wkt) use PEM and include `prg_exec_monitor.hpp` directly;
- all the rest use PEM and do include `geometry_test_common.hpp`.

PR #264, opened 2015-03-12
Fix: remove deprecated include paths

mkaravel:
This PR is about removing and/or fixing include paths that refer to the deprecated `include/boost/geometry/multi/` directory.
In some cases the includes are not fixed blindly, but also polished a bit.

The removal is the deprecated paths is not finished with this PR. There is more to be done, but I thought that splitting it into smaller chunks will cause fewer problems to other developers.

The is also a practical issue: to avoid potential merge conflicts with PR #256, I would like to have PR #256 merged also. In that PR I indicated that it can wait for boost 1.59, but it can also be merged (there are no new features/new functionality in there, just re-arranging and polishing code).


awulkiew:
I propose to postpone it to the time when everybody has more free time and there is no pressure (after the release).

mkaravel:
This PR is ready for merging. See also PR #265 and the PR's description in there.

test/algorithms/multi_length.cpp
```diff
@@ -8,15 +8,8 @@
 
 #include <algorithms/test_length.hpp>
 
-#include <boost/geometry/multi/multi.hpp>
-
 #include <boost/geometry/geometries/geometries.hpp>
 #include <boost/geometry/geometries/point_xy.hpp>
-#include <boost/geometry/multi/geometries/multi_linestring.hpp>
-
-#include <boost/geometry/io/wkt/wkt.hpp>
```
awulkiew: Should this include be removed?

test/algorithms/multi_simplify.cpp
```diff
@@ -1,17 +1,12 @@
...
+// Copyright (c) 2007-2015 Barend Gehrels, Amsterdam, the Netherlands.
 // Use, modification and distribution is subject to the Boost Software License,
 // Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
 // http://www.boost.org/LICENSE_1_0.txt)
 
 #include <geometry_test_common.hpp>
 
-#include <boost/geometry/multi/algorithms/simplify.hpp>
-
-#include <boost/geometry/io/wkt/wkt.hpp>
```
awulkiew: Should this include be removed?

test/algorithms/multi_unique.cpp
```diff
@@ -8,18 +8,10 @@
 #include <geometry_test_common.hpp>
 
 #include <boost/geometry/algorithms/unique.hpp>
-#include <boost/geometry/multi/algorithms/unique.hpp>
-
-#include <boost/geometry/io/wkt/wkt.hpp>
```
awulkiew: And here.

test/algorithms/multi_simplify.cpp
barendgehrels: Good point, don't think so

mkaravel: It is included by `test_simplify.hpp` which is why I removed it.
Typically when we have a supporting `.hpp` file for a unit test, we do not the includes twice.
At least this is my understanding and experience with the unit tests I have seen so far.

test/algorithms/multi_length.cpp
mkaravel: `io/wkt/read.hpp` was included in `test_length.hpp`.
So what I did was to replace `io/wkt/read.hpp` by `io/wkt/wkt.hpp` in test_length.hpp` and removed it from this file.
See also my other comment on the `multi_simplify.cpp` unit test.

test/algorithms/multi_unique.cpp
mkaravel: `io/wkt/wkt.hpp` is included in `test_unique.hpp`.
In that respect it is okay to remove from here.

test/algorithms/multi_length.cpp
barendgehrels: OK for me. It was probably originally included as ".../multi/io/wkt" which is why it was still there. Having it in test_length (etc) is OK

barendgehrels:
I'm OK with merging

PR #268, opened 2015-03-13
Extensions: add ttmath support for fmod

mkaravel:
Add support for the `fmod()` free function.

### DIFF
--- a/include/boost/geometry/extensions/contrib/ttmath_stub.hpp
+++ b/include/boost/geometry/extensions/contrib/ttmath_stub.hpp
@@ -1,8 +1,13 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
-// Copyright (c) 2007-2012 Barend Gehrels, Amsterdam, the Netherlands.
-// Copyright (c) 2008-2012 Bruno Lalande, Paris, France.
-// Copyright (c) 2009-2012 Mateusz Loskot, London, UK.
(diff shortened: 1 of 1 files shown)


mkaravel:
As this is a modification on extensions, I think it makes sense to merge it after we are done with boost 1.58.

include/boost/geometry/extensions/contrib/ttmath_stub.hpp
```diff
@@ -103,6 +107,12 @@ namespace ttmath
         return two * ATan((sqrt(x * x + y * y) - x) / y);
     }
 
+    // needed in order to work with boost::geometry::math::fmod
+    template <uint Exponent, uint Mantissa>
+    inline Big<Exponent, Mantissa> fmod(Big<Exponent, Mantissa> const& x, Big<Exponent, Mantissa> const& y)
```
awulkiew: After the changes done in the other PR here the name should be changed to `mod()` as well.

mkaravel: I am aware of that.
I was waiting to have the okay for the other PRs before modifying this.
I was also considering closing this one and creating a new one so as to keep the history simple.
Will re-act on it once the other PRs are okay/done.

barendgehrels:
OK for me (with the name mod)

mkaravel:
Function renamed from `fmod()` to `mod()`. See commit aac92cb.

PR #269, opened 2015-03-13
Feature : support for fmod free function from within the math namespace

mkaravel:
Add support for calls to `boost::geometry::math::fmod`.
* For floating-point numbers it calls `std::fmod`.
* For fundamental integral numbers it calls `operator%`.
* For user defined number types the free function `fmod` is called (lookup is activated in the global namespace, the `std` namespace and the type's namespace.

### DIFF
--- a/include/boost/geometry/util/math.hpp
+++ b/include/boost/geometry/util/math.hpp
@@ -30,6 +30,7 @@
 #include <boost/math/special_functions/round.hpp>
 #include <boost/numeric/conversion/cast.hpp>
 #include <boost/type_traits/is_fundamental.hpp>
+#include <boost/type_traits/is_integral.hpp>
 
 #include <boost/geometry/util/select_most_precise.hpp>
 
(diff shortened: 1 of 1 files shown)


awulkiew:
Since the intention is to add a function working for arbitrary numerical types maybe it should have more generic name, e.g. `mod()`?

include/boost/geometry/util/math.hpp
```diff
@@ -191,6 +192,66 @@ struct square_root<T, true>
...
+    typename T,
+    bool IsFundemantal = boost::is_fundamental<T>::value /* false */
+>
+struct modulo
+{
+    typedef T return_type;
+
+    static inline T apply(T const& value1, T const& value2)
+    {
+        // for non-fundamental number types assume that sqrt is
```
awulkiew: `sqrt` should be renamed to `fmod`, `function` or something similar.

include/boost/geometry/util/math.hpp
```diff
@@ -191,6 +192,66 @@ struct square_root<T, true>
...
+        // 2) at global scope, or
+        // 3) in namespace std
+        using ::fmod;
+        using std::fmod;
+
+        return fmod(value1, value2);
+    }
+};
+
+template <typename Fundamental, bool IsIntegral>
```
awulkiew: So here IsIntegral could be initialized to boost::is_integral<Fundamental>::value by default.

But actually, I think in general we're specializing math functions for floating-point types. I'm aware that this is done this way in order to support `ttmath::Big` above in the `modulo` struct. So for user-defined types it's required that there should be a function `fmod()`. I think that we should consider the implementation of 3 versions:
- for built-in integrals - using `operator%`
- for built-in floating-point types - calling `fmod()`
- for user-defined types - calling `mod()` because it's more generic name (and maybe additionally a default-version of this function which could be found by the ADL, which'd use operator% or alternativaly call a member function of user-defined type)

include/boost/geometry/util/math.hpp
mkaravel: Thanks for catching this. Typo introduced by careless copy/paste....
Will fix it.

include/boost/geometry/util/math.hpp
mkaravel: True: `IsIntergral` can be initialized as you propose. Will do it your way.

All three versions you mention are already implemented, the difference being that for user defined functions I assume the presence of `fmod()` rather than `mod()`.
I agree that `mod()` is a more generic name. The only instance of user-defined number type that I have and have used is `ttmath::Big<>`, and for those I have added myself the missing function (see PR #268). So we are pretty flexible here. Using `mod()` as the required function is okay with me and will do the change.

mkaravel:
I used `fmod()` to basically stress that this would work with floating-point numbers as well; it is true that `fmod()` is misleading for intergal types.
Using `mod()` as the generic name is okay with me, and will make this change.

mkaravel:
@awulkiew All your comments have been addressed in commit d12f80f.

mkaravel:
@awulkiew I had to overwrite the last commit. So the commit that implements the changes is bf96857.

PR #277, opened 2015-04-01
Workarounds for SunOS

awulkiew:
The proposed changes are workarounds, the code compiles on other platforms. Though I think merging them could be beneficient. Plus, there would be more green in the matrix.

There are a few more failing tests for which workarounds (2 issues) could be added:
* Spherical [distance](http://www.boost.org/development/tests/develop/developer/output/oracle-intel-S2-stlport4-boost-bin-v2-libs-geometry-test-algorithms-distance-distance_se_pl_pl-test-sun-stlport4-release-address-model-32-architecture-x86-threading-multi.html) tests - the use of `comparable_type` private member in the definition of public member struct template `haversine<>::calculation_type<>`, it's hard to believe that this fails only in haversine.
* [point_iterator](http://www.boost.org/development/tests/develop/developer/output/oracle-intel-S2-stlport4-boost-bin-v2-libs-geometry-test-iterators-point_iterator-test-sun-stlport4-release-address-model-32-architecture-x86-threading-multi.html) test - the implicit conversion of reverse iterator to const reverse iterator in the test itself. E.g. only assignment/explicit conversion and comparison could be tested, without implicit conversion.

The workarounds would be trivial, what do you think?

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
@@ -548,9 +548,8 @@ struct buffered_piece_collection
         // reliable integer-based ring. All turns can be compared (inside) to this
         // rings to see if they are inside.
 
-        for (typename piece_vector_type::iterator it = boost::begin(m_pieces);
-            it != boost::end(m_pieces);
-            ++it)
+        for (typename boost::range_iterator<piece_vector_type>::type
+                it = boost::begin(m_pieces); it != boost::end(m_pieces); ++it)
         {
             piece& pc = *it;
             int piece_segment_index = pc.first_seg_id.segment_index;
@@ -562,9 +561,9 @@ struct buffered_piece_collection
                 }
                 // Walk through them, in reverse to insert at right index
                 int index_offset = pc.robust_turns.size() - 1;
-                for (typename std::vector<robust_turn>::const_reverse_iterator
-                        rit = pc.robust_turns.rbegin();
-                    rit != pc.robust_turns.rend();
+                for (typename boost::range_reverse_iterator<const std::vector<robust_turn> >::type
+                        rit = boost::const_rbegin(pc.robust_turns);
+                    rit != boost::const_rend(pc.robust_turns);
                     ++rit, --index_offset)
                 {
                     int const index_in_vector = 1 + rit->seg_id.segment_index - piece_segment_index;
--- a/include/boost/geometry/algorithms/detail/distance/geometry_to_segment_or_box.hpp
+++ b/include/boost/geometry/algorithms/detail/distance/geometry_to_segment_or_box.hpp
@@ -220,7 +220,8 @@ public:
         comparable_return_type cd_min1(0);
         point_iterator_type pit_min;
         seg_or_box_iterator_type it_min1 = seg_or_box_points.begin();
-        seg_or_box_iterator_type it_min2 = ++seg_or_box_points.begin();
+        seg_or_box_iterator_type it_min2 = it_min1;
+        ++it_min2;
         bool first = true;
 
         for (point_iterator_type pit = points_begin(geometry);
--- a/include/boost/geometry/algorithms/detail/distance/geometry_to_segment_or_box.hpp
+++ b/include/boost/geometry/algorithms/detail/distance/geometry_to_segment_or_box.hpp
@@ -185,10 +185,10 @@ public:
                 Geometry const
             > segment_iterator_type;
 
-        typedef typename std::vector
+        typedef typename boost::range_const_iterator
             <
-                segment_or_box_point
-            >::const_iterator seg_or_box_iterator_type;
+                std::vector<segment_or_box_point>
+            >::type seg_or_box_const_iterator;
 
         typedef assign_new_min_iterator<SegmentOrBox> assign_new_value;
 
@@ -219,8 +219,8 @@ public:
         // segment or box
         comparable_return_type cd_min1(0);
         point_iterator_type pit_min;
-        seg_or_box_iterator_type it_min1 = seg_or_box_points.begin();
-        seg_or_box_iterator_type it_min2 = it_min1;
+        seg_or_box_const_iterator it_min1 = boost::const_begin(seg_or_box_points);
+        seg_or_box_const_iterator it_min2 = it_min1;
         ++it_min2;
         bool first = true;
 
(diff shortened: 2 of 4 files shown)


include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -562,9 +561,9 @@ struct buffered_piece_collection
...
                 int index_offset = pc.robust_turns.size() - 1;
-                for (typename std::vector<robust_turn>::const_reverse_iterator
-                        rit = pc.robust_turns.rbegin();
-                    rit != pc.robust_turns.rend();
+                for (typename boost::range_reverse_iterator<const std::vector<robust_turn> >::type
+                        rit = boost::const_rbegin(pc.robust_turns);
+                    rit != boost::const_rend(pc.robust_turns);
                     ++rit, --index_offset)
                 {
                     int const index_in_vector = 1 + rit->seg_id.segment_index - piece_segment_index;
```
barendgehrels: What was wrong with this code? This is standard vector code - I cannot imagine that either layout changes or std code is causing issues.

Anyway, this code will go so I don't object to the change.

mkaravel:
@awulkiew I would like to take a closer look at it. Will probably do this Monday, if not earlier.

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
awulkiew: It fails in the `rit != pc.robust_turns.rend()` line. AFAIU it's because the compiler doesn't explicitly convert the mutable reverse iterator to const reverse iterator before comparison. `rbegin()` and `rend()` return mutable reverse iterators when the container is mutable, which is this case. Those mutable iterators must then be converted into const iterators in the original code. So in the above case the implicit conversions are ommited, const reverse iterators are created explicitly, which should help.

barendgehrels: Thanks for explaining - I'm OK with merging

include/boost/geometry/algorithms/detail/distance/geometry_to_segment_or_box.hpp
```diff
@@ -220,7 +220,8 @@ class geometry_to_segment_or_box
         comparable_return_type cd_min1(0);
         point_iterator_type pit_min;
         seg_or_box_iterator_type it_min1 = seg_or_box_points.begin();
-        seg_or_box_iterator_type it_min2 = ++seg_or_box_points.begin();
+        seg_or_box_iterator_type it_min2 = it_min1;
+        ++it_min2;
```
mkaravel: Following the explanation you gave to Barend for the issue in `buffered_piece_collection.hpp`, my suggestion would be to re-write lines 222 and 223 in the file as:
```
seg_or_box_iterator_type it_min1 = boost::const_begin(seg_or_box_points);
seg_or_box_iterator_type it_min2 = ++boost::const_begin(seg_or_box_points);
```
Please remember to write these in two lines, to be conformant with the 80-chars-per-line requirement.

include/boost/geometry/algorithms/detail/is_valid/has_duplicates.hpp
```diff
@@ -48,7 +48,8 @@ struct has_duplicates
         geometry::equal_to<typename boost::range_value<Range>::type> equal;
 
         iterator it = boost::begin(view);
-        iterator next = ++boost::begin(view);
+        iterator next = it;
+        ++next;
```
mkaravel: Similarly here, I would write like 50 and 51 as:
```
iterator it = boost::const_begin(view);
iterator next = ++boost::const_begin(view);
```
Also in the for-loop at line 52 I would use `next != boost::const_end(view)`.

include/boost/geometry/iterators/flatten_iterator.hpp
```diff
@@ -215,7 +215,8 @@ class flatten_iterator
                 --m_outer_it;
             }
             while ( empty(m_outer_it) );
-            m_inner_it = --AccessInnerEnd::apply(*m_outer_it);
+            m_inner_it = AccessInnerEnd::apply(*m_outer_it);
+            --m_inner_it;
```
mkaravel: For this one I think we should leave it as it is and modify how this is called. The `AccessInnerEnd` class should return a const iterator instead of a mutable iterator, and this should take care of it.

mkaravel: I will check where this is called from and suggest a change.

mkaravel: I propose to re-write `decrement()` as follows:
```
    inline void decrement()
    {
        if ( m_outer_it == m_outer_end
             || m_inner_it == AccessInnerBegin::apply(*m_outer_it) )
        {
            do
            {
                --m_outer_it;
            }
            while ( empty(m_outer_it) );
            m_inner_it = AccessInnerEnd::apply(*m_outer_it);
        }
        --m_inner_it;
    }
```

include/boost/geometry/algorithms/detail/distance/geometry_to_segment_or_box.hpp
awulkiew: Ok, I'll take a look at it later. Using const iterators explicitly shoudl of course be safer, though this is a different change. Have in mind that the conversion from mutable to const iterator is not a problem here and below. This is a different issue than the one I explained.

Here the compiler complains about calling pre-increment or post-increment operator for temporary objects. In general this works when an operator is a member function of a class. But wouldn't work e.g. if the iterator was a pointer, or if the user-defined operator was a free function because rvalue couldn't be converted to non-const reference. And AFAIU on Solaris Compiler it doesn't work in any case.

awulkiew: So I think that your proposal should still fail on Solaris. It could be:

    seg_or_box_iterator_type it_min1 = boost::const_begin(seg_or_box_points);
    seg_or_box_iterator_type it_min2 = it_min2;
    ++it_min2;

or

    seg_or_box_iterator_type it_min1 = boost::const_begin(seg_or_box_points);
    seg_or_box_iterator_type it_min2 = boost::prior(boost::const_begin(seg_or_box_points));

I prefer the first one because it's one dependency less, less functions to debug, probably faster for some iterators which musn't be created two times, boost::prior() call doesn't look intuitive enough for me etc. But I can be persuaded otherwise :)

include/boost/geometry/algorithms/detail/is_valid/has_duplicates.hpp
awulkiew: And then the line 52 would fail from the reason I described before. For this to work also `iterator next` type should be changed to `const_iterator`.

include/boost/geometry/algorithms/detail/distance/geometry_to_segment_or_box.hpp
awulkiew: Actually the above may not be true. It's possible that only `std::vector<>::const_reverse_iterator` fails on Solaris this way. Either way, the code should compile now.

PR #278, opened 2015-04-02
Rename occurrences of CS in extensions.

awulkiew:
This should fix several tests on Android and SunOS platforms where CS is a macro name.

### DIFF
--- a/include/boost/geometry/extensions/gis/projections/proj/hatano.hpp
+++ b/include/boost/geometry/extensions/gis/projections/proj/hatano.hpp
@@ -49,8 +49,8 @@ namespace boost { namespace geometry { namespace projections
             static const int NITER = 20;
             static const double EPS = 1e-7;
             static const double ONETOL = 1.000001;
-            static const double CN = 2.67595;
-            static const double CS = 2.43763;
+            static const double CN_ = 2.67595;
+            static const double CS_ = 2.43763;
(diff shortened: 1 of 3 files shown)


include/boost/geometry/extensions/gis/projections/proj/hatano.hpp
```diff
@@ -49,8 +49,8 @@ namespace boost { namespace geometry { namespace projections
             static const int NITER = 20;
             static const double EPS = 1e-7;
             static const double ONETOL = 1.000001;
-            static const double CN = 2.67595;
-            static const double CS = 2.43763;
+            static const double CN_ = 2.67595;
+            static const double CS_ = 2.43763;
             static const double RCN = 0.37369906014686373063;
```
barendgehrels: This is generated code.
Anyway, I will update the generator (it is not yet in git)

barendgehrels:
I'm OK, thanks

include/boost/geometry/extensions/gis/projections/proj/hatano.hpp
awulkiew: Ok, then I suggest to use lower-case letters for those names to avoid problems with macros and be in line with the Boost guidelines.

Btw, where are those numbers from? They look like single-precision, probably not good enough for double or more precise type.

PR #279, opened 2015-04-09
Feature/models upgrades

awulkiew:
This PR adds some C++11 upgrades for geometries models:
- support for std::initializer_list,
- defaulted ctor for non-complex geometries (point, box, segment),

compile-time checks:
- segment's conept check Point
- point's DimensionCount check using MPL_ASSERT_MSG

and docs updates:
- missing descriptions of ctors,
- consistent description of ctors of non-complex geometries.

### DIFF
--- a/doc/src/examples/geometries/point.cpp
+++ b/doc/src/examples/geometries/point.cpp
@@ -19,11 +19,12 @@ int main()
 {
     bg::model::point<double, 2, bg::cs::cartesian> point1;
     bg::model::point<double, 3, bg::cs::cartesian> point2(1.0, 2.0, 3.0); /*< Construct, assigning three coordinates >*/
-    point1.set<0>(1.0); /*< Set a coordinate. [*Note]: prefer using `bg::set<0>(point1, 1.0);` >*/
-    point1.set<1>(2.0);
 
-    double x = point1.get<0>(); /*< Get a coordinate. [*Note]: prefer using `x = bg::get<0>(point1);` >*/
-    double y = point1.get<1>();
+    bg::set<0>(point1, 1.0); /*< Set a coordinate. >*/
+    bg::set<1>(point1, 2.0);
+
+    double x = bg::get<0>(point1); /*< Get a coordinate. >*/
+    double y = bg::get<1>(point1);
 
     std::cout << x << ", " << y << std::endl;
     return 0;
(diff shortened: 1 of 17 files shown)


awulkiew:
The last commit simplifies the `model::point` example. Since the user should prefer `bg::get` and `bg::set` there is no need to mention member functions. Actually it may be harmful for the user.

The next step would be to add the examples for other geometries. In particular showing how the unified initialization syntax is supported in complex geometries models like `model::polygon`. As you probably remember, it looks similar to WKT:

    // clockwise, closed polygon containing an exterior and interior ring
    polygon_t, poly{{{0,0},{0,5},{5,5},{5,0},{0,0}},{{1,1},{4,1},{4,4},{1,4},{1,1}}};

I'll wait with docs upgrades in case you had something against the changes proposed so far.

doc/src/examples/geometries/point.cpp
```diff
@@ -19,11 +19,12 @@ int main()
...
-    point1.set<0>(1.0); /*< Set a coordinate. [*Note]: prefer using `bg::set<0>(point1, 1.0);` >*/
-    point1.set<1>(2.0);
 
-    double x = point1.get<0>(); /*< Get a coordinate. [*Note]: prefer using `x = bg::get<0>(point1);` >*/
-    double y = point1.get<1>();
+    bg::set<0>(point1, 1.0); /*< Set a coordinate. >*/
+    bg::set<1>(point1, 2.0);
+
+    double x = bg::get<0>(point1); /*< Get a coordinate. >*/
+    double y = bg::get<1>(point1);
```
barendgehrels: This sample explains the class point.cpp, not the point concept. Therefore it was written as it was. The note is I think clear enough: prefer use bg::set. But it is good to have an example of the methods of the class itself too.
Like point_xy which would give the usage of .x() and .y() (only there is no example there...)

barendgehrels:
I'm OK with the changes and new examples for initializer lists, as long as is clearly indicated that it is C++11.
As noted, I prefer keeping the point.cpp example as it was... Otherwise, the constructor should also go (prefer geometry::make), etc. But maybe it is good to mention (also in the example) that this example is specific for this class and not for the generic way of working.

awulkiew:
I restored the example using get/set members for one of the coordinates. I also added an example for box, with C++11 feature. Is it clear enough?

For further docs updates I'll create a separate PR to not obscure more important changes of this PR which are the C++11 features.

PR #281, opened 2015-04-21
Detection of access to uninitialized or destroyed geometries.

awulkiew:
If `BOOST_GEOMETRY_ENABLE_ACCESS_DEBUGGING` is defined getters and setters checks if the data being read was properly initialized and if the object was properly created and wasn't destroyed.

With this change it's easy to detect e.g. the use of references to destroyed temporaries or the use of uninitialized objects.


barendgehrels:
Good idea. Can you rename m_magic to, for example, m_assigned ? (or, longer but clearer, m_access_debugging_assigned)?
For the rest I'm OK with merging this.

PR #282, opened 2015-04-23
[extensions][wkb] Fix GCC compilation errors and warnings.

awulkiew:


include/boost/geometry/extensions/gis/io/wkb/detail/writer.hpp
```diff
@@ -126,7 +126,8 @@ namespace detail { namespace wkb
 
             typedef typename point_type<L>::type point_type;
 
-            for(boost::range_iterator<const L>::type point_iter = boost::begin(linestring);
+            for(typename boost::range_iterator<const L>::type
```
mkaravel: `const` should go after `L`.

include/boost/geometry/extensions/gis/io/wkb/detail/writer.hpp
```diff
@@ -165,7 +166,8 @@ namespace detail { namespace wkb
             uint32_t num_exterior_ring_points = boost::geometry::num_points(exterior_ring);
             value_writer<uint32_t>::write(num_exterior_ring_points, iter, byte_order);
 
-            for(boost::range_iterator<const ring_type>::type point_iter = boost::begin(exterior_ring); 
+            for(typename boost::range_iterator<const ring_type>::type
```
mkaravel: Same here: `const` should go after `ring_type`.

include/boost/geometry/extensions/gis/io/wkb/detail/writer.hpp
```diff
@@ -178,14 +180,16 @@ namespace detail { namespace wkb
 
             interior_rings_type interior_rings = boost::geometry::interior_rings(polygon);
 
-            for(boost::range_iterator<const interior_rings_type>::type interior_ring_iter = boost::begin(interior_rings); 
+            for(typename boost::range_iterator<const interior_rings_type>::type
```
mkaravel: Same issue here.

include/boost/geometry/extensions/gis/io/wkb/detail/writer.hpp
```diff
@@ -178,14 +180,16 @@ namespace detail { namespace wkb
...
+                    interior_ring_iter = boost::begin(interior_rings);
                 interior_ring_iter != boost::end(interior_rings);
                 ++interior_ring_iter)
             {
                 uint32_t num_interior_ring_points = boost::geometry::num_points(*interior_ring_iter);
                 value_writer<uint32_t>::write(num_interior_ring_points, iter, byte_order);
 
-                for(boost::range_iterator<const ring_type>::type point_iter = boost::begin(*interior_ring_iter); 
+                for(typename boost::range_iterator<const ring_type>::type
```
mkaravel: And here.

include/boost/geometry/extensions/gis/io/wkb/detail/writer.hpp
```diff
@@ -178,14 +180,16 @@ namespace detail { namespace wkb
 
             interior_rings_type interior_rings = boost::geometry::interior_rings(polygon);
```
mkaravel: It seems here that interior rings are copied.
Is that a good idea? I think we should avoid the copying. We have already faced similar issues in the library and there is a sort-of standard way for addressing this.

include/boost/geometry/extensions/gis/io/wkb/write_wkb.hpp
```diff
@@ -25,7 +25,7 @@ template <typename G>
 struct write_wkb<point_tag, G>
 {
     template <typename OutputIterator>
-    static inline bool write(const G& geometry, OutputIterator& iter, 
+    static inline bool write(const G& geometry, OutputIterator iter,
```
mkaravel: Here and below.
Why the `OutputIterator` is passed by value rather then by reference as in the original code?
I prefer the reference approach, as with the value approach it is not possible to recover the iterator value after the call to this method.

include/boost/geometry/extensions/gis/io/wkb/detail/writer.hpp
mkaravel: The general paradigm we have used for handling this is as follows.
```
struct my_struct
{
    template <typename InteriorRingIterator>
    static void do_the_work(InteriorRingIterator first, InteriorRingIterator last)
    {
        for (InteriorRingIterator it = first; it != last; ++it)
        {
            // do what is needed
        }
    }

    template <typename InteriorRings>
    static void do_the_work(InteriorRings const& interior_rings)
    {
        do_the_work(boost::begin(interior_rings), boost::end(interior_rings));
    }

    static void apply(...)
    {
        // this is the initial caller
        do_the_work(geometry::interior_rings(polygon));
    }
};
```

include/boost/geometry/extensions/gis/io/wkb/detail/writer.hpp
```diff
@@ -178,14 +180,16 @@ namespace detail { namespace wkb
...
             {
                 uint32_t num_interior_ring_points = boost::geometry::num_points(*interior_ring_iter);
                 value_writer<uint32_t>::write(num_interior_ring_points, iter, byte_order);
 
-                for(boost::range_iterator<const ring_type>::type point_iter = boost::begin(*interior_ring_iter); 
+                for(typename boost::range_iterator<const ring_type>::type
+                        point_iter = boost::begin(*interior_ring_iter);
                     point_iter != boost::end(*interior_ring_iter); 
                     ++point_iter)
                 {
```
mkaravel: Line 197 is too long and should be wrapped.

include/boost/geometry/extensions/gis/io/wkb/write_wkb.hpp
awulkiew: It's because otherwise it wouldn't work on a standard-conformant compiler for a r-value type passed as the iterator. In the tests `std::back_insert_iterator` returned by `std::back_inserter()` is passed into this function which doesn't compile e.g. on GCC. This is why the iterators in the STL algorithms are always passed by value and if needed returned from the function.

In this particular case I guess the iterator could be passed by reference into the `dispatch::write_wkb()` but still it must be passed by value into the `geometry::write_wkb()`. I passed it by value also in the internals the same way how it's done in the STL.

include/boost/geometry/extensions/gis/io/wkb/detail/writer.hpp
awulkiew: Good catch!
I just used the correct `..._return_type` for them to not modify too much.

mkaravel: Great! Thanks for making the change.
`interior_return_type` is indeed better and should avoid the problem I had in mind.

mkaravel:
@awulkiew Thanks for the fixes.

I understand your line of thought about the output iterator.
Most probably the corresponding methods would need to be re-designed and the output iterator returned somehow (either as is and the boolean passed as an argument by reference, or returning an `std::pair<OutputIterator, bool>`, or something else). I fear this is going too far for this right now, and most probably it is not really required as I do not see how the output iterator returned would be used in this case anyway. Let's leave it as it is for now, and if the need comes up we can do the re-design.

I am okay with merging.

awulkiew:
Ok, thanks!

include/boost/geometry/extensions/gis/io/wkb/detail/ogc.hpp
```diff
@@ -107,99 +107,61 @@ struct ewkt_policy
...
-    typename Tag = typename tag<Geometry>::type
->
-struct geometry_type : not_implemented<Tag>
-{
-};
-
-template <typename Geometry, typename CheckPolicy>
-struct geometry_type<Geometry, CheckPolicy, point_tag>
+template <typename Geometry,
+          geometry_type_ogc::enum_t OgcType,
```
barendgehrels: enum_t is not introduced here, but it does not follow any of our conventions

include/boost/geometry/extensions/gis/io/wkb/detail/parser.hpp
```diff
@@ -117,7 +118,9 @@ struct geometry_type_parser
     }
 };
 
-template <typename P, int I, int N>
+template <typename P,
+          std::size_t I = 0,
+          std::size_t N = dimension<P>::value>
 struct parsing_assigner
```
barendgehrels: Indentation is changed. If so, please follow conventions then, angle-bracket new line, last one too

include/boost/geometry/extensions/gis/io/wkb/detail/parser.hpp
```diff
@@ -153,10 +156,7 @@ struct parsing_assigner<P, N, N>
                 byte_order_type::enum_t order)
     {
         // terminate
-        boost::ignore_unused_variable_warning(it);
-        boost::ignore_unused_variable_warning(end);
-        boost::ignore_unused_variable_warning(point);
-        boost::ignore_unused_variable_warning(order);
+        boost::ignore_unused(it, end, point, order);
```
barendgehrels: Looks good

include/boost/geometry/extensions/gis/io/wkb/detail/writer.hpp
```diff
@@ -58,30 +67,36 @@ namespace detail { namespace wkb
...
-        static void run(const P& point, OutputIterator& iter, byte_order_type::enum_t byte_order)
+        static void run(P const& point,
+                        OutputIterator& iter,
+                        byte_order_type::enum_t byte_order)
         {
             // terminate
-            boost::ignore_unused_variable_warning(point);
-            boost::ignore_unused_variable_warning(iter);
-            boost::ignore_unused_variable_warning(byte_order);
+            boost::ignore_unused(point, iter, byte_order);
```
barendgehrels: Instead, we usually leave parameter names out here

include/boost/geometry/extensions/gis/io/wkb/detail/writer.hpp
```diff
@@ -110,8 +126,9 @@ namespace detail { namespace wkb
     struct linestring_writer
     {
         template <typename OutputIterator>
-        static bool write(const L& linestring, OutputIterator& iter,
-            byte_order_type::enum_t byte_order)
+        static bool write(L const& linestring,
```
barendgehrels: Linestring

include/boost/geometry/extensions/gis/io/wkb/detail/ogc.hpp
awulkiew: What do you mean? `enum_t` is a name of a type defined in `geometry_type_ogc`. Do you think that the name of a type should be changed to something else? Or should it be defined in a global scope? Is this is a question to the original author?

barendgehrels: I know it is not part of this review (therefore I wrote "not introduced here"). So yes, it is a remark to the original author (probably Mateusz or Mats).
OK, inspecting the code it is probably part of the scoped enumerations from Boost. You can ignore this remark.

awulkiew: Ok, I addressed all of your suggestions (and more) in this commit: https://github.com/boostorg/geometry/commit/a48df2a15338cc4e38c30496f1678435c2c6faca

barendgehrels: Thanks!

PR #283, opened 2015-04-24
Fix point arithmetic functions and centroid

awulkiew:
The intention of this PR is to fix this ticket: https://svn.boost.org/trac/boost/ticket/11236

In this particular case, at the end of centroid calculation the resulting point is moved using the origin (taken from the input geometry). The point arithmetic functions internally was calculating the results using always the type of the second operand (in one of the STL binary predicates like `std::plus`). In this particular case the resulting Point is using coordinates of type `double` and the input Polygon coordinates of type `int`. `bg::add_point()` internally calls `std::plus<int>` which truncates the `double` value before addition. Ths PR changes it by using the more precise type.

### DIFF
--- a/include/boost/geometry/arithmetic/arithmetic.hpp
+++ b/include/boost/geometry/arithmetic/arithmetic.hpp
@@ -22,6 +22,7 @@
 #include <boost/geometry/core/coordinate_type.hpp>
 #include <boost/geometry/geometries/concepts/point_concept.hpp>
 #include <boost/geometry/util/for_each_coordinate.hpp>
+#include <boost/geometry/util/select_most_precise.hpp>
 
 
 namespace boost { namespace geometry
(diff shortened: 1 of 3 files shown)


include/boost/geometry/arithmetic/arithmetic.hpp
```diff
@@ -51,17 +52,24 @@ struct value_operation
         : m_value(value)
     {}
 
-    template <typename P, int I>
-    inline void apply(P& point) const
+    template <typename PointDst, int I>
+    inline void apply(PointDst& dest_point) const
     {
```
barendgehrels: Abbreviations used (Dst, dest) are not the same.
While you are at it, you might change I to Index and make type std::size_t

What is C ?

barendgehrels:
For the rest I'm OK

include/boost/geometry/arithmetic/arithmetic.hpp
awulkiew: My primary goal was was to fix the issue so I've taken the names that were already used in this code. So `C` is a type of a value passed as the second argument to the `XXX_value()` functions. But sure, I'll fix the names.

awulkiew:
Thanks!

I now saw that a type of a parameter taken by `XXX_value()` functions is always `coordinate_type<Point>`. So my change wasn't needed in this case. So currently my fix doesn't modify the behavior of the function, but we could consider taking an arbitrary type as the second parameter. Then the intermediate result would be calculated using the most precise type. Though this is probably not needed in general because those functions aren't used in the library like this, at least for now.

PR #284, opened 2015-04-27
Add two-pi and half-pi constants

mkaravel:
This PR:
* Adds to more constants in `boost::geometry::math` namespace: `2*π` and `π/2`.
* `ttmath_big` and `ttmath::Big<Exponent, Mantissa>` are modified to support these two new constants.
* Strategies that compute these two constants via `math::pi<>()`, now use the corresponding functions directly.

### DIFF
--- a/include/boost/geometry/extensions/contrib/ttmath_stub.hpp
+++ b/include/boost/geometry/extensions/contrib/ttmath_stub.hpp
@@ -190,15 +190,49 @@ namespace detail
     {
         static inline ttmath::Big<Exponent, Mantissa> apply()
         {
+            static ttmath::Big<Exponent, Mantissa> const half_pi(
+                "1.57079632679489661923132169163975144209858469968755291048747229615390820314310449931401741267105853399107404325664115332354692230477529111586267970406424055872514205135096926055277982231147447746519098");
+            return half_pi;
+        }
+    };
+
+    // Partial specialization for ttmath
+    template <ttmath::uint Exponent, ttmath::uint Mantissa>
+    struct define_half_pi<ttmath::Big<Exponent, Mantissa> >
+    {
+        static inline ttmath::Big<Exponent, Mantissa> apply()
+        {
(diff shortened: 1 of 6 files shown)


awulkiew:
Looks good. How precise should be constants defined for `ttmath::Big`? The user could need something less precise (it could be faster) or more precise than we have. AFAIU all `ttmath::Big` numbers after e.g. multiplying them by `pi` will have the same precision as `pi` if they were less precise. This could kill the performance. Is there an "optimal" precision of `ttmath::Big`?

mkaravel:
Good question and this is also related to how we define π, which I have not changed.

I think what happens is that the additional digits provided by our string are discarded, i.e., `ttmath_big` and `ttmath::Big<>` keep only the digits that are necessary to fill the mantissa and throw away the rest.
`ttmath` numbers are fixed-precision, so the number of bits stored is determined ahead of time and not changed (dynamically or otherwise). In that respect the only overhead that I see in the current implementation is that we provide many more digits than needed for `ttmath_big`, but this overhead is encountered only once when `the_pi` static variable is constructed.

Another solution would be to implement, e.g. `define_pi`, as follows:
```
template <ttmath::unit Exponent, ttmath::unit Mantissa>
struct define_pi<ttmath::Big<Exponent, Mantissa>
{
    static inline ttmath::Big<Exponent, Mantissa> apply()
    {
        ttmath::Big<Exponent, Mantissa> pi_value(0);
        pi_value.SetPi();
        return pi_value;
    }
};
```
but this would mean that π would be constructed any time the `apply()` method is called.
In fact as far as I can understand `SetPi()` does the same as we do: it stores 400 digits ahead of time and uses only those needed to fill the mantissa.
**Note:** apart from `SetPi()`, `ttmath::Big<>` defines also `Set2Pi()` and `Set05Pi()` for 2*π and π/2, respectively.

I think the current implementation is okay as it is.

awulkiew:
All of your points are valid. And I of course like that the values are created only once.

PR #287, opened 2015-04-29
Fix/area huiller

awulkiew:
Inspired by: http://stackoverflow.com/questions/9439644/bug-in-boostgeometrys-area-huiller-algorithm

### DIFF
--- a/include/boost/geometry/strategies/spherical/area_huiller.hpp
+++ b/include/boost/geometry/strategies/spherical/area_huiller.hpp
@@ -6,6 +6,7 @@
 // Modifications copyright (c) 2015, Oracle and/or its affiliates.
 
 // Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
+// Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
 
 // Use, modification and distribution is subject to the Boost Software License,
 // Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
@@ -36,9 +37,14 @@ namespace strategy { namespace area
 \tparam PointOfSegment point type of segments of rings/polygons
 \tparam CalculationType \tparam_calculation
 \author Barend Gehrels. Adapted from:
-- http://www.soe.ucsc.edu/~pang/160/f98/Gems/GemsIV/sph_poly.c
+- http://webdocs.cs.ualberta.ca/~graphics/books/GraphicsGems/gemsiv/sph_poly.c
+- http://tog.acm.org/resources/GraphicsGems/gemsiv/sph_poly.c
 - http://williams.best.vwh.net/avform.htm
-\note The version in Gems didn't account for polygons crossing the 180 meridian.
+\note The version in Graphics Gems IV (page 132-137) didn't account for
+polygons crossing the 0 and 180 meridians. The fix for this algorithm
+can be found in Graphics Gems V (pages 45-46). See:
+- http://kysmykseka.net/koti/wizardry/Game%20Development/Programming/Graphics%20Gems%204.pdf
+- http://kysmykseka.net/koti/wizardry/Game%20Development/Programming/Graphics%20Gems%205.pdf
 \note This version works for convex and non-convex polygons, for 180 meridian
 crossing polygons and for polygons with holes. However, some cases (especially
 180 meridian cases) must still be checked.
@@ -117,6 +123,8 @@ public :
             calculation_type const half = 0.5;
             calculation_type const two = 2.0;
             calculation_type const four = 4.0;
+            calculation_type const pi
+                = geometry::math::pi<calculation_type>();
             calculation_type const two_pi
                 = geometry::math::two_pi<calculation_type>();
             calculation_type const half_pi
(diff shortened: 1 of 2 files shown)


include/boost/geometry/strategies/spherical/area_huiller.hpp
```diff
@@ -142,20 +150,14 @@ public :
 
             E = geometry::math::abs(E);
 
-            // In right direction: positive, add area. In left direction: negative, subtract area.
-            // Longitude comparisons are not so obvious. If one is negative, other is positive,
-            // we have to take the dateline into account.
```
mkaravel: I would keep these three lines of comments I think they are very useful.

include/boost/geometry/strategies/spherical/area_huiller.hpp
```diff
@@ -142,20 +150,14 @@ public :
...
-            calculation_type lon1 = geometry::get_as_radian<0>(p1) < 0
-                ? geometry::get_as_radian<0>(p1) + two_pi
-                : geometry::get_as_radian<0>(p1);
-
-            calculation_type lon2 = geometry::get_as_radian<0>(p2) < 0
-                ? geometry::get_as_radian<0>(p2) + two_pi
-                : geometry::get_as_radian<0>(p2);
+            calculation_type lon = geometry::get_as_radian<0>(p2)
+                                 - geometry::get_as_radian<0>(p1);
+            if ( lon <= 0 )
```
mkaravel: Although it does not make much of a difference, I would write this as `lon < 0`. If `lon` is 0 then the segment line along a meridian and the area of the triangle formed by `p1`, `p2` and the south pole is identically equal to zero.

My suggestions would be:

1. Check if the two longitudes are the same. Then in this case do not compute the excess, but rather set it directly to 0.
2. Use `math::smaller(lon, 0)` rather than `lon < 0`.
3. Use a different name for the variable `lon`; it refers to the difference of longitudes and the current name is somehow misleading.
4. Follow our coding rules and do not put spaces after `(` and before `)`.

include/boost/geometry/strategies/spherical/area_huiller.hpp
```diff
@@ -142,20 +150,14 @@ public :
...
+            {
+                lon += two_pi;
+            }
 
-            if (lon2 < lon1)
+            if ( lon > pi )
```
mkaravel: According to our coding conventions this should read `if (lon > pi)`.

include/boost/geometry/strategies/spherical/area_huiller.hpp
```diff
@@ -117,6 +123,8 @@ public :
             calculation_type const half = 0.5;
             calculation_type const two = 2.0;
             calculation_type const four = 4.0;
+            calculation_type const pi
+                = geometry::math::pi<calculation_type>();
```
mkaravel: You do not need to create a local variable for π. It is used only once.

mkaravel:
Adam, thanks for going over this strategy.

AFAIU there is also a problem when one of the vertices of the areal geometry is the south pole. Based on some experiments of my own, it seems that if the south pole is a vertex of the ring/polygon/multipolygon, then the `area()` method of the strategy should return `sum * radius * radius` rather than `-sum * radius * radius`; consider for example the clockwise closed ring (in degrees) `POLYGON((50 0,100 0,100 -90,50 0))`, which gives me a negative area, whereas it should clearly have been positive. The way to fix this is to keep track of whether the south pole is a vertex of the geometry considered and properly return what should be returned.

mkaravel:
There are at least two more issues with our implementation of this strategy:

1. It does not handle correctly areal geometries that have segments that contain the south pole in their interior (this could be due to numerical errors, but still....).
2. It implicitly assumes that the coordinates of the points passed are normalized. If you pass out-of-range longitudes for example, it can easily fail.

I already have a prototype that fixes most of these issues. I should be able to finish it soon and be able to push it.

include/boost/geometry/strategies/spherical/area_huiller.hpp
awulkiew: I've done it because of the convention used. Any of the above variables is not needed. And the compiler should optimize them out anyway.

include/boost/geometry/strategies/spherical/area_huiller.hpp
awulkiew: Yes I was considering using `lon < 0`, it gives good results. I just decided to keep the approach used in the book mainly because below `lon` is compared with `pi` so both checks must be consistent and I didn't analyse if the result should be ok. The naming is also taken from the book. In the same code there is a variable called `E` probably also taken from some source. In general I noticed that in Geometry names of variables are taken from papers even if they are against our guidelines. But sure, I already altered the algorithm so I can change the naming as well.

AFAIU using `math::equals()` or `math::smaller()` could add an error because the longitudes would be compared according to the EPS. So currently a value of `E` may be calculated and it may be not fully consistent with the value of `lon` and `pi`. But do you think this is a problem? Probably only if the value of `E` can changes rapidly and the later checks are there to reverse this change, but I think that this isn't the case. I'll think about it.

awulkiew:
My intention was to fix the issue that is currently there for a very basic case. The problems around poles are a known issue. AFAIU this algorithm in general is not suitable for polygons around poles so for them we'd be forced to implement some workaround. But I'm not sure if it should be a part of this PR.

We didn't yet agree how the normalization of coordinate should look like. Maybe we shoudn't normalize the coordinates at all?

include/boost/geometry/strategies/spherical/area_huiller.hpp
barendgehrels: Yes, please don't remove these kind of comments

barendgehrels:
Thanks - I'm OK with merging after comments are addressed

mkaravel:
I am not okay with merging BTW. I will give more info soon.

include/boost/geometry/strategies/spherical/area_huiller.hpp
awulkiew: Sorry, I got an impression that it's a comment made explicitly for the approach used below. But indeed it's more general.

PR #288, opened 2015-05-01
Feature: New algorithms relate() and relation()

awulkiew:
This PR adds the OGC SFA Relate functionality. The use case looks as follows:

    // within
    bool a = bg::relate(geom1, geom2, bg::de9im::mask("T*F**F***"));
    // touches
    bool b = bg::relate(geom1, geom2, bg::de9im::mask("FT*******")
                                   || bg::de9im::mask("F**T*****")
                                   || bg::de9im::mask("F***T****"));

Masks can also be create at compile-time (they can also be ORed with `operator||`):

    bool c = bg::relate(geom1, geom2,
                        bg::de9im::static_mask<'T','*','F','*','*','F' /* *** */>());

This PR also adds a non-standard function caluclating the relation for a pair of geometries:

    bg::de9im::matrix m = bg::relation(geom1, geom2);

It's implemented as a separate function because IMO it's more clear. Furthermore in case if we wanted to add a Strategy in the future and if both variants had the same name `relate` then the third argument would have to be `MaskOrStrategy` which could be confusing.

The documentation for both functions is added in the "Algorithms" section with examples and a picture (taken from within). A new section DE9IM is added to the Reference and now contains `de9im::matrix`, `de9im::mask` and `de9im::static_mask`.


meastp:
That's a nice and easy to use interface :)

mloskot:
:+1:

include/boost/geometry/algorithms/detail/relate/de9im.hpp
```diff
@@ -0,0 +1,422 @@
...
+\details This mask can be used to check spatial relations as defined in
+         Dimensionally Extended 9-Intersection Model.
+
+\qbk{[heading See also]}
+\qbk{* [link geometry.reference.algorithms.relate relate]}
+ */
+class mask
+    : public detail::relate::mask<3, 3>
+{
+    typedef detail::relate::mask<3, 3> base_t;
```
mkaravel: I think we typically use `_type` in such cases. So `base_t` should rather be `base_type` I think.

include/boost/geometry/algorithms/detail/relate/de9im.hpp
```diff
@@ -0,0 +1,422 @@
...
+// EQUALS
+template <typename Geometry1, typename Geometry2>
+struct static_mask_equals_type
+{
+    typedef geometry::de9im::static_mask<'T', '*', 'F', '*', '*', 'F', 'F', 'F', '*'> type; // wikipedia
+    //typedef geometry::de9im::static_mask<'T', 'F', 'F', 'F', 'T', 'F', 'F', 'F', 'T'> type; // OGC
+};
+
+// DISJOINT
+typedef geometry::de9im::static_mask<'F', 'F', '*', 'F', 'F', '*', '*', '*', '*'> static_mask_disjoint;
```
mkaravel: Why is this defined as a typedef, while `static_mask_equals_type` via a struct?
I think it would be better to have a unified way for defining these static masks.

include/boost/geometry/algorithms/detail/relate/implementation.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+//template <typename Point, typename Box, int TopDim2>
+//struct relate<Point, Box, point_tag, box_tag, 0, TopDim2, false>
+//    : detail::relate::point_box<Point, Box>
+//{};
+//
+//template <typename Box, typename Point, int TopDim1>
+//struct relate<Box, Point, box_tag, point_tag, TopDim1, 0, false>
+//    : detail::relate::box_point<Box, Point>
+//{};
```
mkaravel: Please either remove these comments, or explain why they are commented (possibly with a `TODO` explanation).

include/boost/geometry/algorithms/detail/relate/relate_base.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+#ifndef DOXYGEN_NO_DETAIL
+namespace detail { namespace relate {
+
+struct implemented_tag {};
+
+template <template <typename, typename> class StaticMaskTrait,
+          typename Geometry1,
+          typename Geometry2>
+struct relate_base
```
mkaravel: I do not particularly like the name `relate_base` (I think we rarely use `_base` in names). Maybe `relate_impl`?

include/boost/geometry/algorithms/detail/relate/result.hpp
```diff
@@ -982,31 +1024,34 @@ struct static_check_matrix
 
 template <typename StaticMask, bool Interrupt>
 class static_mask_handler
-    : private matrix<3>
+    : private matrix_handler< matrix<3> >
 {
-    typedef matrix<3> base_t;
+    typedef matrix_handler< relate::matrix<3> > base_t;
```
mkaravel: Same comment here about `base_t`.

include/boost/geometry/strategies/intersection_result.hpp
```diff
@@ -26,7 +31,7 @@ namespace boost { namespace geometry
     \ingroup overlay
     \see http://gis.hsr.ch/wiki/images/3/3d/9dem_springer.pdf
 */
-struct de9im
+/*struct de9im
```
mkaravel: Why keeping this commented?

mkaravel:
@awulkiew Thanks for the new functionality!!!

What I would really like to see is predefined documented masks that would be accessible by the user. To be more precise, as a user I would like to be able to write something like `bg::de9im::static_mask_covered_by` or `bg::de9im::mask_covered_by` (I think we need both, but I am not 100% sure) and use it with `relate()`.
What do you think?

include/boost/geometry/algorithms/detail/relate/de9im.hpp
awulkiew: I think because disjoint/intersects doesn't use relate so I just ommited this one. Another good explanation could be laziness :) Even if those are details I guess those predefined masks indeed should probably have the same interface.

include/boost/geometry/algorithms/detail/relate/implementation.hpp
awulkiew: I started implementing them but didn't finished because for this we should first agree how should the Box having the same min and max corner points be treated.

include/boost/geometry/strategies/intersection_result.hpp
awulkiew: Yes, I think it could be removed though I'd like to hear from Barend what I should do with it because AFAIU he was the one who contributed this code.

awulkiew:
Those masks could be added. Static versions are already in the details. The interface would have to be more complicated because a mask can depend on the dimensionality of the Geometries, so e.g. be different for Linestrings than for Polygons. That said, I'm not sure if they'd be needed since we have other algorithms for this. I think we could keep them in the details unless some user explicitly requested them.

Furthermore I think we could consider adding a mechanism for "full" run-time composition (ORing) of masks. Currently the user must pass them and connect together with `operator||`. This way he can connect only as many mask as there are in the expression. In other words the number of the masks is defined at compile-time. The most naiive thing to do would be to allow passing a Range of masks. But in this case the connection between masks stored in a range wouldn't be so straighforward. We could also define our own run-time `de9im::or` container but I'd prefer something more elegant.

mkaravel:
Regarding the interface, I guess you only need to templatize the mask by the geometries. Less elegant than what I original thought, but still, I think, useful. Not to mention that this way it will be easier at either compile time or run-time disallow masks that are not defined for certain geometry combinations.

Other than that, I am okay with leaving it aside for now until a concrete need for this appear.
I also like the idea of run-time composition of masks.

mkaravel:
BTW, the reason I proposed the masks at the user level is very simple: I wanted to have an easy interface through `relate()` and `relation()` to implement relational operations. Given such an interface I would have a hard time remembering the DE9IM definitions of the standard relations, so it would make sense to me that we provide these as part of the functionality of `relate()` and `relation()`.

awulkiew:
Yes, I understand your concerns. I've unified the predefined static masks. I prefer to keep them in the details for now. We should be able to easily move them if users requested it. Besides I think that the run-time predefined masks could probably be more useful for some of the users. Instead of using static ones they could just call an algorithm.

mkaravel:
@awulkiew Okay, let's leave them in details for now.
Thanks for taking care of my other comments.

mkaravel:
I am okay with merging this PR.

doc/src/examples/algorithms/relate.cpp
```diff
@@ -0,0 +1,57 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+// QuickBook Example
+
+// Copyright (c) 2011-2012 Barend Gehrels, Amsterdam, the Netherlands.
```
barendgehrels: Maybe you want to update the copyright ;-)

include/boost/geometry/algorithms/detail/relate/de9im.hpp
```diff
@@ -0,0 +1,434 @@
...
+#include <boost/geometry/algorithms/not_implemented.hpp>
+#include <boost/geometry/core/topological_dimension.hpp>
+#include <boost/geometry/core/tag.hpp>
+
+// TEMP - move this header to geometry/detail
+#include <boost/geometry/index/detail/tuples.hpp>
+
+namespace boost { namespace geometry {
+    
+namespace de9im {
```
barendgehrels: curly brace goes on next line

include/boost/geometry/algorithms/detail/relate/de9im.hpp
```diff
@@ -0,0 +1,434 @@
...
+#include <boost/mpl/vector_c.hpp>
+#include <boost/static_assert.hpp>
+#include <boost/tuple/tuple.hpp>
+
+#include <boost/geometry/algorithms/detail/relate/result.hpp>
+#include <boost/geometry/algorithms/not_implemented.hpp>
+#include <boost/geometry/core/topological_dimension.hpp>
+#include <boost/geometry/core/tag.hpp>
+
+// TEMP - move this header to geometry/detail
```
barendgehrels: /algorithms/detail ? or util ?

include/boost/geometry/algorithms/detail/relation/interface.hpp
```diff
@@ -0,0 +1,183 @@
...
+    template <typename Matrix>
+    static inline Matrix
+    apply(Geometry1 const& geometry1,
+          boost::variant<BOOST_VARIANT_ENUM_PARAMS(T)> const& geometry2)
+    {
+        return boost::apply_visitor(visitor<Matrix>(geometry1), geometry2);
+    }
+};
+
+template <
```
barendgehrels: I don't know why suddenly all braces not go on next line, but this is not the style of the library...

include/boost/geometry/strategies/intersection_result.hpp
barendgehrels: Yes, I think so, already 7 or 8 years ago.
I'm very fine with removing it now.

The new code looks very good.

test/algorithms/relational_operations/relate/relate_pointlike_xxx.cpp
```diff
@@ -92,6 +92,8 @@ void test_all()
```
barendgehrels: What does xxx mean? 
Should it be renamed to e.g. "other" ?

doc/doxy/Doxyfile
```diff
@@ -190,6 +190,7 @@ INPUT                  = . .. ../../../../boost/geometry/core \
 			../../../../boost/geometry/algorithms/detail/is_valid \
 			../../../../boost/geometry/algorithms/detail/overlay \
 			../../../../boost/geometry/algorithms/detail/relate \
+            ../../../../boost/geometry/algorithms/detail/relation \
```
barendgehrels: indentation looks wrong

doc/src/examples/algorithms/relate.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+    polygon_type poly;
+    boost::geometry::read_wkt(
+        "POLYGON((2 1.3,2.4 1.7,2.8 1.8,3.4 1.2,3.7 1.6,3.4 2,4.1 3,5.3 2.6,5.4 1.2,4.9 0.8,2.9 0.7,2 1.3)"
+            "(4.0 2.0, 4.2 1.4, 4.8 1.9, 4.4 2.2, 4.0 2.0))", poly);
+
+    point_type p(4, 1);
+
+    boost::geometry::de9im::mask mask("T*F**F***"); // within
+    bool check = boost::geometry::relate(p, poly, mask);
```
barendgehrels: This looks very good!

doc/src/examples/algorithms/relation.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+    polygon_type poly;
+    boost::geometry::read_wkt(
+        "POLYGON((2 1.3,2.4 1.7,2.8 1.8,3.4 1.2,3.7 1.6,3.4 2,4.1 3,5.3 2.6,5.4 1.2,4.9 0.8,2.9 0.7,2 1.3)"
+            "(4.0 2.0, 4.2 1.4, 4.8 1.9, 4.4 2.2, 4.0 2.0))", poly);
+
+    point_type p(4, 1);
+
+    boost::geometry::de9im::matrix matrix = boost::geometry::relation(p, poly);
+    std::string code = matrix.str();
```
barendgehrels: This looks very good too.

Minor note:
.str() is compatible with std::ostringstream
Boost.Filesystem uses .string()

Don't know what to prefer.

include/boost/geometry/algorithms/detail/relate/de9im.hpp
```diff
@@ -0,0 +1,434 @@
...
+\details This matrix can be used to express spatial relations as defined in
+         Dimensionally Extended 9-Intersection Model.
+
+\qbk{[heading See also]}
+\qbk{* [link geometry.reference.algorithms.relation relation]}
+ */
+class matrix
+    : public detail::relate::matrix<3, 3>
+{
+#ifdef DOXYGEN_INVOKED
```
barendgehrels: What is this for? To generate documentation?

include/boost/geometry/algorithms/detail/relate/de9im.hpp
```diff
@@ -0,0 +1,434 @@
...
+\tparam EE Exterior/Exterior intersection mask element
+
+\qbk{[heading See also]}
+\qbk{* [link geometry.reference.algorithms.relate relate]}
+ */
+template
+<
+    char II = '*', char IB = '*', char IE = '*',
+    char BI = '*', char BB = '*', char BE = '*',
+    char EI = '*', char EB = '*', char EE = '*'
```
barendgehrels: What if people specify a non-recognized pattern? For example: 
static_mask<'A', 'B', 'C'>
do they get a compile time warning?

include/boost/geometry/algorithms/detail/relate/de9im.hpp
```diff
@@ -0,0 +1,434 @@
...
+
+// CROSSES
+// dim(G1) < dim(G2) - P/L P/A L/A
+template
+<
+    typename Geometry1,
+    typename Geometry2,
+    std::size_t Dim1 = geometry::topological_dimension<Geometry1>::value,
+    std::size_t Dim2 = geometry::topological_dimension<Geometry2>::value,
+    bool D1LessD2 = (Dim1 < Dim2)
```
barendgehrels: Looks good, but where is this (D1LessD2) used? Don't see any usage

include/boost/geometry/algorithms/crosses.hpp
```diff
@@ -30,7 +30,8 @@
 
 #include <boost/geometry/geometries/concepts/check.hpp>
 
-#include <boost/geometry/algorithms/detail/relate/relate.hpp>
+#include <boost/geometry/algorithms/relate.hpp>
+#include <boost/geometry/algorithms/detail/relate/relate_impl.hpp>
```
barendgehrels: Do we use the suffix _impl more often?
There is also an impl folder in projections, so it is not the only usage of this abbreviation
But is there not an alternative? Thought I saw already some correspondence about this going by but cannot find it in this PR

test/algorithms/relational_operations/relate/test_relate.hpp
```diff
@@ -67,6 +70,44 @@ std::string matrix_format(std::string const& matrix1, std::string const& matrix2
          + ( !matrix2.empty() ? " || " : "" ) + matrix2;
 }
 
+template <typename M>
+char get_ii(M const& m)
```
barendgehrels: inline is missing (it is a hpp)

test/algorithms/relational_operations/relate/test_relate.hpp
```diff
@@ -67,6 +70,44 @@ std::string matrix_format(std::string const& matrix1, std::string const& matrix2
...
+{
+    using bg::detail::relate::interior;
+    return m.template get<interior, interior>();
+}
+
+template <typename M>
+char get_ee(M const& m)
```
barendgehrels: same here and below

barendgehrels:
Thanks a lot - it looks very good.
I like the separate relation function

I'm OK with merging this.

include/boost/geometry/algorithms/detail/relate/de9im.hpp
awulkiew: Indeed, `util` would be a better place. I'll do it aside from this PR.

include/boost/geometry/algorithms/detail/relation/interface.hpp
awulkiew: Sorry, missed this one during refactoring.

test/algorithms/relational_operations/relate/relate_pointlike_xxx.cpp
awulkiew: `xxx` stands for "any geometry" so I'll change the name to `relate_pointlike_geometry`.

doc/src/examples/algorithms/relation.cpp
awulkiew: I picked `str()` because it is in the standard, and corresponds to `c_str()`.
We could have both but it's probably not required.

barendgehrels: One is indeed enough.
I'm OK with str()

include/boost/geometry/algorithms/detail/relate/de9im.hpp
awulkiew: Yes, it's for documentation. Doxygen is able to copy the docs of the members of the base class only if those members are virtual functions. I don't like the approach I've used but I don't see any other way to have the members in the documentation.

include/boost/geometry/algorithms/detail/relate/de9im.hpp
awulkiew: In general the result is undefined but in practice AFAIR other characters than `F`, `T` and [`0`, `9`] are treated as `*`.

include/boost/geometry/algorithms/crosses.hpp
awulkiew: It's sometimes used, yes. I have nothing against changing that name, previously it was `relate_base`. In this file there is a base class for a specialization/dispatched algorithm which could be implemented by falling back to relate. So it could e.g. be `algorithm_relate` or `relate_dispatch_base` but I don't have an idea which I think is the best. If you have an idea for a better name I'll change it.

include/boost/geometry/algorithms/detail/relate/de9im.hpp
barendgehrels: I think it is possible and not so difficult to give a compile time error.

Something like:
```
template <char C>
struct check_char : not_implemented < ... > {};

// Specialize what should compile (maybe give it an apply message)
template<> struct check_char<'*'> {} ;
template<> struct check_char<'T'> {} ;
template<> struct check_char<'F'> {} ;
```
and then in this struct:
```
check_char<II>::apply(); // with a static apply - but maybe just a declaration is fine too
check_char<IB>::apply();
```
etc

or something like that (I did not try it, but I ever did an experiment with a similar thing)

test/algorithms/relational_operations/relate/test_relate.hpp
awulkiew: It's for unit testing only. We don't use `inline` keyword in the test hpp files.

barendgehrels: Indeed. Just checked test_buffer and there it is never done...
In general I think a free function in any hpp should be inlined, so maybe I will change buffer too.
But as long as the hpp is used in one cpp at at time it should be no problem.

include/boost/geometry/algorithms/detail/relate/de9im.hpp
awulkiew: Of course it would be possible to generate a compiler error but do you think that it should be done? The same question is valid for run-time mask. Should the characters be checked in run-tme and an exception generated? Basically I wanted to avoid additional checks and exceptions, and to be consistent with run-time masks I decided to also not check the characters for compile-time masks. Do you think that the masks should be checked?

Actually implementation could be more generic because internally a `static_mask` uses `boost::mpl::vector_c` of chars.

Btw, AFAIR in DE9IM the specific dimensions numbers can be stored in masks too.

barendgehrels: > Do you think that the masks should be checked?
Of course... it was actually the first question I did have about the new interface. I doubted between suggesting an enum (because there you cannot specify unsupported options) or leave it as you proposed it. I like the proposal because it is clear, but I think the options should be checked at compiletime (if possible) and at runtime to avoid undefined behaviour.

Yes, using the vector_c functionality avoids checking them individually, agreed

awulkiew: Ok, another thing. Currently if there is not enough elements (either template parameters or run-time string/array elements) the mask is filled with stars (`*`). If there is too many elements in array passed into the run-time mask, the redundant elements are ommited. Should I keep this behavior or also throw an exception for too small and/or to great number of elements? So require always 9 elements or always lesser or equal number of elements than 9.

barendgehrels: Good point.

I think * is a very reasonable default, but I also would find it clear if there were no defaults... It's a de-9-im matrix, and it might be an unnoticed error if users fill it e.g. with 8 values instead of the supposed 9...
If so (no defaults), for runtime I think an exception is then appropriate indeed. Especially too much is indeed always a kind of error-situation.

Curious to other opinions.

awulkiew: Well, the last character corresponds to overlapping of Exterior/Exterior which will always be set to 2 for any 2d geometries combination (in general probably to the dimension of the input) so typically noone would check that. But I get your point.

An alternative to the exception could be a run-time assertion (for static_mask a compile-time assertion). I guess such input might be considered as a programming error.

Though indeed, in general an exception is thrown in such cases, e.g. std::invalid_argument in the STL.

awulkiew: I added checks for both types of mask. `static_mask` is checked using `BOOST_MPL_ASSERT_MSG`, `mask` ctors may throw a newly added `bg::invalid_input_exception`.

I decided to not remove the possibility to pass a shorter or longer string. It's because similar semantics is also allowed in `std::bitset`. If a smaller number of bits is passed the remaining ones are filled with zeros. It's also possible to pass a string containing more characters than the number of bits in `std::bitmap`. It's even possible to pass a starting position and the number of characters/bits which are used to fill the object, though this last option is not implemented for `de9im::mask` right now.

PR #289, opened 2015-05-04
New feature: envelope and expand for spherical equatorial coordinate system

mkaravel:
This PR implements (or re-implements) the algorithms `bg::envelope` and `bg::expand`.

As far as the old implementation is concerned, the new approach does use `bg::assign_inverse` as it might not be well-defined for user-defined number types.

Regarding the new implementation, the following are implemented for the spherical equatorial and possibly geographic coordinate systems (see lists below for which combinations support geographic coordinate systems; "spherical" below refers to "spherical equatorial"):

#### expand algorithm
* `expand(box, point)` (spherical and geographic)
* `expand(box, segment)` (spherical only)
* `expand(box, box)` (spherical and geographic)

#### envelope algorithm
* `envelope(point, mbr)` (spherical and geographic)
* `envelope(segment, mbr)` (spherical only)
* `envelope(box, mbr)` (spherical and geographic)
* `envelope(linestring, mbr)` (spherical only)
* `envelope(multilinestring, mbr)` (spherical only)
* `envelope(multipoint, mbr)` (spherical and geographic)

This PR depends on PR #271, as it requires the coordinate normalization functionality introduced in that PR.

**Note 1:** due to the changes introduced in this PR, and in particular how MBRs are computed on the sphere/spheroid, one of the distance unit tests fails. This is a known issue and will be addressed after this PR is reviewed and merged.

**Note 2:** what is still missing is the implementation of the `envelope` algorithms for areal geometries (rings, polygons and multipolygons). This is still under development and will either be added to this PR, or through a new PR at a later point.


include/boost/geometry/algorithms/detail/envelope/intersects_antimeridian.hpp
```diff
@@ -0,0 +1,103 @@
...
+                CoordinateType, Units
+            > constants;
+    
+        return
+            math::equals(math::abs(lat1), constants::max_latitude())
+            ||
+            math::equals(math::abs(lat2), constants::max_latitude())
+            ||
+            math::larger(math::abs(lon1 - lon2), constants::half_period());
+    }
```
awulkiew: Is `apply()` taking 4 coordinates used somewhere outside this struct? Is there a reason why the coordinates aren't normalized in this overload? This way it works slightly differently than the other 2 overloads which can be confusing. It should be consistent with the two, moved somewhere else as a separate (low-level) tool or probably most preferably made private.

include/boost/geometry/algorithms/detail/envelope/intersects_antimeridian.hpp
```diff
@@ -0,0 +1,103 @@
...
+                units_type, coordinate_type
+            >(lon1, lat1);
+
+        math::normalize_spheroidal_coordinates
+            <
+                units_type, coordinate_type
+            >(lon2, lat2);
+
+        return apply<units_type>(lon1, lat1, lon2, lat2);
+    }
```
awulkiew: Here `detail::indexed_point_view<>` could be used and point view objects passed into `apply()` taking 2 points. To not duplicate the implementation.

include/boost/geometry/algorithms/detail/envelope/linestring.hpp
```diff
@@ -0,0 +1,221 @@
...
+             ++seg_it, first = false)
+        {
+            typename point_type<Linear>::type p[2];
+            detail::assign_point_from_index<0>(*seg_it, p[0]);
+            detail::assign_point_from_index<1>(*seg_it, p[1]);
+
+            calculation_type lon1 = geometry::get<0>(p[0]);
+            calculation_type lat1 = geometry::get<1>(p[0]);
+            calculation_type lon2 = geometry::get<0>(p[1]);
+            calculation_type lat2 = geometry::get<1>(p[1]);
```
awulkiew: Could `detail::indexed_point_view` be used here instead?

include/boost/geometry/algorithms/detail/envelope/linestring.hpp
```diff
@@ -0,0 +1,221 @@
...
+                    linear_units_type
+                >(lon1, lat1);
+
+            math::normalize_spheroidal_coordinates
+                <
+                    linear_units_type
+                >(lon2, lat2);
+
+            math::convert_coordinates<linear_units_type, radian>(lon1, lat1);
+            math::convert_coordinates<linear_units_type, radian>(lon2, lat2);
```
awulkiew: Here we have the same pattern as above (in `intersects_antimeridian`):
1. create variables contain coordinates
2. normalize the coorinates
3. convert to radians (this isn't done above)

I'm guessing that this pattern can be used often. Would it be a good idea to have one well-defined function for this instead of several small ones?

Why does `convert_coordinates` was added? We already have `get_as_radian`. I'm guessing that `normalize_spheroidal_coordinates` are normalized using the native units (radians or degrees). Is it safe? Let's say the coordinates are normalized using degrees and then converted to radians, are the radians normalized? I'm guessing that in some cases this may not hold. I think that first the coordinates should be converted to radians and then normalized. Furthermore I think it should be done this way in all algorithms for consistency. Therefore we probably doesn't need those generic utils you added for conversion, even more, they can be harmful. We should be forced to convert the units always the same way in all functions. Most preferably we could have single one function doing it.

include/boost/geometry/algorithms/detail/envelope/linestring.hpp
```diff
@@ -0,0 +1,221 @@
...
+            <
+                radian, typename coordinate_system<Box>::type::units
+            >(lon_max, lat_max);
+
+        assign_values(mbr,
+                      boost::numeric_cast<box_coordinate_type>(lon_min),
+                      boost::numeric_cast<box_coordinate_type>(lat_min),
+                      boost::numeric_cast<box_coordinate_type>(lon_max),
+                      boost::numeric_cast<box_coordinate_type>(lat_max));
+    }
```
awulkiew: And here `set_from_radian` could be used, so a tool which we already have and was exactly designed for this purpose. The code above could use `detail::indexed_point_view` and `set_from_radian`.

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
```diff
@@ -0,0 +1,362 @@
...
+{
+private:
+    template <typename T>
+    struct lon_lat_pair
+        : std::pair<T, T>
+    {
+        lon_lat_pair(T const& first, T const& second)
+            : std::pair<T, T>(first, second)
+        {}
+    };
```
awulkiew: I'm wondering, why was this struct added? Do you think that using `std::pair` is not clear enough or is there some other reason?

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
```diff
@@ -0,0 +1,362 @@
...
+             ++it)
+        {
+            coordinate_type lon = geometry::get<0>(*it);
+            coordinate_type lat = geometry::get<1>(*it);
+
+            math::normalize_spheroidal_coordinates
+                <
+                    typename coordinate_system<MultiPoint>::type::units,
+                    coordinate_type
+                >(lon, lat);
```
awulkiew: The same pattern again, get coordinates and normalize.

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
```diff
@@ -0,0 +1,362 @@
...
+
+            if (math::equals(lat, Constants::min_latitude()))
+            {
+                has_south_pole = true;
+            }
+            else if (math::equals(lat, Constants::max_latitude()))
+            {
+                has_north_pole = true;
+            }
+            else
```
awulkiew: Can the result of those comparisons be different for degrees and radians? I'm wondering, maybe we should always use radians internally to be consistent with other functions always converting degrees to radians?

include/boost/geometry/algorithms/detail/envelope/point.hpp
```diff
@@ -0,0 +1,106 @@
...
+
+        box_coordinate_type lon = geometry::get<0>(box_point);
+        box_coordinate_type lat = geometry::get<1>(box_point);
+
+        math::normalize_spheroidal_coordinates
+            <
+                typename coordinate_system<Box>::type::units
+            >(lon, lat);
+
+        assign_values(mbr, lon, lat, lon, lat);
```
awulkiew: The same pattern again. Maybe we should have `bg::detail::normalize()` or `bg::detail::return_normalized()` algorithm working for Points, Boxes and Segments?
This code could be:

    detail::indexed_point_view<Box, min_coordinate> p_min(mbr);
    detail::indexed_point_view<Box, max_coordinate> p_max(mbr);
    bg::convert(bg::return_normalized<box_point_type>(point), p_min);
    bg::convert(bg::return_normalized<box_point_type>(point), p_max);

or something similar.

Or maybe `convert()` could be modified this way, to convert point into box in the right way for all coordinate systems?

include/boost/geometry/algorithms/detail/envelope/point.hpp
```diff
@@ -0,0 +1,106 @@
...
+{
+
+#ifndef DOXYGEN_NO_DETAIL
+namespace detail { namespace envelope
+{
+
+struct envelope_point_on_spheroid
+{
+    template<typename Point, typename Box>
+    static inline void apply(Point const& point, Box& mbr)
```
awulkiew: Should we use `mbr` name? It isn't widely used in the library, isn't it? Usually we use the name `box`. Besides a type is called Box, all types and variables below contains a word `box`.

include/boost/geometry/algorithms/detail/envelope/implementation.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+
+struct exterior_ring_expand_policy
+{
+    template <typename Box, typename Geometry>
+    static inline void apply(Box& mbr, Geometry const& geometry)
+    {
+        Box ring_mbr;
+        envelope_range<>::apply(exterior_ring(geometry), ring_mbr);
+        geometry::expand(mbr, ring_mbr);
```
awulkiew: Are the expand policies really needed? AFAIU a single call to `expand()` for a Polygon should do exactly the same thing. The same for `default_expand_policy` below. It's only purpose is to run `expand()`. Or is there some specific case where a single call of `expand()` wouldn't work?

awulkiew: Please ignore the above comment. I see now that what is needed is to expand a box with an envelope of a Geometry. But since this code implements the `envelope()` this algorithm can't be used like a user would do this:

    expand(box, return_envelope<Box>(geometry));

Btw, shouldn't it be called `polygon_expand_policy` since it must take a Polygon?

include/boost/geometry/algorithms/detail/envelope/range_of_boxes.hpp
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        coordinate_type lon_min(0), lon_max(0);
+        envelope_range_of_longitudes
+            <
+                units_type
+            >::apply(intervals, lon_min, lon_max);
+
+        // TODO(?): convert units?
+        assign_values(mbr,
```
awulkiew: TODOs here and above.

The input Boxes can be different than the output (units, coordinate_type) so I'm guessing that the conversion should be made at some point. Currently the coordinate_type and units of the input Boxes is used but I'm not sure if it wouldn't be better to convert the input coordinates to the output coordinate_type and use it for the intervals. Or to be consistent with other algorithms always use radians and the most precise type?

But it depends, will this algorithm be publicly available? Or is it only for internal use e.g. in partition and rtree?

include/boost/geometry/algorithms/detail/envelope/intersects_antimeridian.hpp
mkaravel: Actually, the code in this file is not used within the rest of code currently submitted in this PR.
It is used right now in my prototype (not pushed) implementation for rings, and was accidentally pushed as part of this PR.
Please disregard it for now, and we can come back later to it.

include/boost/geometry/algorithms/detail/envelope/intersects_antimeridian.hpp
mkaravel: Will do. Thanks for the tip.

include/boost/geometry/algorithms/detail/envelope/linestring.hpp
mkaravel: I would say yes. I do not really see why it should not work.
I will try it and get back to you.

include/boost/geometry/algorithms/detail/envelope/linestring.hpp
mkaravel: I will get back to these comments later.

include/boost/geometry/algorithms/detail/envelope/linestring.hpp
mkaravel: I think the reason I did not use `set_from_radian` is that:

1. I think it does not work with boxes (it takes only one index template parameter, and most probably we should change that...)
2. I think it does not do any conversion *or* the conversion done is implicit, which I wanted to avoid.

In any case I will check again if it can/should be done the way you suggest,

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
mkaravel: I think the code is more readable with a properly named struct for the pair, but this is just my personal point of view.

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
mkaravel: I do not agree with using radians all over the place.
My personal preference would be to stick to the original coordinate type if possible, and only convert to radians if needed.
For example here, the way I have implemented it, it works perfectly well with points whose coordinate type is an integer and the units are degrees: all computations are done with integers.

include/boost/geometry/algorithms/detail/envelope/point.hpp
mkaravel: The old code was using `mbr` in several places, and this is what inspired me.
I like the name `mbr` as it clearly indicates that the output is a minimum bounding rectangle.
I can change it if you want, though.

include/boost/geometry/algorithms/detail/envelope/point.hpp
mkaravel: I wanted to avoid the creation of new points, which is why I consistently work with the coordinates.
I do agree though that `bg::convert` may have to be changed and basically use that instead.

include/boost/geometry/algorithms/detail/envelope/range_of_boxes.hpp
mkaravel: This is meant to be part of the internals and not exposed.
We can discuss if we really need to expose it.
Given its current usage, no conversion is necessary at this point; it happens somewhere else.

mkaravel: As I mentioned in another comment, I would like to avoid conversions to radians if not needed. This is certainly one case where we do not need the conversion to radians for the algorithm to work, so I prefer to keep/use the input coordinate type.

include/boost/geometry/algorithms/detail/envelope/linestring.hpp
mkaravel: I will try to answer your comments/questions one-by-one.

*I'm guessing that this pattern can be used often. Would it be a good idea to have one well-defined function for this instead of several small ones?*

I will think about it, and yes it would make sense. I would even argue that the correct place to put all these is in `bg::convert()`.

*Why does `convert_coordinates` was added? We already have `get_as_radian`.*

I want to first normalize and then convert. I find this more robust.

*I'm guessing that `normalize_spheroidal_coordinates` are normalized using the native units (radians or degrees).*

Correct.

*Is it safe?*

Why do you think it would not be safe? What kind of safety do you refer to?

*Let's say the coordinates are normalized using degrees and then converted to radians, are the radians normalized? I'm guessing that in some cases this may not hold.*

Yes, the radians are normalized. Normalization has nothing to do with the specific units you use, and you can express it in either degrees or radians (or any other units, like seconds of degrees). Normalization basically tells you that latitudes go from `-period/4` to `+period/4` and longitudes go from `-period/2` to `period/2`. The actual value for `period` depends on the units you choose.

*I think that first the coordinates should be converted to radians and then normalized.*

Why do you think so? When you convert to radians first you immediately enter the floating-point world, even for integral coordinates. I would argue that you should delay that as long as possible.

*Furthermore I think it should be done this way in all algorithms for consistency. Therefore we probably doesn't need those generic utils you added for conversion, even more, they can be harmful. We should be forced to convert the units always the same way in all functions. Most preferably we could have single one function doing it.*

So far we have been converting to radians because we needed to feed coordinates to trigonometric functions; this is absolutely understandable. But if you do not need to use trigonometric functions why bother converting?
I do not follow you regarding the *harmful* part. I agree that we could/should encapsulate conversion into a single function, but I do not (not yet at least) get what is problematic with my approach.

include/boost/geometry/algorithms/detail/envelope/implementation.hpp
mkaravel: It seemed to me more natural to name it according to how the geometry is used in the policy (taking its exterior ring).

include/boost/geometry/algorithms/detail/envelope/range_of_boxes.hpp
mkaravel: The same comments apply to both `TODOs` in the code: if we are to expose the algorithm implemented in `envelope_range_of_boxes` further, then both normalization and conversion might be needed explicitly here. as it is currently part of the internal implementation, the `TODOs` are not really needed.

mkaravel:
@awulkiew Thanks for your comments. Looking forward to your replies and further suggestions.

include/boost/geometry/algorithms/detail/envelope/implementation.hpp
awulkiew: Ok, I get it. For me personally it isn't important how the expansion is implemented. It uses only the exterior ring because this is how we generally implement things for Polygon but it might do something else. It'd still expand a boxy using an envelope of a Polygon.

There is also `envelope_polygon` which also internally analyses only the exterior ring of a Polygon and `envelope_range_expand_policy` which expands a box using an envelope of a range. Well, the last name can also be confusing because this policy works on a Range of Points, not Range of envelopes (so boxes?).

include/boost/geometry/algorithms/detail/envelope/linestring.hpp
awulkiew: Indeed it should work better in the case of integral degrees, but AFAIU only in this case. For FP degrees normalized and converted to FP radians it shouldn't change anything.

My point is, it probably isn't possible to represent true 90 as it certainly isn't possible to represent true pi/2. And while converting between both units we can loose information. It's just a guess but in some cases, for some input in FP degrees or some user-defined type the value after conversion could be outside the range e.g. [-pi/2, pi/2] for latitude. I just think it'd be safer to normalize exactly those values that can be passed into some trigonometric functions. The radians are exactly those values. Maybe the radians should also be normalized (2nd time) or maybe only the integral degrees should be normalized at the beginning (and FP radians always normalized at the end)?

include/boost/geometry/algorithms/detail/envelope/linestring.hpp
awulkiew: This is why I proposed to use `indexed_point_view`. I just don't like the idea of implementation of several tools for the same thing. There should be one method widely used across the whole library.

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
awulkiew: Yes, indeed using the original coordinates wherever possible seems to be a better idea.

include/boost/geometry/algorithms/detail/envelope/point.hpp
awulkiew: Ok, I don't have a strong opinion in this case so let's see what others will say.

include/boost/geometry/algorithms/detail/envelope/point.hpp
awulkiew: Indeed, the example above uses 2 temporary points. But only one Point could be used and reused, so also 2 coordinates (similarly to your code). Or do you fear that a user-defined Point could contain some additional members? Anyway, how about this:

    detail::indexed_point_view<Box, min_coordinate> p_min(mbr);
    detail::indexed_point_view<Box, max_coordinate> p_max(mbr);
    bg::convert(bg::return_normalized<box_point_type>(point), p_min);
    bg::convert(p_min, p_max);

which directly initializes Box's min corner and then copies it to max. So no temporaries at all.

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
barendgehrels: Agreed, a std::pair is (on longer term) often confusing

include/boost/geometry/algorithms/detail/envelope/point.hpp
barendgehrels: I don't object to mbr
But the OGC term is "envelope"

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
```diff
@@ -0,0 +1,362 @@
...
+        typedef typename coordinate_type<MultiPoint>::type coordinate_type;
+
+        // analyze point coordinates:
+        // (1) normalize point coordinates
+        // (2) check if any point is the north or the south pole
+        // (3) put all non-pole points in a container
+        //
+        // notice that at this point in the algorithm, we have at
+        // least two points on the spheroid
+        has_south_pole = has_north_pole = false;
```
barendgehrels: We normally don't do onlined multiple assignments in the library, I prefer to split

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
```diff
@@ -0,0 +1,362 @@
...
+            return dispatch::envelope
+                <
+                    typename boost::range_value<MultiPoint>::type
+                >::apply(range::front(multipoint), mbr);
+        }
+
+        // analyze the coordinates and put the non-pole ones in the
+        // coordinates vector
+        std::vector<pair_type> coordinates;
+        bool has_north_pole(false), has_south_pole(false);
```
barendgehrels: Please assign as bool har_north_pole = false;

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
```diff
@@ -0,0 +1,362 @@
...
+        analyze_point_coordinates<constants>(multipoint,
+                                             has_south_pole, has_north_pole,
+                                             std::back_inserter(coordinates));
+
+        coordinate_type lon_min, lat_min, lon_max, lat_max;
+        if (coordinates.size() == 1)
+        {
+            // we have one non-pole point and at least one pole point
+            lon_min = lon_max = coordinates.begin()->first;
+            lat_min = (has_south_pole)
```
barendgehrels: brackets are redundant

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
```diff
@@ -0,0 +1,362 @@
...
+
+        analyze_point_coordinates<constants>(multipoint,
+                                             has_south_pole, has_north_pole,
+                                             std::back_inserter(coordinates));
+
+        coordinate_type lon_min, lat_min, lon_max, lat_max;
+        if (coordinates.size() == 1)
+        {
+            // we have one non-pole point and at least one pole point
+            lon_min = lon_max = coordinates.begin()->first;
```
barendgehrels: As above please avoid = .. =

barendgehrels: Now that you introduced a lon_lat_pair, we should not use first/second but lon/lat
Why it is actually not a geometry point_type that is used here?

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
```diff
@@ -0,0 +1,362 @@
...
+            lat_min = (has_south_pole)
+                ? constants::min_latitude()
+                : constants::max_latitude();
+            lat_max = (has_north_pole)
+                ? constants::max_latitude()
+                : constants::min_latitude();
+        }
+        else if (coordinates.size() == 0)
```
barendgehrels: (minor) coordinates.empty()

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
```diff
@@ -0,0 +1,362 @@
...
+                                    boost::end(range),
+                                    lon_less);
+
+        lon_min = min_max_longitudes.first->first;
+        lon_max = min_max_longitudes.second->first;
+
+        // if the longitude span is "large" compute the true maximum gap
+        if (math::larger(lon_max - lon_min, Constants::half_period()))
+        {
+            std::sort(boost::begin(range), boost::end(range), lon_less);
```
barendgehrels: Is it possible to do calculate the max gap without sorting?

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
```diff
@@ -0,0 +1,362 @@
...
+
+            CoordinateType max_gap_left(0), max_gap_right(0);
```
barendgehrels: = 0, = 0

mkaravel: I am responding to this, but it also applies to other similar comments: AFAIR,
```
some_type x = 0;
```
and
```
some_type x(0);
```
are treated in exactly the same way by the C++ standard: in both cases the constructor is called with the value `0` as argument.
Is your comment related to BG's coding style or is it something else?

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
mkaravel: Is that a matter of coding style, or do you have in mind something else?

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
mkaravel: Is your question really about sorting, or whether we can do it in linear time as opposed to linearithmic time?

I will try to answer both:

1. Yes we can do away without sorting. For example, we can use the sweeping framework I have developed as part of this PR. Instead of sorting all longitudes, I can push them in a priority queue and successively pop them, in order to compute the maximum gap. But I am not really sure we gain something using the sweep approach as opposed to sorting.
2. Yes, it is possible to compute the maximum gap in linear time. There is a bin-based algorithm, that also requires the use of the `floor()` function. You may take a look at [these slides](http://cgm.cs.mcgill.ca/~godfried/teaching/dm-reading-assignments/Maximum-Gap-Problem.pdf), explaining this bin-based linear time algorithm. I am not convinced that in practice it would be faster. I can certainly implement it, either now or at a later time and do benchmarks.

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
barendgehrels: It is coding style

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
barendgehrels: Thanks for explaining - I'm OK by doing it like you did and benchmark later

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
barendgehrels: There have been quite some correspondence about this.
The BG style was: CoordinateType max_gap_left = CoordinateType()
That can be necessary in template environments (to support big numeric types) but not always.
But using elementary types as if it is a constructor (bool, int) suddenly popped up once.
What does C++11, C++14, C++17 prefer?

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
```diff
@@ -0,0 +1,362 @@
...
+            CoordinateType max_gap
+                = maximum_gap(range, max_gap_left, max_gap_right);
```
barendgehrels: E.g. here you also don't write:
```CoordinateType max_gap ( maximum_gap ( ... ) ) ```
Which would be indeed confusing. So I prefer (personally) to don't do it for 0 neither

File: include/boost/geometry/algorithms/detail/envelope/point.hpp (hunk 9 above)
Comment (barendgehrels) [inline include/boost/geometry/algorithms/detail/envelope/point.hpp]: I agree with making and using such a xxx normalized for this

File: include/boost/geometry/algorithms/detail/envelope/implementation.hpp [hunk 21]
```diff
@@ -0,0 +1,104 @@
...
+struct envelope<Polygon, polygon_tag>
+    : detail::envelope::envelope_polygon
+{};
+
+
+template <typename MultiPolygon>
+struct envelope<MultiPolygon, multi_polygon_tag>
+    : detail::envelope::envelope_range
+        <
+            detail::envelope::exterior_ring_expand_policy
```
barendgehrels: Looks good

include/boost/geometry/algorithms/detail/envelope/interface.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+#include <boost/geometry/geometries/concepts/check.hpp>
+
+#include <boost/geometry/algorithms/dispatch/envelope.hpp>
+
+
+namespace boost { namespace geometry
+{
+
+namespace resolve_variant {
```
barendgehrels: curly brace goes next

include/boost/geometry/algorithms/detail/sweep.hpp
```diff
@@ -0,0 +1,87 @@
...
+    typename InitializationVisitor,
+    typename EventVisitor
+>
+inline void sweep(Range const& range, PriorityQueue& queue,
+                  InitializationVisitor& initialization_visitor,
+                  EventVisitor& event_visitor)
+{
+    sweep(range, queue, initialization_visitor, event_visitor,
+          detail::sweep::no_interrupt_policy());
+}
```
barendgehrels: This is in a file detail, but in namespace boost::geometry
Is that intentional? Can it in the end be called by an end-user?

Maybe the sweep requires more documentation?

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
mkaravel: I do not like the initialization with the default constructor you mention in another comment.

I am okay with either solution (using `=` or pass value to constructor). I guess from the readability point of view using `=` is better, and I would argue that this should be part of our coding rules.

include/boost/geometry/algorithms/detail/sweep.hpp
mkaravel: Yes, it is intentional. `bg::sweep()` is something like `bg::partition()` in my mind. At some point I would like to document it, but not yet.

barendgehrels: OK, good

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
barendgehrels: I would like to see it in our coding rules too. I'm OK with the good old standard =

mkaravel:
@barendgehrels Thank you for your comments.

I will update the code and ask for a second round of review once I am done.

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
awulkiew: Value initialization as you wrote it (with small differences between standards) means in this case: initialize with 0 for built-in types and call default ctor for user-defined types. This is a wrong approach since the user-defined type could implement the default ctor not initializing the value to 0. In particular this is true for `ttmath_big`. It we want to initialize with `0` we should explicitly initialize with 0, in the whole library. Just like we're doing for other numbers 2, 4, 5243, etc. Yes, we're already initializing explicitly. Why do you think 0 is different?

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
awulkiew: I'm ok with both since it's valid C++. I agree that using assignment is probably more clear since the other version looks like a function declaration. And in some cases it can be treated as such by standard-conformant compiler, then it requires additional parentheses, etc. On the other hand by using assignment operator we dissalow to use user-defined classes with explicit copy constructor. Though such classes won't be usable in Geometry anyway right now since various methods of initializations are mixed across the library.

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
awulkiew: Indeed, I agree that initialization of multiple variables in separate lines is more clear.

include/boost/geometry/algorithms/detail/envelope/intersects_antimeridian.hpp
mkaravel: Done.

include/boost/geometry/algorithms/detail/envelope/point.hpp
mkaravel: I prefer to stick with `mbr`. I find it very descriptive.

include/boost/geometry/algorithms/detail/envelope/range_of_boxes.hpp
mkaravel: The `TODOs` are gone and replaced by relevant comments.

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
mkaravel: Fixed.

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
mkaravel: Fixed.

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
mkaravel: The `lon_lat_pair` is now gone. I am using points instead.
The code should be more readable now.

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
mkaravel: Done.

include/boost/geometry/algorithms/detail/envelope/multipoint.hpp
mkaravel: Fixed.

mkaravel:
I have tried to address all comments made in relation to this PR.
Major changes to look at:

1. I have introduced a new algorithm `normalize()`/`return_normalized<>()` in `detail` namespace. This algoriothm takes care of the normalization. `normalize()` works for points and boxes right now.
2. I have introduced another new algorithm `convert_units()`; it is responsible for changing the units of a point or a box from degrees to radians or vice versa.
3. Both `envelope` and `expand` have been re-implemented to use these two algorithms. The code should be clearer now (I have to admit I like it much better).

Things to worry about: `envelope` and `expand` may not yet work with points in the spherical equatorial coordinate system that have dimension 3 (longitude, latitude and height for example). I will do this as part of another PR. Having said that, `equals` now works correctly with such points (this is partly the reason for the major changes in commit ba93049 and for the additional test cases added in commit 5ecc576).

Finally, I have not removed the latitude normalization code as suggested in the comments of PR #271, but rather introduced the macro `BOOST_GEOMETRY_NORMALIZE_LATITUDE`. I find this a better solution, in the sense that some user may eventually make use of it. In any case, since the latitude normalization code is guarded by this macro, it does not really hurt to have it in there.

At this point I would like to ask for a second round of review.

include/boost/geometry/algorithms/detail/normalize.hpp
```diff
@@ -0,0 +1,294 @@
...
+    template <typename GeometryIn, typename GeometryOut>
+    static inline void apply(GeometryIn const&, GeometryOut&)
+    {
+    }
+};
+
+
+template <std::size_t Dimension, std::size_t DimensionCount>
+struct assign_loop
+{
```
awulkiew: This functionality could be moved (now or later) somewhere into the `detail::assign` and/or changed into a tool for assigning some subset of coordinates. I recal that I've seen something like this already in the library but I was unable to find it now.

include/boost/geometry/algorithms/detail/convert_units.hpp
```diff
@@ -0,0 +1,177 @@
...
+} // namespace dispatch
+#endif // DOXYGEN_NO_DISPATCH
+
+#ifndef DOXYGEN_NO_DETAIL
+namespace detail
+{
+
+
+template <typename UnitsFrom, typename UnitsTo, typename Geometry>
+inline void convert_units(Geometry& geometry)
```
awulkiew: Typically in Geometry similar functions (`convert`, `transform`) has different semantics, i.e. they take two parameters - input and output Geometries. In general the current use case seems to enforce "erroreous" behavior. AFAIU currently this function can be used to convert the coordinates of some Geometry from one unit to another. This means that at some point the new coordinates are calculated and in some cases this geometry contains coordinates in wrong units. Isn't it confusing? Wouldn't it be more clear if a geometry with specific CS and units was used internally and then at the end converted to the output Geometry if needed?

include/boost/geometry/algorithms/detail/normalize.hpp
mkaravel: Having a general tool for assigning a subset of the coordinates also crossed my mind (as a tool that I would like to have available). I agree with you, it would be nice to have such a tool.
I suggest postponing the move of `assign_loop` to `detail::assign` for another PR, and focus right now on the design of this PR.

include/boost/geometry/algorithms/detail/convert_units.hpp
mkaravel: I do not fully understand why the behavior is erroneous. I have a geometry whose coordinates were computed and assigned somehow. Now I want to change the units to something else (because the units were wrong in the first place, or because I want to convert this geometry to another one with different units). This is what `convert_units()` does.

Even if we follow your suggestion and use geometries with specific units internally (which I do not want to force actually), we would still need a tool like `convert_units()`, since `convert()` does not change units. What I could use is `bg::transform()`. I will investigate this modify the code appropriately.

awulkiew: Yes, it's also my understanding that we need a tool other than `convert()`. I'm complaining about the name, parameters and semantics in general. In Geometry functions that converts or transforms one geometry into another takes 2 parameters. This function corrects the geometry, not converts (in a Geometry sense) so a better name would be `correct_units()`. Then the 2 additional template parameters defining the units doesn't fit. At least one of them isn't needed since the Geometry is known. But for now we could leave it as it is since I don't have a better idea right now, it's a detail after all. Or do you have?

mkaravel: I am trying `bg::transform()`. It seems to work right now, with the "drawback" that I have to define some intermediate helper geometries (that have the correct units).

I will post more once I am done testing the transform approach.

mkaravel:
I have removed the `convert_units<>` utility; the change in coordinate system units is now done using `bg::transform()`. See commit 144fca3 for the details.

include/boost/geometry/geometries/helper_geometry.hpp
```diff
@@ -0,0 +1,154 @@
...
+struct default_units
+{
+    typedef typename coordinate_system<Geometry>::type::units type;
+};
+
+template <typename Geometry>
+struct default_units<Geometry, cartesian_tag>
+{
+    typedef radian type;
+};
```
awulkiew: Please put a comment here explaining why is this specialization defined like this.

mkaravel: Done. See commit b0290dd.

awulkiew:
Thanks! I'm ok with merging.

mkaravel:
@barendgehrels Are you okay with merging too?

barendgehrels:
Sure!

PR #292, opened 2015-05-06
Fix degree_radian_converter

mkaravel:
In this PR I modify the implementation of `degree_radian_converter` to use conversion constants that are consistent with the coordinate type of the geometry passed to the converter.

This is important for keeping the accuracy of user-defined number types (e.g., `ttmath`).

### DIFF
--- a/include/boost/geometry/util/math.hpp
+++ b/include/boost/geometry/util/math.hpp
@@ -493,9 +493,20 @@ inline bool larger(T1 const& a, T2 const& b)
 }
 
 
+template <typename T>
+T d2r()
+{
+    static T const conversion_coefficient = geometry::math::pi<T>() / T(180.0);
+    return conversion_coefficient;
+}
 
-double const d2r = geometry::math::pi<double>() / 180.0;
-double const r2d = 1.0 / d2r;
+template <typename T>
+T r2d()
+{
+    static T const conversion_coefficient = T(180.0) / geometry::math::pi<T>();
+    return conversion_coefficient;
+}
+
 
 /*!
     \brief Calculates the haversine of an angle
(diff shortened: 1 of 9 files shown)


awulkiew:
If that's really needed then I'm guessing it should be changed consistently in the whole library consistently. Shouldn't we rather replace the constants with functions?

    geometry::math::d2r<T>();

mkaravel:
I did a `grep` in `include/boost/geometry` and there are very few places where `math::d2r` and `math::r2d` are used, which is good.

I agree with you. I definitely prefer having functions rather than constants.

Let's wait a bit for other opinions and I would be more than happy to implement `math::d2r<T>()` and `math::r2d<T>()` and change the current use cases within the library with these new functions.

mkaravel:
There are also a few use cases in units tests, which I can also change easily.

mkaravel:
* I have replaced the constants `math::r2d` and `math::d2r` by functions (see commit 1deb1e2).
* I have changed the implementation of the `degree_radian_converter` to use these free functions (see commit 30f9805).
* I have replaced the constants in transform strategies by the corresponding free functions; in these cases I make sure that the type used for the conversion constants is promoted to a floating point number (see commit fbc568e).
* I have changed all other occurrences (in debug code and units tests) of the constants with the corresponding free functions (see commit 535a704).

awulkiew:
Thanks! I'm ok with this change but let's wait for Barend and the others :)

barendgehrels:
I'm OK with merging, thanks for this.
I will (on term) update the projections (on places where similar conversion is done) to make use of this

include/boost/geometry/util/math.hpp
```diff
@@ -493,9 +493,20 @@ inline bool larger(T1 const& a, T2 const& b)
 }
 
 
+template <typename T>
+T d2r()
```
barendgehrels: Oops, inline is missing!
Here and in r2d

mkaravel:
I just created a PR for this: #296. Thanks for spotting this!!!

PR #293, opened 2015-05-11
[projections] fix project_inverse_transformer::apply

jeremy-murphy:
The call to point_to_point was using the wrong namespace and function call.


awulkiew:
Thanks!

PR #297, opened 2015-05-19
Buffer fixes.

awulkiew:
The intention of this PR is to:
- provide a temporary fix for msvc-8.0 caused by the use of Multiprecision. It replaces Multiprecision `cpp_int` with manually implemented solution.
- suppress some compiler warnings.

### DIFF
--- a/include/boost/geometry/strategies/cartesian/side_of_intersection.hpp
+++ b/include/boost/geometry/strategies/cartesian/side_of_intersection.hpp
@@ -52,6 +52,8 @@ struct integral_pow<T, N, 0>
 
 // A tool for multiplication of integers avoiding overflow
 // It's a temporary workaround until we can use Multiprecision
+// The algorithm is based on Karatsuba algorithm
+// see: http://en.wikipedia.org/wiki/Karatsuba_algorithm
 template <typename T>
 struct multiplicable_integral
 {
@@ -63,15 +65,15 @@ struct multiplicable_integral
     static const std::size_t bits = CHAR_BIT * sizeof(T);
     static const std::size_t half_bits = bits / 2;
     typedef typename boost::make_unsigned<T>::type unsigned_type;
-    static const unsigned_type half_base
+    static const unsigned_type base
         = integral_pow<unsigned_type, 2, half_bits>::value;
 
-    int sign;
-    unsigned_type ms;
-    unsigned_type ls;
+    int m_sign;
+    unsigned_type m_ms;
+    unsigned_type m_ls;
 
-    multiplicable_integral(int sign_, unsigned_type ms_, unsigned_type ls_)
-        : sign(sign_), ms(ms_), ls(ls_)
+    multiplicable_integral(int sign, unsigned_type ms, unsigned_type ls)
+        : m_sign(sign), m_ms(ms), m_ls(ls)
     {}
 
     explicit multiplicable_integral(T const& val)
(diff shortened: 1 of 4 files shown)


include/boost/geometry/strategies/cartesian/side_of_intersection.hpp
```diff
@@ -27,6 +35,104 @@ namespace boost { namespace geometry
 namespace strategy { namespace side
 {
 
+namespace detail
+{
+
+template <typename T, T N, std::size_t P>
+struct integral_pow
```
mkaravel: This is definitely not the best implementation of integral powers.
I will try to come up with a better one.
The problem is on the number of instantiations we have to make; it is linear in the power, whereas I believe it should be possible to do it with a logarithmic number.

I am not sure it is that important though for the purposes of this fix.

include/boost/geometry/strategies/cartesian/side_of_intersection.hpp
```diff
@@ -27,6 +35,104 @@ namespace boost { namespace geometry
...
+    // This tool can't be used for non-integral coordinate types
+    // also sign_of_product and sign_of_compare functions must be modified
+    // (comparsons and multiplication)
+    BOOST_STATIC_ASSERT(boost::is_integral<T>::value);
+
+    static const std::size_t bits = CHAR_BIT * sizeof(T);
+    static const std::size_t half_bits = bits / 2;
+    typedef typename boost::make_unsigned<T>::type unsigned_type;
+    static const unsigned_type half_base
+        = integral_pow<unsigned_type, 2, half_bits>::value;
```
mkaravel: This looks like `2^half_bits` which is not the same as `(2^bits)/2`. I find the name a bit misleading.

include/boost/geometry/strategies/cartesian/side_of_intersection.hpp
```diff
@@ -27,6 +35,104 @@ namespace boost { namespace geometry
...
+
+    int sign;
+    unsigned_type ms;
+    unsigned_type ls;
```
mkaravel: These do not follow our coding rules. They should be `m_sign`, `m_ms`, etc.

include/boost/geometry/strategies/cartesian/side_of_intersection.hpp
```diff
@@ -27,6 +35,104 @@ namespace boost { namespace geometry
...
+
+    multiplicable_integral(int sign_, unsigned_type ms_, unsigned_type ls_)
+        : sign(sign_), ms(ms_), ls(ls_)
+    {}
```
mkaravel: See comment about coding rules fir local variables.
I do not like the `_` at the end of the names. We rarely/never use them AFAIU.

include/boost/geometry/strategies/cartesian/side_of_intersection.hpp
```diff
@@ -27,6 +35,104 @@ namespace boost { namespace geometry
...
+
+    explicit multiplicable_integral(T const& val)
+    {
+        unsigned_type val_u = val > 0 ?
+                                unsigned_type(val)
+                              : val == (std::numeric_limits<T>::min)() ?
+                                  unsigned_type((std::numeric_limits<T>::max)()) + 1
```
mkaravel: Is that part of the boost coding rules? I know that `std::min` and `std::max` need to go inside parentheses, but this is the same here.

include/boost/geometry/strategies/cartesian/side_of_intersection.hpp
```diff
@@ -27,6 +35,104 @@ namespace boost { namespace geometry
...
+    static const T value = N * integral_pow<T, N, P-1>::value;
+};
+
+template <typename T, T N>
+struct integral_pow<T, N, 0>
+{
+    static const T value = 1;
+};
+
+// A tool for multiplication of integers avoiding overflow
```
mkaravel: Any reference where the workaround is coming from?

include/boost/geometry/strategies/cartesian/side_of_intersection.hpp
```diff
@@ -27,6 +35,104 @@ namespace boost { namespace geometry
...
+    }
+
+    friend bool operator<(multiplicable_integral const& a,
+                          multiplicable_integral const& b)
+    {
+        if ( a.sign == b.sign )
+        {
+            bool u_less = a.ms < b.ms
+                      || (a.ms == b.ms && b.ls < b.ls);
+            return a.sign > 0 ? u_less : !u_less;
```
mkaravel: Put a space after `!`.

mkaravel:
Without knowing where the algorithm comes from, it looks good.
In order to verify that the algorithm does indeed work, I would need this reference.

The class `multiplicative_integral` should probably go in a different file in `boost/geometry/util`.

include/boost/geometry/strategies/cartesian/side_of_intersection.hpp
awulkiew: If you're asking about `(std::numeric_limits<T>::min)` then the answer is yes, the reason is the same, to avoid macro substitution.

barendgehrels:
@mkaravel Thanks for your review, I agree with your points (also go into util)
@awulkiew  Thanks for this, I'm OK with merging if comments are addressed

awulkiew:
The algorithm is adapted version of Karatsuba multiplication (http://en.wikipedia.org/wiki/Karatsuba_algorithm). I'll add a comment about it.

The results can be verified after defining `BOOST_GEOMETRY_SIDE_OF_INTERSECTION_DEBUG`. Then the same values are calculated using Multiprecision and run-time assertions comparing the results are enabled.

I don't think that it should go into util as it's not intended to be used in other parts of the library, at least in the current form. I fear that if I moved it into the util then we'd forget about it and it'd live there even after possible switching to Multiprecision. Or do you think it could be used somewhere else?

include/boost/geometry/strategies/cartesian/side_of_intersection.hpp
awulkiew: Yes, I think you're right that it should be possible to implement a version with lesser complexity. I also agree that this is not important as long as we don't plan to use it anywhere else in the library. If needed it could be moved e.g. to `util` and refactored.

include/boost/geometry/strategies/cartesian/side_of_intersection.hpp
awulkiew: I renamed it simply to `base` since this is a base of numbers used in Karatsuba algorithm.

awulkiew:
Ok, thanks for the review. I'll merge it to allow testing of the workaround and address any additional issues later.

include/boost/geometry/strategies/cartesian/side_of_intersection.hpp
```diff
@@ -27,6 +35,106 @@ namespace boost { namespace geometry
...
+    // This tool can't be used for non-integral coordinate types
+    // also sign_of_product and sign_of_compare functions must be modified
+    // (comparsons and multiplication)
+    BOOST_STATIC_ASSERT(boost::is_integral<T>::value);
+
+    static const std::size_t bits = CHAR_BIT * sizeof(T);
+    static const std::size_t half_bits = bits / 2;
+    typedef typename boost::make_unsigned<T>::type unsigned_type;
+    static const unsigned_type base
+        = integral_pow<unsigned_type, 2, half_bits>::value;
```
mkaravel: On second thought, you do not really need `integral_pow<>` at all.
Simply do:
```
static const unsigned_type base = (1 << half_bits);
```

awulkiew: Indeed, initially I implemented the pow function because I was using the T type which could be signed type and I feared that bit operations could cause an undefined behavior. But indeed according to the standard they should work as intended for non-negative values and unsigned types. At some point I switched to unsigned type so I could use bit shifts now. Furthermore I could now use bit shifts instead of division.

PR #300, opened 2015-06-01
Disable buffer test cases involving empty geometries

mkaravel:
Some buffer test cases failed when `BOOST_GEOMETRY_ENABLE_ACCESS_DEBUGGING`. These test cases involve testing empty geometries (e.g., `LINESTRING()`, `POLYGON(())`, etc.).
This PR disables these test cases.

### DIFF
--- a/test/algorithms/buffer/buffer_linestring.cpp
+++ b/test/algorithms/buffer/buffer_linestring.cpp
@@ -174,7 +174,9 @@ void test_all()
     test_one<linestring, polygon>("field_sprayer1", field_sprayer1, join_round, end_round, 718.761877, 16.5, 6.5);
     test_one<linestring, polygon>("field_sprayer1", field_sprayer1, join_miter, end_round, 718.939628, 16.5, 6.5);
 
+#ifdef BOOST_GEOMETRY_TEST_BUFFER_TEST_EMPTY_GEOMETRIES
     test_one<linestring, polygon>("degenerate0", degenerate0, join_round, end_round, 0.0, 3.0);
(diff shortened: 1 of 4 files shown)


mkaravel:
Let me just say that this should be considered as an intermediate solution.
What I prefer to have is to have `buffer()` throw an exception if the input is empty, and then the unit test should basically check that the exception is properly thrown.

awulkiew:
So the problem is what should be the result of a buffer() for an empty geometry input. We have at least two possibilities:
1. generate an empty geometry (MultiPolygon containing 0 points)
2. throw and exception
Both solutions are reasonable. Do I understand correctly that the 1. is the original behavior?
With that said this PR doesn't solve the problem and without the debugging enabled the problem is not manifested so the tests passes (though on some compilers there are warnings about the use of uninitialized variables).

mkaravel:
@awulkiew As I wrote in my comment, I prefer solution 2 (among the two you suggest).
Barend?

barendgehrels:
A buffer of an empty geometry (which is valid) should not throw an exception. So it should return an empty polygon. And it did. So behaviour was OK.

So please close this PR - it should be solved differently.

mkaravel:
@barendgehrels Okay, fair enough. I will try to think of another way to solve the problem then.

barendgehrels:
Buffer (boost::geometry::buffer) now returns empty if input is empty.

The problems where in the test case (testing boost::geometry::buffer_inserter), which now assigns the envelope with inverse if input is empty (as it was before). I did not get the error message of the regression, but it restores previous behaviour so it is expected that that is OK now.

There was actually no unit test for the boost::geometry::buffer with strategies (wrapping buffer_inserter), so I added that and check for empty geometries too

PR #301, opened 2015-06-01
Fix: initialize the rescale policy only if the input geometries are non-empty

mkaravel:
This fixes unit test failures (union, intersection, difference) when `BOOST_GEOMETRY_ENABLE_ACCESS_DEBUGGING is enabled`.


mkaravel:
Let me just say that this should be considered as an intermediate solution.
What I prefer to have is to have `buffer()` throw an exception if the input is empty, and then the unit test should basically check that the exception is properly thrown.

include/boost/geometry/policies/robustness/get_rescale_policy.hpp
```diff
@@ -99,10 +105,30 @@ static inline void init_rescale_policy(Geometry1 const& geometry1,
...
+    if (geometry::num_points(geometry1) == 0
+        && geometry::num_points(geometry2) == 0)
+    {
+        return;
+    }
+    else if (geometry::num_points(geometry1) == 0)
+    {
+        geometry::envelope(geometry2, env);
+    }
+    else if (geometry::num_points(geometry2) == 0)
```
awulkiew: Emptiness is checked two times for the same geometry.

mkaravel: Good point thanks!
I will fix this.

include/boost/geometry/policies/robustness/get_rescale_policy.hpp
```diff
@@ -99,10 +105,30 @@ static inline void init_rescale_policy(Geometry1 const& geometry1,
...
+        geometry::envelope(geometry1, env);
+    }
+    else
+    {
+        geometry::envelope(geometry1, env);
+        model::box<Point> env2 = geometry::return_envelope
+            <
+                model::box<Point>
+            >(geometry2);
+        geometry::expand(env, env2);
```
awulkiew: Will in this else-branch the optimal envelope be generated in the spherical nor geographical CS? AFAIR the envelope should be calculated at once for all of the involved geometries.

include/boost/geometry/policies/robustness/get_rescale_policy.hpp
mkaravel: Fixed. See commit 87052dc.

include/boost/geometry/policies/robustness/get_rescale_policy.hpp
mkaravel: This is absolutely true: the final envelope we get this way for spherical/geographic CS may not be optimal.

What we need to do to address this issue is to overload `bg::envelope()` to take two geometries as input, and properly compute their combined envelope. This requires a lot more work though, and although I agree, I consider it as outside the scope of this PR.

We should do it though, that is be able to compute the envelope of two (or possibly more geometries), and we should put it in our TODO list as part of the on-going work for supporting spherical/geographic CS.

include/boost/geometry/policies/robustness/get_rescale_policy.hpp
```diff
@@ -99,10 +105,31 @@ static inline void init_rescale_policy(Geometry1 const& geometry1,
...
 {
-    // Get bounding boxes
-    model::box<Point> env = geometry::return_envelope<model::box<Point> >(geometry1);
-    model::box<Point> env2 = geometry::return_envelope<model::box<Point> >(geometry2);
-    geometry::expand(env, env2);
+    // Get bounding boxes (when at least one of the geometries is not empty)
+    model::box<Point> env;
+    bool is_empty1 = (geometry::num_points(geometry1) == 0);
+    bool is_empty2 = (geometry::num_points(geometry2) == 0);
+    if (is_empty1 == 0 && is_empty2)
```
barendgehrels: I think you mean just "is_empty1" here, not == 0

include/boost/geometry/policies/robustness/get_rescale_policy.hpp
```diff
@@ -99,10 +105,31 @@ static inline void init_rescale_policy(Geometry1 const& geometry1,
         RobustPoint& min_robust_point,
         Factor& factor)
 {
-    // Get bounding boxes
-    model::box<Point> env = geometry::return_envelope<model::box<Point> >(geometry1);
-    model::box<Point> env2 = geometry::return_envelope<model::box<Point> >(geometry2);
-    geometry::expand(env, env2);
+    // Get bounding boxes (when at least one of the geometries is not empty)
+    model::box<Point> env;
+    bool is_empty1 = (geometry::num_points(geometry1) == 0);
```
barendgehrels: brackets can be omitted and bool can be const

include/boost/geometry/policies/robustness/get_rescale_policy.hpp
barendgehrels: Agreed writing it into the TODO - this PR is for fixing the rescale, not supporting spherical. But good point of course.

barendgehrels:
I'm OK with this PR, if comments are addressed

include/boost/geometry/policies/robustness/get_rescale_policy.hpp
```diff
@@ -99,10 +105,31 @@ static inline void init_rescale_policy(Geometry1 const& geometry1,
         RobustPoint& min_robust_point,
         Factor& factor)
 {
-    // Get bounding boxes
-    model::box<Point> env = geometry::return_envelope<model::box<Point> >(geometry1);
-    model::box<Point> env2 = geometry::return_envelope<model::box<Point> >(geometry2);
-    geometry::expand(env, env2);
+    // Get bounding boxes (when at least one of the geometries is not empty)
+    model::box<Point> env;
```
barendgehrels: I think this declaration can go to below (but don't see enough context here to be sure)

mkaravel:
@barendgehrels I believe I have addressed all your comments. In particular, regarding where to place the declaration of `env`, I have moved it a bit down.

@awulkiew @barendgehrels Please let me know if it okay to merge this PR.

barendgehrels:
OK for me! Thanks for fixing.

mkaravel:
@barendgehrels You are welcome. Thank you for the comments.

awulkiew:
I have nothing against merging this PR. However I think that in general it should be handled differently.
IMO the `envelope()` should return bool indicating if the result is correct. And instead of the `num_points()` calls those boolean values should be checked. This could be done instead of this PR or later after merging it.

mkaravel:
We have already discussed the possibility of returning a `bool` for envelope. It is at the top of my to-do-list. Once envelope is changed to return a `bool` I will update the code related to this PR.
Regarding the call to `num_points`, I realized that we need a new algorithm `is_empty()` that will determine if a geometry is empty (it should be more efficient than `num_points() == 0`). I am in the process of implementing it, and then check places in BG where calls to `num_points`  should be replaced by `is_empty()`.

barendgehrels:
About is_empty, I'm OK with that new function.

About envelope returning a bool, I would like to see this in a broader context and not do it at hoc. Compare it with centroid. It is void. Both take a (possibly empty) geometry and fill another geometry as a result. That should be the same.
Same for point_on_surface and point_on_border (detail)
Similar but not exactly the same with convex_hull (currently leaves output just empty)

mkaravel:
If everybody is okay with this PR, I would like to merge it.

awulkiew:
I'd be ok with merging if `envelope()` was made consistent with `is_empty()`.
Btw, above the modified `init_rescale_policy()` overload there is one taking only one geometry. It should also be fixed.

mkaravel:
`envelope()` is made inconsistent with `is_empty()` in PR #302. PR #302 has not been merged yet, so I will wait until that PR is merged first, before merging this PR.

Regarding the `init_rescale_policy()` that takes one geometry, there is already a call to `is_empty()` in place. I do not understand what needs to be fixed.

awulkiew:
Did you mean "made consistent"? If that's the case I'm ok with merging. The other PR should be merged soon anyway.

Ah yes, sorry. I was reading the original sources.

mkaravel:
Yes, I meant "made consistent". Sorry for the typo and the misunderstanding.

barendgehrels:
I'm OK with merging

mkaravel:
Although I intended to wait to merge this PR until after PR #302 is done, I merged it in order to (hopefully) get rid of some failures in the regression matrix for the develop branch.

PR #304, opened 2015-06-03
Replace use of num_points by is_empty

mkaravel:
This PR replaces instances of `bg::num_points() == 0` by `bg::is_empty()` and instances of `bg::num_points() > 0` by `! bg::is_empty()`.

This is done both in the `include` and `test` directories.


barendgehrels:
Thnks I'm OK with merging

PR #305, opened 2015-06-05
[buffer] Replace int with signed_index_type.

awulkiew:


barendgehrels:
Thanks for the PR!

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
```diff
@@ -127,23 +127,23 @@ struct buffer_turn_info
     robust_point_type rob_pi, rob_pj, rob_qi, rob_qj;
 #endif
 
-    int count_within;
+    signed_index_type count_within;
```
barendgehrels: counts are not indexes.
Can you make them std::size_t, besides count_in_original which should stay int?

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
```diff
@@ -127,23 +127,23 @@ struct buffer_turn_info
...
+    signed_index_type count_within_near_offsetted;
 #endif
 
     bool remove_on_multi;
 
     // Obsolete:
-    int count_on_occupied;
-    int count_on_multi;
+    signed_index_type count_on_occupied;
+    signed_index_type count_on_multi;
```
barendgehrels: These will be remoed

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
```diff
@@ -127,23 +127,23 @@ struct buffer_turn_info
...
     bool within_original;
-    int count_on_original_boundary;
-    int count_in_original; // increased by +1 for in ext.ring, -1 for int.ring
+    signed_index_type count_on_original_boundary;
+    signed_index_type count_in_original; // increased by +1 for in ext.ring, -1 for int.ring
 
-    int count_on_offsetted;
-    int count_on_helper;
+    signed_index_type count_on_offsetted;
+    signed_index_type count_on_helper;
```
barendgehrels: Will be removed

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -191,8 +191,8 @@ struct buffered_piece_collection
 
         // 1: half, part of offsetted_rings
         segment_identifier first_seg_id;
-        int last_segment_index; // no segment-identifier - it is the same as first_seg_id
-        int offsetted_count; // part in robust_ring which is part of offsetted ring
+        signed_index_type last_segment_index; // no segment-identifier - it is the same as first_seg_id
+        signed_index_type offsetted_count; // part in robust_ring which is part of offsetted ring
```
barendgehrels: offsetted_count is a count but used in calculations with indexes. What do we normally do here?

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -1197,7 +1197,7 @@ struct buffered_piece_collection
         robust_point_type any_point;
         geometry::recalculate(any_point, point, m_robust_policy);
 
-        int count_in_original = 0;
+        signed_index_type count_in_original = 0;
```
barendgehrels: This should IMO stay int. It is not an index and not used as an index.

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
```diff
@@ -227,7 +227,7 @@ public :
                 && point_y >= geometry::get<min_corner, 1>(section.bounding_box) - 1
                 && point_y <= geometry::get<max_corner, 1>(section.bounding_box) + 1)
             {
-                for (int i = section.begin_index + 1; i <= section.end_index; i++)
+                for (signed_index_type i = section.begin_index + 1; i <= section.end_index; i++)
```
barendgehrels: section begin_index/end_index should be made signed_index_type too.
But that can be done separately, indeed.
So this signed_index_type here is OK

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
```diff
@@ -409,7 +409,8 @@ class analyse_turn_wrt_piece
 
         point_type points[4];
 
-        int helper_count = static_cast<int>(piece.robust_ring.size()) - piece.offsetted_count;
+        signed_index_type helper_count = static_cast<signed_index_type>(piece.robust_ring.size())
+                                            - piece.offsetted_count;
```
barendgehrels: should stay int

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
awulkiew: Why it should stay `int`? It's used to count the elements of a vector meeting some criteria. This means that it should be at least of size of `std::vector<robust_original>::size_type`. But since it can also be decremented, to be perfectly safe it could be replaced by 2 variables, one counting the elements meeting the `m_is_interior` condition and the other one counting the elements meeting the `m_has_interior` condition. And at the end both values should be compared which one is greater. Do you agree?

awulkiew:
Btw, why `turn_index` is initialized to `-1` and therefore must be a signed type? Can e.g. `buffer_turn_info` be created when there are no turns?

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
awulkiew: Sure. I didn't know exactly which variables are used to count in one direction and which counts in both so I choose a safe approach.

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
awulkiew: Again, why a value returned by a Range::size() whatever the Range is (probably std::vector) should be assigned to an `int`?

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
awulkiew: Well, it's initialized with `-1` (Is that correct?). So I'm guessing that it should be a signed type (unless it shouldn't be initialized to `-1`). And it's again, iterated in a loop iterating through `pc.robust_turns` elements so it should be at least of size of the `std::vector<robust_turn>::size_type`.

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
awulkiew: Indeed, also the type of `non_duplicate_index` and of various variables in the sectionalize code.

awulkiew:
Since this `signed_index_type` can be used for various things. And since it's defined as `boost::make_signed<std::size_t>::type`. Maybe then should it be renamed to `signed_size_type` or `signed_size_t`? Or maybe should this additional type be added. Then it could be e.g. used for signed counters and be less confusing.

barendgehrels:
I'm in favour of renaming it to signed_size_type because index explicitly names a target which often is not appropriate and indeed confusing.
If that is done, most of your changes would be OK.

awulkiew:
Ok, I changed the name to signed_size_type and used std::size_t for most of the counters. Some of them are indeed not used. The changes that should be done in other parts of the library, e.g. sectionalize shouldn't be a part of this PR so I'd stop here. Could `turn_index` be made `std::size_t` as well or should it stay signed?

include/boost/geometry/algorithms/detail/signed_index_type.hpp
```diff
@@ -21,6 +21,7 @@ namespace boost { namespace geometry
 
 
 typedef boost::make_signed<std::size_t>::type signed_index_type;
+typedef boost::make_signed<std::size_t>::type signed_size_type;
```
barendgehrels: Thanks!
The filename should be renamed too.
I don't understand that I only see buffer-things, if this renaming was part of this PR, then I would expect much more changes? Where are they?

barendgehrels:
I'm OK with merging.
About turn_index, I agree, it is better to make it unsigned and initialize it with 0

include/boost/geometry/algorithms/detail/signed_index_type.hpp
awulkiew: I only changed the names in the buffer right now, since the refactoring of other parts wasn't the purpose of this PR. Therefore I only added a type and didn't rename a file. I thought it'd be better to change the rest in the other PR or to do it directly in the develop since it would be rather automatic. Of course the refactoring of `sectionalize` and posibly other parts of the library would require a PR.

awulkiew:
Done, I amended the commit for clear history.

PR #306, opened 2015-06-08
Replace BOOST_ASSERT with BOOST_GEOMETRY_ASSERT

awulkiew:
This allows to implement user-defined assertion failure handling mechanism specifically for Boost.Geometry assertions, e.g. to replace it with exception throw and/or logging or custom terminate, etc. The implementation is similar to the one in Boost.Assert so interested users should be familiar with it.

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
@@ -12,11 +12,12 @@
 #include <cstddef>
 #include <iterator>
 
-#include <boost/assert.hpp>
+
 #include <boost/core/ignore_unused.hpp>
 #include <boost/numeric/conversion/cast.hpp>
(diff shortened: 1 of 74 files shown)


mkaravel:
I would be nice to have a unit test where a user-defined assertion handling is used.

awulkiew:
@mkaravel I think that if a function definition was defined as inline function in the same file that includes Boost.Geometry headers then the function would be inlined. And AFAIU the forward declaration is here in order to generate an error message during the linking phase.

We could of course always assume that the required functions were declared before the `core/asert.hpp` header was included however this could be confusing and less flexible. Since now the user could define the handler function in a separate unit.

Another alternative would be to define the macros only if they were not defined before. But I don't consider it as being safe enough.

Of course I'll create a test throwing exceptions.

barendgehrels:
Good idea this gives some extra flexibility.
I'm OK with merging
Thanks for the PR

PR #307, opened 2015-06-09
Replace int with more suitable type (std::size_t, signed_size_type, template parameter).

awulkiew:
This affects buffer, overlay, relate and sectionalize.

Please for now treat `signed_index_type` and `signed_size_type` as one type. I still plan to replace the former with the latter consistently in the whole library. This is not the intention of this PR.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/enrichment_info.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/enrichment_info.hpp
@@ -37,13 +37,13 @@ struct enrichment_info
     // vertex to which is free travel after this IP,
     // so from "segment_index+1" to "travels_to_vertex_index", without IP-s,
     // can be -1
-    signed_index_type travels_to_vertex_index;
+    signed_size_type travels_to_vertex_index;
 
     // same but now IP index, so "next IP index" but not on THIS segment
-    int travels_to_ip_index;
+    signed_size_type travels_to_ip_index;
 
     // index of next IP on this segment, -1 if there is no one
-    int next_ip_index;
+    signed_size_type next_ip_index;
 };
 
 
--- a/include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
@@ -143,7 +143,7 @@ class get_turns_in_sections
 
     template <typename Geometry, typename Section>
     static inline bool neighbouring(Section const& section,
-            int index1, int index2)
+            signed_index_type index1, signed_index_type index2)
     {
         // About n-2:
         //   (square: range_count=5, indices 0,1,2,3
@@ -151,7 +151,7 @@ class get_turns_in_sections
         // Also tested for open polygons, and/or duplicates
         // About first condition: will be optimized by compiler (static)
         // It checks if it is areal (box,ring,(multi)polygon
-        int const n = int(section.range_count);
+        signed_index_type const n = static_cast<signed_index_type>(section.range_count);
 
         boost::ignore_unused_variable_warning(n);
         boost::ignore_unused_variable_warning(index1);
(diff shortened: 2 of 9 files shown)


include/boost/geometry/algorithms/detail/overlay/enrichment_info.hpp
```diff
@@ -37,13 +37,13 @@ struct enrichment_info
...
-    signed_index_type travels_to_vertex_index;
+    signed_size_type travels_to_vertex_index;
 
     // same but now IP index, so "next IP index" but not on THIS segment
-    int travels_to_ip_index;
+    signed_size_type travels_to_ip_index;
 
     // index of next IP on this segment, -1 if there is no one
-    int next_ip_index;
+    signed_size_type next_ip_index;
```
barendgehrels: Be aware that this makes the type probably larger. This might influence our maximum number of intersections (memory limited)

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
```diff
@@ -143,7 +143,7 @@ class get_turns_in_sections
 
     template <typename Geometry, typename Section>
     static inline bool neighbouring(Section const& section,
-            int index1, int index2)
+            signed_index_type index1, signed_index_type index2)
```
barendgehrels: I thought we should rename signed_index_type to signed_size_type and that it was already done (at least in buffer),
so I don't understand why signed_index_type is now popping up here again...

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
```diff
@@ -80,12 +81,14 @@ struct section
     ring_identifier ring_id;
     Box bounding_box;
 
-    int begin_index;
-    int end_index;
+    // NOTE: size_type could be passed as template parameter
```
barendgehrels: Please don't.

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
```diff
@@ -80,12 +81,14 @@ struct section
...
+    // NOTE: these probably also could be of type std::size_t
```
barendgehrels: Yes, this might be right.

barendgehrels:
Thanks, I'm basically OK with merging

include/boost/geometry/algorithms/detail/overlay/enrichment_info.hpp
awulkiew: Indeed, was there a reason why for the `travels_to_vertex_index` a different type was used? Are they all absolute indexes? Or are the other two e.g. offsets?

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
awulkiew: Yes, they will be renamed consistently in the whole library. For now in other places in this file the old name is used so I used it here as well to not mix them.

barendgehrels: Okidoki

include/boost/geometry/algorithms/detail/overlay/enrichment_info.hpp
barendgehrels: It is and stays sometimes unassigned. So -1 means unassigned.

awulkiew: Yes I understand that. But is it an absolute index or is it a difference between some relativaly close indexes (so an offset)? Because if the latter was true, then I suppose a smaller type could be used.

Btw, AFAIR source_index in segment_iterator and ring_iterator could be a smaller type if there was a problem, because only 0 or 1 is assigned.

PR #308, opened 2015-06-10
[strategies][distance][cross track] expose the radius type as a typedef

mkaravel:
(same as for projected_point distance strategy)


awulkiew:
It looks good :)

PR #309, opened 2015-06-11
[is_valid] Suppress unused parameter warnings.

awulkiew:


mkaravel:
Presume the warning have to do with the fact that the visitor may have only static methods.

In any case I am okay with merging.

awulkiew:
Yes, msvc reports a warning in this case...

PR #311, opened 2015-06-12
Fix cross_track_point_box distance strategy

mkaravel:
This PR fixes the implementation of the `cross_track_point_box` distance strategy, so that it accounts for the fact that boxes in the spherical equatorial coordinate system are now defined as Cartesian products of intervals of longitudes and latitudes.


include/boost/geometry/strategies/spherical/distance_cross_track_point_box.hpp
```diff
@@ -32,131 +44,143 @@ namespace boost { namespace geometry
...
 
-        // Return 0 if the point isn't on the right side of any segment
-        return return_type(0);
+        // Otherwise determine the distance of the point to the box as
+        // the minimum distance of the point to the western- and
+        // eastern-most edges of the box
+        return_type d1 = m_ps_strategy.apply(point, bottom_left, top_left);
+        return_type d2 = m_ps_strategy.apply(point, bottom_right, top_right);
+
+        return (std::min)(d1, d2);
```
awulkiew: I see a possibility for improvements here:
- The closer edge could be found and then the distance calculated, only one.
- If the point latitude was in [lat_min, lat_max] the distance could be calculated as e.g. `radius() * (plon - lon_max)`. The longitudes would have to be normalized somehow but in general we should be able to avoid calling a strategy.

With that said I think that the PR could be merged as it is now. Those optimizations could be e.g. implemented later at some point since the purpose of this PR is fixing the bug, not optimizing the code. If you decided not to implement them now, could you write a short note or TODO describing that this part can be optimized? Btw, there was a note about optimization in the deleted section of the code, a different one but related.

awulkiew:
I'm ok with merging.

include/boost/geometry/strategies/spherical/distance_cross_track_point_box.hpp
mkaravel: Regarding the part where the point's latitude is in `[lat_min, lat_max]`, your suggestion will not work in all cases. The spherical projection of the point in the box's segment may not be inside the segment, even though the latitude of the point is inside the segment's latitude range. So in general it will not work.

Regarding the other suggestion to first find the ***closest meridian***, I agree that it can be done, and thus we can avoid one strategy call.
Good point! Thanks! I will implement it ASAP.

awulkiew: Yes, you're right. The strategy must be called in all cases since the shortest path from point to the meridian will be a fragment of a great circle, and won't go along a parallel.

mkaravel: Commit 23addc2 implements the suggested optimization.

include/boost/geometry/strategies/spherical/distance_cross_track_point_box.hpp
```diff
@@ -32,131 +42,223 @@ namespace boost { namespace geometry
...
+            else
+            {
+                // the midway meridian is right of the box
+                if (plon > lon_max && plon <= lon_midway)
+                {
+                    return m_ps_strategy.apply(point, bottom_right, top_right);
+                }
+                else
+                {
+                    return m_ps_strategy.apply(point, bottom_left, top_left);
```
awulkiew: Do I understand correctly that those conditions requires that the plon must be in the valid [-pi, pi] range?

I also have a small suggestion. I'm guessing that if some boolean value indicating which corner should be used was calculated at the beginning, and then after all of those conditions and setting this boolean value a strategy was called with corresponding parameters, it'd allow the compiler to optimize the code more easily. Now in each branch there is a potentially inlineable function call which could result in a code bloat and worse instruction cachng. With this boolean only this variable, most certainly optimized out would be calculated and then there would be only one if/else with function calls. So:

    if (...) {
        temp = ...
        if (...)
            use_left = true;
        else
            use_left = false;
    } else {
        temp = ...
        if (...)
            use_left = false;
        else
            use_left = true;
    }
    if ( use_left )
        m_ps_strategy.apply(point, bottom_left, top_left);
    else
        m_ps_strategy.apply(point, bottom_right, top_right);

This theory could of course be verified but still, if those strategies was called in one place, at the end, the code would probably also be more clear and safe since there would be less function calls into which wrong parameters could be passed.

Not to mention that it would probably be possible to calculate this boolean variable using terniary `?:` operators or at least use it for the calculation of partial results. Or maybe even to write everything in one if condition, though this would probably make the code less readable.

mkaravel: Done. See commit 933ed07.

mkaravel:
@barendgehrels Are you okay with merging this PR?

barendgehrels:
Yes I'm OK with merging.

PR #313, opened 2015-07-01
Fix/sym difference areal

mkaravel:
This PR fixes the symmetric difference algorithm for areal geometries.

The problem with the previous algorithm is that it could generate invalid multipolygons (this could happen when the input to the symmetric difference algorithm was two areal geometries with disjoint interiors and touching boundaries).

The fix proposed in this PR is to compute the differences of the first and second, and second and first geometries, and then return their union (instead of concatenating the two results).

For non-areal geometries the behavior of the algorithm remains unchanged.

### DIFF
--- a/include/boost/geometry/algorithms/sym_difference.hpp
+++ b/include/boost/geometry/algorithms/sym_difference.hpp
@@ -1,6 +1,11 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
-// Copyright (c) 2007-2012 Barend Gehrels, Amsterdam, the Netherlands.
+// Copyright (c) 2007-2015 Barend Gehrels, Amsterdam, the Netherlands.
+
+// This file was modified by Oracle on 2015.
+// Modifications copyright (c) 2015 Oracle and/or its affiliates.
+
+// Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
 
 // Use, modification and distribution is subject to the Boost Software License,
 // Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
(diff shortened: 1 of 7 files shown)


mkaravel:
This PR addresses (and fixes) Boost Trac ticket [11436](https://svn.boost.org/trac/boost/ticket/11436).

include/boost/geometry/algorithms/sym_difference.hpp
```diff
@@ -23,6 +31,193 @@ namespace detail { namespace sym_difference
...
+                    >::value
+            >::apply(areal2, areal1, robust_policy, oit21, strategy);
+
+        return geometry::dispatch::union_insert
+            <
+                helper_geometry_type,
+                helper_geometry_type,
+                GeometryOut
+            >::apply(diff12, diff21, robust_policy, out, strategy);
+    }
```
awulkiew: The section calling `intersection_insert` two times can be found in both `sym_difference_generic` and `sym_difference_areal_areal`. This common code could be placed in one function or `generic` version could be called from `areal_areal` version or the `sym_difference_areal_areal` could be removed and this function calling `intersection_insert` could be called directly inside `dispatch::sym_difference_insert` below.

mkaravel: The code is not entirely the same (not the semantics) hence the "duplication".
As you can easily see in the generic version both calls output to the same output iterator.
In the A/A version they output to a different iterator.

test/algorithms/set_operations/sym_difference/sym_difference_areal_areal.cpp
```diff
@@ -0,0 +1,137 @@
...
+// Licensed under the Boost Software License version 1.0.
+// http://www.boost.org/users/license.html
+
+// Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
+
+#include <iostream>
+
+#ifndef BOOST_TEST_MODULE
+#define BOOST_TEST_MODULE test_sym_difference_areal_areal
+#endif
```
barendgehrels: The sym-difference was already unit-tested by test_difference (which did both, a-b, b-a, and sym). Anyway, having this problem tackled individually is also OK

include/boost/geometry/algorithms/sym_difference.hpp
```diff
@@ -23,6 +31,193 @@ namespace detail { namespace sym_difference
...
+                        geometry::point_order<GeometryOut>::value
+                    >::value
+            >::apply(areal2, areal1, robust_policy, oit21, strategy);
+
+        return geometry::dispatch::union_insert
+            <
+                helper_geometry_type,
+                helper_geometry_type,
+                GeometryOut
+            >::apply(diff12, diff21, robust_policy, out, strategy);
```
barendgehrels: Looks good.

include/boost/geometry/algorithms/sym_difference.hpp
```diff
@@ -23,6 +31,193 @@ namespace detail { namespace sym_difference
...
+    static inline OutputIterator apply(Areal1 const& areal1,
+                                       Areal2 const& areal2,
+                                       RobustPolicy const& robust_policy,
+                                       OutputIterator out,
+                                       Strategy const& strategy)
+    {
+        typedef geometry::model::multi_polygon
+            <
+                GeometryOut
+            > helper_geometry_type;
```
barendgehrels: OK - this is probably the first we have a multi_polygon as helper-geometry. But we have to. I'm OK.

barendgehrels:
If all tests pass I'm OK with merging.
Thanks!

include/boost/geometry/algorithms/sym_difference.hpp
awulkiew: Ah yes, I see. Still we should somehow try to avoid the duplication. Do you think that the `detail::intersection::insert()` could be called instead? Probably not since the `reverse_dispatch` is used there internally. What do you think about closing the `intersection_insert` inside a function or altering `detail::intersection::insert()` to take a template parameter indicating if the `reverse_dispatch` should be used. Or just an overload taking an additional Reverse parameter. Then a call would take 1 or 2 lines, not 15.

Btw, the names of those `intersection`... tools may be confusing, that's a side note.

mkaravel:
@barendgehrels I need to check possible unit test failures. I will report once I am done.

include/boost/geometry/algorithms/sym_difference.hpp
mkaravel: I tried to re-factor common code. Please see commit 4bd9607.

mkaravel:
I have checked and updated existing symmetric difference test cases. See commit 921d666.

mkaravel:
All algorithm unit tests pass.
May I merge this PR?

include/boost/geometry/algorithms/sym_difference.hpp
```diff
@@ -23,6 +31,177 @@ namespace detail { namespace sym_difference
...
+    static inline OutputIterator apply(Geometry1 const& geometry1,
+                                       Geometry2 const& geometry2,
+                                       RobustPolicy const& robust_policy,
+                                       OutputIterator out,
+                                       Strategy const& strategy)
+    {
+        out = compute_difference
+            <
+                GeometryOut
+            >::apply(geometry1, geometry2, robust_policy, out, strategy);
```
awulkiew: Yes, this does look a lot better. Thanks!

It could also be implemented as a function instead of a struct with apply() function. It'd then be even shorter. However then the GeometryOut would still have to be passed as a function template parameter and I guess someone could think that it's the type of the return. So I'm ok with the current implementation.

Though this brings another question, why is the GeometryOut type needed in the first place in the `intersection_insert`. But that's a different story.

awulkiew:
Yes, I'm ok with merging. Thanks!

mkaravel:
Barend, are you okay with merging?

barendgehrels:
Sure I am, as I already mentioned. Code was enhanced after that, and I'm also OK with that.

mkaravel:
@barendgehrels Thanks for the immediate feedback. Will merge shortly.

PR #315, opened 2015-07-09
Fix: possible infinite loop in has_spikes

mkaravel:
If the range passed to `has_spikes` consists of points with `NaN` coordinates (this can happen if the range passed as been computed as the result, e.g., of a set operation), then `has_spikes` can enter an infinite loop.

If a point `p` has such `NaN` coordinates, then `p` is considered as not equal to itself, a property/feature assumed in the previous implementation of `has_spikes`), and the following calls (in the older version of `has_spikes`) to `std::find_if` never return:
```
std::find_if(prev, boost::end(view), not_equal(*prev));
```
and
```
std::find_if(cur, boost::end(view), not_equal(*cur));
```

The solution if to essentially call `std::find_if` as `std::find_if(second, first, last, not_equal(*first))`, which avoids the implicit assumption that the `bg::equals(*first, *first)` returns `true`.

***Note:*** the problem appears for geometries with `NaN` coordinates. The point addressed here is to have `has_spikes` return instead of entering an endless loop; the result returned, however, does not (and could not) have any geometric significance.

### DIFF
--- a/include/boost/geometry/algorithms/detail/is_valid/has_spikes.hpp
+++ b/include/boost/geometry/algorithms/detail/is_valid/has_spikes.hpp
@@ -12,11 +12,11 @@
 
 #include <algorithm>
 
-#include <boost/assert.hpp>
 #include <boost/core/ignore_unused.hpp>
 #include <boost/range.hpp>
 #include <boost/type_traits/is_same.hpp>
 
+#include <boost/geometry/core/assert.hpp>
 #include <boost/geometry/core/point_type.hpp>
 #include <boost/geometry/core/tag.hpp>
 #include <boost/geometry/core/tags.hpp>
@@ -83,7 +83,7 @@ struct has_spikes
     {
         typedef not_equal_to<typename point_type<Range>::type> not_equal;
 
-        BOOST_ASSERT(first != last);
+        BOOST_GEOMETRY_ASSERT(first != last);
 
         Iterator second = first;
         ++second;
--- a/include/boost/geometry/algorithms/detail/is_valid/has_spikes.hpp
+++ b/include/boost/geometry/algorithms/detail/is_valid/has_spikes.hpp
@@ -12,6 +12,7 @@
 
 #include <algorithm>
 
+#include <boost/assert.hpp>
 #include <boost/core/ignore_unused.hpp>
 #include <boost/range.hpp>
 #include <boost/type_traits/is_same.hpp>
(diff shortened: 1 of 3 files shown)


include/boost/geometry/algorithms/detail/is_valid/has_spikes.hpp
```diff
@@ -77,13 +78,23 @@ struct not_equal_to
 template <typename Range, closure_selector Closure>
 struct has_spikes
 {
+    template <typename Iterator>
+    static inline Iterator find_not_equal(Iterator first, Iterator last)
+    {
+        typedef not_equal_to<typename point_type<Range>::type> not_equal;
+
+        BOOST_ASSERT(first != last);
```
barendgehrels: BOOST_GEOMETRY_ASSERT ?

barendgehrels:
Can you add a unit-test for this situation?
If Adam is also OK I'm OK with merging

include/boost/geometry/algorithms/detail/is_valid/has_spikes.hpp
mkaravel: You are absolutely right. Will change it.

mkaravel:
I had in mind to add a unit test. Will do before merging (of course I am also waiting for Adam's approval).

include/boost/geometry/algorithms/detail/is_valid/has_spikes.hpp
mkaravel: Done. See commit 78296da.

mkaravel:
Test cases for `bg::is_simple` and `bg::is_valid` added. See commit 73960f3.

awulkiew:
So basically the algorithm now starts searching from the next element. However I find the way how `find_not_equal` works slightly confusing because it takes a Range but ommits the first element. Wouldn't it be more clear to "manually" increment the iterator in the main algorithm and then use `find_if` directly?
I'm of course ok with merging.

mkaravel:
I wrote this method in order to avoid code duplication. What should be done in each case is really what the method does. If you prefer another name for the method, I am okay. I do not see the point though, of removing the method and duplicating the code.

awulkiew:
Yes, I understand that your goal was to avoid duplication. Ok, let's try to find a better name for it. We're searching for one element so a function should indeed contain a word *find*. E.g. `find_mismatch` is one possibility but it's not clear what's this *mismatch*, which elements are compared exactly. So maybe something like `skipping_find` (name similar to `adjacent_find`) or rather `skipping_find_if` skipping some number of elements (1 in this case). But then, something like `skipping_iterator` could be used with standard `find_if` instead, though I think there is no such thing in Boost. Well, `filter_iterator` could probably be used but the code would probably be more confusing.

mkaravel:
How about `find_different_from_first`? Would that work for you?

awulkiew:
A little too descriptive for my taste but indeed more clear, so ok.

mkaravel:
Done: method `find_not_equal` renamed to `find_different_from_first`.

awulkiew:
Thanks!

PR #319, opened 2015-08-14
Take special care of NaN coordinates in relate for MultiLinestring

awulkiew:
In places where a check must be performed, if a Point is one of the endpoints of a Linestring contained in a MultiLinestring, std::sort() and std::equal_range() algorithms are used. With MSVC the assertion in std::equal_range() fails if the elements cannot be reliably compared, i.e. in the case when Points has NaN coordinates.

This PR adds has_nan_coordinate() and used it in relate() implementation.

Geometries containing NaN coordinates can be easily generated with a set operation containing some unreasonable data like very big segments.


mkaravel:
I am in favor of merging this PR.
Thanks a lot Adam.

PR #322, opened 2015-08-26
Support for reading and writing Multi-geometries

meastp:
Adds support for Multi-geometries in the WKB data format.

This is the first draft, and a (better) version of the svn patches for Multi-geometry WKB support. I need help with the equals-algorithm used for the tests, as they currently only check the number of geometries (in the multi-geometry).

@mloskot

### DIFF
--- a/include/boost/geometry/algorithms/equals.hpp
+++ b/include/boost/geometry/algorithms/equals.hpp
@@ -242,6 +242,20 @@ struct equals<P1, P2, point_tag, point_tag, DimensionCount, Reverse>
         >
 {};
 
+template <typename MultiPoint1, typename MultiPoint2, std::size_t DimensionCount, bool Reverse>
+struct equals<MultiPoint1, MultiPoint2, multi_point_tag, multi_point_tag, DimensionCount, Reverse>
+    : detail::equals::equals_by_relate<MultiPoint1, MultiPoint2>
+{};
+
+template <typename MultiPoint, typename Point, std::size_t DimensionCount, bool Reverse>
+struct equals<MultiPoint, Point, multi_point_tag, point_tag, DimensionCount, Reverse>
+    : detail::equals::equals_by_relate<MultiPoint, Point>
+{};
+
+template <typename MultiPoint, typename Point, std::size_t DimensionCount, bool Reverse>
+struct equals<Point, MultiPoint, point_tag, multi_point_tag, DimensionCount, Reverse>
+    : detail::equals::equals_by_relate<Point, MultiPoint>
+{};
 
 template <typename Box1, typename Box2, std::size_t DimensionCount, bool Reverse>
 struct equals<Box1, Box2, box_tag, box_tag, DimensionCount, Reverse>
@@ -285,7 +299,6 @@ struct equals<Segment1, Segment2, segment_tag, segment_tag, DimensionCount, Reve
 
 template <typename LineString1, typename LineString2, bool Reverse>
 struct equals<LineString1, LineString2, linestring_tag, linestring_tag, 2, Reverse>
-    //: detail::equals::equals_by_collection<detail::equals::length_check>
     : detail::equals::equals_by_relate<LineString1, LineString2>
 {};
 
--- a/include/boost/geometry/extensions/gis/io/wkb/detail/parser.hpp
+++ b/include/boost/geometry/extensions/gis/io/wkb/detail/parser.hpp
@@ -15,6 +15,8 @@
 #include <iterator>
 #include <limits>
 
+#include <boost/geometry/core/exception.hpp>
+
 #include <boost/concept_check.hpp>
 #include <boost/cstdint.hpp>
 #include <boost/type_traits/is_integral.hpp>
@@ -32,6 +34,23 @@
 
 namespace boost { namespace geometry
 {
+
+/*!
+\brief Read WKB Exception
+\ingroup core
+\details The read_wkb_exception is thrown when there is an error in wkb parsing
+ */
+class read_wkb_exception : public geometry::exception
+{
+public:
+
+    inline read_wkb_exception() {}
+
+    virtual char const* what() const throw()
+    {
+        return "Boost.Geometry Read WKB exception";
+    }
+};
 
 #ifndef DOXYGEN_NO_DETAIL
 namespace detail { namespace wkb
@@ -191,9 +210,12 @@ struct point_container_parser
         {
             return false;
         }
-
+        
         typedef typename std::iterator_traits<Iterator>::difference_type size_type;
-        BOOST_GEOMETRY_ASSERT(num_points <= boost::uint32_t( (std::numeric_limits<size_type>::max)() ) );
+        if(num_points > (std::numeric_limits<boost::uint32_t>::max)() )
+        {
+            throw boost::geometry::read_wkb_exception();
+        }
 
         size_type const container_size = static_cast<size_type>(num_points);
         size_type const point_size = dimension<point_type>::value * sizeof(double);
@@ -237,7 +259,11 @@ struct linestring_parser
             return false;
         }
 
-        BOOST_GEOMETRY_ASSERT(it != end);
+        if(it == end)
+        {
+            throw boost::geometry::read_wkb_exception();
+        }
+        
         return point_container_parser<L>::parse(it, end, linestring, order);
     }
 };
@@ -259,15 +285,15 @@ struct polygon_parser
         {
             return false;
         }
-
-        typedef typename ring_type<Polygon>::type ring_type;
+        
+        typedef typename boost::geometry::ring_return_type<Polygon>::type ring_type;
 
         std::size_t rings_parsed = 0;
         while (rings_parsed < num_rings && it != end)
         {
             if (0 == rings_parsed)
             {
-                ring_type& ring0 = exterior_ring(polygon);
+                ring_type ring0 = exterior_ring(polygon);
                 if (!point_container_parser<ring_type>::parse(it, end, ring0, order))
                 {
                     return false;
@@ -275,19 +301,16 @@ struct polygon_parser
             }
             else
             {
-                interior_rings(polygon).resize(rings_parsed);
-                ring_type& ringN = interior_rings(polygon).back();
+                boost::geometry::range::resize(interior_rings(polygon), rings_parsed);
+                ring_type ringN = boost::geometry::range::back(interior_rings(polygon));
+                
                 if (!point_container_parser<ring_type>::parse(it, end, ringN, order))
                 {
                     return false;
                 }
             }
-            ++rings_parsed;
-        }
 
-        if (num_rings != rings_parsed)
-        {
-            return false;
+            ++rings_parsed;
         }
 
         return true;
--- a/include/boost/geometry/extensions/multi/gis/io/wkb/detail/parser.hpp
+++ b/include/boost/geometry/extensions/multi/gis/io/wkb/detail/parser.hpp
@@ -1,0 +1,248 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+
+// Copyright (c) 2009-2012 Mateusz Loskot, London, UK.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_MULTI_IO_WKB_DETAIL_PARSER_HPP
+#define BOOST_GEOMETRY_MULTI_IO_WKB_DETAIL_PARSER_HPP
+
+#include <cstddef>
+#include <iterator>
+#include <limits>
+
+#include <boost/geometry/core/exception.hpp>
+
+#include <boost/geometry/extensions/gis/io/wkb/detail/endian.hpp>
+#include <boost/geometry/extensions/gis/io/wkb/detail/parser.hpp>
+#include <boost/geometry/extensions/gis/io/wkb/detail/ogc.hpp>
+
+
+#include <boost/geometry/multi/core/point_type.hpp>
+#include <boost/geometry/multi/core/ring_type.hpp>
+#include <boost/geometry/core/exterior_ring.hpp>
+
+#include <boost/geometry/multi/core/interior_rings.hpp>
+
+namespace boost { namespace geometry
+{
+
+#ifndef DOXYGEN_NO_DETAIL
+namespace detail { namespace wkb
+{
+
+template <typename MultiPoint>
+struct multipoint_parser
+{
+    template <typename Iterator>
+    static bool parse(Iterator& it, Iterator end, MultiPoint& multipoint, byte_order_type::enum_t order)
+    {
+        if (!geometry_type_parser<MultiPoint>::parse(it, end, order))
+        {
+            return false;
+        }
+        
+        boost::uint32_t num_points(0);
+        if (!value_parser<boost::uint32_t>::parse(it, end, num_points, order))
+        {
+            return false;
+        }
+        
+        // Check that the number of double values in the stream is equal
+        // or greater than the number of expected values in the multipoint
+        
+        typedef typename std::iterator_traits<Iterator>::difference_type size_type;
+        
+        typedef typename point_type<MultiPoint>::type point_type;
+        
+        size_type const container_byte_size = dimension<point_type>::value * num_points * sizeof(double)
+                                            + num_points * sizeof(boost::uint8_t)
+                                            + num_points * sizeof(boost::uint32_t);
+        
+        size_type const stream_byte_size = std::distance(it,end);
+        
+        if(stream_byte_size < container_byte_size)
+        {
+            throw boost::geometry::read_wkb_exception();
+        }
+        
+        point_type point_buffer;
+        std::back_insert_iterator<MultiPoint> output(std::back_inserter(multipoint));
+        
+        typedef typename std::iterator_traits<Iterator>::difference_type size_type;
+        if(num_points > (std::numeric_limits<boost::uint32_t>::max)() )
+        {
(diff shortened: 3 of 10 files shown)


meastp:
When implementing write, I noticed that the only way to get the Linestring type from a MultiLinestring that I could find, was to use range_value<T>. Is this correct?

mkaravel:
Yes.

mkaravel:
I have to admit that it might not be the only way, but it is definitely the typical way we get the linestring type from the multilinestring type.

awulkiew:
Thanks!

Actually to get a reference to an element of a Range, because I think this is what we need in most cases, more correct would be to use `range_reference<>`. Or pass a result of `boost::begin(rng) + index` or `bg::range::at(rng, index)` into a function tamplate taking a true reference and let the compiler handle it. I'm encouraging to use `range_reference<>` because some users already mentioned that they return non-true references in some cases. Even if it is not fully correct because then a Range iterator becomes an InputIterator and we require RandomAccessIterators, nevertheless in general Boost.Geometry currently works well with non-true references.

As for the `equals(MultiPoint, MultiPoint)` which AFAIU is not implemented yet. AFAIR the `equals_by_collection` works correctly only for Areal geometries, it's interesting that it compiles for MultiPoints. I think it would give wrong results if there were duplicated Points in the MultiPoint (AFAIR it's allowed), e.g. for

    MULTIPOINT(0 0, 1 1)
    MULTIPOINT(0 0, 0 0, 1 1)

the result of `equals()` should be `true`. But I think `relate()` is implemented for them now so you could try using `equals_by_relate` instead. Point/MultiPoint and MultiPoint/Point are also missing. And then also tests would be needed. Would you prefer if we handled missing `equals()` variants?

mkaravel:
Having multiples of points in a multipoint is definitely allowed in the sense the multipoint is considered as valid. I agree with Adam: `equals(MultiPoint, MultiPoint)` should rather be implemented using `equals_by_relate`.

meastp:
@awulkiew @mkaravel Thanks! I removed the FIXME's (Linestring type from MultiLinestring) and made an attempt to fix the equals algorithm. It seems to work well as the tests now fail when a Point's coordinates are not equal (whereas before the algorithm would only count Points).

@awulkiew As I'm only using range_value<...>::type, I don't see how your comment applies. I'm probably missing something. :)

More comments? :)

BTW: before a merge, I'd want to rebase this branch and fix the commits if that's okay.

include/boost/geometry/algorithms/equals.hpp
```diff
@@ -64,6 +65,16 @@ namespace boost { namespace geometry
...
+struct num_points_check
+{
+    template <typename Geometry1, typename Geometry2>
+    static inline bool apply(Geometry1 const& geometry1, Geometry2 const& geometry2)
+    {
+        return geometry::math::equals(
+                geometry::num_points(geometry1),
+                geometry::num_points(geometry2));
+    }
+};
```
awulkiew: Given that now `equals_by_relate` is used, is `num_points_check` still needed?

awulkiew:
@meastp Indeed you were only asking about the way of getting the type of a single Geometry from a MultiGeometry. I went a little further, because I thought about accessing an object and storing a reference to this object. Sorry, I should rather say something like: "yes, this is the only way, but take care about the references and if you'd like to store a reference then don't use `range_value<>::type&` but rather `range_reference<>::type`".

But I saw some problematic places in the WKB reader code. The problem is that the Geometries are used like STL Containers, but they shouldn't. AFAIU WKB reader should work for any Geometry, so any type adapted to a Concept, not only the Boost.Geometry models located in the `boost::geometry::model` namespace. Rings, Linestrings and MultiPoints are Ranges, not STL Containers. To access a type of an element or a reference `boost::range_value<>::type` or `boost::range_reference<>::type` must be used accordingly, not `value_type` and `reference` member types. For mutable Ranges of Points, `push_back()` or `resize()` member functions shouldn't be used because in general a Geometry may not define it. `bg::traits::push_back<>` and `bg::traits::resize<>` should be used instead, or one of the utilities defined in `geometry/util/range.hpp`, e.g. `bg::range::push_back()` which uses the trait mentioned above. This also means that Geometries in general won't work with `std::back_insertert_iterator<>`.

include/boost/geometry/extensions/gis/io/wkb/detail/parser.hpp
```diff
@@ -277,17 +277,14 @@ struct polygon_parser
             {
                 interior_rings(polygon).resize(rings_parsed);
                 ring_type& ringN = interior_rings(polygon).back();
+
```
awulkiew: This is a comment to the already existing code in a line above this one. Rings may not define `resize()` or `back()`. Please use e.g. `bg::range::resize()` and `bg::range::back()`.

include/boost/geometry/extensions/multi/gis/io/wkb/detail/parser.hpp
```diff
@@ -0,0 +1,237 @@
...
+        size_type const container_byte_size = dimension<point_type>::value * num_points * sizeof(double)
+                                            + num_points * sizeof(boost::uint8_t)
+                                            + num_points * sizeof(boost::uint32_t);
+        
+        size_type const stream_byte_size = std::distance(it,end);
+        
+        assert(stream_byte_size >= container_byte_size);
+        
+        point_type point_buffer;
+        std::back_insert_iterator<MultiPoint> output(std::back_inserter(multipoint));
```
awulkiew: A MultiPoint may not have an interface of a STLContainer and in particular may not define `push_back()` member function, which means that `std::back_insert_iterator<>` may not work.

However in various places of the library, e.g. in set operations like `union_()` `std::back_inserter(out)` is passed into the internals of the algorithm. This implies that an Output must define `push_back()` member function which means that this algorithm (and others) may not work well with an arbitrary MultiGeometry. Since similar code can be found in other places in the library it can stay as it is. Nobody complained about it so I'm guessing the users are either using STL Containers, Boost.Geometry models or just implement `push_back()` member function.

We could think about a solution later, e.g. to implement our own `back_inserter`, but not as a part of this PR.

include/boost/geometry/extensions/multi/gis/io/wkb/detail/parser.hpp
```diff
@@ -0,0 +1,237 @@
...
+    }
+};
+
+template <typename MultiPolygon>
+struct multipolygon_parser
+{
+    template <typename Iterator>
+    static bool parse(Iterator& it, Iterator end, MultiPolygon& multipolygon, byte_order_type::enum_t order)
+    {
+        typedef typename MultiPolygon::value_type polygon_type;
```
awulkiew: Should be `range_value<MultiPolygon>::type`.

include/boost/geometry/extensions/multi/gis/io/wkb/detail/parser.hpp
```diff
@@ -0,0 +1,237 @@
...
+                    
+                    if (!point_container_parser<ring_type>::parse(it, end, ring0, polygon_byte_order))
+                    {
+                        return false;
+                    }
+                }
+                else
+                {
+                    interior_rings(polygon_buffer).resize(rings_parsed);
+                    ring_type& ringN = interior_rings(polygon_buffer).back();
```
awulkiew: Should be `range::resize()` and `range::back()` and instead of `ring_type&` `range_reference<...>::type`.

include/boost/geometry/extensions/multi/gis/io/wkb/detail/parser.hpp
```diff
@@ -0,0 +1,237 @@
...
+            }
+            
+            std::size_t rings_parsed = 0;
+            
+            while (rings_parsed < num_rings && it != end)
+            {
+                
+                if (0 == rings_parsed)
+                {
+                    ring_type& ring0 = exterior_ring(polygon_buffer);
```
awulkiew: `range_reference<...>::type` or `ring_return_type<Polygon>::type`.

include/boost/geometry/extensions/multi/gis/io/wkb/detail/parser.hpp
```diff
@@ -0,0 +1,237 @@
...
+        
+        size_type const stream_byte_size = std::distance(it,end);
+        
+        assert(stream_byte_size >= container_byte_size);
+        
+        point_type point_buffer;
+        std::back_insert_iterator<MultiPoint> output(std::back_inserter(multipoint));
+        
+        typedef typename std::iterator_traits<Iterator>::difference_type size_type;
+        assert(num_points <= boost::uint32_t( (std::numeric_limits<size_type>::max)() ) );
```
awulkiew: Is it a good idea to cast a maximum value of `difference_type` into `uint32_t`? Isn't it a 64bit type on a x64 machine?

meastp:
@mkaravel 

@awulkiew : Finally I had time to address the comments. I have tried to correct all the comments, leaving back_inserter. I fully agree on writing against the Concept interface and not the (STL) implementation/model, of course.

Please take a look, and see if this is ready to merge. Before a merge, I would like to rebase and cleanup the commits/PR.

awulkiew:
@meastp Thanks!

I've prepared PR #340  adding `range::back_insert_iterator` and `range::back_inserter()` calling traits version of push_back. So let's wait for others to comment about it and if they agreed that it's ok you could replace `std::back_inserter()` with `range::back_insterer()`. I guess it could also be done later since now the users are most certainly defining `push_back()` member functions in their geometries because otherwise some of the algorithms doesn't work.

I'm also not sure if assertion failure is a good method of reporting when the format may be not suitable to represent data which is too big, container has too many elements, etc. Shouldn't rather an exception be thrown in this case?

meastp:
@awulkiew Great! Just tell me when it is merged to develop, and i will rebase and change to range::back_inserter() :) 

What exception should I throw, and what message (if there is a string member)?

awulkiew:
It should be an exception defined in Boost.Geometry in this file:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/core/exception.hpp
and should have a name probably corresponding to `std::length_error` and containing a word `exception` at the end because this is how all exceptions are called in the library, so e.g. `bg::length_exception`, `bg::invalid_length_exception` or some similar name.
Do you have some preference?
@barendgehrels what do you think?

There is one more issue. Raw `assert()` should be avoided in Boost code, `BOOST_ASSERT()` or `BOOST_ASSERT_MSG()` should be used instead.

awulkiew:
@barendgehrels Do you prefer some specific exception name?

meastp:
@awulkiew @barendgehrels 
I can see that PR #340 was merged. Did you agree on an exception text?

Should I use `BOOST_ASSERT()`/`BOOST_ASSERT_MSG()`, or an exception where `assert()` is currently used?

mloskot:
I'd like to follow @meastp excellent work, testing it, etc. So, I'm strongly interested in merging this PR.
Is it ready?

I also asked related question about extensions http://lists.boost.org/geometry/2016/03/3502.php as I'm interested in possibility to release the WKB I/O.

awulkiew:
If you ask me, I'm ok with merging the PR after the exception throw is added. This exception would be a part of an interface so we should choose the name wisely. You both are also members of Boost.Geometry community so your opinions are valuable. :)

Furthermore we should wait for @barendgehrels.

As for the name, I'm ok with both `length_exception` and `invalid_length_exception`, but if we decided to use one of them I'd prefer `length_exception` because it's shorter. If you have a different idea I'm open for suggestion. AFAIR `std::length_error` is thrown by STL only during resizing, when the new size is directly passed into a function, so it's an exception signalling that the argument is wrong. Here a container is also resized internally but an argument is geometry, so an exception should refer the geometry, i.e. indicate it contains too many points. `length_exception` might not be a good name, it might be confusing since length WRT geometry means something different. So it might be a better idea to be more specific and throw `write_wkb_exception`, so corresponding to `read_wkt_exception`.
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/io/wkt/read.hpp#L63
What do you think?

mloskot:
I have seen `length_error` is used in lots of places, reported when vector/list/map/string would become too large.

I might be misreading the C++11 standard, but there is this confusing statement that `length_error` is `logic_error` type with intention for "errors presumably detectable before the program executes", to report violations of implementation-defined limits. Whereas I see it in the standard library used in the actual runtime checks (`std::runtime_error` category).

However, since `geometry::exception` is derived from the general base `std::exception`, its subclass `read_wkt_exception` seems outside of the typical C++ exceptions categories. Thus, to be consistent with what is currently defined and for symmetry, I would use `write_wkb_exception : public geometry::exception`.

I hope my reasoning is clear and helpful.

barendgehrels:
Sorry for the late reaction. I'm OK with merging this. I am OK with adding the exception and write_wkb_exception sounds like the best choice for naming it.
Thanks for this work!

mloskot:
@barendgehrels Thanks!

@meastp I'd like to merge your PR. What do you think about the `write_wkb_exception` addition?

meastp:
@mloskot Great! I like the corrections for exceptions. I just haven't found time to do them yet :)

Also, I would like to rebase the pr before merge, to clean the commit log.

meastp:
I have finally converted the asserts to use of exceptions (I added read_wkb_exception and write_wkb_exception). Hopefully it is now ready to rebase and merge? @mloskot

awulkiew:
Thanks!

The new exceptions are added into `geometry/core/exception.hpp`, but WKB is not a part of official release. So either WKB should be moved from extensions or the exceptions should be moved to `wkb/write.hpp` and `wkb/read.hpp` like it is in case of `wkt/read.hpp` containing the definition of `read_wkt_exception`. Do you think WKB is ready to be moved from extensions and released?

Actually it's not clear for me if the exceptions, even in the final release, should be defined in `geometry/core/exception.hpp` because IOs are optional, and as I mentioned in case of WKT the exception is defined in `wkt/read.hpp`. But maybe indeed all exceptions should be moved into a single file. Anyway, this is not a subject of this PR.

Furthermore there is more information in `read_wkt_exception`, e.g. part of WKT string where the error was detected is included in the exception. But something like this could be added later if needed. Btw, is it possible to detect corrupted WKB?

Regarding rebasing. There is one change in this PR made for `bg::equals()`, adding suport for Multi-Geometries. This should be done in a separate commit.

mloskot:
> is it possible to detect corrupted WKB?

In general case, no, not really.
In cases of substantial WKB degradation, it is only possible to verify whether byte stream following declared number of points is at least of length necessary to store `sizeof(double) * number of X + sizeof(double) * number of Y [+ sizeof(double) * number of Z]`

awulkiew:
Or if e.g. geometry IDs are invalid or incompatible with Geometry passed into `read_wkb()`. Speaking of which. I see that `read_wkb()` returns `false` in this case, while `read_wkt()` throws an exception. If we decided to move WKB IO from extensions at some point in the future it should be consistent with WKT IO (one way or another, though in reality there is only one way because WKT IO is already released).

mloskot:
On 30 Jun 2016 19:02, "Adam Wulkiewicz" <notifications@github.com> wrote:
>
> Or if e.g. geometry IDs are invalid or incompatible with Geometry passed
into read_wkb().

Yes in BG case it makes sense.

> Speaking of which. I see that read_wkb() returns false in this case,
while read_wkt() throws an exception. If we decided to move WKB IO from
extensions at some point in the future it should be consistent with WKT

Good point. I agree.

meastp:
@awulkiew @mloskot Thanks -- I have moved the exception into parser.hpp. I will remember to move the bg::equals into a separate commit when rebasing. I think the error handling can be addressed in a separate PR, after this is merged..?

awulkiew:
@meastp Sure.

meastp:
@awulkiew @mloskot I have rebased. I think I've addressed all the issues so far. Is this ready to merge? :)

mloskot:
@awulkiew do you mind to merge? I can only hit the green button ATM, not sure if this is acceptable way to merge,

meastp:
Is there anything missing, preventing this from being merged? :)

awulkiew:
Ah, thanks for the reminder and sorry about delay. There are 2 things mentioned above which were not addressed i.e.:
- throwing an exception on error instead of returning `false`
- use of `bg::range::back_inserter` instead of `std::back_inserter`
but I'm ok with merging them as they are now (it's still an extension).

Barend what do you think?

A side note, this PR enables `bg::equals()` for MultiPoints. Could it be merged now or should it be merged into develop after develop is merged into master for 1.63?

meastp:
I agree with fixing the back_inserter now if you want.

Throwing exceptions will change the non-multi geometry wkb behaviour as well. Perhaps that could be done in a separate PR after this is merged?

meastp:
I agree with fixing the back_inserter now if you want.

Throwing exceptions will change the non-multi geometry wkb behaviour as well. Perhaps that could be done in a separate PR after this is merged?

mloskot:
Since I have failed to take care of completing the merge myself, I don't want to interfere with opinions of the current active maintainesr (Barend, Adam). However, AFAIU we agreed the exceptions can be improved in separate PR.

@awulkiew Thank you for taking care of this PR!

awulkiew:
@mloskot Yes, I said that I'm ok with merging earlier. I should have mentioned it again, sorry for confusion. Yes, I'm ok with merging the PR as it is. Though I'd like to hear from Barend too.

The biggest problem I have is related to the schedule of 1.63 release. Since not only the extension is modified but also `equals` I don't know if we should wait for develop being merged into master or could we merge it now and release the modified `equals` with 1.63. In other words should this change be treated as new functionality or as a bug fix.

barendgehrels:
Sorry that this process takes so long indeed. But yes, we cannot get it in 1.63 anymore, it is new functionality or at least a major change. It is not simply a bug fix. So let's plan to merge it as soon as 1.63 is released, to be sure we get it into 1.64

barendgehrels:
I'm OK with merging this - it is now certainly still possible for 1.64

PR #324, opened 2015-08-30
Split pj units at divisor, support fractional multipliers.

jeremy-murphy:
Is this what you had in mind, with regard to the TODO comment?

### DIFF
--- a/include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
+++ b/include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
@@ -41,6 +41,7 @@
 
 #include <boost/algorithm/string.hpp>
 #include <boost/range.hpp>
+#include <boost/lexical_cast.hpp>
 
 #include <boost/geometry/util/math.hpp>
 
@@ -208,15 +209,22 @@ inline parameters pj_init(R const& arguments, bool use_defaults = true)
         std::size_t const pos = s.find('/');
         if (pos == std::string::npos)
         {
-            pin.to_meter = std::atof(s.c_str());
+            pin.to_meter = lexical_cast<double>(s);
         }
         else
         {
-            std::string const numerator(s.substr(0, divisor)), denominator(s.substr(divisor + 1));
-            pin.to_meter = std::atof(numerator.c_str()) / std::atof(denominator.c_str());
+            double const numerator = lexical_cast<double>(s.substr(0, pos));
+            double const denominator = lexical_cast<double>(s.substr(pos + 1));
+            if (numerator == 0.0 || denominator == 0.0)
+            {
+                throw proj_exception(-99);
+            }
+            pin.to_meter = numerator / denominator;
         }
-        //if (*s == '/') /* ratio number */
-        //    pin.to_meter /= strtod(++s, 0);
+        if (pin.to_meter == 0.0)
+        {
+            throw proj_exception(-99);
+        }
         pin.fr_meter = 1. / pin.to_meter;
     }
     else
--- a/include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
+++ b/include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
@@ -205,9 +205,11 @@ inline parameters pj_init(R const& arguments, bool use_defaults = true)
 
     if (! s.empty())
     {
-        std::size_t const divisor = s.find('/');
-        if (divisor == std::string::npos)
+        std::size_t const pos = s.find('/');
+        if (pos == std::string::npos)
+        {
             pin.to_meter = std::atof(s.c_str());
+        }
         else
         {
             std::string const numerator(s.substr(0, divisor)), denominator(s.substr(divisor + 1));
--- a/include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
+++ b/include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
@@ -35,6 +35,7 @@
 #ifndef BOOST_GEOMETRY_PROJECTIONS_IMPL_PJ_INIT_HPP
 #define BOOST_GEOMETRY_PROJECTIONS_IMPL_PJ_INIT_HPP
 
+#include <cstdlib>
 #include <string>
 #include <vector>
 
@@ -204,8 +205,14 @@ inline parameters pj_init(R const& arguments, bool use_defaults = true)
 
     if (! s.empty())
     {
-        // TODO: IMPLEMENT SPLIT
-        pin.to_meter = atof(s.c_str());
+        std::size_t const divisor = s.find('/');
+        if (divisor == std::string::npos)
+            pin.to_meter = std::atof(s.c_str());
+        else
+        {
+            std::string const numerator(s.substr(0, divisor)), denominator(s.substr(divisor + 1));
+            pin.to_meter = std::atof(numerator.c_str()) / std::atof(denominator.c_str());
+        }
         //if (*s == '/') /* ratio number */
         //    pin.to_meter /= strtod(++s, 0);
         pin.fr_meter = 1. / pin.to_meter;


awulkiew:
Thanks for the PR! It looks good for me, besides that `atof` in general can return undefined value if the input is invalid. I'd consider using e.g. `boost::lexical_cast<>` throwing an exception. But this was also a problem in the original code.

With merging we must wait for Barend because he's the one who's porting the code from Proj4 and contributing back. This is also why he prefers to maintain the lines count, so this addition could be implemented as a free function. If defined in a separate header file it'd be reusable too.

And then instead of this:

    pin.to_meter = atof(s.c_str());

there could simply be:

    pin.to_meter = some_name(s);

But let's wait for Barend's response.

Btw, I'm guessing it's not the only one place where this is needed, right? In case I was right, is it already implemented in other places or are there also TODOs there?

jeremy-murphy:
Sure, throwing an exception is almost certainly better. Not sure off the top of my head if there are locale issues with respect to the decimal point: presumably it would be sufficient to ensure that the conversion uses the C locale and not the user's locale?

I think this is the only place that the conversion is needed, so I'm not sure what the issue of maintaining line count is about?

awulkiew:
Regarding the lines count. This code is a port of Proj4. It's simply convenient to have the algorithm looking exactly the same in Geometry's extensions and the original Proj4. So when the lines of the code of both libraries are traversed one by one it's easy to find correspondence. I can imagine that this simplifies the synchronization of both libraries in the case of updating Boost.Geometry from Proj4 and also contributing the bugfixes back to Proj4. AFAIR this is also what Barend have said in the past.

In this particular line the string is parsed and the number assigned. If the suggestion mentioned above was implemented (parsing function called) the code in both libraries would be very similar, with the only difference that in Boost.Geometry the string would be parsed differently.

I also know that Barend is currently working on projections. Barend would this change interfere with your work?

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
```diff
@@ -204,8 +205,14 @@ inline parameters pj_init(R const& arguments, bool use_defaults = true)
 
     if (! s.empty())
     {
-        // TODO: IMPLEMENT SPLIT
-        pin.to_meter = atof(s.c_str());
+        std::size_t const divisor = s.find('/');
+        if (divisor == std::string::npos)
```
barendgehrels: Please at {} in every if

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
```diff
@@ -204,8 +205,14 @@ inline parameters pj_init(R const& arguments, bool use_defaults = true)
...
+            pin.to_meter = std::atof(s.c_str());
+        else
+        {
+            std::string const numerator(s.substr(0, divisor)), denominator(s.substr(divisor + 1));
+            pin.to_meter = std::atof(numerator.c_str()) / std::atof(denominator.c_str());
```
barendgehrels: Please check for denominator=0

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
```diff
@@ -204,8 +205,14 @@ inline parameters pj_init(R const& arguments, bool use_defaults = true)
...
+        }
```
barendgehrels: Actually check for to_meter=0 should be done here too (not related to your changes)

barendgehrels:
I'm OK with merging. Thanks for the contribution.
@awulkiew  is right  that this all should stay close to the original. But the split was skipped at the time I ported it, and this change makes it closer to the original. 
Some of the files (as this one, all in impl) are manually converted. The rest (in proj) is automatically converted and should stay untouched.

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
barendgehrels: Maybe divisor is not the most clear name for this variable. Just pos like they are usually called?

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
jeremy-murphy: Sure, that of course can be done easily, but does it really make sense
given that the input is text from the source code? I mean it's not like the
usual run-time case where the user enters a bad string and can enter a good
one next time: if this fails it is because the string is wrong in the code,
right?

barendgehrels: It is a library, the string comes from somewhere. The user can enter the string as well.

jeremy-murphy: We're talking about the string that comes from the second column, `to_meter`, of `pj_units[]` in `pj_units.hpp`, right? I wouldn't think of that as a string the user enters though of course it is possible. If someone accidentally adds an entry with `100/0` then is run-time really the right place to catch it? Surely it can be caught with some compile-time check on the string literals?

barendgehrels: It can come from pj_units, but can also come from a line as
```bg::projections::parameters par = bg::projections::detail::pj_init_plus(parameters);```
where a user enters ``` to_meter=1/10```

Comment (jeremy-murphy) [inline include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp]: Ahh, OK, I missed that, thanks. (I also realized that compile-time checks in C++03 are trickier than I thought.) What should happen if either numerator or denominator are zero then? Throw an exception? I see a lot of magic exception numbers thrown throughout the code; are they documented?

**Comment by jeremy-murphy** [comment]
OK it throws `proj_exception(-99)` for zeroes -- that -99 is just a placeholder until I understand what should be there.

**Comment by barendgehrels** [comment]
You are right - the original code uses error-codes which are just converted to exceptions with those codes (also in the generation). To be decided. -99 is fine for now.

I'm OK with merging now.

File: include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp [hunk 4]
```diff
@@ -204,10 +206,23 @@ inline parameters pj_init(R const& arguments, bool use_defaults = true)
...
+        std::size_t const pos = s.find('/');
+        if (pos == std::string::npos)
+        {
+            pin.to_meter = lexical_cast<double>(s.c_str());
+        }
+        else
+        {
+            double const numerator = lexical_cast<double>(s.substr(0, divisor)), 
+                         denominator = lexical_cast<double>(s.substr(divisor + 1));
+            if (numerator == 0.0 || denominator == 0.0)
```
barendgehrels: lexical_cast is indeed nice.
but please add curly braces in every if...

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
```diff
@@ -204,10 +206,23 @@ inline parameters pj_init(R const& arguments, bool use_defaults = true)
...
-        // TODO: IMPLEMENT SPLIT
-        pin.to_meter = atof(s.c_str());
+        std::size_t const pos = s.find('/');
+        if (pos == std::string::npos)
+        {
+            pin.to_meter = lexical_cast<double>(s.c_str());
+        }
+        else
+        {
+            double const numerator = lexical_cast<double>(s.substr(0, divisor)),
```
barendgehrels: divisor is not changed to pos? Did you compile this code before commit?

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
```diff
@@ -204,10 +206,23 @@ inline parameters pj_init(R const& arguments, bool use_defaults = true)
...
-        pin.to_meter = atof(s.c_str());
+        std::size_t const pos = s.find('/');
+        if (pos == std::string::npos)
+        {
+            pin.to_meter = lexical_cast<double>(s.c_str());
+        }
+        else
+        {
+            double const numerator = lexical_cast<double>(s.substr(0, divisor)), 
+                         denominator = lexical_cast<double>(s.substr(divisor + 1));
```
barendgehrels: Sorry, did not mention this earlier, but in case of these kind of initializations we always declare on a separate line. So ```double const denumerator =``` etc

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
```diff
@@ -204,10 +206,23 @@ inline parameters pj_init(R const& arguments, bool use_defaults = true)
 
     if (! s.empty())
     {
-        // TODO: IMPLEMENT SPLIT
-        pin.to_meter = atof(s.c_str());
+        std::size_t const pos = s.find('/');
+        if (pos == std::string::npos)
+        {
+            pin.to_meter = lexical_cast<double>(s.c_str());
```
barendgehrels: .c_str() is probably redundant - at least you don't use it a few lines below

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
```diff
@@ -204,10 +206,23 @@ inline parameters pj_init(R const& arguments, bool use_defaults = true)
...
+        {
+            double const numerator = lexical_cast<double>(s.substr(0, divisor)), 
+                         denominator = lexical_cast<double>(s.substr(divisor + 1));
+            if (numerator == 0.0 || denominator == 0.0)
+                throw proj_exception(-99);
+            pin.to_meter = numerator / denominator;
+        }
         //if (*s == '/') /* ratio number */
         //    pin.to_meter /= strtod(++s, 0);
+        if (pin.to_meter == 0.0)
```
barendgehrels: Also here please add {}

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
```diff
@@ -204,10 +206,23 @@ inline parameters pj_init(R const& arguments, bool use_defaults = true)
...
+        }
+        else
+        {
+            double const numerator = lexical_cast<double>(s.substr(0, divisor)), 
+                         denominator = lexical_cast<double>(s.substr(divisor + 1));
+            if (numerator == 0.0 || denominator == 0.0)
+                throw proj_exception(-99);
+            pin.to_meter = numerator / denominator;
+        }
         //if (*s == '/') /* ratio number */
```
barendgehrels: This comment can go now

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
jeremy-murphy: Sure, OK.

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
jeremy-murphy: Ah, yes, I was a bit rushed, thanks.

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
jeremy-murphy: OK, I'll try to remember but I'll probably forget a lot just because it is not my habit. Why do you prefer it that way?

include/boost/geometry/extensions/gis/projections/impl/pj_init.hpp
jeremy-murphy: Ahhh, haha, I *thought* I compiled it but it is the recurring I problem I have with working on Boost that my development copy is not my 'working' copy. I don't like bjam all that much so I never learnt to use the test suites but I probably should.

jeremy-murphy:
I think this is finished now -- was there anything left that you're unsure of or want changed?

barendgehrels:
Yes, it is finished, thanks a lot for the PR and processing all the comments. Will merge.

PR #325, opened 2015-09-03
Fix/difference intersection linear areal

mkaravel:
The existing `bg::intersection(L, A)` and bg::difference(L, A)` algorithm discard spikes present in the linear geometry passed to the algorithm. This is not only inconsistent with the fact that Boost.Geometry considers linear geometries with spikes as valid, but can also produce invalid output.

To give an example of one such case where the output can be invalid, consider the linestring `LINESTRING(0 5,-5 5,0 5)` and the polygon `POLYGON((0 0,0 10,10 10,10 0,0 0))`. The existing code returns as output of `bg::difference()` the linestring `LINESTRING(0 5,0 5)` which is invalid (it is 0-dimensional instead of 1-dimensional).

Another interesting case is the test case where we have the linestring `LINESTRING(0 5,5 5,0 5)` and the polygon `POLYGON((0 0,0 10,10 10,10 0,0 0))`. The expected output is the empty multilinestring, but instead the linestring `LINESTRING(0 5,0 5)` is again reported.

As far as spikes are concerned, this is handled by specifying the last template parameter of `bg::detail::overlay::follow<>` to be `false` instead of `true` (the default value is `true` and signifies that spikes should be discarded).

As far as the second case is concerned, it is handled now by performing an initial scan of the turns computed, to determine is there are any turns that are considered as *crossing* turns, i.e., correspond to intersection points at which the linestring leaves the closure of the areal geometry and goes towards the free space, or vice versa. If no such turns are found, then the algorithm determines if the entire linear geometry lies inside the areal geometry or not, in exactly the same way as if no turns where found.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/intersection_insert.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/intersection_insert.hpp
@@ -208,11 +208,13 @@ struct intersection_of_linestring_with_areal
     class is_crossing_turn
     {
         // return true is the operation is intersection or blocked
-        template <std::size_t I, typename Turn>
+        template <std::size_t Index, typename Turn>
         static inline bool has_op_i_or_b(Turn const& t)
         {
-            return t.operations[I].operation == overlay::operation_intersection
-                || t.operations[I].operation == overlay::operation_blocked;
+            return
+                t.operations[Index].operation == overlay::operation_intersection
+                ||
+                t.operations[Index].operation == overlay::operation_blocked;
         }
 
         template <typename Turn>
@@ -238,7 +240,7 @@ struct intersection_of_linestring_with_areal
         }
 
         template <typename Turn>
-        static inline bool is_tab_or_mab(Turn const& t)
+        static inline bool has_i_or_b_ops(Turn const& t)
         {
             return
                 (t.method == overlay::method_touch
@@ -257,7 +259,7 @@ struct intersection_of_linestring_with_areal
         static inline bool apply(Turn const& t)
         {
             bool const is_crossing
-                = has_method_crosses(t) || is_cc(t) || is_tab_or_mab(t);
+                = has_method_crosses(t) || is_cc(t) || has_i_or_b_ops(t);
 #if defined(BOOST_GEOMETRY_DEBUG_FOLLOW)
             debug_turn(t, ! is_crossing);
 #endif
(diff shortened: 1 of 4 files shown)


mkaravel:
I am still working on this PR, as there seem to be some problems with my solution.
I will post a message once the PR is ready for review.

mkaravel:
The PR is now ready for review.

awulkiew:
Thanks! I'm of course ok with merging.

mkaravel:
@barendgehrels What do you think?

awulkiew:
AFAIU this PR is ready for merging. Barend are you ok with this?

include/boost/geometry/algorithms/detail/overlay/intersection_insert.hpp
```diff
@@ -175,21 +176,113 @@ template
...
+                 t.method == overlay::method_collinear)
+                &&
+                t.operations[0].operation == t.operations[1].operation
+                &&
+                t.operations[0].operation == overlay::operation_continue
+                ;
+        }
+
+        template <typename Turn>
+        static inline bool is_tab_or_mab(Turn const& t)
```
barendgehrels: What do you mean by tab or mab? touch/blocked/...

include/boost/geometry/algorithms/detail/overlay/intersection_insert.hpp
```diff
@@ -175,21 +176,113 @@ template
...
+                  << "; non-crossing? "
+                  << std::boolalpha << non_crossing << std::noboolalpha
+                  << std::endl;
+    }
+#endif
+
+    class is_crossing_turn
+    {
+        // return true is the operation is intersection or blocked
+        template <std::size_t I, typename Turn>
```
barendgehrels: Index

barendgehrels:
I'm OK with merging

include/boost/geometry/algorithms/detail/overlay/intersection_insert.hpp
mkaravel: This was a name used in an earlier version of this method.
I changed the name to something more descriptive.

include/boost/geometry/algorithms/detail/overlay/intersection_insert.hpp
mkaravel: Done.

mkaravel:
I have merged the PR. If the latest (naming) changes need further improvement please let me know.

barendgehrels:
Thanks, it is OK.

PR #326, opened 2015-09-16
Debug print

jeremy-murphy:
The use of `debug_print_complement_graph()` in `polygon.hpp`, even when debugging is turned off, requires `std::cout` and therefore `<iostream>` (which is not included). This how I have plugged the hole but I understand if you want to do it differently.

### DIFF
--- a/include/boost/geometry/algorithms/detail/is_valid/debug_complement_graph.hpp
+++ b/include/boost/geometry/algorithms/detail/is_valid/debug_complement_graph.hpp
@@ -55,11 +55,7 @@ debug_print_complement_graph(OutputStream& os,
     }
 }
 #else
-template <typename OutputStream, typename TurnPoint>
-inline void debug_print_complement_graph(OutputStream&,
-                                         complement_graph<TurnPoint> const&)
-{
-}
+#define debug_print_complement_graph(A, B)
 #endif
 
 


mkaravel:
Jeremy, I am not sure I fully understand the problem. If debugging is turned off, then the empty body function `debug_print_complement_graph()` is called, but this function does not need `std::cout` to be defined.

Could you please explain a bit more what the problem is?

mkaravel:
What I do understand though is that in regular mode, the definition of the class `complement_graph<>` needs to be defined and in fact it is not included. For this issue I would prefer to solve the problem (which is another problem) in a different way (instead of using a macro I mean).

jeremy-murphy:
Hi Menelaos. Yeah the problem is in combination with `boost/geometry/algorithms/detail/is_valid/polygon.hpp:330`:
```C++
debug_print_complement_graph(std::cout, g);
```
Even though the body of the debug function is empty, `std::cout` is still expected to be declared.

So whenever I include `boost/geometry/algorithms/is_valid.hpp`, I either must include `iostream` before it or disable this debug function somehow, otherwise I get a compilation error on `std::cout`:
```
In file included from .../boost/geometry/algorithms/is_valid.hpp:14:
In file included from .../boost/geometry/algorithms/detail/is_valid/implementation.hpp:15:
.../boost/geometry/algorithms/detail/is_valid/polygon.hpp:330:47: error: no member named 'cout' in namespace 'std'; did you mean 'count'?
```
Alternatively, I could have just commented out the call to `debug_print_complement_graph`, which would make no difference to me, but I thought this would at least not break for people actually using the function.

mkaravel:
@jeremy-murphy The problem is crystal clear now. I have created PR #327 for fixing the issue you reported in another way. If you agree you may close this PR and I will merge into develop PR #327.

jeremy-murphy:
Great, thanks!

PR #327, opened 2015-09-17
Fix issue with use of std::cout only in debug mode

mkaravel:
This is a follow up of PR #326 proposed by @jeremy-murphy.
The problem is that when `debug_print_complement_graph()` is called, `std::cout` needs to be defined
which requires the inclusion of `<iostream>` even in non-debug mode.

With this PR the call to `debug_print_complement_graph()` is guarded by the appropriate
macro and the use inclusion of `<iostream>` is no longer needed in non-debug mode.

### DIFF
--- a/include/boost/geometry/algorithms/detail/is_valid/polygon.hpp
+++ b/include/boost/geometry/algorithms/detail/is_valid/polygon.hpp
@@ -11,6 +11,9 @@
 #define BOOST_GEOMETRY_ALGORITHMS_DETAIL_IS_VALID_POLYGON_HPP
 
 #include <cstddef>
+#ifdef BOOST_GEOMETRY_TEST_DEBUG
+#include <iostream>
+#endif // BOOST_GEOMETRY_TEST_DEBUG
 
 #include <algorithm>
 #include <deque>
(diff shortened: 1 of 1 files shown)


jeremy-murphy:
If you're conditionally compiling the call of `debug_print_complement_graph()` then I think you don't need the non-debug definition of it in `debug_print_complement_graph.hpp` any more.

mkaravel:
@jeremy-murphy I thought about that before making the changes and decided to leave it as it is; if `debug_print_complement_graph()` is to be used somewhere else (admittedly not very likely) it is safer to have the non-debug definition available as well.

jeremy-murphy:
Sure, no worries. Looks good! Thanks for acting on it so quickly.

mkaravel:
@jeremy-murphy I will wait today in case somebody else wants to comment, and then merge tomorrow. Hope this is okay with you.

jeremy-murphy:
Still waiting for comment from others?

jeremy-murphy:
Thanks, Menelaos.

mkaravel:
@jeremy-murphy You are welcome. I am sorry for the delay.

PR #328, opened 2015-09-27
[transform][strategies] Don't clobber BOOST_UBLAS_TYPE_CHECK macro.

jeremy-murphy:
This macro is also defined in boost/numeric/ublas/detail/config.hpp.

I assume that this is the behaviour you want, rather than checking the previous value?


awulkiew:
Ok thanks!

PR #329, opened 2015-10-06
[algorithms][intersection] Fixes liang_barsky for integral data

springmeyer:
I've noticed that `bg::intersection` fails to produce correct results when the input geometry and clipping box use integral types for coordinates.

The reduced testcase at https://github.com/springmeyer/boost-geometry-line-clipping-testcase/blob/master/bg-testcase.cpp produces this result with current boost geometry:

```
LINESTRING(-10 0,10 0)
LINESTRING(-20 0)
```

And produces the correct results (they should be the same) with the fix:

```
LINESTRING(-10 0,10 0)
LINESTRING(-10 0,10 0)
```

Per https://github.com/mapbox/mapnik-vector-tile/pull/102 I found and have been using a workaround for this problem: using `linear_ring` as the clipping shape which invokes a different code path entirely.

This is my first time contributing to `boost::geometry` so let me know if you'd like this report somewhere else or any other things I can do to help. I presume that clipping a polygon with a `box` instead of a `linear_ring` will be slightly faster?

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/clip_linestring.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/clip_linestring.hpp
@@ -51,14 +51,14 @@ class liang_barsky
 private:
     typedef model::referring_segment<Point> segment_type;
 
-    template <typename T>
-    inline bool check_edge(T const& p, T const& q, T& t1, T& t2) const
+    template <typename PointType, typename CalcType>
+    inline bool check_edge(PointType const& p, PointType const& q, CalcType& t1, CalcType& t2) const
     {
         bool visible = true;
 
         if(p < 0)
         {
-            T const r = q / p;
+            CalcType const r = static_cast<CalcType>(q) / p;
             if (r > t2)
                 visible = false;
             else if (r > t1)
@@ -66,7 +66,7 @@ private:
         }
         else if(p > 0)
         {
-            T const r = q / p;
+            CalcType const r = static_cast<CalcType>(q) / p;
             if (r < t1)
                 visible = false;
             else if (r < t2)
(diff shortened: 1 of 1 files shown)


awulkiew:
Welcome and thanks for the fix!

So the problem is that the intermediate values are truncated to integral type, is that correct?

There is no need to report the issue elsewhere. We greately appreciate the PRs as the review process is very convenient.

Regarding the performance it looks like it should be faster since in the case of Polygon^Linestring the intersections of segments of both geometries would have to be calculated. This operation would have O(NlogN) complexity, where N is the number of points of a Polygon, the number of points of a Linestring would be constant. In the case of Polygon^Box it should be done in O(N).

include/boost/geometry/algorithms/detail/overlay/clip_linestring.hpp
```diff
@@ -51,14 +51,14 @@ class liang_barsky
 private:
     typedef model::referring_segment<Point> segment_type;
 
-    template <typename T>
-    inline bool check_edge(T const& p, T const& q, T& t1, T& t2) const
+    template <typename PointType, typename CalcType>
+    inline bool check_edge(PointType const& p, PointType const& q, CalcType& t1, CalcType& t2) const
```
awulkiew: I suggest to change the name of PointType to CoordinateType since neither `p` nor `q` are not Points.

include/boost/geometry/algorithms/detail/overlay/clip_linestring.hpp
```diff
@@ -86,9 +86,10 @@ class liang_barsky
     inline bool clip_segment(Box const& b, segment_type& s, bool& sp1_clipped, bool& sp2_clipped) const
     {
         typedef typename select_coordinate_type<Box, Point>::type coordinate_type;
+        typedef double calc_type;
```
awulkiew: Do here type `double` is always used as an intermediate calculation type. This may be a problem in the case of user-defined coordinate types which could have greater precision than `double`, in such a case the precision would be lost during casting done above. In such cases we use `select_most_precise<>` utility (see the `geometry/util` directory). Here it could be used like this:

    typedef typename select_most_precise<coordinate_type, double>::type calc_type;

springmeyer:
@awulkiew - thank you very much for the code review. I'm made both of the suggested changes in https://github.com/springmeyer/geometry/commit/6f214c54c61cac2863c295e4825eea3c4e9332a7 and confirmed that the testcase still returns the correct (fixed) results. Let me know if anything else needs to happen to land this?

awulkiew:
It's fine, thanks! Are everyone ok with merging?

barendgehrels:
Thanks for the fix! I'm OK with merging.

PR #330, opened 2015-10-08
Fix assertion failure in distance(Pt, Box) for NaN coordinates

awulkiew:
If there are NaN coordinates the assertions in distance(Pt, Box) cannot be reasonably checked, therefore this PR changes the conditions so if the original condition isn't met an additional check is performed if NaN coordinates were not involved. This is consistent with the general policy WRT invalid geometries, the algorithm doesn't fail but may generate invalid result.


mkaravel:
Looks good to me.
I am in favor of merging.

barendgehrels:
I'm OK too

PR #332, opened 2015-10-21
Fix special case of intersects(Segment, Box)

awulkiew:
In a special case when the difference between points' coordinates of a Segment is 0 and the Segment is disjoint the functio returns wrong result. E.g. for:

    SEGMENT(2 1 0,2 1 10)
    BOX(0 0 0,10 0 10)

See also:
http://stackoverflow.com/questions/32457920/boost-rtree-of-box-gives-wrong-intersection-with-segment

### DIFF
--- a/include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
+++ b/include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
@@ -128,7 +128,9 @@ struct disjoint_segment_box_impl
                   && t_min.first > ti_max)
                  ||
                  (geometry::math::equals(t_max.second, 0)
-                  && t_max.first < ti_min) )
+                  && t_max.first < ti_min)
+                 ||
+                 (math::sign(ti_min) * math::sign(ti_max) > 0) )
             {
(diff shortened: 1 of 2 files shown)


mkaravel:
I have not really looked into the details, but I am okay with merging this.

awulkiew:
I'd appreciate if you double checked me with this.

The t parameters present in the original algorithm are stored as numerator and denominator pairs (rational). If a t parameter is infinite the denominator equals 0 which is checked above the code I altered.

As for the special case. If a segment is parallel to the box's face and box's edge perpendicular to this face has length = 0 (in other words it's a flat box and segment parallel to it) then t parameters will be equal to positive or negative infinity depending on the side on which the segment lies WRT box. The conditions added in this PR checks if both infinity values has the same sign and returns if they have.

mkaravel:
@awulkiew Okay. Will take a closer look and get back to it.

awulkiew:
@barendgehrels are you ok with merging it?

barendgehrels:
Yes, I am OK

PR #333, opened 2015-10-21
[geometry][algorithm] For Bug Ticket #11710 

gagan1kumar:
__is_simple__ - now return __true__ for all empty geometry

### DIFF
--- a/include/boost/geometry/algorithms/detail/is_simple/interface.hpp
+++ b/include/boost/geometry/algorithms/detail/is_simple/interface.hpp
@@ -31,6 +31,8 @@ struct is_simple
     static inline bool apply(Geometry const& geometry)
     {
         concept::check<Geometry const>();
+        if(boost::geometry::is_empty(geometry))
+            return true;
         return dispatch::is_simple<Geometry>::apply(geometry);
     }
 };


mkaravel:
@gagan1kumar  Thank you for the PR. AFAIR, only empty multi-geometries are considered valid, and in that respect it would make sense to me to be consistent with that approach. I would return `true` only for empty multi-geometries.

mkaravel:
@gagan1kumar I have addressed the issue in PR #334 (only for multi-geometries as I describe in my previous comment). Empty linestrings, rings and polygons are currently considered as invalid by `bg::is_valid()` which means that the behavior of `bg::is_simple()` for those inputs is undefined. This is a general rule we follow in BG: algorithms (with the obvious exception of `bg::is_valid()`) are assuming valid input; if invalid input is passed through, the behavior of the BG algorithm is defined, and in general we do our best to produce a sensible result.

Could you please close this PR?

barendgehrels:
OK for me to close.
Besides this (note to @gagan1kumar ): never create PR from/for master, always for develop branch
But thanks anyway for this PR (even if it is fixed otherwise)

PR #334, opened 2015-11-05
Boost.Trac tickets 11709 and 11710

mkaravel:
This PR addresses two Boost.Trac tickets: [11709](https://svn.boost.org/trac/boost/ticket/11709) and [11710](https://svn.boost.org/trac/boost/ticket/11710).

###### Ticket [11709](https://svn.boost.org/trac/boost/ticket/11709)
`bg::is_simple()` segfaults if given an empty linestring. In this PR the linestring is first checked for emptiness (before any further computations are performed) and `false` is returned is the linestring is found to be empty. Notice that the actual return value does not really matter, as empty linestrings are considered invalid, which means that the behavior of `bg::is_simple()` is undefined for them (the same holds actually for any BG algorithm operating on invalid input).

###### Ticket [11710](https://svn.boost.org/trac/boost/ticket/11710)

`bg::is_simple()` used to return `false` for empty multi-geometries. This was inconsistent with the behavior of `bg::is_valid()` which accepts empty multi-geometries as valid. In this PR we change this behavior and `bg::is_simple()` returns `true` for empty multi-geometries.

### DIFF
--- a/include/boost/geometry/algorithms/detail/is_simple/multipoint.hpp
+++ b/include/boost/geometry/algorithms/detail/is_simple/multipoint.hpp
@@ -40,9 +40,9 @@ struct is_simple_multipoint
 {
     static inline bool apply(MultiPoint const& multipoint)
     {
-        if ( boost::size(multipoint) == 0 )
+        if (boost::empty(multipoint))
         {
-            return false;
+            return true;
         }
 
         MultiPoint mp(multipoint);
--- a/include/boost/geometry/algorithms/detail/is_simple/areal.hpp
+++ b/include/boost/geometry/algorithms/detail/is_simple/areal.hpp
@@ -40,11 +40,11 @@ struct is_simple_ring
     static inline bool apply(Ring const& ring)
     {
         simplicity_failure_policy policy;
-        return
-            !detail::is_valid::has_duplicates
-                <
-                    Ring, geometry::closure<Ring>::value
-                >::apply(ring, policy);
+        return ! boost::empty(ring)
+            && ! detail::is_valid::has_duplicates
+                    <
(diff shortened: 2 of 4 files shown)


awulkiew:
I'm of course ok with merging. But I'm wondering, maybe we should return `true` for all empty "Range-based" geometries so also Linestrings, Rings and Polygons in case if we decided at some point to consider them as valid?

mkaravel:
I deliberated (before your comment) about that and decided that `false` is a bit safer: if `bg::is_simple()` were to return `true` you would be more inclined to believe/expect that they are also valid. This way would entirely miss that `bg::is_valid()` returns `false` for these three geometries and this could create a confusion.

I agree that if we switch to accept the empty linestring, ring and polygon as valid geometries, we would also have to make sure that `bg::is_simple()` is changed accordingly (so more work at that point). I would prefer to postpone this change for that moment rather that do it now for the reason I explain above.

include/boost/geometry/algorithms/detail/is_simple/multipoint.hpp
```diff
@@ -40,9 +40,9 @@ struct is_simple_multipoint
 {
     static inline bool apply(MultiPoint const& multipoint)
     {
-        if ( boost::size(multipoint) == 0 )
+        if (boost::empty(multipoint))
         {
-            return false;
+            return true;
```
barendgehrels: I agree with this, an empty multi-geometry should be considered as simple.

barendgehrels:
I agree with your reasoning and the commit. OK with merging. Thanks.

mkaravel:
@awulkiew @barendgehrels I will wait a bit more in case there are other comments and will merge into develop this weekend.

PR #335, opened 2015-11-06
Boost.Trac ticket 11711

mkaravel:
This PR partially addresses Boost.Trac ticket [11711](https://svn.boost.org/trac/boost/ticket/11711): "is_simple and is_valid - Points that are NaN or Infinity".

**Problem:** The problem with the existing `bg::is_valid()` implementation is that it may not handle properly geometries that have invalid floating-point coordinates, namely coordinates that are NaN or infinite. For such geometries `bg::is_valid()` may return `true`, which should not be the case.

**Fix:** In this PR the coordinates of a geometry whose coordinate type is a fundamental floating-point number are checked. If at least one coordinate is found to be NaN or infinite, the geometry is considered as invalid.
Note that for non-fundamental number types or fundamental non-floating-point types the check is essentially deactivated: fundamental non-floating-point types cannot have NaN or infinite values, whereas for non-fundamental number types a mechanism similar to `std::numeric_limits<>` needs to be present to perform similar checks (and such a mechanism is not present currently).

**is_simple:** The Boost.Trac ticket asks for `bg::is_simple()` to return `false` for geometries that have invalid coordinates. This is not implemented in this PR. As it is the case with almost all BG algorithms (the only exception is `bg::is_valid()`) the geometric input is assumed to be valid, and validity of the input is not checked internally in the algorithm. According to this rule, `bg::is_simple()` should not check for invalid coordinates as suggested in the Boost.Trac ticket; when called with an invalid geometry as input (as it is the case in the ticket) the behavior of `bg::is_simple()` is simply undefined.

### DIFF
--- a/include/boost/geometry/algorithms/detail/is_valid/has_invalid_coordinate.hpp
+++ b/include/boost/geometry/algorithms/detail/is_valid/has_invalid_coordinate.hpp
@@ -1,0 +1,154 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+
+// Copyright (c) 2014-2015, Oracle and/or its affiliates.
+
+// Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
+// Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
+
+// Licensed under the Boost Software License version 1.0.
+// http://www.boost.org/users/license.html
+
+#ifndef BOOST_GEOMETRY_ALGORITHMS_DETAIL_IS_VALID_HAS_INVALID_COORDINATE_HPP
+#define BOOST_GEOMETRY_ALGORITHMS_DETAIL_IS_VALID_HAS_INVALID_COORDINATE_HPP
+
+#include <cstddef>
+
+#include <boost/type_traits/is_floating_point.hpp>
+
+#include <boost/geometry/core/coordinate_type.hpp>
+#include <boost/geometry/core/point_type.hpp>
+
+#include <boost/geometry/util/has_nan_coordinate.hpp>
+#include <boost/geometry/util/has_infinite_coordinate.hpp>
+
+#include <boost/geometry/iterators/point_iterator.hpp>
+#include <boost/geometry/views/detail/indexed_point_view.hpp>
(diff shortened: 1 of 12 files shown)


include/boost/geometry/algorithms/detail/is_valid/has_invalid_coordinate.hpp
```diff
@@ -0,0 +1,154 @@
...
+            (! visitor.template apply<failure_invalid_coordinate>())
+            :
+            (! visitor.template apply<no_failure>());
+    }
+
+    template <typename Point>
+    static inline bool apply(Point const& point)
+    {
+        return geometry::has_nan_coordinate(point)
+            || geometry::has_infinite_coordinate(point);
```
awulkiew: If `isfinite()` was used here instead, internally only 2 conditions would have to be checked per coordinate (`x >= min` and `x <= max`). With `isnan()` and `isinf()` more operations have to be performed (at least 3 condition checks and additional bitwise operations for ieee floats). Would it have sense to replace this?

mkaravel: Done. See commit 348b7f7.

awulkiew:
Ok, thanks. I'm only wondering if instead of `has_non_finite_coordinate` the name `has_finite_coordinates` would be a better pick. It's because it would be a direct reference to the C++11 standard and Boost.Math `isfinite`. Besides that, I'm ok with merging.

mkaravel:
@awulkiew It was the easiest modification to do. Implementing `has_finite_coordinates` would imply quite a few more changes in the algorithm logic, which is why I chose to do it this way.
Let me know what you prefer. I can implement `has_finite_coordinates`, but this would mean even more code, and I am not sure it is worth it. But it is really up to you.

awulkiew:
@mkaravel I'm ok with `has_non_finite_coordinate` :)

mkaravel:
@barendgehrels Are you okay with merging?

mkaravel:
@barendgehrels I would like to merge this PR so that fix gets into 1.60
Okay to merge?

barendgehrels:
I'm OK, thanks for the PR

PR #336, opened 2015-11-11
[algorithms][centroid] Fix crash when multi-geometry contains empty geometry

talaj:
Fixes seg-faulting of centroid algorithm on multi-geometries when sub-geometry is empty. Test case is included.

### DIFF
--- a/include/boost/geometry/algorithms/centroid.hpp
+++ b/include/boost/geometry/algorithms/centroid.hpp
@@ -220,19 +220,22 @@ struct centroid_range_state
         iterator_type it = boost::begin(view);
         iterator_type end = boost::end(view);
 
-        typename PointTransformer::result_type
-            previous_pt = transformer.apply(*it);
-
-        for ( ++it ; it != end ; ++it)
+        if (it != end)
         {
             typename PointTransformer::result_type
-                pt = transformer.apply(*it);
+                previous_pt = transformer.apply(*it);
 
-            strategy.apply(static_cast<point_type const&>(previous_pt),
-                           static_cast<point_type const&>(pt),
-                           state);
-            
-            previous_pt = pt;
+            for ( ++it ; it != end ; ++it)
+            {
+                typename PointTransformer::result_type
+                    pt = transformer.apply(*it);
+
+                strategy.apply(static_cast<point_type const&>(previous_pt),
+                               static_cast<point_type const&>(pt),
+                               state);
+
+                previous_pt = pt;
+            }
         }
     }
 };
--- a/test/algorithms/centroid.cpp
+++ b/test/algorithms/centroid.cpp
@@ -175,6 +175,21 @@ void test_exceptions()
     test_centroid_exception<bg::model::linestring<P> >();
     test_centroid_exception<bg::model::polygon<P> >();
     test_centroid_exception<bg::model::ring<P> >();
+
+    // Empty exterior ring
+    test_centroid_exception<bg::model::polygon<P> >(
+        "POLYGON((), ())");
+    test_centroid_exception<bg::model::polygon<P> >(
+        "POLYGON((), (0 0, 1 0, 1 1, 0 1, 0 0))");
+}
+
+template <typename P>
+void test_empty()
+{
+    // Empty interior ring
+    test_centroid<bg::model::polygon<P> >(
+        "POLYGON((0 0, 1 0, 1 1, 0 1, 0 0), ())",
+        0.5, 0.5);
 }
 
 void test_large_integers()
@@ -255,6 +270,7 @@ int test_main(int, char* [])
     test_large_doubles();
 
     test_exceptions<bg::model::d2::point_xy<double> >();
+    test_empty<bg::model::d2::point_xy<double> >();
 
     return 0;
 }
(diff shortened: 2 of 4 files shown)


awulkiew:
Thanks!

It's a case when an element of multi-geometry is invalid/empty. Typically centroid() throws for empty input (if BOOST_GEOMETRY_CENTROID_NO_THROW is not defined), though in this case only a part of an input geometry is "empty". Strictly speaking the multi-geometry is invalid but of course for invalid input a fuction shouldn't crash. So such empty multi-geometry elements would be simply ignored. I'm wondering, maybe the function should throw an exception also in this case?

talaj:
@awulkiew - I think the centroid algorithm should not throw on empty sub-geometries, because if we understand centroid as synonym for [center of mass](https://github.com/boostorg/geometry/blob/1500dc364452c4219e00b6d241cfe71362b88b33/include/boost/geometry/algorithms/centroid.hpp#L586) it makes sense to just ignore it as it doesn't add any weight to the calculation.

mkaravel:
Suppose we have a polygon that has a valid exterior ring and an empty interior ring; the polygon is obviously invalid, but we can really compute its centroid.

What I am trying to say here is that the prevailing approach of BG with respect to invalid geometries is of course not to crash, but also compute something as meaningful as possible. Following this "rule" I think that, in principle, when computing the centroid of a geometry with empty sub-geometries, the empty sub-geometries should be ignored.

However, let's look the above "rule" in more detail. Below I am considering the possible cases with respect to empty sub-geometries, and state my proposal regarding how we should treat them (please note that all geometries below might be invalid, but can still come up as input in centroid):
* If the number of points of a geometry is 0, an exception should be thrown (for example when we have `MULTILINESTRING((),())').
* If a multi-linestring with at least one point has one or more empty linestrings, ignore them.
* If a polygon has an empty exterior ring, throw an exception (I think this is the current behavior actually).
* If a polygon has a non-empty exterior ring and at least one empty interior ring, ignore the interior ring(s).
* If a multi-polygon has one or more empty polygons (i.e., polygons with WKT: `POLYGON(())`) then the polygon(s) should be ignored.
* For all other cases, the behavior of the centroid algorithm on multi-polygons really depends on the behavior of the centroid algorithm on the individual polygons contained in the multi-polygon (please correct me if I am wrong on that...).

talaj:
@mkaravel - I have added commit with tests covering your list. Only one test case is not passing - multi-polygon with empty exterior ring. Such geometry should throw, but does not currently. Interesting is that single polygon with empty exterior is throwing.

awulkiew:
@talaj Yes, that's because the emptiness is checked on a higher level. I guess we could change it and therefore check conditions deeper in the algorithm. Though I'm ok with merging the PR as it is.

talaj:
It's questionable whether polygon with empty exterior ring should throw because is empty or because is invalid.

Should following multi-polygon throw?

```
MULTIPOLYGON(((4 1,4 3,8 3,8 1,4 1)), ((), (0 0, 1 0, 1 1, 0 1, 0 0)))
```

mkaravel:
@talaj According to the rules I wrote it should throw (and this was the intention). I think the exception to be thrown is the `centroid_exception` (which does not really specify anything else other than the fact that the centroid algorithm cannot compute the centroid).

awulkiew:
@mkaravel And "cannot compute the centroid" means that nothing was assigned to the output. I'm wondering, are there any reasons why doesn't this function return `bool` indicating if the result was calculated correctly?

mkaravel:
@awulkiew Good question. I guess the only answer I can come up with is that it would complicate the return type for `return_centroid()` as it would have to be an `std::pair<Point, bool>` in that case. But this is just a guess.

awulkiew:
@mkaravel Well, `return_centroid()` could check the result and throw. The same is true for `envelope()`. I guess a design decision was made some time ago that all functions should work like this. The only exception I found is `intersection()` which does return `bool` however AFAIU it always returns `true` for combinations of geometries other than Box/Box. Anyway, we probably shouldn't change the behavior of `centroid()` now.

mkaravel:
@awulkiew You mean not change the behavior with respect the return type; I agree.
I think we should still change the behavior for multi-polygons that have polygons with empty exterior rings.

awulkiew:
@mkaravel Yes, I was talking about the return type. As for the empty exterior rings, I have nothing against it. My point is that it could be done as a part of this PR or later. For now this PR is a bugfix for segfult. Should the behavior change be a part of it too? @talaj would you like to do it?

talaj:
> would you like to do it?

@awulkiew - Yes, I would like. I will have time next week if you don't mind. I think it should be part of this pull request as tests I have added here are failing because of it.

barendgehrels:
@awulkiew @talaj @mkaravel First, to be clear, please don't change the return type indeed. If it is about exterior rings, OK for me if everyone agrees.
Second: be aware that the branch closes soon (for bugfixes only, other changes are not allowed anymore) - if this change has to be in, and it is considered as a bugfix, it has to be soon. 
Anyway, I prefer merging the bugfix first, and do any new actions in a separte PR

barendgehrels:
@talaj Thanks for the bugfix!

talaj:
@barendgehrels - Ok, I removed failing test case with empty exterior ring so all tests are passing now and this pull request should be ready to merge.

I will try to solve the exterior ring issue in a separate pull request.

barendgehrels:
OK, thanks - everyone OK with merging?

awulkiew:
@barendgehrels Yes.

mkaravel:
@barendgehrels @talaj @awulkiew 
Yes for me as well.

barendgehrels:
OK, merged, thanks

PR #338, opened 2016-01-17
Fix for andoyer antipodal distance, Ticket #11917

kenba:
Calculate distance between antipodal points, i.e.  when C is zero.

### DIFF
--- a/include/boost/geometry/strategies/geographic/distance_andoyer.hpp
+++ b/include/boost/geometry/strategies/geographic/distance_andoyer.hpp
@@ -132,13 +132,21 @@ private :
         CT const c2 = 2;
         CT const c3 = 3;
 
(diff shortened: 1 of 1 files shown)


awulkiew:
Thanks for the PR!

There is an implementation of Andoyer inverse formula capable to calculate both distance and azimuth:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/andoyer_inverse.hpp
I also handled antipodal points there in #339.

So an alternative to this PR could be to use thie fixed `andoyer_inverse` formula in `andoyer` strategy.

awulkiew:
The implementation of `andoyer` distance strategy was replaced with the new one so this fix is not needed.

PR #339, opened 2016-01-18
Fix andoyer

awulkiew:
This PR is related to #338

It fixes `andoyer_inverse` formula for points in antipodal areas. This formula can be used to calculate distances and azimuths between points using first order Andoyer-Lambert-Forsyth type approximation. Besides fixed distances the azimuths for points in the vincinity of antipodal point are convergent to the azimuth calculated for antipodal point.

This formula may be used in the Andoyer strategy which currently uses separate, older implementation.


awulkiew:
`andoyer_inverse` formula is now used in the implementation of `andoyer` distance strategy instead of the original implementation.

PR #341, opened 2016-02-04
Enable rescale policy only in cartesian coordinate system.

awulkiew:
This PR fixes the regression reported here:
https://svn.boost.org/trac/boost/ticket/11789

An assertion fails in `expand()` called from `sectionalize()` for non-cartesian geometries. It's because it is called for Points containing very big integral coordinates. The reason is that currently, by default, floating-point coordinates are rescaled into integral coordinates internally in set operations and these integral coordinates are passed into expand() to calculate bounding boxes. This also means that since these coordinates are normalized, the result is not correct. 

After merging this PR rescaling will be enabled only for geometries in cartesian CS.

### DIFF
--- a/include/boost/geometry/policies/robustness/get_rescale_policy.hpp
+++ b/include/boost/geometry/policies/robustness/get_rescale_policy.hpp
@@ -252,9 +252,15 @@ struct rescale_policy_type
         false
 #else
         boost::is_floating_point
-        <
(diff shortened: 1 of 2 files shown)


awulkiew:
@barendgehrels are you ok with this change?

barendgehrels:
I am OK with merging this, thanks for the change

PR #342, opened 2016-02-09
[io] Support all geometries in svg() and use it in svg_mapper.

awulkiew:
SVG tag for Segment in `svg()` is `<line>`. This function is called by `svg_mapper`. The implementation of `svg_map` previously used only for Box is now used for Segment as well. Previously it was a separate implementation transforming Segments into Linestrings. This means that `svg_mapper` now outputs Segments as SVG Lines (`<line>`) instead of PolyLines (`<polyline>`).


barendgehrels:
Thanks! It looks good, I am ok with merging

PR #343, opened 2016-03-03
Fix winding strategy for spherical and geographic system.

awulkiew:
In spherical and geographic CS if a point has the same latitude as both points of a segment it doesn't mean that it lies on the segment. This PR fixes this case by scanning dimension 0 (X) instead of dimension 1 (Y). In spherical and geographic it's guaranteed that if longitude is the same a point lies on a segment.

Furthermore special checks are added for points lying on poles and segments going through poles.

IMPORTANT: The change of dimensions mentioned above influences buffer implementation which uses winding strategy directly and assume that this strategy scans in a specific dimension.


barendgehrels:
Thanks Adam! Also for adapting the buffer.
I'm OK with merging this.

PR #345, opened 2016-03-09
Point/Box and Box/Box spatial relations in non-cartesian coordinate systems.

awulkiew:
This PR adds or changes the following:
- support spherical/geographic CS in `box_in_box` strategy and enable it as the default one
- support spherical/geographic CS in `point_in_box` strategy and enable it as the default one
- fix `disjoint()`/`intersects()` for point/box in spherical/geographic CS (using tools implemented for `point_in_box` strategy)
- fix `disjoint()`/`intersects()` for box/box in spherical/geographic CS (separate implementation)

This PR changes the way how `within(point, box)` and `covered_by(point, box)` is calculated by default. The reason is that a Box in spherical and geographical CS is not a Polygon. The differences are:
- the horizontal edge of a Box may be longer than 180 deg
- the horizontal edge of a Box is not defined by geodesic but by small circle

`point_in_box_by_side` is no longer used by default. It returns wrong result for Boxes with edges longer than 180 deg. It's because when a side is calculated the closer direction is picked so for those long edges (length in (180, 360]) some sides are the opposite than expected.

This PR also fixes the ticket: https://svn.boost.org/trac/boost/ticket/11987

Since `point_in_box` and `box_in_box` now support all CSes I plan to move them to `strategies/agnostic` directory. This is not a part of this PR to make the review of the code more convenient, to avoid displaying the unchanged parts of implementations as new ones on GitHub.

### DIFF
--- a/include/boost/geometry/algorithms/detail/disjoint/box_box.hpp
+++ b/include/boost/geometry/algorithms/detail/disjoint/box_box.hpp
@@ -104,16 +104,10 @@ struct box_box<Box1, Box2, 0, DimensionCount, spherical_tag>
         calc_t const diff2 = b2_max - b2_min;
 
         // check the intersection if neither box cover the whole globe
-        if (math::smaller(diff1, constants::period())
-         && math::smaller(diff2, constants::period()) ) // < period
+        if (diff1 < constants::period() && diff2 < constants::period())
         {
             // calculate positive longitude translation with b1_min as origin
-            calc_t const c0 = 0;
-            calc_t diff_min = b2_min - b1_min;
-            math::normalize_longitude<units_t, calc_t>(diff_min);
-            if (diff_min < c0) // [-180, 180] -> [0, 360]
-                diff_min += constants::period();
-        
+            calc_t const diff_min = math::longitude_distance_unsigned<units_t>(b1_min, b2_min);
             calc_t const b2_min_transl = b1_min + diff_min; // always right of b1_min
 
             if (b2_min_transl > b1_max                                // b2_min right of b1_max
--- a/include/boost/geometry/algorithms/detail/disjoint/box_box.hpp
+++ b/include/boost/geometry/algorithms/detail/disjoint/box_box.hpp
@@ -5,8 +5,8 @@
 // Copyright (c) 2009-2015 Mateusz Loskot, London, UK.
 // Copyright (c) 2013-2015 Adam Wulkiewicz, Lodz, Poland.
 
-// This file was modified by Oracle on 2013-2015.
-// Modifications copyright (c) 2013-2015, Oracle and/or its affiliates.
+// This file was modified by Oracle on 2013-2016.
+// Modifications copyright (c) 2013-2016, Oracle and/or its affiliates.
 
 // Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
 // Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
@@ -28,6 +28,9 @@
 
 #include <boost/geometry/algorithms/dispatch/disjoint.hpp>
 
+#include <boost/geometry/util/normalize_spheroidal_coordinates.hpp>
+#include <boost/geometry/util/select_most_precise.hpp>
+
 
 namespace boost { namespace geometry
 {
@@ -39,7 +42,13 @@ namespace detail { namespace disjoint
 template
 <
     typename Box1, typename Box2,
-    std::size_t Dimension, std::size_t DimensionCount
+    std::size_t Dimension = 0,
+    std::size_t DimensionCount = dimension<Box1>::value,
+    typename CSTag = typename tag_cast
+                        <
+                            typename cs_tag<Box1>::type,
+                            spherical_tag
+                        >::type
 >
 struct box_box
 {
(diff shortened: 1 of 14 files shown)


include/boost/geometry/algorithms/detail/disjoint/box_box.hpp
```diff
@@ -72,6 +81,57 @@ struct box_box<Box1, Box2, DimensionCount, DimensionCount>
...
+
+        // check the intersection if neither box cover the whole globe
+        if (math::smaller(diff1, constants::period())
+         && math::smaller(diff2, constants::period()) ) // < period
+        {
+            // calculate positive longitude translation with b1_min as origin
+            calc_t const c0 = 0;
+            calc_t diff_min = b2_min - b1_min;
+            math::normalize_longitude<units_t, calc_t>(diff_min);
+            if (diff_min < c0) // [-180, 180] -> [0, 360]
```
barendgehrels: Please please add parentheses here...

include/boost/geometry/strategies/cartesian/box_in_box.hpp
```diff
@@ -61,9 +65,94 @@ struct box_covered_by_range
...
+        typedef typename coordinate_system<Geometry>::type::units units_t;
+        typedef math::detail::constants_on_spheroid<calc_t, units_t> constants;
+
+        // min <= max <=> diff >= 0
+        calc_t const diff_ed = bed_max - bed_min;
+        calc_t const diff_ing = bing_max - bing_min;
+        // if containing is smaller it cannot contain
+        // or interiors doesn't overlap (no interior in contained)
+        if (diff_ing < diff_ed || diff_ed == 0)
+            return false;
```
barendgehrels: Parentheses missing

include/boost/geometry/strategies/cartesian/box_in_box.hpp
```diff
@@ -61,9 +65,94 @@ struct box_covered_by_range
...
+
+        // if containing covers the whole globe it contains all
+        if (!math::smaller(diff_ing, constants::period())) // >= period
+            return true;
```
barendgehrels: Space missing after !, parameters missing

include/boost/geometry/strategies/cartesian/box_in_box.hpp
```diff
@@ -61,9 +65,94 @@ struct box_covered_by_range
...
+
+        // calculate positive longitude translation with bing_min as origin
+        calc_t const c0 = 0;
+        calc_t diff_min = bed_min - bing_min;
+        math::normalize_longitude<units_t, calc_t>(diff_min);
+        if (diff_min < c0) // [-180, 180] -> [0, 360]
```
barendgehrels: Parantheses missing

include/boost/geometry/strategies/cartesian/box_in_box.hpp
```diff
@@ -61,9 +65,94 @@ struct box_covered_by_range
...
+                BoxContainingValue
+            >::type calc_t;
+        typedef typename coordinate_system<Geometry>::type::units units_t;
+        typedef math::detail::constants_on_spheroid<calc_t, units_t> constants;
+
+        // min <= max <=> diff >= 0
+        calc_t const diff_ed = bed_max - bed_min;
+        calc_t const diff_ing = bing_max - bing_min;
+        // if containing is smaller it cannot contain
+        if (diff_ing < diff_ed)
```
barendgehrels: Also here, please check all your if's!

include/boost/geometry/strategies/cartesian/box_in_box.hpp
```diff
@@ -61,9 +65,94 @@ struct box_covered_by_range
...
+
+
+template <typename Geometry>
+struct box_covered_by_range<Geometry, 0, spherical_tag>
+{
+    template <typename BoxContainedValue, typename BoxContainingValue>
+    static inline bool apply(BoxContainedValue const& bed_min,
+                             BoxContainedValue const& bed_max,
+                             BoxContainingValue const& bing_min,
+                             BoxContainingValue const& bing_max)
```
barendgehrels: Sorry, I have no idea what you mean by bed / bing. Can you change the names to something more meaningful?

include/boost/geometry/strategies/cartesian/point_in_box.hpp
```diff
@@ -53,9 +56,75 @@ struct covered_by_range
...
+        // calculate positive longitude translation with min_value as origin
+        calc_t const c0 = 0;
+        calc_t diff_min = value - min_value;
+        math::normalize_longitude<units_t, calc_t>(diff_min);
+        if (diff_min < c0) // [-180, 180] -> [0, 360]
+            diff_min += constants::period();
+
+        return min_value + diff_min <= max_value;
+    }
+};
```
barendgehrels: So this code is duplicated, the only difference apparently being the last line? It must be possible to share more functionality?

test/algorithms/relational_operations/within/within_sph_geo.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+
+int test_main( int , char* [] )
+{
+    test_cs<bg::model::point<double, 2, bg::cs::spherical_equatorial<bg::degree> > >();
+    test_cs<bg::model::point<double, 2, bg::cs::geographic<bg::degree> > >();
+
+#if defined(HAVE_TTMATH)
+    test_cs<bg::model::point<ttmath_big, 2, bg::cs::spherical_equatorial<bg::degree> > >();
+    test_cs<bg::model::point<ttmath_big, 2, bg::cs::geographic<bg::degree> > >();;
```
barendgehrels: Duplicate ;

barendgehrels:
Thanks for this PR! Overall it looks very good. Can you adhere more to the code conventions and check if the code duplication(s) can be factored out?
I could not dive into the very details or test it. Can you give a short explanation of the substrategy?

Did you plan to have this in 1.61? Would be very good, so we might merge it today allready and fix the comments later, that is OK for me.

include/boost/geometry/strategies/cartesian/box_in_box.hpp
awulkiew: Those are the original names used in this file, see:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/strategies/cartesian/box_in_box.hpp#L43
I think they are short versions of `box_contained_xxx` and `box_containing_xxx`. So if I wanted to change that I should change it consistently in the whole file. Should I do that? Do you prefer some specific name?

barendgehrels: Ah yes that will it main. Did I create that? Anyway, the names you propose sound very clear, so yes, if you can change it, thanks

awulkiew: I'd prefer something shoter. I checked those names and they are less readable in expressions and make the lines a lot longer. So I'd prefer leaving them. Though indeed words `bed` and `bing` may be confusing. So if you strongly believed that they should be changed I'd prefer picking some different, short names, e.g. `box_ed_min`, `ed_min` or `min_ed`.

awulkiew:
I refactored the code to avoid duplications and added missing braces.

The `SubStrategy` was originally used as a method of injecting part of the code different for `within` and `covered_by`. I added template arguments in order to allow dispatching it by `Dimension` and CS tag. Then I specialized it for a family of spherical systems and Dimension == 0. This means that a longitude is handled differently than other cordinates.

Yes, I'd like to have it in 1.61. The tests are passing locally with msvc14 and gcc 4.9.1. The only one issue that is left are the confusing names of arguments. And later I also plan to add a corresponding test for the rtree, update the release notes, etc. So sure, I think it may be merged.

include/boost/geometry/strategies/cartesian/box_in_box.hpp
barendgehrels: OK, no problem.

barendgehrels:
OK, thanks the adaptions and the explanation - it is clear now (also the remark about that only dimension==0 is handled differently). I'm OK with merging.

PR #347, opened 2016-04-13
Silence unused parameter warnings (GCC/Clang -Wextra).

plopresti:
For all I know, some of these might even be real bugs...

But in any event, it's nice when Boost does not generate warnings for those of us who like to enable lots and lots of warnings.

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
@@ -58,14 +58,14 @@ public :
     static inline void apply(std::size_t size_at_start,
                 Rings& rings, typename boost::range_value<Rings>::type& ring,
                 Turns& turns,
-                typename boost::range_value<Turns>::type const& turn,
+                typename boost::range_value<Turns>::type const& /*turn*/,
                 Operation& operation,
-                detail::overlay::traverse_error_type traverse_error,
+                detail::overlay::traverse_error_type /*traverse_error*/,
                 Geometry const& ,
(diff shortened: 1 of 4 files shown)


mloskot:
Why not `boost::ignore_unused_variable_warning` which is already used across the codebase?

awulkiew:
There is shorter version: `boost::ignore_unused()`.
But I also prefer removing or commenting out unused parameters, I think workarounds should be used only if there is no other way.

mloskot:
@awulkiew Sure, whatever is the current and preferred convention. I vaguely recall the explicit `ignore_used` was preferred as it also was an explicit comment visible in the code.

plopresti:
Removing or commenting out a parameter name is 100% standard C++, it's the standard way to silence this warning in any code base, and it can only mean one thing (so it functions perfectly as "an explicit comment visible in the code").

Why would anyone prefer something both more verbose and specific to Boost?

Thanks, Adam.

mloskot:
@plopresti I'm not saying it is not standard way, but sometimes project-specific guidelines recommend different ways. For the same reason some accept `T* p=0`, other projects will require `T* p=nullptr`. I'm not arguing with you here, I've just (mis)recalled some very old discussions.

awulkiew:
Another way would be removing some of these parameters entirely but AFAIU it cannot be done.
Barend what do you think?

awulkiew:
Barend, I'd like to hear your opinion before merging this PR.

barendgehrels:
I'm OK with merging - the parameters can also be removed, but commenting is also fine for me.

PR #349, opened 2016-05-25
Feature/perimeter

vissarion:
Perimeter tests for cartesian, spherical and geographic coordinate systems.

### DIFF
--- a/test/algorithms/perimeter/perimeter_geo.cpp
+++ b/test/algorithms/perimeter/perimeter_geo.cpp
@@ -9,7 +9,7 @@
 // http://www.boost.org/LICENSE_1_0.txt)
 
 #include <algorithms/test_perimeter.hpp>
-#include <algorithms/perimeter/polygon_cases.hpp>
+#include <algorithms/perimeter/perimeter_polygon_cases.hpp>
 
 #include <boost/geometry/geometries/geometries.hpp>
 #include <boost/geometry/geometries/point_xy.hpp>
@@ -21,7 +21,7 @@ struct geo_strategies
     // point radius type
     typedef typename bg::promote_floating_point
     <
-    typename bg::coordinate_type<P>::type
+        typename bg::coordinate_type<P>::type
     >::type rtype;
 
     typedef bg::srs::spheroid<rtype> stype;
@@ -33,7 +33,7 @@ struct geo_strategies
 template <typename P>
 void test_default() //this should use andoyer strategy
 {
-    for(std::size_t i = 0; i < Poly_data_geo.size(); ++i)
+    for (std::size_t i = 0; i < Poly_data_geo.size(); ++i)
     {
         test_geometry<bg::model::polygon<P> >(Poly_data_geo[i],
                                               1116814.237 + 1116152.605);
@@ -52,7 +52,7 @@ void test_default() //this should use andoyer strategy
 template <typename P, typename N, typename Strategy>
 void test_with_strategy(N exp_length, Strategy strategy)
 {
-    for(std::size_t i = 0; i < Poly_data_geo.size(); ++i)
+    for (std::size_t i = 0; i < Poly_data_geo.size(); ++i)
     {
         test_geometry<bg::model::polygon<P> >(Poly_data_geo[i],
                                               exp_length,
--- a/test/algorithms/perimeter/perimeter_geo.cpp
+++ b/test/algorithms/perimeter/perimeter_geo.cpp
@@ -1,0 +1,114 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+// Unit Test
+
+// Copyright (c) 2016 Oracle and/or its affiliates.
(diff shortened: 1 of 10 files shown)


barendgehrels:
Thanks a lot Vissarion for this new PR! It is looking very good. Only a few minor details.

test/algorithms/perimeter/polygon_cases.hpp
```diff
@@ -0,0 +1,46 @@
...
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef POLYGON_CASES_HPP
+#define POLYGON_CASES_HPP
+
+#include <string>
+
+static std::string Poly_data_geo_arr[] = {
```
barendgehrels: Should not start with a capital, and curly brace is normally placed on next line in our conventions.

Maybe POLYGON_CASES_HPP can be preceded by perimeter, for in case there is another file like this (might be probable)

The _arr suffix is not necessary in my opinion, I usually have things like:

static std::string case_recursive_boxes_28[2] =
{
    "MULTIPOLYGON(((5 1,5 0,4 0,4 2,5 3,5 1)),((4 2,3 2,3 3,4 3,4 2)))",
    "MULTIPOLYGON(((2 2,2 3,3 3,4 2,3 1,2 1,2 2)),((3 4,4 3,3 3,3 4)),((4 2,5 2,4 1,4 2)))"
};

The poly prefix (which I did not use) is actually quite useful.

test/algorithms/perimeter/polygon_cases.hpp
```diff
@@ -0,0 +1,46 @@
...
+    "POLYGON((0 90,1 80,1 70))",
+    "POLYGON((0 90,1 80,1 80,1 80,1 70,1 70))",
+    "POLYGON((0 90,1 80,1 79,1 78,1 77,1 76,1 75,1 74,\
+              1 73,1 72,1 71,1 70))"
+};
+
+std::vector<std::string> Poly_data_geo(Poly_data_geo_arr, Poly_data_geo_arr +
+                             sizeof(Poly_data_geo_arr) / sizeof(std::string));
```
barendgehrels: Right, I did not do it like this, but it makes sense.

test/algorithms/perimeter/perimeter_geo.cpp
```diff
@@ -0,0 +1,114 @@
...
+    // Geometries with length zero
+    test_geometry<P>("POINT(0 0)", 0);
+    test_geometry<bg::model::linestring<P> >("LINESTRING(0 0,0 1,1 1,1 0,0 0)",
+                                             0);
+}
+
+template <typename P, typename N, typename Strategy>
+void test_with_strategy(N exp_length, Strategy strategy)
+{
+    for(std::size_t i = 0; i < Poly_data_geo.size(); ++i)
```
barendgehrels: Minor thing, after for we always put a space

test/algorithms/perimeter/perimeter_geo.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include <boost/geometry/geometries/point_xy.hpp>
+
+template <typename P>
+struct geo_strategies
+{
+    // Set radius type, but for integer coordinates we want to have floating
+    // point radius type
+    typedef typename bg::promote_floating_point
+    <
+    typename bg::coordinate_type<P>::type
```
barendgehrels: Usually indented

barendgehrels:
I'm OK with merging

vissarion:
Thanks a lot Barend for the comments.

test/algorithms/perimeter/perimeter_polygon_cases.hpp
```diff
@@ -0,0 +1,46 @@
...
+std::vector<std::string> Poly_data_sph(Poly_data_sph_1, Poly_data_sph_1 +
+                              sizeof(Poly_data_sph_1) / sizeof(std::string));
+
+
+static std::string Multipoly_data_1[] = {
+    "MULTIPOLYGON(((0 0,180 0,0 0)), ((0 0, 0 90, 90 90, 0 0)))"
+};
+
+std::vector<std::string> Multipoly_data(Multipoly_data_1, Multipoly_data_1 +
+                               sizeof(Multipoly_data_1) / sizeof(std::string));
```
awulkiew: Names should not start with capital letters.

Ok so you use vectors to call the cases in a loop (only for Polygons) assuming that the expected results are always the same. So if some different results were expected it should be changed. Then you should either write additional cases by yourself or instead of storing only WKTs, store also expected results. I see the intent but in a long run this approach is either overcomplicated or not good enough depending on how you look at it. Though it could stay as it is for now.

vissarion:
Thanks Adam for the feedback. I was also skeptical about that. I made it a bit simpler.

PR #351, opened 2016-06-06
Fix/cart intersect for different RobustPoint types

awulkiew:
This fixes the ticket: https://svn.boost.org/trac/boost/ticket/12189
and issue reported in this thread: http://lists.boost.org/geometry/2016/05/3550.php

`relate_cartesian_segments::apply()` expects single `RobustPoint` type. This causes compilation error in e.g. relational operations when geometries using different point types are passed.


barendgehrels:
Looks all good to me, I'm OK with this, thanks!

PR #353, opened 2016-07-25
Fix int vs. signed_size_type mismatch

jhunold:
Introduced in f5b4e4a9fa0e491fea4644227f9e7499e0f8f36e
For a failing regression test take a look at [clang-linux-3.8~c++1z](http://www.boost.org/development/tests/develop/developer/output/teeks99-02-dc3-8-1z-Docker-64on64-boost-bin-v2-libs-geometry-test-algorithms-buffer-algorithms_buffer-test-clang-linux-3-8~c++1z-debug.html) for example.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/traversal_switch_detector.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/traversal_switch_detector.hpp
@@ -87,7 +87,7 @@ struct traversal_switch_detector
         return it == m_regions.end() ? -1 : it->second;
     }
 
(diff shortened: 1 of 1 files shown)


awulkiew:
Thanks!
Barend are you ok with this fix?

In `traversal_switch_detector` there is also regions map keeping region ids of type `int`. Should these IDs also be changed to `signed_size_type`?

barendgehrels:
Sure - did notice it too, but it was also fixed in another Pr so did not fix it myself

Thanks!

PR #354, opened 2016-07-26
Prevent integer overflow during segment touch ratio calculation

mcquay239:
I use boost/geometry clipping operations after snap rounding (to prevent robustness issues). So I have no intersecting segments in my areals, segments are intersecting with their endpoints. Integers are large (they can be about 2**30), but I use 64-bit integer coordinate type. So sides (side_info) are calculating perfectly, but when segment ratio calculation leads to integer overflow.

But we don't actually need to calculate these ratios, they are trivial.

### DIFF
--- a/include/boost/geometry/strategies/cartesian/intersection.hpp
+++ b/include/boost/geometry/strategies/cartesian/intersection.hpp
@@ -339,24 +339,57 @@ struct cartesian_segments
         // (only calculated for non-collinear segments)
         if (! collinear)
         {
+            math::detail::equals_factor_policy<robust_coordinate_type>
+                policy(robust_dx_a, robust_dy_a, robust_dx_b, robust_dy_b);
+
+            robust_coordinate_type const zero = 0;
+
             robust_coordinate_type robust_da0, robust_da;
             robust_coordinate_type robust_db0, robust_db;
 
-            cramers_rule(robust_dx_a, robust_dy_a, robust_dx_b, robust_dy_b,
-                get<0>(robust_a1) - get<0>(robust_b1),
-                get<1>(robust_a1) - get<1>(robust_b1),
-                robust_da0, robust_da);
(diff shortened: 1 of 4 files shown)


barendgehrels:
Thanks @mcquay239 ! 
The compilation fixes can certainly be applied and will help some regressions in the regression matrix.
The other fix can most probably be applied too - (have to spend more time on it to confirm, which I now don't have) but to be sure, did you run the unit tests?

mcquay239:
I did and I cought a regression in multiline/multiline intersection.
Unfortunately I noticed it after the pull request, because there already
were several test failures. I beleive it's something easy-to-fix and I'll
try to debug it.

вт, 26 июля 2016 г., 23:11 Barend Gehrels <notifications@github.com>:

barendgehrels:
OK - thanks again. If necessary we can also split this PR

mcquay239:
I've just fixed tests. The only notice is that a linestring, consisting of an isolated point (two equal points, actually), is not simple (really?). But as I understand it's an expected behaviour.

Anyway all tests except broken earlier are passed.

mcquay239:
I don't know if it's necessary to split this PR: #353 is equivalent to https://github.com/boostorg/geometry/pull/354/commits/b5c07c67ba53c3916d59ade6b005806acde75325, you can apply #353 and then think about this one.

mcquay239:
Updated PR, checked tests once more, please apply 
@barendgehrels

awulkiew:
@barendgehrels Considering you've made many changes in the strategy since 2016 is this PR still relevant?

barendgehrels:
> @barendgehrels Considering you've made many changes in the strategy since 2016 is this PR still relevant?

Indeed, I probably missed this. No, I don't think it's still relevant. Let's close it - or maybe still compare it

PR #355, opened 2016-08-02
Geographic formulas

awulkiew:
This PR changes/adds formulas which could be used in the implementation of algorithms and strategies for geometries in geographic CS. It's the result of experiments with various methods of solving geodesic problems. Below is the list of changes:

* existing geographic formulas are moved from `algorithms/detail`/`bg::detail` to `formulas`/`formula` directory/namespace:
    - `andoyer_inverse`
    - `thomas_inverse`
    - `vincenty_inverse`
    - `vincenty_direct`

* inverse and direct formulas in addition to distance/azimuth (inverse) and coordinates of the second point (direct) can now calculate reverse azimuth, reduced length and geodesic scale.

* new formulas are added:
    - `thomas_direct` - solution of direct geodesic problem with Andoyer-Lambert-Forsyth type of approximation with 2nd order terms in flattening as proposed by Thomas.
    - `differential_quantities` - formula calculating reduced length and geodesic scale from segment endpoints and azimuths as proposed by Karney, it is used in inverse and direct formulas
    - `sjoberg_intersection` - formula calculating the intersection of two geodesics as proposed by Sjoberg, it takes inverse formula and the order of approximation of geodesic integral as template parameters 
    - `gnomonic_spheroid` - gnomonic projection on spheroid (ellipsoid of revolution) as proposed by Karney, it takes inverse and direct formulas as template parameters
    - `gnomonic_intersection` - formula calculating the intersection of two geodesics iteratively projecting, finding the intersection point and projecting back, as proposed by Karney, it takes inverse and direct formulas as template parameters

Various formulas has different accuracy and performance.

More testing has to be done but I think that the PR could be merged now because parts of it could be used in implementation of other formulas and strategies, e.g. differential quantities in the implementation of geographic area (as proposed by Karney).

In the near future I plan to implement inverse, direct and intersection formulas based on great elliptic arcs (in cartesian 3d, so similar to what we have already for spherical). Some researchers claim that the results are better than in case of Andoyer's formula and the method is faster (*Tseng at al - A COMPARISON OF GREAT CIRCLE, GREAT ELLIPSE, AND GEODESIC SAILING, 2013*). Taking into account that currently we use Andoyer's formula in geographic distance by default a potentially better and simple alternative method could be the default for all geographic algorithms.


barendgehrels:
Thanks! Will comment later, hopefully today

include/boost/geometry/formulas/differential_quantities.hpp
```diff
@@ -0,0 +1,269 @@
...
+                           CT const& cos_alp0_sqr, CT const& f)
+    {
+        if (Order == 0)
+        {
+            return 0;
+        }
+
+        CT const sig_12 = atan2(cos_sig1 * sin_sig2 - sin_sig1 * cos_sig2,
+                                cos_sig1 * cos_sig2 + sin_sig1 * sin_sig2);
+        CT const sin_2sig1 = 2 * cos_sig1 * sin_sig1; // sin(2sig1)
```
vissarion: Should be CT(2) instead of 2. This should hold in general for all constants in the code in order to support other number types like ttmath.

include/boost/geometry/formulas/sjoberg_intersection.hpp
```diff
@@ -0,0 +1,372 @@
...
+        CT const t01_t02 = t01 * t02;
+        CT const t01_t02_2 = c2 * t01_t02;
+        CT const sqr_t01_sqr_t02 = math::sqr(t01) + math::sqr(t02);
+
+        CT t = tan_beta1;
+        int t_id = 0;
+
+        // find the initial t using simplified spherical solution
+        // though not entirely since the reduced latitudes and azimuths are spheroidal
+        // paper 2007
```
vissarion: Instead of paper 2007, paper, etc it is more standard to use something like [Sjoberg02].

vissarion:
Thank you Adam for this useful PR! I had some minor comments. I am ok for merging for now, and test it more later.

include/boost/geometry/formulas/andoyer_inverse.hpp
```diff
@@ -142,64 +150,93 @@ struct andoyer_inverse
...
+    static inline void normalize_azimuth(CT & azimuth, CT const& A, CT const& dA)
+    {
+        CT const c0 = 0;
+        
+        if (A >= c0) // A indicates Eastern hemisphere
+        {
+            if (dA >= c0) // A altered towards 0
+            {
+                if (azimuth < c0)
+                    azimuth = c0;
```
barendgehrels: Please either add parentheses, or putting it in one line is also OK for me

include/boost/geometry/formulas/differential_quantities.hpp
```diff
@@ -0,0 +1,269 @@
...
+\brief The solution of a part of the inverse problem - differential quantities.
+\author See
+- Charles F.F Karney, Algorithms for geodesics, 2011
+https://arxiv.org/pdf/1109.4448.pdf
+*/
+template <
+    typename CT,
+    bool EnableReducedLength,
+    bool EnableGeodesicScale,
+    unsigned Order = 2,
```
barendgehrels: unsigned int? std::size_t ?

include/boost/geometry/formulas/differential_quantities.hpp
```diff
@@ -0,0 +1,269 @@
...
+    static inline void apply(CT const& dlon,
+                             CT const& sin_lat1, CT const& cos_lat1,
+                             CT const& sin_lat2, CT const& cos_lat2,
+                             CT const& azimuth, CT const& reverse_azimuth,
+                             CT const& b, CT const& f,
+                             CT & reduced_length, CT & geodesic_scale)
+    {
+        CT const one_minus_f = 1 - f;
+        CT const sin_bet1 = one_minus_f * sin_lat1;
+        CT const sin_bet2 = one_minus_f * sin_lat2;
```
barendgehrels: BOOST_ASSERT(one_minus_f != 0) (for dvisions below) ?

include/boost/geometry/formulas/differential_quantities.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+        CT const L3 = (9 * sin_4sig_12 + 4 * sin3_2sig_12 - 48 * sin_2sig_12 + 60 * sig_12) / 512;
+
+        // Order 3 and higher
+        return c2a0ep2 * (L1 + c2a0ep2 * (L2 + c2a0ep2 * L3));
+    }
+
+    static inline void normalize(CT & x, CT & y)
+    {
+        CT const l = math::sqrt(math::sqr(x) + math::sqr(y));
```
barendgehrels: Please change l into len, because l/1 are difficult to distinguish - and we divide by it, how sure are we that it is non zero?

include/boost/geometry/formulas/gnomonic_intersection.hpp
```diff
@@ -0,0 +1,147 @@
...
+{
+
+/*!
+\brief The intersection of two geodesics using spheroidal gnomonic projection
+       as proposed by Karney.
+\author See
+    - Charles F.F Karney, Algorithms for geodesics, 2011
+      https://arxiv.org/pdf/1109.4448.pdf
+    - GeographicLib forum thread: Intersection between two geodesic lines
+      https://sourceforge.net/p/geographiclib/discussion/1026621/thread/21aaff9f/
```
barendgehrels: It is MIT-licensed, also open and we change license here to Boost. It is reworked to templates and Karney suggested this.  I don't know how to exactly formulate that here

include/boost/geometry/formulas/gnomonic_intersection.hpp
```diff
@@ -0,0 +1,147 @@
...
+*/
+template <
+    typename CT,
+    template <typename, bool, bool, bool, bool ,bool> class Inverse,
```
barendgehrels: small things, comma after space, and at template < the < should go to next line

include/boost/geometry/formulas/gnomonic_intersection.hpp
```diff
@@ -0,0 +1,147 @@
...
+                             Spheroid const& spheroid)
+    {
+        CT const lon_a1 = lona1;
+        CT const lat_a1 = lata1;
+        CT const lon_a2 = lona2;
+        CT const lat_a2 = lata2;
+        CT const lon_b1 = lonb1;
+        CT const lat_b1 = latb1;
+        CT const lon_b2 = lonb2;
+        CT const lat_b2 = latb2;
```
barendgehrels: Why? Why not just pass the variables? Are the types conversed?

include/boost/geometry/formulas/gnomonic_spheroid.hpp
```diff
@@ -0,0 +1,118 @@
...
+        CT distance = a * atan(rho / a);
+
+        bool found = false;
+        for (int i = 0 ; i < 10 ; ++i)
+        {
+            direct_result d_res = direct_quantities_type::apply(lon0, lat0, distance, azimuth, spheroid);
+            CT const& m = d_res.reduced_length;
+            CT const& M = d_res.geodesic_scale;
+            
+            CT const drho = m / M - rho; // rho = m / M
```
barendgehrels: Non zero?

test/formulas/direct_cases.hpp
```diff
@@ -0,0 +1,452 @@
...
+        { -0.58943065613921030721, -2.59810442273856878614, -134.95009296257936171060, 249935.55921406237757764757, 0.99922675162293117523 }
+    },{
+        { 1, -1 }, 250000, -180,
+        { 1.00000000000000000000, -3.26088938011404927408, -180.00000000000000000000, 249935.55991226507467217743, 0.99922675429255258805 },
+        { 0.99999999999999977796, -3.26088938009322193423, -180.00000000000000000000, 249935.55988221566076390445, 0.99922675435769237051 },
+        { 1.00000000000000000000, -3.26088936183515265554, -180.00000000000000000000, 249935.55786483851261436939, 0.99922675437017982603 }
+    }
+};
+
+size_t expected_size = sizeof(expected) / sizeof(expected_results);
```
barendgehrels: std::size_t const

barendgehrels:
Thanks a lot Adam! This is a considerable extension of our geographic functionality. I'm OK with merging.

barendgehrels:
Do you target 1.62 for these enhancements?

include/boost/geometry/formulas/differential_quantities.hpp
awulkiew: I'll change it to `unsigned int` to avoid confusion.
FYI `unsigned` is short for `unsigned int` (http://en.cppreference.com/w/cpp/language/types).

include/boost/geometry/formulas/differential_quantities.hpp
awulkiew: `1-f = 0 <=> f=1 <=> b=0` so it's a disk ("flat" ellipsoid). And `b` may be passed by the user so maybe it should be an exception. But since it's such unreasonable scenario it shouldn't be checked here each time a formula is used because in the vast majority of cases `f` will be reasonable. Instead it could be checked in `srs::spheroid`. What do you think?

include/boost/geometry/formulas/differential_quantities.hpp
barendgehrels: I know ;-)
Thanks for changing

include/boost/geometry/formulas/differential_quantities.hpp
barendgehrels: Indeed the scenario is improbable but not impossible.
OK for me to check it in spheroid, and add a comment here that it is checked there.

include/boost/geometry/formulas/gnomonic_intersection.hpp
awulkiew: The `\author` is the author of the papers/formulas, not author of the code. I've taken this description template from Andoyer and Vincenty which was first formulas implemented in BG. And there the `\author` mentions papers containing the original formulas.

I haven't used the code of GeographicLib directly which indeed is MIT licensed. However indeed I've seen the code and used it as a complementary source of formulas to the paper because some things were not obvious in the paper (in particular in case of differential quantities and gnomonic projection).

This particular formula (intersection using gnomonic projection) is not implemented in GeographicLib. A simple example was attached by Karney in the forum thread but I haven't used it directly either (plus there was no license there).

include/boost/geometry/formulas/gnomonic_intersection.hpp
awulkiew: They might be conversed if they were different than `CalculationType`.

awulkiew:
Done. Regarding 1.62, these formulas aren't used in any strategy/algorithm for now, they may need a few minor tweaks too, so it doesn't matter if they're released. However I know that Vissarion is working on geographic area strategy which use these formulas. So if he plans to propose a PR for 1.62 then these formulas should be released as well. I don't know if I'm going to implement geographic intersection strategy (I'm experimenting with a different approach now), there is one more week left before master branch is closed for major changes and week after that for bugfixes, so maybe.

PR #357, opened 2016-08-15
Add C++11 std::array point adaptor

norbertwenzel:
Add point adaptor based on existing Boost.Array adaptor for C++11 std::array. I'm currently not very happy with the preprocessor checks in the test, but since the whole test is really small I think the preprocessor checks are not too bad. Anyway, I'm open to suggestions if changes are required.

### DIFF
--- a/include/boost/geometry/geometries/adapted/std_array.hpp
+++ b/include/boost/geometry/geometries/adapted/std_array.hpp
@@ -1,0 +1,115 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+
+// Copyright (c) 2010 Alfredo Correa
+// Copyright (c) 2010-2012 Barend Gehrels, Amsterdam, the Netherlands.
+// Copyright (c) 2016 Norbert Wenzel
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_GEOMETRIES_ADAPTED_STD_ARRAY_HPP
+#define BOOST_GEOMETRY_GEOMETRIES_ADAPTED_STD_ARRAY_HPP
+
+
+#define BOOST_GEOMETRY_ADAPTED_STD_ARRAY_TAG_DEFINED
+
+
+#include <cstddef>
+
+#include <boost/type_traits/is_arithmetic.hpp>
+
+#include <boost/geometry/core/access.hpp>
+#include <boost/geometry/core/cs.hpp>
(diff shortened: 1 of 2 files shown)


awulkiew:
Thanks!
It looks good to me.

norbertwenzel:
Since the merge window is now open again I just wanted to ask if there is anything that needs to be done on my side, before this one is ready.

include/boost/geometry/geometries/adapted/std_array.hpp
```diff
@@ -0,0 +1,114 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+
+// Copyright (c) 2010 Alfredo Correa
```
barendgehrels: The copyright is not up to date

test/geometries/boost_array_as_point.cpp
```diff
@@ -31,6 +35,12 @@ int test_main(int, char* [])
     std::clog << bg::distance(p1, p2) << std::endl;
     std::clog << bg::distance(p2, p3) << std::endl;
     std::clog << bg::distance(p3, p4) << std::endl;
+
+#ifndef BOOST_NO_CXX11_HDR_ARRAY 
+    std::array<double, 3> p5{13,14,15};
+    std::clog << bg::distance(p4, p5) << std::endl;
+#endif //BOOST_NO_CXX11_HDR_ARRAY
```
barendgehrels: Nice. This is not about your change, but we should update this test. Writing to clog is not really a unit test, apparently this was never really finished. Anyway, that is not an objective for merging.

barendgehrels:
And thanks for your contribution and notification. If it is merged it will be merged to master too.
We also have to update the release notes for this.

include/boost/geometry/geometries/adapted/std_array.hpp
norbertwenzel: This code is basically copied from the `boost::array` adaptor and that's why I did not change it. I don't think my minimal changes are above a threshold of creativity to substantiate any copyright. But I will add my name to this file if you consider it necessary.

test/geometries/boost_array_as_point.cpp
norbertwenzel: At first I thought this test is incomplete too, but now I think it's not too bad. Since this test merely tests if the new point type is recognized by using the relatively simple `bg::distance()` algorithm and hope this test compiles. So the only thing I'd maybe add to these adaptor tests would be a check for the distance results in order to test if the point layout has been recognized correctly. But that's probably a question of taste.

norbertwenzel:
@barendgehrels I've amended the copyright info in `std_array.hpp`.

barendgehrels:
@norbertwenzel  thanks! I will merge it.

PR #360, opened 2016-09-02
Feature/max latitude

vissarion:
### Problem: compute the vertex latitude of a segment

Given a segment compute the point with the highest latitude, called vertex latitude. Note that on sphere/ellipsoid this could be higher than the maximum latitude of the segment's endpoints. 
### Solution

This PR implements formulas proposed in 

`Wood - Vertex Latitudes on Ellipsoid Geodesics, SIAM Rev., 38(4), 637–644, 1996` 

It is implemented in formulas/vertex_latitude.hpp for both geographic and spherical cases. The method computes both maximum and minimum latitude points of the segment. Several tests are also added.
### Comments
- The solution is expected to be much faster than computing the shortest distance between the segment and the North (or South) pole. Note that the later is a correct (but slower) solution for the problem. 
- This PR is going to be used by the algorithm that computes the envelope of a segment. 
- Azimuth algorithm for spherical segments was altered in order to use an azimuth formula. This is more consistent to geographic azimuth, i.e. the algorithm calls a formula. Then the azimuth formula is directly be used by the vertex latitude formula.

### DIFF
--- a/include/boost/geometry/formulas/vertex_latitude.hpp
+++ b/include/boost/geometry/formulas/vertex_latitude.hpp
@@ -54,16 +54,16 @@ public:
         inverse_result i_res = inverse_type::apply(lon1, lat1, lon2, lat2,
                                                    spheroid);
 
-        CT const alp1 = i_res.azimuth;
-        CT const alp2 = i_res.reverse_azimuth;
+        CT const alp1 = std::abs(i_res.azimuth);
+        CT const alp2 = std::abs(i_res.reverse_azimuth);
 
-        std::cout << "a1=" << alp1 * 180 / math::pi<CT>()
-                  << " a2="
-                  << alp2 * 180 / math::pi<CT>()
-                  << std::endl;
+        //std::cout << "a1=" << alp1 * 180 / math::pi<CT>()
+        //          << " a2="
+        //          << alp2 * 180 / math::pi<CT>()
+        //          << std::endl;
 
         // if the segment does not contain the vertex of the geodesic
-        // then the vertex is one of the endpoints
+        // then return the endpoint of max (min) latitude
         if ((alp1 < half_pi && alp2 < half_pi)
             || (alp1 > half_pi && alp2 > half_pi)
             || (north == sign))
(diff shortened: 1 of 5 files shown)


vissarion:
I am working in merging this implementation with the one of envelope in BG. I think it is clearer to write a new PR on envelope and delete the current one.

include/boost/geometry/formulas/vertex_latitude.hpp
```diff
@@ -0,0 +1,149 @@
...
+            CT const a_c2 = math::sqr(a / c);
+            CT const vertex_lat = std::asin(math::sqrt((a_c2 - 1) / (a_c2 - e2)));
+
+            sign_adjastment(lat1, lat2, vertex_lat, vrt_result);
+        }
+        return vrt_result;
+    }
+
+    template <typename T>
+    inline static void sign_adjastment(CT lat1, CT lat2, CT vertex_lat, T& vrt_result)
```
barendgehrels: I think these values should be passed as const&. For built-in types this does not make any difference, probably. But for user-defined numeric types it might make a difference. Also below in vertex_on_segment.

include/boost/geometry/formulas/vertex_latitude.hpp
```diff
@@ -0,0 +1,149 @@
...
+    {
+        vertex_lat_result vrt_result;
+
+        geometry::formula::result_spherical<CT> result = geometry::formula::
+                spherical_azimuth<CT, true>(lon1, lat1, lon2, lat2);
+
+        CT const alp1 = std::abs(result.azimuth);
+        CT const alp2 = std::abs(result.reverse_azimuth);
+
+        if(vertex_on_segment(alp1, alp2, lat1, lat2, vrt_result))
```
barendgehrels: minor: I think we have spaces after if, so if ( 
(contrary to function calls where we don't do that)

include/boost/geometry/formulas/vertex_latitude.hpp
```diff
@@ -0,0 +1,149 @@
...
+        {}
+
+        CT north, south;
+    };
+
+    template <
+            typename T1,
+            typename T2
+            >
+    static inline vertex_lat_result spherical(T1 const& lon1,
```
barendgehrels: Also minor, if it fits in one line (80 chars), you may write it on one line (template <typename T1, typename T2>
If it does not fit on one line, it should be indented like functions do, so < goes on next line, aligned under template

include/boost/geometry/formulas/vertex_latitude.hpp
```diff
@@ -0,0 +1,149 @@
...
+            CT const sin2_lat1 = math::sqr(sin(lat1));
+            CT const cos_lat1 = math::sqrt(CT(1) - sin2_lat1);
+
+            // normal radius at point p1(lon1,lat1)
+            CT const n_b1 = a / (math::sqrt(CT(1) - e2 * sin2_lat1));
+
+            // the invariant of the geodesic
+            CT const c = n_b1 * cos_lat1 * sin_alp1;
+
+            CT const a_c2 = math::sqr(a / c);
```
barendgehrels: who guarantees c is non-zero?

include/boost/geometry/formulas/vertex_latitude.hpp
```diff
@@ -0,0 +1,149 @@
...
+            CT const vertex_lat = std::asin(math::sqrt((a_c2 - 1) / (a_c2 - e2)));
```
barendgehrels: Same here - is there chance that denumerator gets 0 ?

vissarion:
This PR is superfluous because of PR "Feature/envelope expand geo #367", thus I am closing it.

PR #367, opened 2016-10-27
Feature/envelope expand geo

vissarion:
### Problem: compute the envelope and expand functions on segments on spheroid

The core procedure when computing the envelope of a segment is to compute the segment's vertex latitude. That is, given a segment compute the point with the highest latitude, called vertex latitude. Note that on sphere/ellipsoid this could be higher than the maximum latitude of the segment's endpoints. 

### Solution

This PR implements formulas proposed in 

`Wood - Vertex Latitudes on Ellipsoid Geodesics, SIAM Rev., 38(4), 637–644, 1996` 

It is implemented in formulas/vertex_latitude.hpp for both geographic and spherical cases. It also adds strategies for azimuth computation for all coordinate systems. Then the algorithm envelope (or expand) will call the right azimuth strategy and the right vertex latitude formula depending on the coordinate system. Several tests are also added.

### Comments
- The solution is expected to be much faster than computing the shortest distance between the segment and the North (or South) pole. Note that the later is a correct (but slower) solution for the problem. 
- Azimuth algorithm for spherical segments was altered in order to use an azimuth formula. This is more consistent to geographic azimuth, i.e. the algorithm calls a formula. Then the azimuth formula is directly used by the vertex latitude formula. 
- Rings that cover more than half of the earth are not supported which is the same as in area i.e. if we alter the orientation of an exterior ring then the area will be negative and not area(earth) - area(ring)

*** EDIT / REVISION ***
Instead of [Wood96] formula we now use a simpler formula (that utilize the auxiliary sphere) cf. commit 3fda39d


barendgehrels:
I'm OK with merging

include/boost/geometry/algorithms/detail/envelope/interface.hpp
```diff
@@ -94,7 +127,18 @@ struct envelope<boost::variant<BOOST_VARIANT_ENUM_PARAMS(T)> >
 template<typename Geometry, typename Box>
 inline void envelope(Geometry const& geometry, Box& mbr)
 {
-    resolve_variant::envelope<Geometry>::apply(geometry, mbr);
+    // TODO put this into a resolve_strategy stage
+    //      (and take the return type from resolve_variant)
```
awulkiew: Yes, this should be done (TODO). So in 2 parameter envelope pass object of type `bg::default_strategy` into the `resolve_variant::envelope<>` and specialize for this type above.

include/boost/geometry/algorithms/detail/expand/interface.hpp
```diff
@@ -106,21 +119,51 @@ inline void expand(Box& box, Geometry const& geometry,
...
+added to the box
 
 \qbk{[include reference/algorithms/expand.qbk]}
  */
 template <typename Box, typename Geometry>
 inline void expand(Box& box, Geometry const& geometry)
 {
-    resolve_variant::expand<Geometry>::apply(box, geometry);
+    // TODO put this into a resolve_strategy stage
+    //      (and take the return type from resolve_variant)
```
awulkiew: TODO

include/boost/geometry/formulas/vertex_latitude.hpp
```diff
@@ -0,0 +1,157 @@
...
+        //apply Clairaut relation
+        CT const betv =  vertex_latitude_on_sphere<CT>::apply(bet1, alp1);
+
+        //return the spheroid latitude
+        return atan( tan(betv) / one_minus_f );
+    }
+
+    /*
+    template <typename T>
+    inline static void sign_adjastment(CT lat1, CT lat2, CT vertex_lat, T& vrt_result)
```
awulkiew: Is the word "adjastment" ok?

include/boost/geometry/strategies/azimuth.hpp
```diff
@@ -0,0 +1,36 @@
...
+{
+
+/*!
+    \brief Traits class binding a default azimuth strategy to a coordinate system
+    \ingroup azimuth
+    \tparam Tag tag of coordinate system
+*/
+template <typename Tag, typename CalculationType>
+struct default_strategy
+{};
```
awulkiew: Here there should be static assert (see other algorithms), e.g.:

    BOOST_MPL_ASSERT_MSG
        (
            false, NOT_IMPLEMENTED_FOR_THIS_TAG
            , (types<Tag>)
    );

include/boost/geometry/strategies/cartesian/azimuth_cartesian.hpp
```diff
@@ -0,0 +1,64 @@
...
+    {}
+
+    inline void apply(CalculationType const& lon1_rad,
+                      CalculationType const& lat1_rad,
+                      CalculationType const& lon2_rad,
+                      CalculationType const& lat2_rad,
+                      CalculationType& a1,
+                      CalculationType& a2) const
+    {
+        // not implemented
```
awulkiew: If `azimuth::services::default_strategy` had static/mpl assert this whole file wouldn't be needed.

As an alternative we could simply implement it. Wouldn't it be something like:

    a1 = a2 = atan2(y2 - y1, x2 -x1);

?

awulkiew:
So currently `envelope()` expects azimuth strategy. I'd rather expect it taking a strategy calculating an envelope of a Segment, plus specific/simple versions for Point and Box (which could be not implemented as it is now). From user's perspective passing azimuth strategy into the envelope() doesn't make sense. He/she should rather pass some envelope strategy, just like distance strategy is passed into the distance algorithm and area strategy into the area algorithm.

Plus with the segment envelope strategy the interface would be the same if we decided to support bezier or arc segments at some point. Not that I'm planning to do this in the near future, this simply tells me that this is more correct level of abstraction.

On the other hand azimuth strategy could also be used by `azimuth()` algorithm if we decided to add it (I remember users asking about it).

What do you think?

vissarion:
Thanks Adam for the comments. I agree about the azimuth strategy. What I did was to make the current BG code work with as little changes as possible. For a more convenient interface I propose a strategy e.g. segment_mbr(Point p1, Point p2, Box& mbr) that compute the envelope of segment (p1,p2) and stores it to mbr. This is not far from the interface of distance strategies that take two points and return the distance. Similarly, azimuth strategies could take points as parameters instead of coordinates (as it is in the current code of this PR).

What do you think?

awulkiew:
I think this is the correct approach.

It'd have to take the Spheroid and CalculationType as template parameters, correct?

TLDR: I think that either 2 input points and 1 mbr or 2 input points and 2 output corners are ok. It's because if you have at least points (as opposed to plain coordinates) then the information about the CS units is there for you. In some cases it may be beneficial to know it (e.g. for normalization purposes). The former (the one you proposed) seems to be slightly better because we already have strategies taking Boxes (e.g. within Box/Box Pt/Box or distance Box/Box) and because in this case more compile-time information is preserved.

vissarion:
Yes, there are two template parameters: Spheroid and CalculationType. Then there is a choice of having one more, namely the inverse formula, e.g. geometry::formula::thomas_inverse or do it as in distance and create one strategy per inverse formula. The pos of the former is that it is more compact but exposes the user to use formulas for parametrization.

vissarion:
As discussed above I moved all the CS related computation from the algorithm level to the strategy level. Any comments?

include/boost/geometry/strategies/cartesian/envelope_segment.hpp
```diff
@@ -0,0 +1,64 @@
...
+#include <boost/geometry/algorithms/detail/envelope/segment.hpp>
+
+namespace boost { namespace geometry
+{
+
+namespace strategy { namespace envelope
+{
+
+template<typename CalculationType>
+class segment_cartesian
```
awulkiew: `CalculationType = void`

Regarding the consistency of the names, we'd have `strategy::area::geographic` and `strategy::azimuth::geographic` so my guess as a user would be to expect the name of the CS right after the name of the strategy "category" or algorithm name. So I was thinking about renaming it to `envelope::cartesian_segment` and `envelope::geographic_segment` (we also have intersection::relate_cartesian_segments). And this way it'd also be more pleasant for english speakers. What do you think?

vissarion:
I moved the envelope computation back to algorithm level and also changed the names `segment_cartesian` to `cartesian_segment`.

awulkiew:
@vissarion Unfortunately after merging area PR https://github.com/boostorg/geometry/pull/359 there are some conflicts.

vissarion:
@awulkiew Thanks for the message, I resolved the conflicts.

PR #371, opened 2016-12-22
handle_colocations.hpp always includes <iostream>

nre-ableton:
This file contains a few stray `std::cout` statements which do not seem to be related to `BOOST_GEOMETRY_DEBUG_IDENTIFIER`, and hence cause a compiler error if that symbol is undefined.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/handle_colocations.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/handle_colocations.hpp
@@ -11,6 +11,7 @@
 
 #include <cstddef>
 #include <algorithm>
+#include <iostream>
 #include <map>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


awulkiew:
The library shouldn't depend on iostream. All occurances of `cout` statements are for debugging. In `handle_colocations.hpp` they are all guarded by `BOOST_GEOMETRY_DEBUG_HANDLE_COLOCATIONS`. Or am I missing something? Where have you seen the stray ones?

nre-ableton:
My bad, it looks like this was fixed in https://github.com/boostorg/geometry/commit/7fc45bef3febd449a38736a198809e9574c5efa3#diff-4461eb3f155f150363a79d411e287c53, which was not present in boost 1.62. I see now that boost 1.63 has been released, and it contains that fix. Sorry for the noise!

PR #373, opened 2017-01-03
Fix tests

Ironholds:
Something I forgot to do with #372  - amend the tests too so the change doesn't break em when a different error is emitted!


awulkiew:
Thanks!
In the future please use more descriptive commit messages putting tags at the beginning marking modified parts of the library. In this case it could've been something like: `[test][io] Change expected wkt exception message.`

Ironholds:
Can do!

PR #384, opened 2017-03-08
Throw exceptions using BOOST_THROW_EXCEPTION.

awulkiew:
BOOST_THROW_EXCEPTION was used before in the library to throw exceptions but not in all places. This PR adds it consistently in all cases besides one place being `self_section_visitor` defined in `algorithms/detail/overlay/self_turn_points.hpp`. It's because in this case the exception should be removed entirely. This requires the changes in `partition` which will be proposed in another PR.

Related ticket: https://svn.boost.org/trac/boost/ticket/12885


awulkiew:
Since noone has any objections and it's a simple fix I'm merging it.

PR #385, opened 2017-03-09
Interruptable partition()

awulkiew:
This PR makes partition() algorithm interruptable. The partition is interrupted when `visitor.apply(...)` returns `false`.

`self_ip_exception` is no longer needed, so removed. It's related to https://github.com/boostorg/geometry/pull/384

A side effect is performance increase. Below I show times gathered using this benchmark:
https://github.com/awulkiew/benchmark-geometry/blob/master/benchmarks/algorithms.cpp
data generated like this:
https://github.com/awulkiew/benchmark-geometry/blob/master/benchmarks/data.hpp
below:
 - gcc is mingw-w64 4.9.1 -O2
 - msvc is msvc-14 /O2
 - develop - times for develop branch, interrupt - times for this branch
 - relate-disj is relate() with disjoint mask
 - zero - there is no overlap between geometries
 - min - there is minimal overlap btween geometries (touching)
 - max - each segment of one geometry overlaps two segments of other geometry
 - zigzag - each segment of one geometry overlaps all segments of other geometry
 - time of one function call in seconds
 - zero/min/max geometries contain 100001 points, zigzag linestrings contain 400 points

| test name           | gcc develop | gcc interrupt | msvc develop | msvc interrupt |
| ------------------- | ----------- | ------------- | ------------ | -------------- |
relate-disj-poly-zero | 0.00780 | 0.00780 | 0.00839 | 0.00876 |
relate-disj-poly-min  | **0.00557** | 0.00452 | **0.00616** | 0.00457 |
relate-disj-poly-max  | **0.02275** | 0.00450 | **0.02758** | 0.00442 |
relate-disj-ls-zero   | 0.00446 | 0.00448 | 0.00455 | 0.00439 |
relate-disj-ls-min    | **0.00560** | 0.00450 | **0.00594** | 0.00436 |
relate-disj-ls-max    | **0.02248** | 0.00455 | **0.02775** | 0.00440 |
relate-disj-ls-zigzag | **0.11100** | 0.00009 | **0.13010** | 0.00008 |
intersects-poly-zero  | 0.00775 | 0.00784 | 0.00854 | 0.00812 |
intersects-poly-min   | **0.00553** | 0.00450 | **0.00603** | 0.00440 |
intersects-poly-max   | **0.02095** | 0.00442 | **0.02766** | 0.00450 |
intersects-ls-zero    | 0.00441 | 0.00447 | 0.00479 | 0.00441 |
intersects-ls-min     | **0.00547** | 0.00447 | **0.00635** | 0.00442 |
intersects-ls-max     | **0.02106** | 0.00447 | **0.02828** | 0.00447 |
intersects-ls-zigzag  | **0.09510** | 0.00006 | **0.11216** | 0.00008 |
touches-poly-zero     | 0.00779 | 0.00784 | 0.00837 | 0.00832 |
touches-poly-min      | 0.00545 | 0.00547 | 0.00604 | 0.00621 |
touches-poly-max      | 0.44652 | 0.44052 | 0.45616 | 0.45228 |
touches-poly          | 0.00620 | 0.00610 | 0.00852 | 0.00863 |
intersects-poly       | 0.00619 | 0.00613 | 0.00869 | 0.00855 |
is_valid-poly         | 0.02374 | 0.02306 | 0.03384 | 0.03265 |

### DIFF
--- a/include/boost/geometry/algorithms/detail/partition.hpp
+++ b/include/boost/geometry/algorithms/detail/partition.hpp
@@ -1,6 +1,7 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
 // Copyright (c) 2011-2015 Barend Gehrels, Amsterdam, the Netherlands.
+// Copyright (c) 2017 Adam Wulkiewicz, Lodz, Poland.
 
 // This file was modified by Oracle on 2015, 2017.
 // Modifications copyright (c) 2015-2017 Oracle and/or its affiliates.
@@ -106,11 +107,11 @@ inline void expand_with_elements(Box& total, IteratorVector const& input,
 
 // Match forward_range with itself
 template <typename IteratorVector, typename VisitPolicy>
-inline void handle_one(IteratorVector const& input, VisitPolicy& visitor)
+inline bool handle_one(IteratorVector const& input, VisitPolicy& visitor)
 {
     if (boost::size(input) == 0)
     {
-        return;
+        return true;
     }
 
     typedef typename boost::range_iterator<IteratorVector const>::type it_type;
@@ -121,9 +122,14 @@ inline void handle_one(IteratorVector const& input, VisitPolicy& visitor)
         it_type it2 = it1;
         for (++it2; it2 != boost::end(input); ++it2)
         {
-            visitor.apply(**it1, **it2);
+            if (! visitor.apply(**it1, **it2))
+            {
+                return false; // interrupt
+            }
         }
     }
+
+    return true;
 }
 
 // Match forward range 1 with forward range 2
@@ -133,7 +139,7 @@ template
     typename IteratorVector2,
     typename VisitPolicy
 >
-inline void handle_two(IteratorVector1 const& input1,
+inline bool handle_two(IteratorVector1 const& input1,
                        IteratorVector2 const& input2,
                        VisitPolicy& visitor)
 {
@@ -149,7 +155,7 @@ inline void handle_two(IteratorVector1 const& input1,
 
     if (boost::size(input1) == 0 || boost::size(input2) == 0)
     {
-        return;
+        return true;
     }
 
     for(iterator_type1 it1 = boost::begin(input1);
@@ -160,9 +166,14 @@ inline void handle_two(IteratorVector1 const& input1,
             it2 != boost::end(input2);
             ++it2)
         {
-            visitor.apply(**it1, **it2);
+            if (! visitor.apply(**it1, **it2))
+            {
+                return false; // interrupt
+            }
         }
     }
+
+    return true;
 }
 
 template <typename IteratorVector>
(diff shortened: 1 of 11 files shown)


barendgehrels:
Looking very good - also the performance increases!

include/boost/geometry/algorithms/detail/partition.hpp
```diff
@@ -408,17 +421,17 @@ public :
...
+                             IteratorVector1 const& input1,
+                             IteratorVector2 const& input2,
+                             std::size_t level,
+                             std::size_t min_elements,
+                             VisitPolicy& visitor,
+                             ExpandPolicy1 const& expand_policy1,
+                             OverlapsPolicy1 const& overlaps_policy1,
+                             ExpandPolicy2 const& expand_policy2,
+                             OverlapsPolicy2 const& overlaps_policy2,
+                             VisitBoxPolicy& box_policy)
```
barendgehrels: Next time, if possible, please don't mix layout changes with other changes. It makes the review more work.

About the layout change itself, this is how it is more and more done. But it is not how we started. We started by just intending 3 or 4 tabs (units 4 spaces). So independent on method name.

A drawback of this new convention is that, if the method name changes, or the return type (in length), all the lines have to change.

Another, larger, drawback is with long method names, it is all very intended, has to be split more (so into more lines), and in the end is less readable.

Especially because of the last reason, I don't know if I agree. However, sometimes I see myself taking it over also because QtCreator nearly forces it... that is maybe how this appears here too.

awulkiew: Yes, it looks better but requires more work unless the environment is able to format automatically somehow. In this case I've done it manually while traversing the file. I agree that I shouldn't have done it as a part of this PR, sorry about that.

awulkiew:
I noticed that on linux gcc 4.9.2 x64 the performance decreased slightly for "zero" cases:
https://awulkiew.github.io/benchmark-geometry/
But I was able to compensate this with a simple optimization: https://github.com/boostorg/geometry/commit/51f85b6fe24c2a71de7ddc4ecc7aac8435c95ce8

awulkiew:
It seems the change is smaller than I though. So the issue remains and it looks like it depends on the compiler.  I don't know why it doesn't manifest itself with other compilers I tested (esspecially mingw-w64 gcc).

Is it a good tradeof for immediate interruption? It all depends how probable is that a function is called for cases when it can be interrupted. My guess is that if we consider arbitrary geometries scatered around the coordinate system then the probability that they overlap is low, so a user would notice performance decrease. But if we consider that in practice a user would test bounding boxes first, e.g. by using the rtree, then the probability that the geometries indeed overlap increases.

It doesn't manifest itself in `get_turns` tests when all turns has to be calculated, `no_interrupt_policy` is used which causes the visitor to always return `true`. So my guess is that in such cases when the visitor always returns the same value the conditions can be optimized away and the performance is unaffected (should happen e.g. in `relation()` and `buffer()`).

include/boost/geometry/algorithms/detail/partition.hpp
barendgehrels: OK - no problem

PR #386, opened 2017-03-22
Use envelope-seg and disjoint-seg/box strategies in various algorithms.

awulkiew:
- add envelope and disjoint strategies getters into side, winding and intersection strategies
- pass envelope and disjoint strategies into sectionalize and/or partition in get_turns, disjoint, is_valid and overlay.

One more thing should be done, i.e. forcing `point_on_border` to return the vertex in at least non-cartesian coordinate systems. `point_on_surface` seems to be not used anywhere so there is no problem here. Do you have any thoughts regarding `point_on_border`? Second parameter `true` seems to be passed only in 2 places in the library, i.e. `overlay/intersection_insert.hpp` and `overlay/ring_properties.hpp`.

### DIFF
--- a/include/boost/geometry/algorithms/detail/point_on_border.hpp
+++ b/include/boost/geometry/algorithms/detail/point_on_border.hpp
@@ -4,6 +4,10 @@
 // Copyright (c) 2008-2012 Bruno Lalande, Paris, France.
 // Copyright (c) 2009-2012 Mateusz Loskot, London, UK.
 
+// This file was modified by Oracle on 2017.
+// Modifications copyright (c) 2017 Oracle and/or its affiliates.
+// Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
+
 // Parts of Boost.Geometry are redesigned from Geodan's Geographic Library
 // (geolib/GGL), copyright (c) 1995-2010 Geodan, Amsterdam, the Netherlands.
 
@@ -18,6 +22,7 @@
 #include <cstddef>
 
 #include <boost/range.hpp>
+#include <boost/static_assert.hpp>
 
 #include <boost/geometry/core/tags.hpp>
 #include <boost/geometry/core/point_type.hpp>
@@ -29,6 +34,8 @@
 #include <boost/geometry/algorithms/detail/convert_point_to_point.hpp>
 #include <boost/geometry/algorithms/detail/equals/point_point.hpp>
 
+#include <boost/geometry/util/condition.hpp>
+
 
 namespace boost { namespace geometry
 {
@@ -39,10 +46,10 @@ namespace detail { namespace point_on_border
 {
 
 
-template<typename Point>
 struct get_point
 {
-    static inline bool apply(Point& destination, Point const& source, bool)
+    template <typename Point>
+    static inline bool apply(Point& destination, Point const& source)
     {
         destination = source;
         return true;
@@ -74,13 +81,17 @@ struct midpoint_helper<Point, DimensionCount, DimensionCount>
 };
 
 
-template<typename Point, typename Range>
+template <bool Midpoint>
 struct point_on_range
 {
-    static inline bool apply(Point& point, Range const& range, bool midpoint)
+    template<typename Point, typename Range>
+    static inline bool apply(Point& point, Range const& range)
     {
+        typedef typename geometry::cs_tag<Point>::type cs_tag;
+        BOOST_STATIC_ASSERT((! Midpoint || boost::is_same<cs_tag, cartesian_tag>::value));
+
         const std::size_t n = boost::size(range);
-        if (midpoint && n > 1)
+        if (Midpoint && n > 1)
         {
             typedef typename boost::range_iterator
                 <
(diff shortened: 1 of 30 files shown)


barendgehrels:
Thanks for the PR. 
About point_on_border: it is implemented like that for overlay purposes. Returning the last vertex instead will most probably need some changes there, as soon as we want to do overlays with non-cartesian (and we want...). It probably needs a bit more research.

awulkiew:
I tried to disable the `midpoint` in `point_on_border()` and some tests started to fail: buffer, intersection, union, difference, overlay.

For now I only checked overlay and would like to get some feedback regarding my findings. There are 2 calls to `point_on_border()` with `midpoint = true` in the library:
1. https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/intersection_insert.hpp#L341
2. https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/ring_properties.hpp#L65
In the case of overlay the 2. is relevant (`ring_properties` is also used in buffer so it's possible it's related). The point calculated there is used in 3 places in overlay:
a) https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/select_rings.hpp#L264
b) https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/assign_parents.hpp#L50
c) https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/assign_parents.hpp#L143
In the failing case `case_recursive_boxes_4` union is tested and after disabling `midpoint` the check at place a) starts to give different results, this is how it looks like (midpoint enabled on the left), the point is tested for `within` with the green geometry:

overlay_union

Obviously the result of `within()` is different in both cases. But is `within()` really the correct check? What if the green geometry had a boundry there, a vertex, etc? Would it be possible to prepare such a case where the midpoint would lie on a boundary? Anyway, maybe I'll play with it later and try to break this test case.

For now I tried replacing `within()` with `covered_by()` in all 3 places (a, b, c) because AFAIU this is how rings should be tested since they may touch each other and in c) `within()` Point/Box is checked so it really doesn't make any difference because EPS is not used anyway. This fixes the overlay test. I didn't check other algorithms yet. Before that I'd like to know what do you think? Is there anything I'm missing?

Aside from that, point-in-polygon and point-in-box within/covered_by strategies should be passed into these algorithms. Actually this should be done in all places in the library. For all algorithms used internally.

barendgehrels:
Thanks for quick and clear research! I am fine with changing it to coveredby, if that fixes the new failures. I think it is not there because when written, coveredby did not exist yet. Also passing strategies is fine of course.

awulkiew:
Disabling midpoint (even with `covered_by`) breaks buffer_multi_polygon tests:
- `multipoly_zonethru_10` slightly different area but it looks the same
- `multipoly_rt_b` one polygon not included (below)
- `multipoly_rt_k` one polygon not included

obraz

It's caused by the fact that at c) (https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/assign_parents.hpp#L143) The `inner_in_map.point` is tested with `outer.envelope` which seems to be too small to cover (so also contain) the vertex point. Expanding it slightly with `expand_by_epsilon()` here: https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/assign_parents.hpp#L218 fixes the issue.

Btw, we should probably always expand envelopes slightly if they are going to be used with geometries compared with epsilon. Like in this case (c), the purpose is to check `covered_by(point, ring)` and this check is done taking the epsilon into account. If we want to do a simple envelope check before, so `covered_by(point, ring_box)`, then we should expand this envelope to cover the "effective area" of a ring. I've done something like this some time ago in sectionalize to fix an issue. I think it would be possible to generate a buffer case which would fail due to this behavior, even with midpoint enabled.

include/boost/geometry/algorithms/detail/point_on_border.hpp
```diff
@@ -273,31 +268,45 @@ struct point_on_border<multi_linestring_tag, Point, Multi>
...
+    concepts::check<Point>();
+    concepts::check<Geometry const>();
+
+    return dispatch::point_on_border
+            <
+                typename tag<Geometry>::type,
+                Midpoint
+            >::apply(point, geometry);
+}
+*/
```
barendgehrels: This is commented. Maybe we can omit this functionality. Don't think it is ever used, though we can of course not be sure about that. The bool flag was only added for overlay.
For the rest - changes look OK to me, I did not get into the details but I think it is an enhancement indeed. Thanks.

awulkiew: Should I remove the related code from the internals and Midpoint parameter, so remove this feature entirely?

What with related but not used anywhere `point_on_surface`?
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/point_on_surface.hpp

awulkiew:
@barendgehrels Strategies are supported in set operations entirely. There is only one place where `detail::point_on_border::midpoint_helper` is used:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/follow.hpp#L66
which was added in this commit:
https://github.com/boostorg/geometry/commit/e35a3b1610d61c184bbf810f8d73449d92870957
After changing this place to use the segment endpoint instead of point in between the tests both in develop and my branch pass. Do you remember the rationale behind this, or a test case which it fixed? I'd like to use the segment endpoint, is that ok?

awulkiew:
@barendgehrels Not taking into account the change I mentioned above at line:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/follow.hpp#L66
this PR is ready for merging.

PR #398, opened 2017-05-05
Feature/disjoint fix

vissarion:
Fixing disjoint algorithm for geographic computation. A main case in disjoint algorithm relies in computing the vertex longitude of a segment. Thus, tests are also added for vertex longitude computation.


include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
```diff
@@ -214,16 +143,21 @@ struct disjoint_segment_box_sphere_or_spheroid
...
+                                                            alp1,
+                                                            azimuth_strategy);
 
         segment_point_type p_vertex_rad;
         geometry::set_from_radian<0>(p_vertex_rad, vertex_lon);
         geometry::set_from_radian<1>(p_vertex_rad, vertex_lat);
 
+        std::cout << "vertex=" <<  vertex_lon * math::r2d<CT>()
+                      << " , " <<  vertex_lat * math::r2d<CT>()
+                     << std::endl;
```
awulkiew: This should either be removed or disabled by default and enabled with `#ifdef`.

include/boost/geometry/formulas/vertex_longitude.hpp
```diff
@@ -85,6 +89,8 @@ class vertex_longitude_on_spheroid
         // More constants
         CT const f = flattening<CT>(spheroid);
         CT const C1 = CT(1);
+        CT const C4 = CT(4);
```
awulkiew: That's a minor remark. In general variable names should be lower-case names. In other places of the library constants such as these are already lowe-case. In this file below `c0`, `c1` and `c3` are used for constants. So I think that they should be named consistently in this file at least, but it'd probably be good idea to use names consistently across the whole library.

include/boost/geometry/formulas/vertex_longitude.hpp
```diff
@@ -147,10 +153,15 @@ class vertex_longitude_on_spheroid
         CT const C31 = (c1 - c1 * n) * eps + c2 * eps2;
         CT const C32 = c3 * eps2;
```
awulkiew: Also here there are constants `C31` and `C32` which are conceptually somethign different, they are not representations of numbers `31` and `32`. So I think they should be named differently than other constants. So upper-case `C` is probably ok here since it makes the name different. The name could also be something more descriptive (like `sin2_sig2`).

include/boost/geometry/formulas/vertex_longitude.hpp
```diff
@@ -164,27 +175,104 @@ class vertex_longitude_on_spheroid
...
-struct vertex_longitude
+struct compute_vertex_lon
 {
     BOOST_MPL_ASSERT_MSG
-         (
-             false, NOT_IMPLEMENTED_FOR_THIS_COORDINATE_SYSTEM, (types<CS_Tag>)
-         );
+    (
+            false, NOT_IMPLEMENTED_FOR_THIS_COORDINATE_SYSTEM, (types<CS_Tag>)
+            );
```
awulkiew: There are some changes in indentation which seems to be mistakes or something automatic, also above.

awulkiew:
I'm not sure why some my comments are marked as outdated because at least one of them still holds but anyway. I'm ok with merging after addressing the comments.

include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
vissarion: This is removed in commit b7d5f4b84951df8147a570f32b0b5066da419261

PR #400, opened 2017-05-29
Use pointer_traits.to_address instead of addressof(*p)

glenfe:
addressof(*p) before construct() is not well-defined, since p does not alias storage that has an object constructed in it yet.


awulkiew:
Thanks!

PR #402, opened 2017-06-27
Fix wrong results in intersection of linestrings with a common point

vissarion:
This PR proposes a fix for the bug related to ticket https://svn.boost.org/trac10/ticket/13057 

It is about producing wrong results when intersecting two linestrings with a common point. 

EDIT: working towards a more general solution that not only work for doubles


include/boost/geometry/algorithms/detail/relate/topology_check.hpp
```diff
@@ -173,6 +173,8 @@ struct topology_check<MultiLinestring, multi_linestring_tag>
 
             std::size_t count = boost::size(ls);
 
+            m_has_interior = false;
```
awulkiew: This shouldn't be set here, It should be set outside the loop because otherwise a MultiLinestring won't have an interior if an empty Linestring is the last element of MultiLinestring.

include/boost/geometry/algorithms/detail/disjoint/box_box.hpp
```diff
@@ -121,9 +121,16 @@ struct box_box<Box1, Box2, 0, DimensionCount, spherical_tag>
             // calculate positive longitude translation with b1_min as origin
             calc_t const diff_min = math::longitude_distance_unsigned<units_t>(b1_min, b2_min);
             calc_t const b2_min_transl = b1_min + diff_min; // always right of b1_min
+            calc_t b2_max_transl = b2_min_transl - constants::period() + diff2;;
 
-            if (b2_min_transl > b1_max                                // b2_min right of b1_max
-             && b2_min_transl - constants::period() + diff2 < b1_min) // b2_max left of b1_min
+            // if the translation is too close then use the original point
+            if (math::abs(b2_max_transl - b2_max) < 100 * std::numeric_limits<calc_t>::epsilon())
```
awulkiew: This may not work for non-fundamental types for which `epsilon()` is 0. I'd prefer some different solution probably checking additional conditions e.g. if `b2_max` and `b1_min` are lesser than 180 deg, etc.

awulkiew:
Could you add an exact test of disjoint() for this edge-case? Either as a BG test case or, as a comment in this PR or in the ticket. I'd like to experiment with it.

Please don't use vague commit messages such as "Simple workarounds for bugs", it doesn't mean anything.

Also AFAIU these should rather be 2 separate commits because 2 different issues are fixed.

include/boost/geometry/algorithms/detail/disjoint/box_box.hpp
vissarion: AFAIU the unsigned difference `math::abs(b2_max_transl - b2_max)` takes values very close to k*2*pi for k=0,1,2,... so the condition could be changed to  `math::abs(b2_max_transl - b2_max) < constants::period()`. Of course, checking conditions with  `b2_max` and `b1_min` is ok.

awulkiew: How about changing `100 * epsilon() to `half_period()`?

vissarion:
Closing this PR because of many conflicts and create a new one with the desired fix.

PR #403, opened 2017-07-02
Handle degenerated segments near poles in spherical and geographic intersection strategies.

awulkiew:
This PR fixes the ticket: https://svn.boost.org/trac10/ticket/13035.

The problem is that near poles segments may be very short even though the longitudes of the endpoints are different. When the `equals(pt1, pt2)` is used to check if the segment is degenedated it may return false if longitudes are different but still near the pole the distance between the endpoints is close or equal to 0.

The spherical and geographic intersection strategies are modified in a way that distances between the segments are also taken into account when checking if a segment is degenerated to a point.


awulkiew:
I'd like to include this fix in 1.65.

PR #404, opened 2017-07-03
Fix uninitialized variable in topology check of multilinestrings

vissarion:


awulkiew:
Something bad happened here, you pushed some files containing merging artifacts. Maybe it would be easier to create a new PR since only 1 line of code has to be chenged. It could easily be done directly at GitHub.

PR #405, opened 2017-07-04
Add missing propagation of area strategy in buffer, is_valid and overlay.

awulkiew:
The area strategy isn't propagated in all places where it should be, default strategy is used. This PR changes that.

It also adds strategy parameter to `correct()` algorithm allowing to pass area strategy.


barendgehrels:
Good! Do you want to get this on 1.64?

awulkiew:
Yes, I'd like to release it with 1.64.

awulkiew:
And by 1.64 I (and probably you too) mean 1.65 ;)

barendgehrels:
Sure, of course, 1.65 :-)

PR #407, opened 2017-07-04
Fix wrong results in intersection of linestrings with a common point

vissarion:
This PR proposes a fix for the bug related to ticket https://svn.boost.org/trac10/ticket/13057

It is about producing wrong results when intersecting two linestrings with a common point.


include/boost/geometry/algorithms/detail/disjoint/box_box.hpp
```diff
@@ -121,9 +121,18 @@ struct box_box<Box1, Box2, 0, DimensionCount, spherical_tag>
             // calculate positive longitude translation with b1_min as origin
             calc_t const diff_min = math::longitude_distance_unsigned<units_t>(b1_min, b2_min);
             calc_t const b2_min_transl = b1_min + diff_min; // always right of b1_min
+            calc_t b2_max_transl = b2_min_transl - constants::period() + diff2;;
```
awulkiew: Double `;`, but I can fix it later.

awulkiew:
Thanks!

PR #410, opened 2017-07-18
Geographic distance between a point and a segment

vissarion:
This PR proposes a formula for computing the geographic distance between a point and a segment. It also adds the respective strategy. This strategy will be used by point-box, box-box and general geometry-geometry distance computation in future PR(s).

The distance is approximated by an iterative Newton method. The algorithm accepts as template parameters a direct and an inverse geodesic formula. Thus, its accuracy and performance depends on that choice. There is also a choice to compute the closest point on the input segment to the query point.


awulkiew:
Why will it be used in Pt-Box and Box-Box distance computation? Box's horizontal edges aren't geodesics so in this case the distance cannot be calculated with Pt-Seg strategy and vertical edges are meridians so it's trivial to calculate the closest point on an edge and then call Pt-Pt distance strategy. The distance between vertices must be calculated with Pt-Pt distance strategy. Am I missing something?

vissarion:
Sure, for Pt-Box and Box-Box the strategy is called with a point and a meridian segment which is an easy case and there is no need for Newton iteration, so the strategy will be called but the formula not.

include/boost/geometry/formulas/distance_point_segment.hpp
```diff
@@ -0,0 +1,335 @@
...
+namespace boost { namespace geometry { namespace formula
+{
+
+template
+<
+        typename CT,
+        typename Units,
+        typename Inverse_type_azimuth,
+        typename Inverse_type_distance,
+        typename Direct_type,
```
awulkiew: I suggest to take direct and inverse formulas as class template instead of a type here:

    template <typename, bool, bool, bool, bool ,bool> class Inverse,
    template <typename, bool, bool, bool, bool> class Direct,

And then internally define types passing appropriate parameters:

    typedef Inverse<CT, true, false, false, false, false> inverse_distance_type;
    typedef Inverse<CT, false, true, false, false, false> inverse_azimuth_type;

This way this formula would create inverse formula types able to calculate whatever is needed, so it wouldn't rely on a caller. Currently the caller has to know which results of inverse formula will be used internally and pass the correct type.

include/boost/geometry/formulas/distance_point_segment.hpp
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    template <typename Spheroid>
+    result_distance_point_segment
+    static inline apply(CT lon1, CT lat1, //p1
+                        CT lon2, CT lat2, //p2
+                        CT lon3, CT lat3, //query point p3
+                        Spheroid const& spheroid,
+                        CT earth_radius =
+                           geometry::srs::sphere<CT>().get_radius<1>())
```
awulkiew: What's the point of passing WGS84 mean radius here? AFAIU it's redundant and forces the user to pass correct radius, consistent with spheroid. I guess it's for calculating some initial values before iteration. But shouldn't it simply be calculated from passed spheroid internally?

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
```diff
@@ -0,0 +1,308 @@
...
+\tparam Spheroid is the spheroidal model used
+\tparam CalculationType \tparam_calculation
+*/
+template
+<
+    typename FormulaPolicy = strategy::andoyer,
+    typename Spheroid = srs::spheroid<double>,
+    typename CalculationType = void
+>
+class cross_track_geo
```
awulkiew: The naming scheme for geographic strategies should be consistent. Since the intersection strategy is named `geographic_segments` I propose to rename this strategy e.g. to `geographic_cross_track`.

include/boost/geometry/strategies/geographic/intersection.hpp
```diff
@@ -118,7 +118,6 @@ struct geographic_segments
         return strategy_type(m_spheroid);
     }
 
-    template <typename Geometry>
```
awulkiew: What's the reason of changing the interface of geographic intersection strategy? This way it has interface inconsistent with other intersection strategies. Plus it cannot be passed into algorithms (`equals()`) getting distance strategy from intersection strategy without changing the algorithms as well.

include/boost/geometry/strategies/geographic/parameters.hpp
```diff
@@ -24,6 +26,23 @@ namespace boost { namespace geometry { namespace strategy
 
 struct andoyer
 {
+    //TODO: this should be replaced by an andoyer direct formula
```
awulkiew: Right, implementing andoyer direct formula would be preferable to using thomas.

vissarion: This needs some research. I propose to do this in a separate PR. It is expected that using thomas direct it should be a bit slower and with small deviations in the computed result compared to the future andoyer direct formula.

vissarion:
Thanks for the review comments, I did some changes addressing them.

include/boost/geometry/formulas/distance_point_segment.hpp
```diff
@@ -0,0 +1,345 @@
...
+
+    template <typename Spheroid>
+    result_distance_point_segment
+    static inline apply(CT lon1, CT lat1, //p1
+                        CT lon2, CT lat2, //p2
+                        CT lon3, CT lat3, //query point p3
+                        Spheroid const& spheroid)
+    {
+        CT earth_radius = (CT(2) * spheroid.get_radius<1>()
+                           + spheroid.get_radius<2>()) / CT(3);
```
awulkiew: I just added `mean_radius()` formula to develop which could be used here, also the variable could be const. But this could be changed later after the PR is merged.

include/boost/geometry/formulas/distance_point_segment.hpp
```diff
@@ -0,0 +1,345 @@
...
+        typename Units,
+        template <typename, bool, bool, bool, bool ,bool> class Inverse,
+        template <typename, bool, bool, bool, bool> class Direct,
+        bool EnableClosestPoint = false
+>
+class distance_point_segment{
+
+public:
+
+    typedef Inverse<CT, true, false, false, true, true> inverse_distance_type;
```
awulkiew: Here the name doesn't fully reflect the parameters since besides distance quantities are calculated but I guess that's fine.

awulkiew:
Thanks! I'm ok with merging if you don't plan to add anything else to this PR.

include/boost/geometry/formulas/distance_point_segment.hpp
vissarion: I used the mean_radius() formula here, thanks. A side remark: is there a reason to call the argument of mean_radius() "geometry"; as far as I understand it is a coordinate system reference model, i.e. a sphere or spheroid?

vissarion:
Added some changes related to last comments. I am ok now.

include/boost/geometry/formulas/distance_point_segment.hpp
awulkiew: It's just a general name for parameter of multiple types. It could've been called `Model` as well. But a geometry could also be passed into this function, like, e.g. nsphere from extensions or some n-dimensional ellipsoid in the future, though very unlikely we'll have that.

awulkiew:
The last commit is very big and contains wierd changes. Would it be possible to make it in a clean way? I.e. reset the HEAD to the previous one (4e1e18f), merge the latest develop in your branch and after that make `mean_radius()` commit and force-push?

vissarion:
Still working on this PL, some comments below:

1. Regarding comment above on distance point to meridian segment, note that it is not trivial to compute the closest point. The misleading comes from the fact that parallels create 90 degree angles with meridians so one might think that the closest point on a meridian segment has the same latitude as the query point. But parallels are not shortest paths apart from equator so the closest point is always closer to the pole. This is the case for both spherical and geographic CS.

2. There was an issue in spherical cross_track distance computation algorithm that is fixed here. There was a non symmetric behaviour. That is, given for example a parallel segment (a,b)--(c,b) the distance from point (a,b) is not exactly the same as from point (c,b). The longer the segment the larger the error. The issue also holds for non-parallel segments. This is fixed by using the correct azimuths to compute the projection of the point on (or out of) the segment. Also formulas are used instead of strategies for azimuths that should improve performance a bit.

awulkiew:
1. You're correct, this is not trivial. Though since we know the longitude of the result I guess we can use this information in iterative solution which could simplify it.

2. Thanks! So I guess `course` include and the function itself could be removed now.

vissarion:
1. This makes sense and needs some investigation.
2. `course` has been removed from the whole library, I kept the function for backward compatibility; but I have no idea if this is used by any users.

vissarion:
@awulkiew I will create a new PL with a solution for (1). So I am done with this PL for now.

awulkiew:
Thanks!

PR #411, opened 2017-07-20
Fix geographic area strategy for sphere.

awulkiew:
In this case `e2` is `0` which resulted in `c2` being `NaN` and at the end the final result being `NaN`. This PR sets `c2` to `a2` if `e2` is `0`.

Plus unnecessary `sqrt()` call is removed.


awulkiew:
@vissarion Could you have a look if everything is ok?

vissarion:
Fine regarding the bug-fix. Thanks.
Regarding performance the strategy computes a lot of useless stuff in the a==b case. Spherical strategy should be used somehow or avoid computing quantities not used in a==b case. Anyway, this should be done in separate PR.

PR #412, opened 2017-07-21
Side strategy passing and spike test fix

awulkiew:
Changes done in this PR:
 - passing of side strategy into internals of various algorithms (e.g. `buffer`, `is_simple`, `is_valid`, `remove_spikes`).
 - support for user-defined strategy in `is_convex`.
 - fix for spike detection:
   - removed redundant code in `point_is_spike_or_equal`, replaced with `direction_code`
   - in some edge cases of very small coordinates differences function was returning invalid results, so now instead of checking both axes, the axis with greater difference determines the direction
   - `direction_code` is dispatched by coordinate system and calculated differently for non-cartesian points

### DIFF
--- a/include/boost/geometry/algorithms/detail/point_is_spike_or_equal.hpp
+++ b/include/boost/geometry/algorithms/detail/point_is_spike_or_equal.hpp
@@ -19,10 +19,13 @@
 
 #include <boost/geometry/algorithms/detail/direction_code.hpp>
 #include <boost/geometry/algorithms/detail/recalculate.hpp>
+#include <boost/geometry/core/cs.hpp>
 #include <boost/geometry/policies/robustness/robust_point_type.hpp>
 #include <boost/geometry/strategies/side.hpp>
 #include <boost/geometry/util/condition.hpp>
 #include <boost/geometry/util/math.hpp>
+#include <boost/geometry/util/normalize_spheroidal_coordinates.hpp>
+#include <boost/geometry/util/select_coordinate_type.hpp>
 
 namespace boost { namespace geometry
 {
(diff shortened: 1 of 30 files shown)


include/boost/geometry/algorithms/detail/point_is_spike_or_equal.hpp
```diff
@@ -42,33 +65,29 @@ namespace detail
...
-        // Check if it moves forward
-        int const sgn_x2 = sign_of_difference<0>(segment_b, segment_a);
-        int const sgn_y2 = sign_of_difference<1>(segment_b, segment_a);
-
-        return sgn_x1 != sgn_x2 || sgn_y1 != sgn_y2;
+#ifdef BOOST_GEOMETRY_ENABLE_POINT_IS_SPIKE_OR_EQUAL_TEST
+        bool r1 = collinear_point_is_spike_or_equal(last_point, segment_a, segment_b);
+        bool r2 = direction_code(segment_a, segment_b, last_point) < 1;
+        if (r1 != r2)
+            std::cout << "spike detection failure with: " << r1 << " " << r2 << std::endl;
```
awulkiew: Temporarily I added code comparing the previous and current version of spike detection enabled when defined `BOOST_GEOMETRY_ENABLE_POINT_IS_SPIKE_OR_EQUAL_TEST`. In case you wanted to check what's the difference, i.e. where the result was wrong. I noticed several differences in some buffer tests (AFAIR multi_polygon), union and more.

awulkiew:
I fixed more bugs related to the wrong handling of non-cartesian CS in get_turns/sectionalize, namely special handling of longitudes (wrapping) in `preceding`/`exceeding` functions and sectionalize segment direction calculation.

There is still one problematic thing though. In sectionalize the segment direction is calculated and while for longitudes it's possible to calculate it for geographic segments, for latitudes it probably is not possible because geographic segment may e.g. go first north and then south. Bounding box of such segment may have both min and max latitude both below and above the first segment's endpoint. For now it's ok because directions in dimension greater than 0 are used only in `buffer` and buffer doesn't support non-cartesian geometries anyway but in the future we should think about a replacement. Assuming my understanding is correct, i.e. it's used in buffer in similar way to get_turns.

PR #417, opened 2017-08-21
Fix winding strategy for specific cases in spherical and geographic.

awulkiew:
This PR fixes sinding strategy by adding separate strategies per CS: `cartesian_winding`, `spherical_winding` and `geographic_winding`. The last one taking geodesic `FormulaPolicy` and `Spheroid` (interface consistent with other geographic strategies).

The agnostic agnostic winding strategy is left for backward compatibility (with Boost 1.63 and below because SideStrategy parameter added in 1.64 was removed). This agnostic version internally is using one of the CS-specific strategies mentioned above based on `cs_tag<>` of `Point` parameter.


barendgehrels:
Looks good to me! Thanks.

barendgehrels:
Sorry, wrong button, reopened

vissarion:
I am ok with it too. Thanks

PR #422, opened 2017-09-08
enrich_intersection_points: fix to_index0/to_index1 confusion

MaxKellermann:
Found by Coverity:

 "pointless_expression: The expression to_index1 >= 0 && to_index1 >=
 0 does not accomplish anything because it evaluates to either of its
 identical operands, to_index1 >= 0."


barendgehrels:
Thanks a lot, this was indeed wrong. And the fix is right. Can be merged - let's do it immediately.

PR #436, opened 2017-11-24
Use to_address free function instead of pointer_traits member

glenfe:
The free function was accepted and voted into C++20 (P0653R2).


awulkiew:
Thanks! I think I'll merge it after 1.66 is released. Is that ok for you?

glenfe:
Yes, no rush. :) Thank you.

glenfe:
The corresponding changes to Boost.Core were merged to its master branch today, so if this change could also be merged to your master, it would be appreciated.

awulkiew:
@glenfe Done.

PR #444, opened 2017-12-30
Get rid of iterator.hpp

DanielaE:
Boost's iterator.hpp is deprecated (just like std::iterator in C++17). It does nothing but pulling std::iterator into namespace boost and including standard headers 'iterator' and 'cstddef'. Therefore get rid of including iterator.hpp (it had no effect in Boost.Geometry anyway) and replace inheritance by lifting std::iterator's members into the derived class. Instantiating std::iterator is granted with lengthy warning messages by latest MSVC when compiling in C++17 mode.

Signed-off-by: Daniela Engert <dani@ngrt.de>

### DIFF
--- a/example/c08_custom_non_std_example.cpp
+++ b/example/c08_custom_non_std_example.cpp
@@ -14,7 +14,6 @@
 
 #include <boost/assert.hpp>
 
-#include <boost/iterator.hpp>
 #include <boost/iterator/iterator_adaptor.hpp>
 #include <boost/iterator/iterator_categories.hpp>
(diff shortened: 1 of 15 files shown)


barendgehrels:
Thanks. Did you test this with C++03?

DanielaE:
Nope, but I can spin-up a test-run with msvc 9 if you like. If only this wouldn't take hours to finish ...

DanielaE:
Ok, ran it. I can see no problem with this patch and C++03.

awulkiew:
Thanks!

PR #445, opened 2017-12-30
Most members of std::allocate are deprecated in C++17

DanielaE:
Replace them by their cousins from std::allocator_traits.

Signed-off-by: Daniela Engert <dani@ngrt.de>

### DIFF
--- a/include/boost/geometry/index/detail/rtree/node/scoped_deallocator.hpp
+++ b/include/boost/geometry/index/detail/rtree/node/scoped_deallocator.hpp
@@ -21,7 +21,8 @@ class scoped_deallocator
     scoped_deallocator(scoped_deallocator const&);
     scoped_deallocator & operator=(scoped_deallocator const&);
 public:
-    typedef typename Alloc::pointer pointer;
+    typedef typename ::boost::container::allocator_traits<Alloc>::pointer pointer;
+
     inline scoped_deallocator(pointer p, Alloc & a)
         : m_ptr(p), m_alloc(a)
     {}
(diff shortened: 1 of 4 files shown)


awulkiew:
Thanks!

However there are 2 issues:
1. The R-tree already partially use `boost::container::allocator_traits<>` so this should be used consistently instead of `std::allocator_traits<>`. The reason is that R-tree just like Boost.Container containers emulates C++11 fetures in C++03.
2. The changes that were done in `node/variant_static.hpp` should also be done in `node/variant_dynamic.hpp`

Would you like to change the PR and propose modified version or would you prefer me to do it?

DanielaE:
Well, ok - I will bite the bullet and rework the PR. My biggest grief is the enormous time it takes to run the test-suite after each change for both C++03 and C++17. This literally takes hours 😢

awulkiew:
Thanks. I think in this case it is sufficient to run only tests relevant for the R-tree to confirm it compiles properly. E.g. from `BOOST_ROOT/libs/geometry` you can run `../../b2 index/test/rtree` or even `../../b2 index/test/rtree/generated`. Furthremore you can run multiple tests at once by passing `-j` parameter to `b2`. E.g. to run 4 R-tree tests at once: `../../b2 -j4 index/test/rtree/generated`. This should take a fraction of time compared to running all of the tests one by one if you have a machine suitable for it.

DanielaE:
Hi Adam, Happy New Year!

I've reworked the patch accordingly. I hope you like it more, now!

I'm still puzzled how the other two source files could slip through my fingers. It must have been my "test speed optimization" which didn't work as expected. Therefore I missed `boost::container::allocator_traits<>`, too. I've never heard about that before. Anyway, your tip with testing just the affected parts is just great - thanks for that! This running-tests-in-parallel thing is old news to me: I run all tests on 6 (out of my 8) cpu cores dedicated to testing.

awulkiew:
Hi Daniela, Happy New Year!

Thanks for the patch.

PR #455, opened 2018-01-23
Support both RadiusType and Sphere as template parameter in spherial strategies.

awulkiew:
For consistency with old spherical strategies and new geographic strategies allow specifying the type of member radius as either numeric radius type or sphere model. Make `double` the default radius type.


vissarion:
Thanks! I am ok for merging.

PR #460, opened 2018-02-14
Feature/simplify

barendgehrels:
The simplify feature has been improved such that:

- closing point of a polygon can be part of simplification process
- simplified output rings with zero area, or reversed area, are omitted from output
- simplification with resulting duplicate point is corrected, duplicated point is omitted
- consistent behaviour for small polygons (e.g. triangles) and pentagons (they could be reduced to two points earlier)
- updated tests 
- added unit test using countries (cf buffer), including SVG. This is optional (not compiled by default). It gives good insight in the simplification process, and showed the earlier version was not yet good enough (most islands became spikey polygons)


include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -208,9 +208,13 @@ struct simplify_ring
...
             // start/end are modified and a corresponding slice will be used
             // for simplification
+
+            // Take only 10% of the simplify distance, to avoid to aggressive
+            // behaviour at closing points (it is known as a "open problem")
+
             // TODO: for open polygons, implementation should be modified
             simplify_closure_inspector::get_start_end(start, end, ring,
-                    max_distance, strategy);
+                    max_distance / 10.0, strategy);
```
vissarion: Where does value 10% come from? Is it a well known "open problem" or it comes from some testcase?

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -116,34 +118,203 @@ struct simplify_range
...
+            // Verify area around closing point, if that can be simplified,
+            // start/end are modified and a corresponding slice will be used
+            // for simplification
+
+            // Take only a part of simplify distance, to avoid aggressive
+            // behaviour at closing points (it is known as a "open problem")
+
+            // TODO: for open polygons, implementation should be modified
+            simplify_closure_inspector::get_start_end(start, end, ring,
+                    max_distance / 2.0, strategy);
```
vissarion: Where does value 2.0 come from? Is it empirical (it was 10.0, 1000.0 etc in previous versions)? Is it a well known "open problem" or it comes from some testcase?

vissarion:
Thanks! I reviewed the code and I am ok for merging. 

A more general question: is there an example in tests where simplify destroys validity?

include/boost/geometry/algorithms/simplify.hpp
barendgehrels: The "open problem" is described here: http://cs.smith.edu/~jorourke/TOPP/P24.html#Problem.24
I did not look in more detail to those cited sources. But I understand the problem, you have to have a starting point (previously the closing point) and it is more or less an arbitrary starting point. If you rotate the polygon and do the simplification again, you might get something different.
This solution mitigates this problem by potentially simplifying the area around the closing point - but that is a different simplification, it still does not guarantee the optimal solution.

The 1000.0 earlier was actually hiding another bug in the first version (forgot to close the result), and I wanted to drastically reduce that later (10), but that is also arbitrary. And it did show some artefacts (but that can also be random). The value of 2.0 is still arbitrary too, it must be "some" value smaller than specified distance. But this did not give artefacts.

I will send you a picture later or maybe we can inlude them even in this review (Greece is one of the testcases!)

barendgehrels:
Yes, simplify can indeed destroy validity, and it does (I have some testcases). This PR enhances this (polygons with zero area omitted, empty polygons omitted) but apart from that, rings previously valid can start to self-cross. Users an call "dissolve" afterwards (if released), and PostGIS has a ST_SimplifyPreserveTopology algorithm, which we could also think about in the future.

barendgehrels:
simplify_enhancement_feb_2018

Image showing enhanced (left) and original (right)

barendgehrels:
BTW Boost.Geometry can do a better job if you use the known method of inflating a multi-polygon, followed by deflating it (+optionally finish with simplifying that result - and selecting on area). This way it stays valid (and it takes much longer).
Also an advantage is that islands close together can be merged - or merged to mainland

simplify_with_buffer_feb_2018

vissarion:
Thanks for the nice pictures of Greece! 

Regarding validity indeed postgis (actually geos) has a function that preserves topology by modifying the Douglas-Peucker algorithm. I guess that Boost.Geometry's implementation is the one of the standard algorithm, or have some heuristics?

Also that link here is broken: https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/strategies/agnostic/simplify_douglas_peucker.hpp#L260

include/boost/geometry/algorithms/simplify.hpp
vissarion: But this is a problem of the strategy (i.e. Douglas-Peucker algorithm) that offer no guarantees for the maximum distance between the original and the simplified curve. Therefore, should this heuristic of decreasing `max_distance` be inside that strategy?

barendgehrels:
Thanks for your comments @vissarion 

Yes, the buffer solution is completely different indeed. It uses the same specified distance, but not for the distance between the points, but it is then the distance to inflate/deflate.

I currently have not yet a concrete idea how to keep output valid using Douglas-Peucker (we indeed use the std implementation), especially because in a multipolygon, or with interiors, the simplified lines might easily start to cross each other (I have many samples of this).

Then about the smaller distance for the closing area. Your comment did let me rethink about this and I now realize that the new implementation might cause new artefacts (as is in fact visible at the island Kos). That can be solved by rotating the iterator, instead of current slicing, because the artefact might appear at the slice begin-end (which might be mitigated but not completely solved). I did initially think about the rotational iterator/range, but because we don't have it, and it is not in Boost (as far as I know), I took the slicing. Alternatively we can rotate the whole range, and maybe we should do that initially (to verify that behaviour), but it is less performant because it copies the range.

So this (the artefacts) can be seen as a regression, we can postpone this merge, I have to fix it. Selecting another rotation (in the sense of iterators) point will not cause such artefacts (because it is just as if the polygon starts/closes there - each polygon can be rotated over all its point while staying spatially equal). 

Rotation still does not guarantee the "optimal" solution - but that can be accepted.

barendgehrels:
Based on Vissarions comment and my reply, new implementation using rotate. This is again improving the result and removing the artifacts.
Actually it is not so bad that it is not (yet) a rotational view - also we should have the closing point there, vector is fast, and implementation is now faster anyway because I check before if the result will be empty.
Rotation is also safer to use, as I also stated above.

simplify_using_rotate

The new result where the artefact in Kos is gone, also Limnos is know valid (but this does still not mean that validity is guaranteed)

barendgehrels:
One more update, I realized that certain cases (having another collinear series of points right at end of closing area) would not be solved. This changes the solution: I now start with the farthest point (from closing point). If I'm right that is never such a collinear point (though in a circle, it nearly is), so that fixes that, and the "closing area" in general (which is now removed).
Also, if I'm right, any farthest is always on the convex hull (at least most probably), so it also fixes cases such as that spike on the Lesbos island (which was the starting point) in earlier pictures.
Also it reduces artifacts (because the starting point might not be optimal and result in an elongated triangle - but if it is the farthest point, that is much less probable).

It now looks like:
simplify_rotate2

Additionally, to avoid empty output (which can happen, it is just a kind of random process), the algorithm now tries once more at 1/4 of the polygon (and if then not found, 2 times more). That is actually optional but takes (most probably) care that polygons which could have any output, with any starting point, will have output (though it does not guarantee it - to guarantee optimal solution, all starting points on convex hull should be tried...)

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -71,41 +81,32 @@ struct simplify_range_insert
...
-            Note also that, especially if max_distance is too large,
-            the output ring might be self intersecting while the input ring is
-            not, although chances are low in normal polygons
-
-            Finally the inputring might have 3 (open) or 4 (closed) points (=correct),
-                the output < 3 or 4(=wrong)
-        */
-
-        if (boost::size(range) <= int(Minimum) || max_distance < 0.0)
+        if (boost::size(range) <= static_cast<int>(MinimumToUseStrategy) || max_distance < 0)
```
awulkiew: `boost::size` returns unsigned type so my guess is that casting is not needed and can cause a comparison warning.

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -116,34 +117,174 @@ struct simplify_range
                     range, geometry::range::back_inserter(out), max_distance, strategy
                 );
         }
+
+        // Verify the two remaining points are equal. If so, remove one of them.
+        // This can cause the output being under the minimum size
+        if (boost::size(out) == 2
+            && geometry::equals(geometry::range::front(out),
+                    geometry::range::back(out)))
```
awulkiew: This is the second time this condition is present in the code. This check could be implemented as a free function.

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -57,7 +61,13 @@ struct simplify_range_insert
...
         boost::ignore_unused(strategy);
 
-        if (boost::size(range) <= 2 || max_distance < 0)
+        if (boost::size(range) == 2
+            && geometry::equals(geometry::range::front(range),
+                                geometry::range::back(range)))
+        {
+            std::copy(boost::begin(range), boost::begin(range) + 1, out);
+        }
+        else if (boost::size(range) <= 2 || max_distance < 0)
```
awulkiew: Do these conditions work the same way for closed and open rings? Shouldn't the size be `3` in case of closed ring? Or is some particular closure assumed?

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -116,34 +117,174 @@ struct simplify_range
...
+
+            // Close the rotated copy
+            rotated.push_back(range::at(ring, index));
+
+            simplify_range<0>::apply(rotated, out, max_distance, strategy);
+
+            // Verify that what was positive, stays positive (or goes to 0)
+            // and what was negative stays negative (or goes to 0)
+            int const output_sign = area_sign(geometry::area(out));
+            if (output_sign == input_sign)
```
awulkiew: Is this the check for the same orientation? If that's the case maybe we should have a function for that? To not use area explicitly because the actual area is not used here.

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -154,12 +295,11 @@ struct simplify_polygon
                     InteriorRingsOut& interior_rings_out,
                     Distance const& max_distance, Strategy const& strategy)
     {
-        traits::resize<InteriorRingsOut>::apply(interior_rings_out,
-            boost::size(interior_rings_in));
+        traits::clear<InteriorRingsOut>::apply(interior_rings_out);
```
awulkiew: FYI, there is `geometry::range::clear(interior_rings_out)`.

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -194,16 +327,19 @@ struct simplify_multi
...
-             it_in != boost::end(multi);
-             ++it_in, ++it_out)
+                it = boost::begin(multi); it != boost::end(multi); ++it)
         {
-            Policy::apply(*it_in, *it_out, max_distance, strategy);
+            single_type single_out;
+            Policy::apply(*it, single_out, max_distance, strategy);
+            if (! geometry::is_empty(single_out))
+            {
+                traits::push_back<MultiGeometry>::apply(out, single_out);
```
awulkiew: `geometry::range::push_back(out, single_out);`

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -194,16 +327,19 @@ struct simplify_multi
     static inline void apply(MultiGeometry const& multi, MultiGeometry& out,
                     Distance const& max_distance, Strategy const& strategy)
     {
-        traits::resize<MultiGeometry>::apply(out, boost::size(multi));
+        traits::clear<MultiGeometry>::apply(out);
```
awulkiew: `range::clear(out);`

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -116,34 +117,174 @@ struct simplify_range
...
+            index = get_opposite<Strategy>(index, ring);
+
+            if (visited_indexes.count(index) > 0)
+            {
+                // Avoid trying the same starting point more than once
+                continue;
+            }
+
+            std::rotate_copy(boost::begin(ring), boost::begin(ring) + index,
+                             boost::end(ring), rotated.begin());
```
awulkiew: `boost::begin(ring) + index` could be replaced with `range::pos(ring, index)`. My guess is that because index is of unsigned type and the random access iterator is advanced with difference_type which is signed there can be a conversion warning generated here.

barendgehrels:
@awulkiew  thanks for your review. I applied your comments and will answer your questions and merge.

include/boost/geometry/algorithms/simplify.hpp
barendgehrels: Indeed, it checks for orientation only, that what was positive should stay positive, and what was negative should stay negative.

barendgehrels:
Answer on this question: _Do these conditions work the same way for closed and open rings? Shouldn't the size be 3 in case of closed ring? Or is some particular closure assumed?_
At that point, yes, it works the same way there.
In general, open polygons should still be visited because I now realize well that these could create artefacts in the past (because it simplified from start to end, which might be located very close together). Now that is solved automatically. The thing to be done is that they should be reopened (correct closure). I will do that later (though it is one line), because I want to add a unit test for that case.

include/boost/geometry/algorithms/simplify.hpp
awulkiew: I rather was thinking about making the code more self-descriptive. The area is not important here and may be confusing. The orientation is checked so if the function name reflected that then the coment wouldn't be needed because the code itself would make it clear. Furthermore enum type would be more appropriate than `int` to store the result. Therefore it could be something like `get_ring_orientation(a)` returning value of type `bg::order_selector` so `bg::clockwise`, `bg::counterclockwise ` or `bg::order_undetermined`.

awulkiew:
Btw, douglas_peucker is an agnostic strategy which takes Pt/Seg distance strategy. So theoretically it should work with other coordinate systems if non-cartesian Pt/Seg distance strategy was passed. In that case when calculating the orientation the area strategy should be derived from simplify strategy which should derive it from distance strategy. But this could be done later.

include/boost/geometry/algorithms/simplify.hpp
barendgehrels: That is a good idea indeed.

barendgehrels:
Right. Yes. BTW, I verified that it works, adding another test (simplify on a more complex latlong geometry in spherical coordinates), but I did not commit that yet (there is already a spherical test - but we should add geographic too, and indeed with a large coverage for the area check).
Thanks again for your comments.

PR #468, opened 2018-04-07
Improve projections' parameters handling.

awulkiew:
This PR makes the same changes as closed https://github.com/boostorg/geometry/pull/461 but it's implemented on top of https://github.com/boostorg/geometry/pull/465.

In short it allows to replace this:

    pj_param(pvalues, "sdatum").s; //old

with this:

    pj_get_param_s(pvalues, "datum"); //new

and this:

    //old
    if (pj_param(parameters, "tR").i)
        a = pj_param(parameters, "dR").f;

with this:

    //new
    pj_param_f(parameters, "R", a);

So the information about the parameter type is expressed in compile-time as function name instead of passed as character in cstring. But most importantly the function doesn't have to be called twice to check if the parameter is set and then get it.


vissarion:
Looks ok to me, thanks!

PR #469, opened 2018-04-07
Add support for horizontal grid shift (nadgrids) in transformation.

awulkiew:
This PR adds support for nadgrids. Relevant proj4 code was converted to C++ and Boostified. So all grids formats are supported (ctable, ctable2, ntv1, ntv2), also vertical geoid grid GTX format but it is not used right now. There are some differences though:
- proj4 stores loaded grids in global storage and pointers to a relevant subset of grids in projection parameters structure. In Boost.Geometry grids are moved outside transformation to allow users to place grids storage(s) wherever they like.
- in proj4 the grids are loaded implicitly when there is `+nadgrids` parameter passed. In Boost.Geometry an object representing a subset of grids explicitly has to be initialized and then passed into transforming function.
- in proj4 grids has to be "installed" into certain directories. In Boost.Geometry the user can implement StreamPolicy opening any input stream having unformatted input interface. The default one uses `std::ifstream` opening files having the same names as the ones in `+nadgrids` parameter in working directory.

I'm open to suggestion, in particular regarding the interface, names, etc.

This is how it looks like right now by default:

    std::string from = "+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs";
    std::string to = "+proj=aea +lat_1=34 +lat_2=40.5 +lat_0=0 +lon_0=-120 +x_0=0 +y_0=-4000000 +ellps=clrk66 +datum=NAD27 +units=m +no_defs";
    // +datum=NAD27 uses +nadgrids=@conus,@alaska,@ntv2_0.gsb,@ntv1_can.dat

    bg::srs::transformation<> tr = {bg::srs::proj4(from), bg::srs::proj4(to)};

    bg::srs::grids_storage<> gs;

    // reads headers of grids, adds them to grids_storage
    // and creates representation of grids that are used by this transformation
    bg::srs::transformation_grids<> tg = tr.initialize_grids(gs);

    point_geo san_fran(-122.419167, 37.779167);
    point_xy xy;

    tr.forward(san_fran, xy);
    std::cout << "without grids: " << bg::wkt(xy) << std::endl;

    // loads content of relevant grid if needed and stores it in grids_storage
    tr.forward(san_fran, xy, tg);
    std::cout << "with grids: " << bg::wkt(xy) << std::endl;

Now let's say a user wants to pass his own StreamPolicy opening files from directory "/abc" and share single grid storage between multiple threads:

    struct my_stream_policy
    {
        typedef std::ifstream stream_type;

        static inline void open(stream_type & is, std::string const& gridname)
        {
            is.open(std::string("/abc/") + gridname, std::ios::binary);
        }
    };

    typedef bg::srs::grids_storage
        <
            my_stream_policy,
            bg::srs::shared_grids
        > my_grids_storage;

    typedef bg::srs::transformation_grids
        <
            my_grids_storage
        > my_transformation_grids;

    // the rest is similar
    my_grids_storage gs;

    boost::thread t1{[&](){
        bg::srs::transformation<> tr = {bg::srs::proj4(from1), bg::srs::proj4(to1)};
        my_transformation_grids tg = tr.initialize_grids(gs);    
        for (int i = 0 ; i < 1000 ; ++i)
            tr.forward(san_fran, xy, tg);
    }};
    boost::thread t2{[&](){
        bg::srs::transformation<> tr = {bg::srs::proj4(from2), bg::srs::proj4(to2)};
        my_transformation_grids tg = tr.initialize_grids(gs);
        for (int i = 0 ; i < 1000 ; ++i)
            tr.forward(san_fran, xy, tg);
    }};
    t1.join();
    t2.join();

Note that sharing the same transformation object between threads should work too because neither `transformation` nor `transformation_grids` do not change while the transformation is performed:

    my_grids_storage gs;
    bg::srs::transformation<> tr = {bg::srs::proj4(from), bg::srs::proj4(to)};
    my_transformation_grids tg = tr.initialize_grids(gs);
    boost::thread t1{[&](){
        for (int i = 0 ; i < 1000 ; ++i)
            tr.forward(san_fran, xy, tg);
    }};
    boost::thread t2{[&](){
        for (int i = 0 ; i < 1000 ; ++i)
            tr.forward(san_fran, xy, tg);
    }};
    t1.join();
    t2.join();

This PR is implemented on top of https://github.com/boostorg/geometry/pull/468 and while the other PR is not merged this one will contain some unrelated changes.

### DIFF
--- a/include/boost/geometry/srs/projections/grids.hpp
+++ b/include/boost/geometry/srs/projections/grids.hpp
@@ -1,0 +1,188 @@
+// Boost.Geometry
+
+// This file was modified by Oracle on 2018.
+// Modifications copyright (c) 2018, Oracle and/or its affiliates.
+// Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_SRS_PROJECTIONS_GRIDS_HPP
+#define BOOST_GEOMETRY_SRS_PROJECTIONS_GRIDS_HPP
+
+
+#include <boost/geometry/srs/projections/impl/pj_gridinfo.hpp>
+
+#include <fstream>
+
+
+namespace boost { namespace geometry
+{
+
+namespace srs
+{
+
+// Forward declarations for functions declarations below
+class grids;
+
+template <typename GridsStorage>
+class projection_grids;
+
+} // namespace srs
+namespace projections { namespace detail
+{
+
+// Forward declaratios of grids friends
+template <typename StreamPolicy>
+inline bool pj_gridlist_merge_gridfile(std::string const& gridname,
+                                       StreamPolicy const& stream_policy,
+                                       srs::grids & grids,
+                                       std::vector<std::size_t> & gridindexes);
+template <bool Inverse, typename CalcT, typename StreamPolicy, typename Range>
+inline bool pj_apply_gridshift_3(StreamPolicy const& stream_policy,
+                                 Range & range,
+                                 srs::grids & grids,
+                                 std::vector<std::size_t> const& gridindexes);
+
+// Forward declaratios of projection_grids friends
+template <typename Par, typename GridsStorage>
+inline void pj_gridlist_from_nadgrids(Par const& defn,
+                                      srs::projection_grids<GridsStorage> & grids);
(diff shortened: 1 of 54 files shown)


vissarion:
It seems ok to me. Thanks!
One general comment, how Boost.Geometry is going to keep track of updates on proj4 (or proj5 or any later version)?

awulkiew:
Currently all updates have to be made manually.

When projections were first converted to Boost.Geometry (AFAIR it was proj4 4.9.1) at least the code containing projections was converted automatically. However the code in proj4 was changed since then so I'd expect the old converter does not work anymore. Besides, I don't know where it is and how easy would it be to update it. Plus the rest of the code has to be updated manually anyway as it was in the past.

Issue #471, opened 2018-04-09
Invalid envelope of spherical polygon


awulkiew:
It seems envelope() function return incorrect bounding box for spherical polygon in Boost 1.67.0.

To reproduce:

    #include <boost/geometry.hpp>
    #include <iostream>

    using coordinate_system = boost::geometry::cs::spherical_equatorial<boost::geometry::degree>;
    using lonlat = boost::geometry::model::point<double, 2, coordinate_system>;
    using box = boost::geometry::model::box<lonlat>;
    using polygon = boost::geometry::model::polygon<lonlat>;
    using linestring = boost::geometry::model::linestring<lonlat>;

    int main()
    {
        polygon p{{{
                {2.4, 48.9021},
                {2.4, 48.89},
                {2.3, 48.89},
                {2.3, 48.9021},
                {2.4, 48.9021}
        }}};
        linestring l{
                {2.4, 48.9021},
                {2.4, 48.89},
                {2.3, 48.89},
                {2.3, 48.9021},
                {2.4, 48.9021}
        };

        box mbr1 = boost::geometry::return_envelope<box>(p);
        box mbr2 = boost::geometry::return_envelope<box>(l);

        std::cout << std::setprecision(7) << std::fixed;
        std::cout << boost::geometry::wkt(mbr1.min_corner()) << " " << boost::geometry::wkt(mbr1.max_corner()) << std::endl;
        std::cout << boost::geometry::wkt(mbr2.min_corner()) << " " << boost::geometry::wkt(mbr2.max_corner()) << std::endl;
    }

Result:

    POINT(2.3000000 48.8900000) POINT(2.4000000 48.9021000)
    POINT(2.3000000 48.8900000) POINT(2.4000000 48.9021108)

The reason is that in the case of polygon envelope is calculated by iterating through points while in the case of linestring segments are taken into account.

awulkiew:
I'd expect the same issue with geographic CS.

PR #472, opened 2018-04-12
Fix envelope polygon

vissarion:
This PR fixes the bug reported on issue https://github.com/boostorg/geometry/issues/471.


awulkiew:
I'm ok with merging, thanks.

One remark regarding the implementation of envelope, not your fix but in general. Currently algorithm is dispatched with coordinate system. This is wrong, only strategies should depend on CS. So this is something which should be changed in the future.

awulkiew:
Btw, you could add info in release notes about fixing the issue.

awulkiew:
Thanks.

PR #478, opened 2018-04-24
Box-segment distance for spherical and geographic coordinate systems

vissarion:
This PL implements the box-segment distance computation for spherical and geographic coordinate systems (CS). 

The CS independent functionality is still in the algorithm which is generalized to support the new  coordinate systems. The CS dependent functionality is implemented in a set of new strategies. 

Along the way few fixes in envelope of segment and point-segment distance are performed. 

Now, distance should be able to be computed between any pair of geometries and for any CS. A new PL will test this.


include/boost/geometry/strategies/cartesian/distance_pythagoras.hpp
```diff
@@ -105,6 +105,24 @@ public :
...
+        BOOST_CONCEPT_ASSERT( (concepts::ConstPoint<Point1>) );
+        BOOST_CONCEPT_ASSERT( (concepts::ConstPoint<Point2>) );
+
+        assert_dimension_equal<Point1, Point2>();
+
+        return geometry::get<Index>(p1) - geometry::get<Index>(p2);
+    }
+
+    template <typename CT>
+    inline CT vertical_or_meridian(CT const& lat1, CT const& lat2) const
```
awulkiew: Similar story here. There are no meridians in cartesian. It looks like you wanted to move responsibility for some calculations from Seg-Box or Pt-Seg strategy. Are `vertical_or_meridian()` or `coordinate()` needed in distance Pt/Pt algorithm?

include/boost/geometry/strategies/cartesian/mirror_box.hpp
```diff
@@ -0,0 +1,48 @@
...
+#define BOOST_GEOMETRY_STRATEGIES_CARTESIAN_MIRROR_BOX_HPP
+
+namespace boost { namespace geometry
+{
+
+
+namespace strategy { namespace mirror_box
+{
+
+struct cartesian
```
awulkiew: What is this strategy? Is there mirror_box() algorithm? It doesn't do anything for cartesian. If there is no algorithm needing it and it doesn't do anything in castesian then my guess is that it's a utility used in some spherical and/or geographic strategy. If that's the case this shouldn't be a separate strategy.

include/boost/geometry/strategies/cartesian/segment_below_of_box.hpp
```diff
@@ -0,0 +1,64 @@
...
+                >(p1, p0, bottom_right, ps_strategy);
+    }
+};
+
+#ifndef DOXYGEN_NO_STRATEGY_SPECIALIZATIONS
+namespace services
+{
+
+template <typename LessEqual, typename ReturnType, typename PPStrategy>
+struct default_strategy<LessEqual, ReturnType, PPStrategy, cartesian_tag>
```
awulkiew: Same story here. Why is it a separate strategy and why is it possible to get default one for specific CS? Is it going to be used as a standalone strategy? In what scenario?

include/boost/geometry/strategies/geographic/azimuth.hpp
```diff
@@ -52,41 +52,52 @@ public :
         return m_spheroid;
     }
 
-    template <typename T>
+    template
+    <
+        bool EnableAzimuth,
+        bool EnableReverseAzimuth,
+        typename T
+    >
```
awulkiew: As mentioned before. I'm not sure that this is a good function design because the user needs to have template arguments and function arguments synchronized. I'd rather implement this as functions taking exactly these output arguments which will be returned. In case of overloads returning single value it could be returned from the function instead of taken as output argument. I.e.:

    strategy.apply(..., azi1, azi2);
    T azi1 = strategy.apply(...);
    T azi2 = strategy.apply_reverse(...);

include/boost/geometry/strategies/geographic/distance.hpp
```diff
@@ -125,9 +125,26 @@ public :
...
+        else
+        {
+            return apply(p1, p2);
+        }
+    }
+
     // points on a meridian not crossing poles
     template <typename CT>
-    inline CT meridian(CT lat1, CT lat2) const
+    inline CT vertical_or_meridian(CT lat1, CT lat2) const
```
awulkiew: Same here.

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
```diff
@@ -105,6 +111,30 @@ public :
...
+        typedef typename azimuth_strategy::type azimuth_type;
+        return azimuth_type(m_spheroid);
+    }
+
+    struct envelope_segment_strategy
+    {
+        typedef envelope::geographic_segment<FormulaPolicy, Spheroid, CalculationType> type;
+    };
+
+    inline typename envelope_segment_strategy::type get_envelope_segment_strategy() const
```
awulkiew: Same story if not needed by the algorithm but needed by upper-level strategy then should be implemented there. Unless there is no Seg-Box strategy and this Pt-Seg strategy is passed into distance algorithm for Seg-Box arguments. But then it'd mean that the algorithm has to be dispatched by CS because cartesian Pt-Seg strategy returns dummy strategies here.

include/boost/geometry/strategies/geographic/segment_below_of_box.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+namespace boost { namespace geometry
+{
+
+
+namespace strategy { namespace segment_below_of_box
+{
+
+template <typename LessEqual, typename ReturnType, typename PPStrategy>
+struct geographic
```
awulkiew: AFAIU this should not be a separate strategy but an internal detail of other strategy.

include/boost/geometry/strategies/spherical/azimuth.hpp
```diff
@@ -36,43 +36,48 @@ public :
...
-    static inline void apply(T const& lon1_rad, T const& lat1_rad,
-                             T const& lon2_rad, T const& lat2_rad,
-                             T& a1, T& a2)
+    template
+    <
+        bool EnableAzimuth,
+        bool EnableReverseAzimuth,
+        typename T
+    >
+    inline void apply(T const& lon1_rad, T const& lat1_rad,
```
awulkiew: The interface here as well.

include/boost/geometry/strategies/spherical/distance_cross_track_box_box.hpp
```diff
@@ -162,15 +162,15 @@ public :
                     <
                         Strategy, box_point_type1, box_point_type2
                     >::apply(ps_strategy, ps_strategy.get_distance_strategy()
-                               .meridian(lat_min1, lat_max2));
+                               .vertical_or_meridian(lat_min1, lat_max2));
```
awulkiew: Right, this function should be implemented in this strategy or some details, not in P/P strategy. It's neither needed in P/P algorithm nor P/S algorithm. It's only needed here in S/B strategy.

include/boost/geometry/strategies/spherical/distance_cross_track.hpp
```diff
@@ -358,6 +358,28 @@ public :
...
+        typedef typename azimuth_strategy::type azimuth_type;
+        return azimuth_type();
+    }
+
+    struct envelope_segment_strategy
+    {
+        typedef envelope::spherical_segment<CalculationType> type;
+    };
+
+    inline typename envelope_segment_strategy::type get_envelope_segment_strategy() const
```
awulkiew: AFAIU this is not the place for these functions.

include/boost/geometry/strategies/spherical/distance_cross_track.hpp
```diff
@@ -532,9 +554,42 @@ public :
...
```
awulkiew: AFAIU this is not the place for these functions.

include/boost/geometry/strategies/spherical/distance_cross_track_point_box.hpp
```diff
@@ -100,14 +100,16 @@ public :
                 return geometry::strategy::distance::services::result_from_distance
                         <
                             Strategy, Point, box_point_type
-                        >::apply(ps_strategy, ps_strategy.get_distance_strategy().meridian(plat, lat_max));
+                        >::apply(ps_strategy, ps_strategy.get_distance_strategy()
+                                 .vertical_or_meridian(plat, lat_max));
```
awulkiew: Same here.

include/boost/geometry/strategies/spherical/distance_haversine.hpp
```diff
@@ -94,8 +94,20 @@ public :
...
+        BOOST_CONCEPT_ASSERT( (concepts::ConstPoint<Point2>) );
+
+        assert_dimension_equal<Point1, Point2>();
+
+        return geometry::get<Index>(p1) - geometry::get<Index>(p2);
+    }
+
     template <typename T1, typename T2>
-    inline radius_type meridian(T1 lat1, T2 lat2) const
+    inline radius_type vertical_or_meridian(T1 lat1, T2 lat2) const
```
awulkiew: Same here. Is this element of the interface requried by the distance P/P algorithm? If it's a CS specific optimization used in some other strategy it should be implemented there, not here.

include/boost/geometry/algorithms/detail/envelope/segment.hpp
```diff
@@ -334,7 +333,7 @@ class envelope_segment_impl
         CalculationType lon2_rad = math::as_radian<Units>(lon2);
         CalculationType lat2_rad = math::as_radian<Units>(lat2);
         CalculationType alp1, alp2;
-        strategy.apply(lon1_rad, lat1_rad, lon2_rad, lat2_rad, alp1, alp2);
+        strategy.apply<true, true>(lon1_rad, lat1_rad, lon2_rad, lat2_rad, alp1, alp2);
```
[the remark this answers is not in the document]
vissarion: I prefer the first proposed interface (which I didn't implement before because I wanted to stick on single "apply" name but I agree it is cleaner). The second interface above call two times the azimuth formula when both alph1 and alph2 needed. Regarding switched endpoints it is not always true and that is the reason for this change here. Sometimes you have to subtract/add pi etc but this is formula's job.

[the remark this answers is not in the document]
awulkiew: The second interface only presents the two latter functions that are different. The one taking 2 output parameters would also be there. So no double calculation also in this case.

vissarion:
@awulkiew  thanks for the comments.

I created an S/B distance strategy that provides all the functionality needed from the S/B distance algorithm; e.g. it provides the utilities [segment_below_of_box](https://github.com/vissarion/geometry/blob/4a237e6878bc396557d8fe3b8c329c519da34b50/include/boost/geometry/strategies/spherical/distance_segment_box.hpp#L34) and [mirror](https://github.com/vissarion/geometry/blob/4a237e6878bc396557d8fe3b8c329c519da34b50/include/boost/geometry/strategies/spherical/distance_segment_box.hpp#L133). 

Vertical_or_meridian method is moved to P/S strategy from P/P. This is the right place since it is used by higher order strategies like P/B and S/B that derive a P/S strategy.

Regarding the interface of strategies I changed it a bit in order to have a unique interface for spherical distance strategies (apart from P/P one) i.e. the number type as a first parameter and a P/P strategy as the second. In general I think it would make things clearer to have an "as unique as possible" interface for all distance strategies but that should be done in a separate PR.

include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
```diff
@@ -64,10 +65,38 @@ struct disjoint_segment_box_sphere_or_spheroid
...
+    {
+        enum type
+        {
+            intersect,
+            disjoint_no_vertex,
+            disjoint_vertex
+        };
+        disjoint_info(type t) : t_(t){}
+        operator type () const {return t_;}
+        type t_;
```
awulkiew: Typically we put `m_` before member variable.

include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
```diff
@@ -64,10 +65,38 @@ struct disjoint_segment_box_sphere_or_spheroid
...
+    struct disjoint_info
+    {
+        enum type
+        {
+            intersect,
+            disjoint_no_vertex,
+            disjoint_vertex
+        };
+        disjoint_info(type t) : t_(t){}
+        operator type () const {return t_;}
```
awulkiew: Where is this needed? is this class needed if you're converting it to enum? Why not just use enum instead?

awulkiew:
Thanks! I'm ok with merging afer resolving conflicts.

include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
awulkiew: I saw that you did this change in merge commit. Not the best way but ok. Btw my point was not to name it `m_` but `m_t` but ok this can be done later.

Issue #475, opened 2018-04-17
Wrong union result in spherical coordinate system


remster:
Consider the test case:
```
namespace cartesian {
  using point = boost::geometry::model::d2::point_xy<double, boost::geometry::cs::cartesian>;
  using linestring = boost::geometry::model::linestring<point>;
  using polygon = boost::geometry::model::polygon<point>;
  using multipolygon = boost::geometry::model::multi_polygon<polygon>;
  using mpoint = boost::geometry::model::multi_point<point>;
}

namespace spherical {
  using point = boost::geometry::model::d2::point_xy<double, boost::geometry::cs::spherical_equatorial<boost::geometry::degree> >;
  using linestring = boost::geometry::model::linestring<point>;
  using polygon = boost::geometry::model::polygon<point>;
  using multipolygon = boost::geometry::model::multi_polygon<polygon>;
  using mpoint = boost::geometry::model::multi_point<point>;
  const double earthRadiusMts = 6371000;
}

BOOST_AUTO_TEST_CASE(testUnionBug) {
  auto p1wkt = "POLYGON((-78.4072265625001 43.06652924482626,-78.4072265625 43.06740063068311,-78.4063141178686 43.06653210403569,-78.4072265625001 43.06652924482626))";
  auto p2wkt = "POLYGON((-78.55968743491499 43.06594969590624,-78.55036227331367 43.07380195109801,-78.53503704605811 43.08248347074284,-78.51769210872999 43.08880392487917,-78.49899564953199 43.09251971058174,-78.47966844278045 43.09348761253013,-78.46045580120891 43.09167037120638,-78.44209853911326 43.08713812460473,-78.42530412309867 43.08006566649393,-78.41071917768537 43.07072563376782,-78.40631359930124 43.06653210565861,-78.55968743491499 43.06594969590624))";

  {
    spherical::polygon p1,p2;
    boost::geometry::read_wkt(p1wkt, p1);
    boost::geometry::read_wkt(p2wkt, p2);

    spherical::multipolygon out;
    boost::geometry::union_(
          p1,
          p2,
          out);
    std::cout << "Spherical union: " << boost::geometry::wkt(out) << std::endl;
  }
  {
    cartesian::polygon p1,p2;
    boost::geometry::read_wkt(p1wkt, p1);
    boost::geometry::read_wkt(p2wkt, p2);

    cartesian::multipolygon out;
    boost::geometry::union_(
          p1,
          p2,
          out);
[...]

PR #479, opened 2018-05-03
Fix for empty spherical union

awulkiew:
This PR contains fixes for some of the issues raised here: https://github.com/boostorg/geometry/issues/475.


vissarion:
thanks, I am ok with it.

awulkiew:
This PR fixes more or les half of the issues raised. In my esstimation handling the rest requires more severe changes in overlay. I'd rather do this in a separate PR.

awulkiew:
@barendgehrels What do you think? There are some changes made in this PR which are similar to the ones you made recently, e.g. greater enlargement of sections in `secionalize()`.

include/boost/geometry/policies/robustness/segment_ratio.hpp
```diff
@@ -53,8 +53,10 @@ struct less<Type, false>
     {
         BOOST_GEOMETRY_ASSERT(lhs.denominator() != 0);
         BOOST_GEOMETRY_ASSERT(rhs.denominator() != 0);
-        return lhs.numerator() * rhs.denominator()
-             < rhs.numerator() * lhs.denominator();
+        Type const l = lhs.numerator() / lhs.denominator();
```
barendgehrels: Please avoid using `l` as a variable name

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
```diff
@@ -778,8 +778,49 @@ struct sectionalize_multi
     }
 };
 
-template <typename Sections>
-inline void enlarge_sections(Sections& sections)
+// TODO: If it depends on CS it should probably be made into strategy.
```
barendgehrels: I don't think it should be a strategy.

As far as I can see, just increasing the sections by a larger amount than just an epsilon is OK for all coordinate systems. Because otherwise some intersections can be missed in rare cases.
Enlarging the sections does no harm, other than that it might compare a few segment pairs more.

Can't you just expand by a larger value, regardless of rescaling or coordinate system? Did you try that? To me that sounds better than introducing this strategy.

awulkiew: Yes right now, enlarging them by greater value should work. Providing various strategies allows to tweak enlargement values for specific formulas. It is not needed to pay the performance price in cartesian because andoyer formula has low accuracy in geographic.

However, I'm considering a possibility of passing epsilon used e.g. in point comparisons or distance comparisons in a strategy. It's because various formulas have different accuracy. If this was possible then a user would be able to define the epsilon too to something even greater and then in order to be consistent it should be possible to change the enlargement value as well.

include/boost/geometry/policies/robustness/segment_ratio.hpp
awulkiew: Is `a` and `b` clear enough?

barendgehrels: Sure, fine for me.

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
barendgehrels: OK, that sounds interesting and more generic.

awulkiew:
@barendgehrels I addressed the changes you requested. Are you ok with merging?

awulkiew:
> What is the state of that?

It's an idea for the future. We'd need umbrella strategies framework to allow the users to pass epsilon in a reasonable way. I.e. it should be done in a standalone strategy. Otherwise (right now) we'd have to take it in all strategies of operations where e.g. internally points are checked for equality, so most of them.

> I'm not so enthousiastic about this epsilon-box-enlarging strategy, per coordinate-system different. It is still there. Did you try enlarging the box by just a value? And what is the largest value you need?

I think enlarging by min/max scaled by epsilon is correct, or at least more correct than picking a constant value. It's because other operations are done like that, e.g. point equality is checked by scaling the coordinates by epsilon. If you wanted to pick a single largest value it would still depend on coordinate system. In cartesian the coordiantes can be arbitrary so in the worst case, at the end of the `double` range, e.g. when points are compared for equality this comparison is done WRT machine epsilon like that: `numeric_limits<double>::epsilon() * numeric_limits<double>::max()` [1] so around: `2.2e-16 * 1.8e+308` ~ `4.0e+292`. In order to represent points correctly (at the edges of bounding box) bounding box should be enlarged by a number from `2.2e-16` to `4.0e+292` depending on the coordinates.

And this is only when we consider point comparisons. In the library we have many different operations having different accuracies. We also compare distances which errors depend on coordinates, the distance itself and formula used. And many more.

My guess is that constant number will fail with sufficiently large geometries. E.g. considering only point comparisons I think that because for buffer the bounding box is enlarged by `0.001` this algorithm should start giving wrong results when coordinates are around `4503599627370`.

[1] See:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/util/math.hpp#L186
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/util/math.hpp#L118

awulkiew:
Btw, it could actually be an interesting experiment for all algorithms, not only buffer, to scale up the input several times and see at which point the output is be different or the algorithm fails.

awulkiew:
@barendgehrels thanks for the review.

PR #482, opened 2018-05-14
[srs] Add missing cstring include.

e-kwsm:
`geometry/algorithms/detail/relate/result.hpp` uses `memset` and `memcpy` but `cstring` is not included.


e-kwsm:
BTW, there are variants of `memcpy`, `::memcpy`, `std::memcpy`, and `::memset`, which should be qualified as `std::`.

awulkiew:
Thanks!
Actually the code would be more clear if these functions were replaced with `std::copy` and `std::fill` which internally should call `memcpy` and `memset` in this case.

PR #486, opened 2018-05-31
Introduce formula for Karney's direct geodesic method

adl1995:
This pull request introduces the direct geodesic method as proposed by [Karney (2011)](https://arxiv.org/pdf/1109.4448.pdf). The following are the highlights of the changes introduced:
- Implementation of the direct geodesic method.
- Series expansions of integrals given in the paper. All of these were calculated using Maxima with the [geod.mac](https://sourceforge.net/p/geographiclib/code/ci/release/tree/maxima/geod.mac) script (associated with GeographicLib), except for co-efficients of C3. I noticed there was a variation in the results produced by GeographicLib and the function generated through Maxima, which caused `lon2` to be inaccurate. Replacing the function with [`Geodesic::C3coeff()`](https://sourceforge.net/p/geographiclib/code/ci/release/tree/src/Geodesic.cpp#l1442) from GeographicLib removed this inaccuracy, so I have used that instead.
- Utility functions in [math.hpp](https://github.com/BoostGSoC18/geometry/blob/feature/geodesic_direct/include/boost/geometry/util/math.hpp), such as normalization, Clenshaw's summation, and Horner's evaluation. Some of these overlap with [formulas/area_formulas.hpp](https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/formulas/area_formulas.hpp). I will merge them in a single file as part of a separate PR.
- Test cases with existing dataset and antipodal points dataset, collected from [GeodTest](https://zenodo.org/record/32156), associated with GeographicLib. These were converted into C++ array format using this [Python script](https://github.com/adl1995/boost-geometry-extra/blob/master/geographicLib-dataset-parse.py). Geodesic scale (M12) is missing from the GeodTest dataset, so I generated it manually using this [C++ script](https://github.com/adl1995/boost-geometry-extra/blob/master/geographicLib-direct-antipodal.cpp).

However, the tests fail when comparing geodesic scale (M12) with Boost implementation of the direct method. Also, I noticed that using the C++ header `GeographicLib/Geodesic.hpp` and `GeodSolve` CLI tool both provided different results. As an example, using these points:

```
lat1: 31.863784754278001, lon1: 0, azi1: 29.572313410210999, s12: 19993324.8682601
```
with GeographicLib, GeodSolve CLI, and Boost gave the following results:

```
GeographicLib version:      -0.99598315084972932
GeodSolve tool:             -0.99588800900114727
Boost version:              -0.99625199787331664
```

I'm not sure why this happens, because `GeodSolve` uses the same C++ code at the back-end. The difference might not seem much, but since only a tolerance within `1e-7` is acceptable, these tests fail.


include/boost/geometry/formulas/karney_direct.hpp
```diff
@@ -0,0 +1,290 @@
...
+/*!
+\brief The solution of the direct problem of geodesics on latlong coordinates,
+       after Karney (2011).
+\author See
+- Charles F.F Karney, Algorithms for geodesics, 2011
+https://arxiv.org/pdf/1109.4448.pdf
+*/
+template <
+    typename CT,
+    size_t SeriesOrder = 8,
```
vissarion: maybe place it at the end of parameters' list to be more consistent with similar strategies

include/boost/geometry/formulas/karney_direct.hpp
```diff
@@ -0,0 +1,290 @@
...
+            // longitudinal difference.
+            CT const lon12 = lam12 / math::d2r<T>();
+
+            // Add the longitude at first point to the longitudinal
+            // difference and normalize the result.
+            result.lon2 = math::normalize_angle(math::normalize_angle(lon1) +
+                                                math::normalize_angle(lon12));
+        }
+
+        if (BOOST_GEOMETRY_CONDITION(CalcQuantities))
```
vissarion: this part of code duplicates code from https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/formulas/differential_quantities.hpp is it possible to integrate ?

include/boost/geometry/util/series_expansion.hpp
```diff
@@ -0,0 +1,1099 @@
...
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_UTIL_SERIES_EXPANSION_HPP
+#define BOOST_GEOMETRY_UTIL_SERIES_EXPANSION_HPP
+
+#include <boost/geometry/util/math.hpp>
+
+namespace boost { namespace geometry { namespace series_expansion {
+
+    /*
```
vissarion: there is a lot of duplication in the maxima scripts inside comments, could you try to decrease it?

vissarion:
Thanks for this PR!

> Series expansions of integrals given in the paper. All of these were calculated using Maxima with the geod.mac script (associated with GeographicLib), except for co-efficients of C3. I noticed there was a variation in the results produced by GeographicLib and the function generated through Maxima, which caused lon2 to be inaccurate. Replacing the function with Geodesic::C3coeff() from GeographicLib removed this inaccuracy, so I have used that instead.

lon2 is inaccurate with respect to the data set in https://zenodo.org/record/32156 ? How do you know that Geodesic::C3coeff() is more accurate than the function generated from Maxima?

 > Test cases with existing dataset and antipodal points dataset, collected from GeodTest, associated with GeographicLib. These were converted into C++ array format using this Python script. Geodesic scale (M12) is missing from the GeodTest dataset, so I generated it manually using this C++ script.

Could you please add this information in the code as a comment for future reference?

include/boost/geometry/formulas/karney_direct.hpp
adl1995: Thanks! Just updated this.

include/boost/geometry/formulas/karney_direct.hpp
```diff
@@ -0,0 +1,290 @@
...
+                                    Azi const& azimuth12,
+                                    Spheroid const& spheroid)
+    {
+        result_type result;
+
+        CT const lon1 = lo1;
+        CT const lat1 = la1;
+        Azi const azi12 = math::normalize_angle<CT>(azimuth12);
+
+        if (math::equals(distance, Dist(0)) || distance < Dist(0))
```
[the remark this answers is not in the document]
adl1995: Thanks! Just updated this.

include/boost/geometry/formulas/karney_direct.hpp
adl1995: I actually tried using that code. However, almost all the test cases fail if I do. I find that odd, since it seems to be working well with Vincenty and Thomas. I will try and dig a little deeper into the code and see if I'm missing out on a step.

include/boost/geometry/util/series_expansion.hpp
adl1995: I just removed the duplicated code in [this](https://github.com/boostorg/geometry/pull/486/commits/cc19342b4eac30183fd6f40d4d4d0a8b36fed885) commit. Please let me know if I should further reduce the comments.

vissarion:
>Yes, I compared lon2 with the dataset in (https://zenodo.org/record/32156). The way I confirmed this was by running the tests. Using the Maxima generated function, around 35 test cases fail due to lon2 being outside the accepted tolerance value (1e-7).
>The values returned using the Geodesic::C3coeff() function are:
>...

My question is about the place where the error occurs? Is it an error in the dataset or comes from the Maxima script and BG implementation. Is it a case of antipodal points? It seems that there is a big difference in the results so it should be simple to identify which result (dataset/GeographicLib or Maxima/Boost.Geometry) is correct. Could you please paste the link to a complete failing test case (lon,lat,distance,azimuth,etc..).

include/boost/geometry/util/series_expansion.hpp
adl1995: > Also the following script (taken from area formula, right?) is repeated 7 times!
...

Yes, it matches the one from area formula, except the last part `s/epsCT(2)/eps2/g;`. Before that the script was converting `eps2` -> `epsCT(2)`. So, I will just keep a single copy of this script and remove the duplicated ones.

> Finally, is the script taken from http://geographiclib.sourceforge.net/html/geod.mac without modifications? If so you can just add the link in the comments.

The scripts are indeed taken from (http://geographiclib.sourceforge.net/html/geod.mac), however, I had to make minor adjustments e.g. change the function header and data types to keep it consistent with Boost Geometry guidelines. Is it okay to discard those changes? If so, I will only provide a link to `geod.mac`.

vissarion: >The scripts are indeed taken from (http://geographiclib.sourceforge.net/html/geod.mac), however, I had to make minor adjustments e.g. change the function header and data types to keep it consistent with Boost Geometry guidelines. Is it okay to discard those changes? If so, I will only provide a link to geod.mac, otherwise, I will add them in a separate file.

Ok, since they contain changes, I think it is useful to keep them in Boost.Geometry.

adl1995:
Quoting my previous comment:

> ~Yes, I compared lon2 with the dataset in (https://zenodo.org/record/32156).~

Sorry, I just checked again and it seems I confused the two datasets. The inaccuracy in `lon2` is only with dataset/Boost.Geometry. The tests for antipodal points (collected from dataset/GeographicLib) pass regardless of which function is used (i.e. the one generated through Maxima script or [Geodesic::C3coeff()](https://sourceforge.net/p/geographiclib/code/ci/release/tree/src/Geodesic.cpp#l1442)).

However, for dataset/Boost.Geometry, the tests for `lon2` fail (36/88 cases) if I use the Maxima generated function. I have pasted the failing test case here: https://gist.github.com/adl1995/9a9f5dada2b6a083b7b1d79ae35c3ca3

[line#226 in karney_direct.hpp](https://github.com/BoostGSoC18/geometry/blob/feature/geodesic_direct/include/boost/geometry/formulas/karney_direct.hpp#L226) is responsible for the error. As an example, the points:
```
lon1 = 1, lat1 = 1, s12 = 250000, azi12 = -45 
```
give the following results for `lon2`:
```
GeographicLib:        -0.589430658212281
Boost:                -0.589432892643727
Test case (expected): -0.589430658212280
``` 
Also, if I replace the `Geodesic::C3coeff()` function in GeographicLib with the one generated through Maxima, I get the exact result provided through Boost (inaccurate). This leads me to conclude that the issue is indeed with the Maxima function.

PS. the values returned by the `Geodesic::C3coeff()` or Maxima function are not dependent on the dataset. It only takes `n = f / (2 - f)` and the coefficient array as parameter. 

Please let me know if I should provide any further details.

include/boost/geometry/util/series_expansion.hpp
adl1995: I have moved the Maxima scripts to: [doc/other/maxima/geod.mac](https://github.com/BoostGSoC18/geometry/blob/feature/geodesic_direct/doc/other/maxima/geod.mac).

vissarion:
@adl1995  thanks for the details.

Still I do not understand how do you recognize which case is the accurate and which the inaccurate one. I see that you assume that original GeographicLib code gives the accurate result and maybe you are correct but I do not see the reasoning. Is it possible for example that GeographicLib returns an inaccurate result using `Geodesic::C3coeff()` and the use of the Maxima function yields the accurate one?

adl1995:
The reason I am inclined to recognize `Geodesic::C3coeff()` function as accurate is that it produces results that match with dataset/Boost.Geometry.

However, I agree that this alone should not be the basis of identifying the inaccuracy. I have looked through the paper [Algorithms for geodesics](https://arxiv.org/pdf/1109.4448.pdf) in which Eq. (25) seems to describe the C3 series expansion. Also, Section 5 and Eq. (53) in the technical report [Geodesics on an ellipsoid of revolution](https://arxiv.org/pdf/1102.1215) provide a more detailed explanation.

With that said, I am not sure how I should go on about debugging this. I am not much familiar with integrals, so confirming if the series expansions are indeed correct might be difficult.

I did notice that the function [evaluate_coeffs_C3_I3](https://github.com/vissarion/geometry/blob/39813fa732e75116b5498ef4f8ca33e6e585087c/include/boost/geometry/formulas/series_expansion_direct.hpp#L330) in the prototype implementation you provided differs from the one I generated through Maxima ([code_C3.txt](https://github.com/boostorg/geometry/files/2075956/code_C3.txt)). Did you use the same [geod.mac](https://geographiclib.sourceforge.io/html/geod.mac) script for generating this?

vissarion:
I took them from the papers you mentioned, if I remember correctly. The look the same with `Geodesic::C3coeff()` for what order did you yet the inaccuracy issue? What happens with other orders?

adl1995:
I just tried modifying the series order. The tests for `lon2` pass when order is set to `2`, but fail at all other values i.e. `3`, `4`, `5`, `6`, `7`, `8`.

vissarion:
> I just tried modifying the series order. The tests for lon2 pass when order is set to 2, but fail at all other values i.e. 3, 4, 5, 6, 7, 8.

ok this sheds some light. It seems that the Boost.Geometry dataset is created with order 2 series so you can use that order to your tests too.

adl1995:
@vissarion Thanks for the suggestion!

I have now replaced `Geodesic::C3coeff()` with Maxima generated function. All tests are passing with series order 2 with dataset/Boost.Geometry.

For dataset/GeographicLib, however, the tests still fail for geodesic scale (M12), which was calculated manually since it is absent from GeodTest.dat. Previously, this was calculated for series order 6. I have now updated it with series order 7. This reduces the number of failing test cases to 87 (previously 99).

Quoting from [GeodTest Zenodo page](https://zenodo.org/record/32156):

> These [test set values] are computed using high-precision direct geodesic calculations.

The instructions for [building GeographicLib using high-precision arithmetic](https://geographiclib.sourceforge.io/html/highprec.html) state that it requires the quadmath library. But, I'm unable to find it using `apt`. @vissarion, have you tried building the library this way before? I'm not sure it this will resolve the above issue with M12, but it might be worth a try.

vissarion:
>The instructions for building GeographicLib using high-precision arithmetic state that it requires the quadmath library. But, I'm unable to find it using apt. @vissarion, have you tried building the library this way before? I'm not sure it this will resolve the above issue with M12, but it might be worth a try.

There is a `libquadmath0` debian package. I haven't build GeographicLib with high precision arithmetic, but I agree that it worth a try. Before that what about testing with series order 8---instead of 7.

adl1995:
> There is a libquadmath0 debian package. I haven't build GeographicLib with high precision arithmetic, but I agree that it worth a try. Before that what about testing with series order 8---instead of 7.

I already tried testing with series order 8, and all lower values, but I get the same results.

So, I just calculated [geodesic scale (M12) using high precision arithmetic](https://github.com/boostorg/geometry/pull/486/commits/2cba2fa83fe62c52efd5fc2d7fcb1401264701cf). I only had to change the compiler to `gcc-7`, which comes pre-packaged with `libquadmath`. However, that did not resolve the issue, as the tests are still failing. It seems the issue must be with the library implementation. I'll try debugging it again.

If you have any suggestions, please let me know.

adl1995:
@vissarion The issue above is now fixed! It turned out to be a minor calculation mistake which led to the inaccuracy in M12. I had multiplied `(sin_sigma2 * sin_sigma1)` instead of adding them `(sin_sigma2 + sin_sigma1)`.

All tests are passing now. Please let me know if this pull request requires further changes.

include/boost/geometry/util/math.hpp
```diff
@@ -771,6 +771,125 @@ inline Result rounding_cast(T const& v)
     return detail::rounding_cast<Result, T>::apply(v);
 }
 
+/*!
+\brief Evaluate the sine and cosine function with the argument in degrees
```
vissarion: why is it needed to compute the trigonometric functions with the argument in degrees and not in radians. The input of direct formulas are in radians

include/boost/geometry/util/math.hpp
```diff
@@ -771,6 +771,125 @@ inline Result rounding_cast(T const& v)
...
+
+    x /= h; y /= h;
+}
+
+
+/*!
+\brief Normalize a given angle.
+*/
+template<typename T>
+    inline T normalize_angle(T x)
```
vissarion: tab

include/boost/geometry/util/math.hpp
```diff
@@ -771,6 +771,125 @@ inline Result rounding_cast(T const& v)
...
+    while (it != begin);
+    return result;
+}
+
+/*
+\brief Evaluate the polynomial.
+*/
+template<typename CT>
+inline CT polyval(int N,
+                         const CT coeff[],
```
vissarion: tab

doc/other/maxima/geod.mac
```diff
@@ -0,0 +1,232 @@
+/*
```
vissarion: copyright info

vissarion: similarly add your name to the contributors in the begging of every file that you change

include/boost/geometry/util/math.hpp
```diff
@@ -771,6 +771,125 @@ inline Result rounding_cast(T const& v)
...
+
+    // z - (z - y) must not be simplified to y.
+    y = y < z ? z - (z - y) : y;
+
+    return x < 0 ? -y : y;
+}
+
+/*!
+\brief Normalize the given values.
+*/
```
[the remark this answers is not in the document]
vissarion: Also, maybe the right place to put normalization functions is util/normalize_spheroidal_coordinates.hpp and try not to duplicate functionality with the already implemented functions there.

include/boost/geometry/util/math.hpp
adl1995: Actually, the input to Karney direct is in degrees. This can be seen in [direct.cpp](https://github.com/BoostGSoC18/geometry/blob/feature/geodesic_direct/test/formulas/direct.cpp#L44) test file.

doc/other/maxima/geod.mac
adl1995: @vissarion Regarding the copyright, one whose behalf should I add this? And since `geod.mac` file is taken from GeographicLib, shouldn't it contain its respective copyright?

include/boost/geometry/util/math.hpp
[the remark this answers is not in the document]
adl1995: I believe [normalize_angle](https://github.com/BoostGSoC18/geometry/blob/feature/geodesic_direct/include/boost/geometry/util/math.hpp#L848) and [normalize_longitude](https://github.com/BoostGSoC18/geometry/blob/feature/geodesic_direct/include/boost/geometry/util/normalize_spheroidal_coordinates.hpp#L358) serve the same purpose. However, using the latter for normalizing, say azimuth, would be semantically incorrect. In this case, should I keep the `normalize_angle` function?

doc/other/maxima/geod.mac
vissarion: >Regarding the copyright, one whose behalf should I add this?

Just yourself. You can acknowledge gsoc. 

>And since geod.mac file is taken from GeographicLib, shouldn't it contain its respective copyright?

You can add something similar to this:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/srs/projections/proj/aea.hpp

include/boost/geometry/util/math.hpp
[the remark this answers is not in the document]
vissarion: If they are the same you can try to create a general normalize_angle in util/normalize_spheroidal_coordinates.hpp and the call it from  normalize_longitude as well as your code.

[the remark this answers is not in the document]
adl1995: I have added moved the `normalize_angle` function to [util/normalize_spheroidal_coordinates.hpp#L375](https://github.com/BoostGSoC18/geometry/blob/feature/geodesic_direct/include/boost/geometry/util/normalize_spheroidal_coordinates.hpp#L375), where it calls `normalize_longitude` under the hood.

Also, I noticed that `normalize` from `math.hpp` is already present in [formulas/vertex_longitude.hpp#L59](https://github.com/BoostGSoC18/geometry/blob/feature/geodesic_direct/include/boost/geometry/formulas/vertex_longitude.hpp#L59). I will merge these in a separate PR. For now, this has also been moved to `util/normalize_spheroidal_coordinates.hpp`.

doc/other/maxima/geod.mac
adl1995: I've just added the [copyright information and updated the contributors list](https://github.com/boostorg/geometry/pull/486/commits/82c5c4d20c8f51dbd9d9e788b824ad06675a897e). Please let me know if it requires changes.

vissarion: "on behalf of" sounds a bit unclear. What about, "as part of"? Or "partially supported by gsoc18" in a separate sentence?

adl1995: I have replaced it with "... as part of Google Summer of Code 2018 program".

include/boost/geometry/util/math.hpp
adl1995: Currently, the internal functions assume input in degrees. It might be possible to make the switch to radians. However, wouldn't it be simpler if the user could provide input in radians and we convert it into degrees at the beginning? Or, would this be inefficient?

Neither the paper nor GeographicLib documentation go into detail on why this design was chosen. I don't think the computations are any faster because of this, because the angles are manually converted into radians at various points in the code.

vissarion: >Currently, the internal functions assume input in degrees. It might be possible to make the switch to radians. However, wouldn't it be simpler if the user could provide input in radians and we convert it into degrees at the beginning? Or, would this be inefficient?

I am not sure that I understand what you want to say. The user can use either radians or degrees and this is converted to radians and passed to formulas for computations.

>Neither the paper nor GeographicLib documentation go into detail on why this design was chosen. I don't think the computations are any faster because of this, because the angles are manually converted into radians at various points in the code.

If there are a lot of conversions it should be investigated whether this can be avoided by replacing the current function by one that uses only radians. Maybe in a separate PR.

adl1995: @vissarion So, I will perform the conversion to radians in a separate PR. I have created an [issue](https://github.com/BoostGSoC18/geometry/issues/2) as a reminder for myself.

If there are no further changes required with this PR, can we please merge this?

vissarion:
@awulkiew do you have any comments on this PR?

include/boost/geometry/util/series_expansion.hpp
```diff
@@ -0,0 +1,679 @@
...
+    }
+
+    /*
+    \brief Given the set of coefficients coeffs2[] evaluate on
+      C3 and return the set of coefficients coeffs1[].
+
+      Elements coeffs1[1] through coeffs1[SeriesOrder - 1] are set.
+    */
+    template <typename CT, size_t SeriesOrder>
+    static inline void evaluate_coeffs_C3(CT eps, CT coeffs1[], CT coeffs2[])
```
awulkiew: `CT const& eps`

The coeffs parameters are confusing. Their sizes should be part of types. It is not clear which is input and which output.

include/boost/geometry/util/series_expansion.hpp
```diff
@@ -0,0 +1,679 @@
...
+    {
+        CT mult = 1;
+        int offset = 0;
+
+        // l is the index of C3[l].
```
awulkiew: If this was separate type and the code was describing itself you wouldn't need to write this comment.

include/boost/geometry/util/series_expansion.hpp
```diff
@@ -0,0 +1,679 @@
...
+        for (size_t l = 1; l < SeriesOrder; ++l)
+        {
+            // Order of polynomial in eps.
+            int m = SeriesOrder - l - 1;
+            mult *= eps;
+            coeffs1[l] = mult * math::polyval(m, coeffs2 + offset, eps);
```
awulkiew: Why does the first element is not set?

Right, so `polyval` takes a subrange of `coeffs2`. The order of parameters is nonintuitive. Typically length of array is passed after array's name. But I'd encourage you to use STL algorithm kind of interface, and take begin and end iterator of subrange.

    math::polyval(coeffs2 + offset, coeffs2 + offset + m, eps);

But this is not optimal, an improvement would be to move the iterator/pointer instead of keeping offset separately, so:

    math::polyval(coeffs2, coeffs2+m, eps);
    coeffs2 += m + 1;

One variable less. However with `boost::array<>` or other class you'd still have to create separate iterator:

    const_iterator it_C3x = coeffs2.begin()
    //...
    coeffs1[l] = mult * math::polyval(it_C3x, it_C3x + m, eps);
    it_C3x += m + 1;

include/boost/geometry/util/series_expansion.hpp
```diff
@@ -0,0 +1,679 @@
...
+
+     using Clenshaw summation.
+    */
+    template <typename CT, size_t SeriesOrder>
+    static inline CT sin_cos_series(CT sinx, CT cosx, const CT coeffs[])
+    {
+        size_t n = SeriesOrder;
+
+        // Point to one beyond last element.
+        coeffs += (n + 1);
```
awulkiew: Right, so in `karney_direct.hpp` line 220 there is an error in size of coeffs_C3.

If compile-time of coeffs was part of coeffs' type and you had static assertion here the compiler would inform you about it.

include/boost/geometry/util/series_expansion.hpp
```diff
@@ -0,0 +1,679 @@
...
+        CT ar = 2 * (cosx - sinx) * (cosx + sinx);
+
+        CT k0 = n & 1 ? *--coeffs : 0;
```
awulkiew: Is this test for odd value?

include/boost/geometry/formulas/karney_direct.hpp
```diff
@@ -0,0 +1,290 @@
...
+        CT const cos_B11 = cos(B11);
+
+        CT const sin_tau1
+            = sin_sigma1 * cos_B11 + cos_sigma1 * sin_B11;
+        CT const cos_tau1
+            = cos_sigma1 * cos_B11 - sin_sigma1 * sin_B11;
+
+        // Index zero element of coeffs_C1p is unused.
+        CT coeffs_C1p[SeriesOrder + 1];
+        series_expansion::evaluate_coeffs_C1p<CT, SeriesOrder>(epsilon, coeffs_C1p);
```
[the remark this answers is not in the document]
adl1995: @awulkiew Yes, epsilon is only a parameter.

Reiterating what you said above, in case I use `boost::array`, I won't have to explicitly provide the template parameters, but they will still be part of the function header, right?

This works fine for most cases, however, in some cases the array size is defined as below:
```cpp
size_t const coeffs_C3_size = (SeriesOrder * (SeriesOrder - 1)) / 2;
```
The array can be defined as:
```cpp
boost::array<CT, coeffs_C3_size> coeffs_C3x;
```
Which is then passed to this function:
```cpp
template <typename CT, size_t SeriesOrder>
static inline void evaluate_coeffs_C3x(CT const& n, boost::array<CT, SeriesOrder>& c)
{ ... }
```
In this case, the `SeriesOrder` template parameter will be the updated array size, hence the `switch` statement will not have any effect, since the value is out of bound.

One solution is to pass an additional argument to the function, which represents `SeriesOrder`. Another way to solve this is by converting the series expansions functions to class methods, and storing `SeriesOrder` as a class member. Please let me know if a there's a better alternative.

I have [this commit](https://github.com/BoostGSoC18/geometry/commit/1972bcda3e5c204f2a0d0bda66239725828f1306), which might make it easier to follow what I am trying to say above.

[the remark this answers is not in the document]
awulkiew: @adl1995 Yes the most simple solution is to pass `SeriesOrder` as separate parameter:

    template <size_T SeriesOrder, typename CT, size_t N>
    inline void evaluate_coeffs_C3x(boost::array<CT, N>& c, CT const& n)
    {
        BOOST_MPL_ASSERT(N == (SeriesOrder * (SeriesOrder - 1)) / 2);
        // ...
    }

but then the caller of this function has to know what's the required size of array for passed `SeriesOrder`. It would be better to have it all in one place. Would it have sense to implement several representations of coefficient containers? So in this case something like:

    template <typename T, size_t SeriesOrder>
    struct coeffs_C1 : boost::array<T, SeriesOrder + 1>
    {
        coeffs_C1(T const& epsilon)
        {
            // the content of evaluate_coeffs_C1()
        }
    };

    template <typename T, size_t SeriesOrder>
    struct coeffs_C3x : boost::array<T, (SeriesOrder * (SeriesOrder - 1)) / 2)>
    {
        coeffs_C3x(T const& n)
        {
            // the content of evaluate_coeffs_C3x()
        }
    };

    // etc.

this way the caller would only have to know the `SeriesOrder`. This would be the same for all coefficients. He wouldn't be forced to know how to calculate the size of array, if he should add 1 to size or not, etc. Everything would be defined in one place.

And then `sin_cos_series()` which AFAIU always calculates the series using all elements of array would take the size from `static_size` member (derived from boost::array<>)

    template <typename CT, typename Coeffs>
    inline CT sin_cos_series(CT const& sinx, CT const& cosx, Coeffs const& coeffs)
    {
        // and then:
        size_t const size = Coeffs::static_size;
        CT el0 = coeffs[0];
    }

Does it have sense or is there something about coeffiients that I don't understand?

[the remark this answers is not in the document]
adl1995: @awulkiew Thanks. Yes, this makes perfect sense. This will also make the code in `karney_direct.hpp` simpler. Also, `series_expansion.hpp` seems like a better place to include size computation logic for coefficient containers.

include/boost/geometry/formulas/karney_direct.hpp
```diff
@@ -0,0 +1,290 @@
...
+            series_expansion::evaluate_coeffs_C3<CT, SeriesOrder>(epsilon, coeffs_C3, coeffs_C3x);
+
+            CT const B31 =
+                series_expansion::sin_cos_series<CT, SeriesOrder>(sin_sigma1, cos_sigma1, coeffs_C3);
+
+            CT const lam12 = omega12 + A3c *
+                             (sigma12 + (series_expansion::sin_cos_series<CT, SeriesOrder>
+                                                           (sin_sigma2,
+                                                            cos_sigma2,
+                                                            coeffs_C3) - B31));
```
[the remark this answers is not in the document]
adl1995: I think the `sin_cos_series` function is fixed now. Previously,
I was passing `SeriesOrder` for all coefficients, but now the
size is queried from `boost::array`'s `static_size` member.

So, if the container size is `9`, `n` is set to `(9 - 1)`. This line:
```
CT k0 = n & 1 ? coeffs[--index] : 0;
```
checks if `n` is odd (not in this case), so `k0` is set to `0`.

The loop runs 4 times because of `n /= 2`.
So, in total 8 elements are accessed (excluding the first one).

If the container size is `8`, then all the container elements are
used for computing the result.

include/boost/geometry/util/math.hpp
```diff
@@ -771,6 +772,104 @@ inline Result rounding_cast(T const& v)
...
+    while (it != begin);
+    return result;
+}
+
+/*
+\brief Evaluate the polynomial.
+*/
+template<typename CT>
+inline CT polyval(int N,
+                  const CT coeff[],
```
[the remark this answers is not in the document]
adl1995: The array size changes with each iteration in `evaluate_coeffs_C3` function, so it is not compile-time constant.

I've just [updated](https://github.com/BoostGSoC18/geometry/commit/9c96bec2bf8521e720a97674b3b93c626769e702) this function to take `std::vector` as the parameter, and obtained its size using `boost::size(coeff)`.

include/boost/geometry/util/series_expansion.hpp
```diff
@@ -0,0 +1,679 @@
...
+    /*
+     The coefficients C3[l] in the Fourier expansion of B3.
+
+     The expansion below is performed in Maxima, a Computer Algebra System.
+     The C++ code (that yields the function evaluate_coeffs_C3 below) is
+     generated by the following Maxima script:
+     geometry/doc/other/maxima/geod.mac
+    */
+    template <typename CT, int SeriesOrder>
+    static inline void evaluate_coeffs_C3x(CT const& n, CT c[]) {
```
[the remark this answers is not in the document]
adl1995: I have changed the function header as:
```cpp
template <typename Coeffs, typename CT>
static inline void evaluate_coeffs_C3x(Coeffs &c, size_t const& SeriesOrder, CT const& n) {
```
I'm currently using:
```cpp
BOOST_GEOMETRY_ASSERT(coeff_size == (SeriesOrder * (SeriesOrder - 1)) / 2);
```
which works fine. However, if I use:
```cpp
BOOST_MPL_ASSERT(coeff_size == (SeriesOrder * (SeriesOrder - 1)) / 2);
```
I get this error:
```cpp
./boost/geometry/util/series_expansion.hpp: In function ‘void boost::geometry::series_expansion::evaluate_coeffs_C3x(Coeffs&, const size_t&, const CT&)’:
./boost/mpl/assert.hpp:60:51: error: expected primary-expression before ‘enum’
 #   define BOOST_MPL_AUX_ASSERT_CONSTANT(T, expr) enum { expr }
                                                   ^
./boost/mpl/assert.hpp:285:1: note: in expansion of macro ‘BOOST_MPL_AUX_ASSERT_CONSTANT’
 BOOST_MPL_AUX_ASSERT_CONSTANT( \
 ^
./boost/geometry/util/series_expansion.hpp:576:9: note: in expansion of macro ‘BOOST_MPL_ASSERT’
BOOST_MPL_ASSERT(coeff_size == (SeriesOrder * (SeriesOrder - 1)) / 2);
         ^
```

Is this because both values have to be compile-time constant?

include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
```diff
@@ -363,6 +364,35 @@ inline void normalize_longitude(CoordinateType& longitude)
...
+/*!
+\brief Short utility to normalize an angle on a spheroid
+       normalized in range (-180, 180].
+\tparam Units The units of the coordindate system in the spheroid
+\tparam CoordinateType The type of the coordinates
+\param angle Angle
+\ingroup utility
+*/
+template <typename Units, typename CoordinateType>
+inline void normalize_angle(CoordinateType& angle)
```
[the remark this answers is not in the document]
adl1995: Would `normalize_coordinate` or `normalize_1d_coordinate` be a better name for this?

include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
```diff
@@ -363,6 +364,35 @@ inline void normalize_longitude(CoordinateType& longitude)
...
+/*!
+\brief Normalize the given values.
+\tparam ValueType The type of the values
+\param x Value x
+\param y Value y
+TODO: adl1995 - Merge this function with
+formulas/vertex_longitude.hpp
+*/
+template<typename ValueType>
+inline void normalize_values(ValueType& x, ValueType& y)
```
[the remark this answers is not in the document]
adl1995: Should I rename this as `normalize_2d_coordinates`?

include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
```diff
@@ -363,6 +364,35 @@ inline void normalize_longitude(CoordinateType& longitude)
...
+{
+    ValueType h = boost::math::hypot(x, y);
+
+    x /= h; y /= h;
```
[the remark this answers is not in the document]
adl1995: I have added the assertion:
```
BOOST_GEOMETRY_ASSERT(h > 0);
```

include/boost/geometry/util/series_expansion.hpp
adl1995: Yes, `k0` is set to `*--coeffs` if `n` is odd.

I have added a comment in the file.

include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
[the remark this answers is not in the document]
awulkiew: AFAIU you're using it to normalize longitude and azimuth so you could use `normalize_longitude()` for the former and call this one `normalize_azimuth()` and use it for the latter.

include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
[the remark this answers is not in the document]
awulkiew: You use it to normalize unit vector so maybe `normalize_unit_vector()` or something along the lines. Something which in the name has a part indicating that the input has to have non-zero length.

include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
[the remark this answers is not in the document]
awulkiew: If you're sure that the input should have length greater than `0` then assertion is fine.

include/boost/geometry/formulas/karney_direct.hpp
```diff
@@ -0,0 +1,266 @@
...
+        math::sin_cos_degrees<CT>(math::round_angle<CT>(azi12), sin_alpha1, cos_alpha1);
+
+        // Find the reduced latitude.
+        CT sin_beta1, cos_beta1;
+        math::sin_cos_degrees<CT>(math::round_angle<CT>(lat1), sin_beta1, cos_beta1);
+        sin_beta1 *= one_minus_f;
+
+        math::normalize_values<CT>(sin_beta1, cos_beta1);
+
+        cos_beta1 = std::max(math::sqrt(c0), cos_beta1);
```
awulkiew: `std::max` won't compile with msvc. Take it in brackets. Furthermore sqrt(0) is 0. So:

    (std::max)(c0, cos_beta1);

or am I missing something?

include/boost/geometry/formulas/karney_direct.hpp
```diff
@@ -0,0 +1,266 @@
...
+                                    Azi const& azimuth12,
+                                    Spheroid const& spheroid)
+    {
+        result_type result;
+
+        CT lon1 = lo1;
+        CT const lat1 = la1;
+
+        Azi azi12 = azimuth12;
+        math::normalize_angle<degree, Azi>(azi12);
```
awulkiew: So if you called it `normalize_azimuth()` then it'd be clear that it has to be in range (-180, 180].

include/boost/geometry/util/math.hpp
```diff
@@ -771,6 +772,106 @@ inline Result rounding_cast(T const& v)
...
+    T remainder; int quotient;
+
+    remainder = std::fmod(x, T(360));
+    quotient = int(std::floor(remainder / 90 + T(0.5)));
+    remainder -= 90 * quotient;
+
+    // Convert to radians.
+    remainder *= d2r<T>();
+
+    T s = std::sin(remainder), c = std::cos(remainder);
```
awulkiew: Before you used math functions without `std::`. If you specify `std::` then the compiler will not use functions from the namespace where a user-defined numeric type is defined. So please remove `std::`. If you like you can bring `std::` math functions into scope:

    using std::floor;
    using std::sin;
    using std::cos;

but they are not used like that in the whole library so I'd say this is not needed.

include/boost/geometry/util/math.hpp
```diff
@@ -771,6 +772,106 @@ inline Result rounding_cast(T const& v)
...
+template<typename T>
+inline void sin_cos_degrees(T const& x,
+                            T & sinx,
+                            T & cosx)
+{
+    // In order to minimize round-off errors, this function exactly reduces
+    // the argument to the range [-45, 45] before converting it to radians.
+    T remainder; int quotient;
+
+    remainder = std::fmod(x, T(360));
```
awulkiew: Use `math::mod()`

include/boost/geometry/util/math.hpp
```diff
@@ -771,6 +772,106 @@ inline Result rounding_cast(T const& v)
...
+template<typename T>
+inline T round_angle(T x) {
+    static const T z = 1/T(16);
+
+    if (x == 0)
+    {
+        return 0;
+    }
+
+    T y = std::abs(x);
```
awulkiew: `math::abs()`

include/boost/geometry/util/series_expansion.hpp
```diff
@@ -0,0 +1,746 @@
...
+
+     The expansion below is performed in Maxima, a Computer Algebra System.
+     The C++ code (that yields the function evaluate_coeffs_C3 below) is
+     generated by the following Maxima script:
+     geometry/doc/other/maxima/geod.mac
+    */
+    template <typename Coeffs, typename CT>
+    inline void evaluate_coeffs_C3x(Coeffs &c, size_t const& SeriesOrder, CT const& n) {
+        size_t const coeff_size = Coeffs::static_size;
+        BOOST_GEOMETRY_ASSERT(coeff_size == (SeriesOrder * (SeriesOrder - 1)) / 2);
```
awulkiew: If you took `SeriesOrder` as template parameter you could check this at compile time:

    static size_t const coeff_size = Coeffs::static_size;
    BOOST_MPL_ASSERT(((coeff_size == (SeriesOrder * (SeriesOrder - 1)) / 2)));

include/boost/geometry/util/series_expansion.hpp
```diff
@@ -0,0 +1,746 @@
...
+        int offset = 1;
+
+        // l is the index of C3[l].
+        for (size_t l = 1; l < Coeffs1::static_size; ++l)
+        {
+            // Order of polynomial in eps.
+            int m = Coeffs1::static_size - l - 1;
+            mult *= eps;
+
+            std::vector<CT> coeffs2_slice(coeffs2.begin(), coeffs2.begin() + offset);
```
awulkiew: You don't have to create a container here, one each iteration btw. This will be horribly wrong. Pass range into `math::polyval()`.

include/boost/geometry/util/math.hpp
```diff
@@ -771,6 +772,106 @@ inline Result rounding_cast(T const& v)
...
+    }
+    while (it != begin);
+    return result;
+}
+
+/*
+\brief Evaluate the polynomial.
+*/
+template<typename CT>
+inline CT polyval(std::vector<CT> coeff,
```
awulkiew: template <typename It>
    inline CT polyval(It first, It last, std::iterator_traits<It>::value_type const& eps)

or just

    template <typename It, typename T>
    inline CT polyval(It first, It last, T const& eps)

include/boost/geometry/util/math.hpp
```diff
@@ -771,6 +772,106 @@ inline Result rounding_cast(T const& v)
...
+                  const CT eps)
+{
+    int N = boost::size(coeff) - 1;
```
awulkiew: std::distance(first, last) - 1

include/boost/geometry/util/math.hpp
```diff
@@ -771,6 +772,106 @@ inline Result rounding_cast(T const& v)
...
+    int index = 0;
+
+    CT y = N < 0 ? 0 : coeff[index++];
```
awulkiew: *(first + (index++))

include/boost/geometry/util/series_expansion.hpp
adl1995: I'm having some trouble passing `SeriesOrder` as the template parameter.

I changed the `evaluate_coeffs_C3x` function header to:
```cpp
template <typename Coeffs, typename CT, size_t SeriesOrder>
inline void evaluate_coeffs_C3x(Coeffs &c, CT const& n) {
```
and called it as:
```cpp
evaluate_coeffs_C3x(*this, n);
```
but I get this error:
```
error: ‘evaluate_coeffs_C3x’ was not declared in this scope
```
I also tried changing the order of parameters as:
```cpp
template <size_t SeriesOrder, typename CT, typename Coeffs>
inline void evaluate_coeffs_C3x(CT const& n, Coeffs &c) {
```
but calling the function as:
```cpp
evaluate_coeffs_C3x(n, *this);
```
or:
```cpp
evaluate_coeffs_C3x<SeriesOrder, CT>(n, *this);
```
doesn't work either.

It produces the error:
```
error: expected primary-expression before ‘>’ token
```
and:
```
couldn't deduce template parameter ‘SeriesOrder’
```
respectively.

awulkiew: Either of these should work:

    template <size_t SeriesOrder, typename CT, typename Coeffs>
    inline void evaluate_coeffs_C3x(CT const& n, Coeffs &c){}

    // this is more strict, it takes only the struct you like
    template <size_t SeriesOrder, typename CT>
    inline void evaluate_coeffs_C3x(CT const& n, coeffs_C3x<SeriesOrder, CT> &c){}

and either of these should work:

    evaluate_coeffs_C3x<SeriesOrder>(n, *this);
    evaluate_coeffs_C3x<SeriesOrder, CT>(n, *this);
    evaluate_coeffs_C3x<SeriesOrder, CT, coeffs_C3x>(n, *this); // assuming it's called in coeffs_C3x's constructor
    
If you're calling these functions inside structs' constructors then they should be defined before structs.

adl1995: Thanks, this worked.

Out of curiosity, why was this working before for the previous function calls? Does this have something to do with passing explicit template parameters? As I only got the error with `evaluate_coeffs_C3x<SeriesOrder>`.

awulkiew: No idea. You'd have to know how exactly type deduction is done in your compiler. Are you using Visual Studio? Its compiler allows things it shouldn't sometimes.

adl1995: No, I'm using `g++` version `7.2.0` on Ubuntu.

adl1995:
Thank you for the detailed reviews. I think all the requested changes have been addressed. I have also merged `develop` into this branch, and resolved all conflicting files.

Please let me know if there are any further changes required.

awulkiew:
Thanks! I'm ok with merging. I'd prefer to wait until after the 1.68 release. Or do you need this PR merged in order to work on the next step of the project?

adl1995:
That's fine. I can work on local branches for the next part i.e. the inverse method.

Btw, when is the next release scheduled for?

awulkiew:
Here is the schedule: https://www.boost.org/development/index.html
The branches for final release are closed 1 Aug however after beta is released we only do bugfixing.

I think the whole month is too long to keep you wait so I'm ok with merging it now. If needed I'll create a branch for bugfixing based on previous develop state.

adl1995:
Okay, as you deem fit. If it is merged before August 14th, I can mention this in my final evaluation report.

PR #487, opened 2018-06-06
Distance between any two geometries for spherical and geographic cs

vissarion:
This PR is implemented on top of https://github.com/boostorg/geometry/pull/478 and tests the functionality of the distance function between any pair of geometries for spherical and geographic CS.


awulkiew:
Thanks!

PR #488, opened 2018-06-07
Fix/closing iterator returns reference to temp

jonasdmentia:
This PR addresses an issue with both closing_iterator and ever_circling_range_iterator where they can return a reference to a temporary object. This happens when using a range where Range::reference is a value type (in my particular case, I hit this with boost::transformed_range). This happens because neither of them were honoring the range's reference type, they were assuming it was 'value_type const&'.

The first commit adds the tests, both of which fail with a SEGFAULT, and the second commit fixes the iterators.

Built and tested on Ubuntu 16.04 with gcc 5.4:
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.9) 5.4.0 20160609

### DIFF
--- a/include/boost/geometry/iterators/closing_iterator.hpp
+++ b/include/boost/geometry/iterators/closing_iterator.hpp
@@ -39,7 +39,7 @@ struct closing_iterator
         closing_iterator<Range>,
         typename boost::range_value<Range>::type const,
         boost::random_access_traversal_tag,
-        typename boost::range_reference<Range>::type const,
+        typename boost::range_reference<Range const>::type,
         typename boost::range_difference<Range>::type
     >
 {
@@ -49,7 +49,7 @@ private:
             closing_iterator<Range>,
             typename boost::range_value<Range>::type const,
             boost::random_access_traversal_tag,
-            typename boost::range_reference<Range>::type const,
+            typename boost::range_reference<Range const>::type,
             typename boost::range_difference<Range>::type
         > base_type;
 
--- a/include/boost/geometry/iterators/closing_iterator.hpp
+++ b/include/boost/geometry/iterators/closing_iterator.hpp
@@ -38,9 +38,12 @@ struct closing_iterator
     <
         closing_iterator<Range>,
         typename boost::range_value<Range>::type const,
-        boost::random_access_traversal_tag
+        boost::random_access_traversal_tag,
+        typename boost::range_reference<Range>::type const,
+        typename boost::range_difference<Range>::type
     >
 {
+    typedef typename boost::range_reference<Range>::type const reference_type;
     typedef typename boost::range_difference<Range>::type difference_type;
 
     /// Constructor including the range it is based on
@@ -71,7 +74,7 @@ struct closing_iterator
 private:
     friend class boost::iterator_core_access;
 
-    inline typename boost::range_value<Range>::type const& dereference() const
+    inline reference_type dereference() const
     {
         return *m_iterator;
     }
--- a/include/boost/geometry/iterators/closing_iterator.hpp
+++ b/include/boost/geometry/iterators/closing_iterator.hpp
@@ -43,8 +43,19 @@ struct closing_iterator
         typename boost::range_difference<Range>::type
     >
 {
-    typedef typename boost::range_reference<Range>::type const reference_type;
-    typedef typename boost::range_difference<Range>::type difference_type;
+private:
+    typedef boost::iterator_facade
+        <
+            closing_iterator<Range>,
+            typename boost::range_value<Range>::type const,
+            boost::random_access_traversal_tag,
+            typename boost::range_reference<Range>::type const,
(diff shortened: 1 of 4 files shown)


awulkiew:
Thanks!

Note that transformed range is SinglePassRange, not RandomAccesRange which is required by the library (see Linestring and Ring concepts). It's because as you noticed non-real-reference has to be returned and the standard allows only InputIterator to do that. AFAIR `iterator_facade` will model InputIterator even though `boost::random_access_traversal_tag` was passed. So in general if you wanted to be safe you should avoid passing non-RandomAccessRanges because someone, somewhere in the code might assume that passed ranges and iterators should work in a certain way, i.e. always return real reference. For instance, instead of `range_reference<const Range>::type` someone might have used `range_value<Range>::type const&`. Reference might have been wrapped inside a class and stored in a container which would result in dangling reference, e.g. consider this:

    template <typename T>
    struct ref_wrapper
    {
        wrapper(T const& r) : ref(r) {}
        T const& ref;
    };

    // ...

    // Range const& range
    ref_wrapper<boost::range_value<const Range>::type> ref(*boost::begin(range));
    // at this point wrapped reference is dangling reference if Range's iterator returns temporary object convertible to value_type

With that said I think the code should work in practice so I'm ok with the change. But if you want to be more complete you could go through the code and see if the iterators you just modified are used as intended.

test/iterators/ever_circling_iterator.cpp
```diff
@@ -74,7 +76,7 @@ void test_geometry(std::string const& wkt)
     // Check the range_iterator-one
     {
         std::ostringstream out;
-        bg::ever_circling_range_iterator<G> it(geo);
+        bg::ever_circling_range_iterator<G const> it(geo);
```
mkaravel: Why did you need to modify this line? I would expect the ever_circling range iterator to work with or with out the ``const``.

include/boost/geometry/iterators/closing_iterator.hpp
```diff
@@ -38,9 +38,12 @@ struct closing_iterator
     <
         closing_iterator<Range>,
         typename boost::range_value<Range>::type const,
-        boost::random_access_traversal_tag
+        boost::random_access_traversal_tag,
+        typename boost::range_reference<Range>::type const,
+        typename boost::range_difference<Range>::type
     >
 {
+    typedef typename boost::range_reference<Range>::type const reference_type;
```
mkaravel: I think the proper way to define those is to use the types from the `iterator_facade` class.
Same for `difference_type`.

mkaravel: It really does not matter: I just checked the ``iterator_facade`` class and the types you provide are those used to define ``iterator_facade::reference`` and ``iterator_facade::difference_type``. So you code is perfectly fine.

My comment should be understood as a stylistic one.

test/iterators/ever_circling_iterator.cpp
jonasdmentia: ever_circling_range_iterator stores a non-const pointer to Range, but if the range you're constructing the iterator with is const then you end up trying to assign a pointer to a const range to the non-const pointer.  I think ever_circling_range_iterator is doing the right thing here, it assumes the const-ness is captured as part of the Range template parameter, it's just that at the call site the G template parameter _doesn't_ capture the const-ness and the variable geo is a 'G const&'....hence the explicit const.

This didn't matter before because geo was a local non-const variable, but now that I've separated the WKT parsing and pass geo in as 'G const&' it does. Note that the same thing was already being done in the iterator typedef used with ever_circling_iterator (i.e. the non-range version), I suspect it wasn't actually needed before for the same reason (geo was non-const) but it is now.

include/boost/geometry/iterators/closing_iterator.hpp
jonasdmentia: Yep, totally agree. Fixed.

include/boost/geometry/iterators/closing_iterator.hpp
```diff
@@ -38,10 +38,24 @@ struct closing_iterator
     <
         closing_iterator<Range>,
         typename boost::range_value<Range>::type const,
-        boost::random_access_traversal_tag
+        boost::random_access_traversal_tag,
+        typename boost::range_reference<Range>::type const,
```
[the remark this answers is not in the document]
jonasdmentia: I just carried forward the same style used with boost::range_value above it on line 40, but thinking about it there could be a difference if the range is using a proxy type instead of a regular reference (which I don't think would be an issue with range's value type?). I'll fix it up.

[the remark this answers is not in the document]
awulkiew: I think for value type it should be ok but we'll see. If anything goes wrong the tests on various platforms should detect that.

awulkiew:
Thanks!

PR #495, opened 2018-06-25
Minor fixes for compilation with VxWorks 7

kuhlenough:
Change \<host-os\> to \<target-os\>  in one Jamfile for cross-compilation.
And change <proj to fully qualified <boost::geometry::srs::par4::proj
to avoid confusing clang when compiling with the Dinkum STL
which has inline proj() in global namespace


awulkiew:
Thanks!

Issue #498, opened 2018-07-02
Unexpected results with expand() and make_inverse() for spherical and geographic CS


vissarion:
The code below (as posted in boost.geometry mailing list) returns 

```
((0, 0), (5, 5))
((0, 0), (5, 5))
((-180, 0), (180, 5))
((-3.14159, 0), (3.14159, 5))
```

which is unexpected. There is an issue in the initialization of the box using `make_inverse`. Note that if the box is initialized by the first point `point_type{0,0}` of the expansion the result is as expected. 

```
#include <boost/geometry.hpp>
#include <boost/geometry/algorithms/expand.hpp>
#include <boost/geometry/core/cs.hpp>
#include <boost/geometry/geometries/box.hpp>
#include <boost/geometry/geometries/point_xy.hpp>

#include <iostream>

// XXX -- default template parameters correspond to the Boost.Geometry
//        example; see
// https://www.boost.org/doc/libs/1_67_0/libs/geometry/doc/html/geometry/reference/algorithms/expand/expand_2.html
//
template < typename Coordinate = short int
           , typename CS = boost::geometry::cs::cartesian
           >
struct expand_example
{
  static void apply ()
  {
    using point_type = boost::geometry::model::d2::point_xy<Coordinate,CS>;
    using box_type = boost::geometry::model::box<point_type>;

    using boost::geometry::expand;

    box_type box = boost::geometry::make_inverse<box_type>();

    expand(box, point_type{0,0});
    expand(box, point_type{1,2});
    expand(box, point_type{5,4});
    expand(box, boost::geometry::make<box_type>(3,3,5,5));

    std::cout << boost::geometry::dsv(box) << std::endl;
  }
};

int main ()
{
  using namespace boost::geometry;
  expand_example<>::apply();
  expand_example<double,cs::cartesian>::apply();
  expand_example<double,cs::geographic<degree>>::apply();
  expand_example<double,cs::geographic<radian>>::apply();
}```

PR #499, opened 2018-07-06
Fix make_inverse and expand issue 

vissarion:
This PR proposes a fix for issue https://github.com/boostorg/geometry/issues/498 by avoiding normalization for boxes created by ```make_inverse``` in spherical and geographic coordinate systems.

### DIFF
--- a/include/boost/geometry/util/is_inverse.hpp
+++ b/include/boost/geometry/util/is_inverse.hpp
@@ -32,10 +32,10 @@ bool is_inverse(Box const& box)
     bound_type high = bounds<bound_type>::highest();
     bound_type low = bounds<bound_type>::lowest();
 
-    return math::equals(geometry::get<0, 0>(box), high) &&
-           math::equals(geometry::get<0, 1>(box), high) &&
-           math::equals(geometry::get<1, 0>(box), low) &&
-           math::equals(geometry::get<1, 1>(box), low);
+    return (geometry::get<0, 0>(box) == high) &&
+           (geometry::get<0, 1>(box) == high) &&
+           (geometry::get<1, 0>(box) == low) &&
+           (geometry::get<1, 1>(box) == low);
 }
 
 }} // namespace boost::geometry
--- a/include/boost/geometry/util/is_inverse.hpp
+++ b/include/boost/geometry/util/is_inverse.hpp
@@ -1,0 +1,43 @@
+// Boost.Geometry
+
+// Copyright (c) 2018 Oracle and/or its affiliates.
+
+// Contributed and/or modified by Vissarion Fysikopoulos, on behalf of Oracle
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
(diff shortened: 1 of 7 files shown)


include/boost/geometry/util/is_inverse.hpp
```diff
@@ -0,0 +1,43 @@
...
+template <typename Box>
+bool is_inverse(Box const& box)
+{
+    typedef typename point_type<Box>::type point_type;
+    typedef typename coordinate_type<point_type>::type bound_type;
+
+    bound_type high = bounds<bound_type>::highest();
+    bound_type low = bounds<bound_type>::lowest();
+
+    return math::equals(geometry::get<0, 0>(box), high) &&
```
awulkiew: Only for 2d boxes?

Btw, is it needed to call `math::equals`? Why not operators `<=` and `>=`? After assignment of the bounds we should be certain that these are the exact numbers. `math::equals` uses machine epsilon scaled by the number's magnitude so it's slower and not exact.

vissarion: >Only for 2d boxes?
its use is intended for spherical and geographic CS

>Btw, is it needed to call math::equals? Why not operators <= and >=? After assignment of the bounds we should be certain that these are the exact numbers. math::equals uses machine epsilon scaled by the number's magnitude so it's slower and not exact.
yes ```==``` is better

awulkiew: Right, it's for testing longitude and latitude since the rest of the coordinates is treated like cartesian and handled in recursive template. And in this case and also for the first two coordinates in case of cartesian box the inverse case is handled automatically on per-coordinate basis. So I think that this tool should NOT be called `is_inverse()` because it is too general. The name should indicate that this is specifically for the spherical/geographic coordinates. Historically we add `spheroidal_coordinates` or something like this to the name.

After valid coordinates are assigned they should no longer contain any of the bounds values. Why not use `OR` in the condition? If any of the coordinates is out of bounds just assign new point? If we wanted to be prepared for the case when only one of the coordinates e.g. max corner's longitude was set then we should check one coordinate at a time, one dimension at a time in expand(). And I think I'd prefer this, i.e. to have a tool in expand's details `set_or_expand<D, I>()` checking the coordinate explicitly for bounds if needed and seting or expanding in place. This way all coordinates would be independent from each other. Currently they are tied together.

Side note: Both longitude and latitude are set at once because the code is prepared for the case of the change of both longitude and latitude during normalization. This is not needed in practice because the normlization is not done separately (normalization of both can be enabled with macro #define but I doubt anybody knows that). So we could simplify the code by agreeing that since now we will always normalize separately. In this case the coordinates could be handled one dimension at a time also in non-cartesian CSes. And in this case this tool should check only one dimension.

vissarion: I changed the name of the utility to reflect the fact that it is used for spheroidal coordinates. 

Regarding comments on checking per dimension I agree but I think this PR is not the right place to implement such a general change and should be implemented in a separate PR that should also take care of other possible cases where spheroidal coordinates are threaded together. Also adding the utility under the ```detail::expand``` namespace is not a good idea as the code is right now since it is also used by ```envelope```.

awulkiew: Ok, I agree that it should be done in separate PR, thanks.

PR #502, opened 2018-07-26
Example: usage of distance formula

adl1995:
This pull request adds a simple example on how to use a formula (in this case, Karney's direct method).

### DIFF
--- a/example/ml03_distance_formula.cpp
+++ b/example/ml03_distance_formula.cpp
@@ -1,0 +1,37 @@
+// Boost.Geometry
+
+// Copyright (c) 2018 Adeel Ahmad, Islamabad, Pakistan.
+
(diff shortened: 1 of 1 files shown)


awulkiew:
I think this PR should be cleaned up before merging. It seems to add code related to `karney_inverse` formula and strategy which AFAIU was merged recently with additional fixes. It adds many unnecessary merge commits, etc. I thought that GitHub should take it into account and show only the differing parts of the code? Am I missing something?

vissarion:
I agree. I asked Adeel to have a look. It seems that only the last 2
commits are needed.

adl1995:
I just rebased this branch against develop. In fact we only need a single commit in this PR.

awulkiew:
@adl1995 Thanks!

Issue #513, opened 2018-10-08
9438-returns: call of overloaded ‘ignore_unused_variable_warning(...)’ is ambiguous


Magnati:
Hello dear team of Boostorg.

First, I am not sure wheter this is the right repository to post this issue, if not, I will grandly repost it somewhere else.

The error is pretty much the same as in the closed issue [#9438](https://svn.boost.org/trac10/ticket/9438#no1) and I am not sure why it reappeared.

- OS: Ubuntu 18.04 Bionic Beaver
- clang: 6.0
- lemon: 1.3.1
- boost: happened under 1.63 and  was reproduced with boost 1.68

```c++
#include <boost/geometry.hpp>
#include <lemon/list_graph.h>

namespace bg = boost::geometry;
namespace bgi = boost::geometry::index;

using Point = boost::geometry::model::point<double, 2, bg::cs::cartesian>;
using Graph = lemon::ListDigraph;
using Node = lemon::ListDigraph::Node;
typedef std::pair<Point, Node> Pair;
using BoostRTree = bgi::rtree<Pair, bgi::quadratic<16>>;
using BoostQueryResult = std::vector<Pair>;

int main()
{
	Point one = Point(10, 51);
	Point two = Point(10, 51);
	BoostRTree rtreePair = BoostRTree();
	Graph graph;
	Node dummy1 = graph.addNode();
	Node dummy2 = graph.addNode();

	rtreePair.insert(std::make_pair(one, dummy1));	
	return 0;
}

```

mloskot:
FYI, a `lemon`-stripped version of your example:

```
#include <boost/geometry.hpp>

namespace bg = boost::geometry;
namespace bgi = boost::geometry::index;

using Point = boost::geometry::model::point<double, 2, bg::cs::cartesian>;
struct Node {};
struct Graph { Node addNode() { return Node(); } };
typedef std::pair<Point, Node> Pair;
using BoostRTree = bgi::rtree<Pair, bgi::quadratic<16>>;
using BoostQueryResult = std::vector<Pair>;

int main()
{
    Point one = Point(10, 51);
    Point two = Point(10, 51);
    BoostRTree rtreePair = BoostRTree();
    Graph graph;
    Node dummy1 = graph.addNode();
    Node dummy2 = graph.addNode();

    rtreePair.insert(std::make_pair(one, dummy1));	
    return 0;
}
```

[...]

PR #514, opened 2018-10-12
Replace boost::ignore_unused_variable_warning with boost::ignore_unused

mloskot:
The former is provided by Boost.ConceptCheck.
The latter is an improved equivalent provided by Boost.Core.

Related to #513

### DIFF
--- a/doc/src/examples/algorithms/create_svg_overlay.hpp
+++ b/doc/src/examples/algorithms/create_svg_overlay.hpp
@@ -14,6 +14,7 @@
 #include <fstream>
 
 #include <boost/foreach.hpp>
+#include <boost/core/ignore_unused.hpp>
 #include <boost/algorithm/string.hpp>
 
 #if defined(HAVE_SVG)
@@ -45,10 +46,7 @@ void create_svg(std::string const& filename, Geometry const& a, Geometry const& 
                     "fill:rgb(0,0,0);font-family:Arial;font-size:10px");
     }
 #else
-    boost::ignore_unused_variable_warning(filename);
-    boost::ignore_unused_variable_warning(a);
-    boost::ignore_unused_variable_warning(b);
-    boost::ignore_unused_variable_warning(range);
+    boost::ignore_unused(filename, a, b, range);
 #endif
 }
 
--- a/include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
@@ -14,6 +14,7 @@
 #ifndef BOOST_GEOMETRY_ALGORITHMS_DETAIL_OVERLAY_GET_TURN_INFO_FOR_ENDPOINT_HPP
 #define BOOST_GEOMETRY_ALGORITHMS_DETAIL_OVERLAY_GET_TURN_INFO_FOR_ENDPOINT_HPP
 
+#include <boost/core/ignore_unused.hpp>
 #include <boost/geometry/core/assert.hpp>
 #include <boost/geometry/algorithms/detail/overlay/get_turn_info.hpp>
 #include <boost/geometry/policies/robustness/no_rescale_policy.hpp>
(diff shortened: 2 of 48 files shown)


doc/src/examples/algorithms/create_svg_overlay.hpp
```diff
@@ -45,10 +46,10 @@ void create_svg(std::string const& filename, Geometry const& a, Geometry const&
...
     }
 #else
-    boost::ignore_unused_variable_warning(filename);
-    boost::ignore_unused_variable_warning(a);
-    boost::ignore_unused_variable_warning(b);
-    boost::ignore_unused_variable_warning(range);
+    boost::ignore_unused(filename);
+    boost::ignore_unused(a);
+    boost::ignore_unused(b);
+    boost::ignore_unused(range);
```
awulkiew: This could be:

        boost::ignore_unused(filename, a, b, range);

What is the intention in using one call per variable? E.g. do you think it is clearer this way?

mloskot: @awulkiew Good point. I've updated the commit.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
```diff
@@ -402,10 +403,7 @@ struct get_turn_info_for_endpoint
     {
         typedef typename cs_tag<typename TurnInfo::point_type>::type cs_tag;
 
-        boost::ignore_unused_variable_warning(i2);
-        boost::ignore_unused_variable_warning(j2);
-        boost::ignore_unused_variable_warning(ip_index);
-        boost::ignore_unused_variable_warning(tp_model);
+        boost::ignore_unused(i2, j2, (ip_index, tp_model);
```
awulkiew: oops.

include/boost/geometry/geometries/concepts/polygon_concept.hpp
```diff
@@ -65,12 +66,12 @@ class Polygon
...
-            boost::ignore_unused_variable_warning(ce);
-            boost::ignore_unused_variable_warning(ci);
-            boost::ignore_unused_variable_warning(poly);
-            boost::ignore_unused_variable_warning(cpoly);
+            boost::ignore_unused(e);
+            boost::ignore_unused(i);
+            boost::ignore_unused(ce);
+            boost::ignore_unused(ci);
+            boost::ignore_unused(poly);
+            boost::ignore_unused(cpoly);
```
awulkiew: One function call.

include/boost/geometry/geometries/concepts/polygon_concept.hpp
```diff
@@ -115,9 +116,9 @@ class ConstPolygon
             ring_const_type ce = traits::exterior_ring<const_polygon_type>::get(*cpoly);
             interior_const_type ci = traits::interior_rings<const_polygon_type>::get(*cpoly);
 
-            boost::ignore_unused_variable_warning(ce);
-            boost::ignore_unused_variable_warning(ci);
-            boost::ignore_unused_variable_warning(cpoly);
+            boost::ignore_unused(ce);
+            boost::ignore_unused(ci);
+            boost::ignore_unused(cpoly);
```
awulkiew: One function call.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
mloskot: oops, thanks!

include/boost/geometry/geometries/concepts/polygon_concept.hpp
mloskot: I skipped those as the list of argument is quite long

```
#ifdef BOOST_NO_CXX11_VARIADIC_TEMPLATES
            boost::ignore_unused(e, i, ce, ci, poly, cpoly);
#else
            boost::ignore_unused(e, i, ce, ci);
            boost::ignore_unused(poly, cpoly);
#endif
```

I'll replace with two calls

awulkiew: Right, the emulation supports up to 5 arguments right now. So it's ok to call twice. I'd remove `#ifdef` because it needlessly complicates the code. With it it's 6 lines, the same number of lines as if you called once per variable.

mloskot: Yes, it's called twice now. The PR has been updated.

awulkiew: Actually, this is the last place where one function could be called. Two calls we're talking about are above.

mloskot: Yes, it's fixed.

Issue #517, opened 2018-10-17
Compilation error for the example of adapted legacy geometry object model


xventura81:
Trying the example code 
[Example model](https://www.boost.org/doc/libs/1_68_0/libs/geometry/doc/html/geometry/examples/example__adapting_a_legacy_geometry_object_model.html)
[Adapted example model](https://www.boost.org/doc/libs/1_68_0/libs/geometry/doc/html/geometry/examples/example_source_code__adapting_a_legacy_geometry_object_model.html)

... I got the following build error (Visual Studio 15.8.6):
```
namespace bg = boost::geometry;

int main()
{
	QPolygon polygon;
	std::vector<QPoint> points = { {0,0}, {0,1}, {1,0} };
	bg::append(polygon, points);
}
```

Error	C2664	'void boost::geometry::detail::append::append_range<QRing,Range>::apply(Geometry &,const Range &,int,int)': cannot convert argument 1 from 'const QRing' to 'Geometry &'	BoostGeometryExample	...\boost.1.68.0.0\lib\native\include\boost\geometry\algorithms\append.hpp	127	

This error is pointing to the following code in append.hpp:

```
template <typename Polygon, typename Range>
struct range_to_polygon
{
    typedef typename ring_type<Polygon>::type ring_type;

    static inline void apply(Polygon& polygon, Range const& range,
                int ring_index, int = 0)
    {
        if (ring_index == -1)
        {
            append_range<ring_type, Range>::apply(
                        exterior_ring(polygon), range);
        }
        else if (ring_index < int(num_interior_rings(polygon)))
        {
            append_range<ring_type, Range>::apply(
                        range::at(interior_rings(polygon), ring_index), range);
        }
    }
};
```

The r-value returned by `interior_rings` might be being linked to a const& and assume that this might be causing the error.

xventura81:
I think those issues should be early detected through the unit tests. Might we be missing some additional tests using adapted geometry models?

PR #518, opened 2018-10-18
Avoid passing temporaries as function parameters when an L-value reference is expected

xventura81:
Solve problem described in #517

### DIFF
--- a/include/boost/geometry/algorithms/append.hpp
+++ b/include/boost/geometry/algorithms/append.hpp
@@ -91,19 +91,23 @@ template <typename Polygon, typename Point>
 struct point_to_polygon
 {
     typedef typename ring_type<Polygon>::type ring_type;
+    typedef typename ring_return_type<Polygon>::type exterior_ring_type;
+    typedef typename interior_return_type<Polygon>::type interior_ring_range_type;
 
     static inline void apply(Polygon& polygon, Point const& point,
                 int ring_index, int = 0)
     {
         if (ring_index == -1)
         {
+            exterior_ring_type ext_ring = exterior_ring(polygon);
             append_point<ring_type, Point>::apply(
-                        exterior_ring(polygon), point);
+                        ext_ring, point);
         }
         else if (ring_index < int(num_interior_rings(polygon)))
         {
+            interior_ring_range_type int_rings = interior_rings(polygon);
             append_point<ring_type, Point>::apply(
-                        range::at(interior_rings(polygon), ring_index), point);
+                        range::at(int_rings, ring_index), point);
         }
     }
 };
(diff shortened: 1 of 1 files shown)


awulkiew:
Boost.Geometry supports C++03 so your proposal has to do it as well.

include/boost/geometry/algorithms/append.hpp
```diff
@@ -91,21 +91,22 @@ template <typename Polygon, typename Point>
...
     static inline void apply(Polygon& polygon, Point const& point,
                 int ring_index, int = 0)
     {
         if (ring_index == -1)
         {
-			auto&& ext_ring = exterior_ring(polygon);
+			ring_type& ext_ring = exterior_ring(polygon);
             append_point<ring_type, Point>::apply(
-                        ext_ring, point);
+				ext_ring, point);
```
mloskot: @xventura81 Has your editor changed spaces vs tabs? Not sure what happened here.

xventura81: Yeah, Visual Studio does it by default.

mloskot: IMHO, that is not good to let accidental style changes happen. Every editor can be controlled to prevent it from replacing spaces with tabs or auto-reformatting. Imagine N authors working on common code base use N editors and each editor has different auto-formatting style as default... :-)

xventura81: You are right. Let's revert those changes in the indentation.

include/boost/geometry/algorithms/append.hpp
```diff
@@ -91,19 +91,22 @@ template <typename Polygon, typename Point>
...
 {
     typedef typename ring_type<Polygon>::type ring_type;
+	typedef typename interior_return_type<Polygon>::type interior_ring_range_type;
 
     static inline void apply(Polygon& polygon, Point const& point,
                 int ring_index, int = 0)
     {
         if (ring_index == -1)
         {
+			ring_type& ext_ring = exterior_ring(polygon);
```
awulkiew: `exterior_ring()` may return not-real reference so you should use `ring_return_type<Polygon>::type` here, similar to `interior_return_type<Polygon>::type` you used. Otherwise you may end up with dangling reference,

include/boost/geometry/algorithms/append.hpp
```diff
@@ -113,19 +116,22 @@ template <typename Polygon, typename Range>
...
 {
     typedef typename ring_type<Polygon>::type ring_type;
+	typedef typename interior_return_type<Polygon>::type interior_ring_range_type;
 
     static inline void apply(Polygon& polygon, Range const& range,
                 int ring_index, int = 0)
     {
         if (ring_index == -1)
         {
+			ring_type& ext_ring = exterior_ring(polygon);
```
awulkiew: `ring_return_type<Polygon>::type`

awulkiew:
I made some comments but they are outdated now. Please check them out. I won't create new ones to avoid spamming.

awulkiew:
Thanks! Now I'd only suggest to squash all of the commits into one. What do you think?

Something like:

    git reset --soft HEAD~4
    git commit
    git push -f

xventura81:
> Thanks! Now I'd only suggest to squash all of the commits into one. What do you think?
> 
> Something like:
> 
> ```
> git reset --soft HEAD~4
> git commit
> git push -f
> ```

Done! Thank for all the observations.

awulkiew:
Great, thanks!

PR #522, opened 2018-10-24
Support python3 in building documentation

vissarion:
Python 3 removed the global `execfile` function (see https://docs.python.org/3.5/whatsnew/3.0.html?highlight=execfile#builtins) and thus `make_qbk.py` script produces the error `name 'execfile' is not defined` (even with `Python 2.7.6` in my case).

This PR proposes a fix (tested with `Python 2.7.6` and `Python 3.4.3`).

### DIFF
--- a/doc/make_qbk.py
+++ b/doc/make_qbk.py
@@ -7,6 +7,7 @@
 #  Copyright (c) 2017 Adam Wulkiewicz, Lodz, Poland
 #
 #  Copyright (c) 2018, Oracle and/or its affiliates.
+#  Contributed and/or modified by Vissarion Fysikopoulos, on behalf of Oracle
 #  Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
 #  Use, modification and distribution is subject to the Boost Software License,
 #  Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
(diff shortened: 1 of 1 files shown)


doc/make_qbk.py
```diff
@@ -198,7 +199,7 @@ def cs_to_quickbook(section):
 class_to_quickbook2("de9im_1_1static__mask", "de9im_static_mask")
 
 os.chdir("index")
-execfile("make_qbk.py")
+exec(compile(open("make_qbk.py", "rb").read(), "make_qbk.py", 'exec'))
```
mloskot: Wouldn't it be sufficient to call `exec(open("make_qbk.py").read())`?

As per https://docs.python.org/3.3/whatsnew/3.0.html#builtins
> Removed execfile(). Instead of execfile(fn) use exec(open(fn).read()).

Generally, catching up with Python 3 is a good idea, I think.

vissarion: I used the call used in `2to3` script; see https://github.com/python/cpython/blob/d4d60134b29290049e28df54f23493de4f1824b6/Lib/lib2to3/fixes/fix_execfile.py 
As far as I understand, that call has at least better error handling.

mloskot: Better errors handling is not entirely clear to me. Perhaps due to the fact `exec` and `compile` make two step action.
I've never bothered to use 2to3 really. Unless Python 2 support is desired, I'd stick to plain Python 3 API. Anyhow, I just wondered.

Issue #523, opened 2018-10-27
Some headers don't compile independently


tonyelewis:
Following on from [Trac#12393](https://svn.boost.org/trac10/ticket/12393)&hellip; some Geometry headers don't compile independently (under Clang) in violation of the [Boost header policy](https://www.boost.org/development/header.html):

> Make sure that a translation unit consisting of just the contents of the header file will compile successfully.

One example is `extensions/algebra/algorithms/reverse.hpp`:

~~~no-highlight
boost/geometry/extensions/algebra/algorithms/reverse.hpp:33:24: error: unknown type name 'vector_tag'; did you mean 'mpl::aux::vector_tag'?
struct reverse<Vector, vector_tag>
                       ^~~~~~~~~~
                       mpl::aux::vector_tag
/opt/include/boost/mpl/vector/aux_/tag.hpp:25:8: note: 'mpl::aux::vector_tag' declared here
struct vector_tag;
       ^
boost/geometry/extensions/algebra/algorithms/reverse.hpp:42:19: error: use of undeclared identifier 'rotation_quaternion_tag'
struct reverse<R, rotation_quaternion_tag>
                  ^
boost/geometry/extensions/algebra/algorithms/reverse.hpp:51:19: error: use of undeclared identifier 'rotation_matrix_tag'
struct reverse<R, rotation_matrix_tag>
                  ^
3 errors generated.
~~~

Thanks.

mloskot:
@tonyelewis Thanks for the report. Apparently, more headers need attention than just this one, see #525

tonyelewis:
NP. Thanks for looking into this so quickly.

PR #525, opened 2018-10-28
Add compile-only tests verifying headers are self-contained

mloskot:
Add Boost.Jam rule generating test target for each header found in specified location.
List test suites (target aliases) for groups of headers.
Build self_contained_headers test only if `CI` environment variable is NOT defined or on-demand, if `TEST_HEADERS` environment variable is defined, or if `b2 test/self_contained_headers` is requested from command line.

Related to issue #523

-----

First, this PR does not attempt to fix any headers, not yet.

I've run all the headers tests on 24 CPUs (48 logical cores) as follows:

```shell
cd libs/geometry
b2 -j 16 test/self_contained_headers >build.log 2>&1
```

AFAICT, it compile-tests *~900* headers in less than *4 minutes*:

```shell
cat build.log | grep -E -o "c\+\+.+\~hpp\.test" | sort | uniq | wc -l
898
cat build.log | grep -E -o "passed.+~hpp\.test" | wc -l
547
cat build.log | grep -E -o "failed.+~hpp\.test" | wc -l
350
```

And, I have only included several extensions, see `alias`-es  and comments in `test/self_contained_headers/Jamfile`.

-----

To run the tests on CI services, it should be enough to define environment variable `TEST_HEADERS=1`  for one just build job. For example, Boost.GIL defines it for first job only https://travis-ci.org/boostorg/gil/jobs/447353675

### DIFF
--- a/test/Jamfile.v2
+++ b/test/Jamfile.v2
@@ -2,7 +2,7 @@
 #
 # Copyright (c) 2007-2012 Barend Gehrels, Amsterdam, the Netherlands.
 # Copyright (c) 2008-2012 Bruno Lalande, Paris, France.
-# Copyright (c) 2009-2012 Mateusz Loskot, London, UK.
+# Copyright (c) 2009-2018 Mateusz Loskot, London, UK.
 #
 # This file was modified by Oracle on 2017, 2018.
 # Modifications copyright (c) 2017-2018 Oracle and/or its affiliates.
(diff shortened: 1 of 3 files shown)


vissarion:
Thanks, this is useful.

mloskot:
Thanks!

awulkiew:
Wow, thanks! Just to be sure, does it work with CircleCI?

mloskot:
I see no reason why it should not work on CircleCI - I mean, those tests should be automatically filtered there as on Travis or AppVeyor.

CircleCI does define `CI` as [built-in environment variable](https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables) and `b2` just reads the environment variables (no `CI` needs to be defined in any `Jamfile`).

awulkiew:
Ok, thanks for the explanation.

awulkiew:
@mloskot This change polluted our regression matrix (https://www.boost.org/development/tests/develop/developer/geometry.html). I'm considering disabling it by default. Or maybe do you have some other ideas?

mloskot:
@awulkiew I'm trying to make those tests on-demand only in #529

I'm not sure how to avoid the long paths. Since the `.cpp` file is generated, I assumed that is not an issue, so ignored it.

awulkiew:
If they are not run by default it is no longer an issue since we can just run them on linux. Thanks!

PR #531, opened 2018-11-28
Feature/get turn info add retrieve policy

barendgehrels:
This PR (actually my first) adds a retrieve policy, and passes it to get_turn_info.

This policy retrieves a point known as "k" (pk / qk), the first non-duplicate point coming consecutively after the points "i" and "j" (two points from two segments p and q, part of the intersection of two segments, so these points are pi,pj and qi,qj).

In the past points pk/qk were just passed. But now we need another point (pl/ql) and it is not convenient to add that. Therefore the policy is introduced.

An additional advantage is that point "k" is now retrieved lazily, on demand. Often that point is not necessary. It is actually only necessary if there are collinearies. And often that is not the case. So in all those cases, performance should increase just a little bit.

This PR adds the policy and retrieves point "k" using the method .get(). In another PR I will add an argument to get, to be able to retrieve also the subsequente point, as described above. But for now this PR is large enough.

It is also related to side calculation, the place where that point is used, to check whether it is located left/right or collinear to a segment.

To my surprise, also non-typical usage of side calculation were introduced in the implementations for linear intersections, and the typical order is changed there. Therefore I copied the originals there - they cannot use the policy because they don't use the "k" point. Actually there should be looked at, because it is looking suspicious, but we can do that afterwards.

I'm curious to your findings.

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
@@ -253,11 +253,13 @@ class piece_turn_visitor
 #endif
                     > turn_policy;
 
+                overlay::retrieve_null_policy retrieve_policy;
+
                 turn_policy::apply(*prev1, *it1, *next1,
                                     *prev2, *it2, *next2,
-                                    false, false, false, false,
                                     the_model,
                                     m_intersection_strategy,
+                                    retrieve_policy, retrieve_policy,
                                     m_robust_policy,
                                     std::back_inserter(m_turns));
             }
--- a/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
@@ -170,6 +170,7 @@ class piece_turn_visitor
         typedef typename boost::range_value<Rings const>::type ring_type;
         typedef typename boost::range_value<Turns const>::type turn_type;
         typedef typename boost::range_iterator<ring_type const>::type iterator;
+        typedef typename point_type<ring_type const>::type point_type;
 
         signed_size_type const piece1_first_index = piece1.first_seg_id.segment_index;
         signed_size_type const piece2_first_index = piece2.first_seg_id.segment_index;
@@ -230,7 +231,9 @@ class piece_turn_visitor
             the_model.operations[1].seg_id = piece2.first_seg_id;
             the_model.operations[1].seg_id.segment_index = index2; // override
 
+            // TODO: next_point can be moved inside a specified retrieve policy for piece points
             iterator next1 = next_point(ring1, it1);
+            overlay::retrieve_null_policy<point_type> retrieve_policy1(*next1);
 
             iterator it2 = it2_first;
             for (iterator prev2 = it2++;
@@ -238,6 +241,7 @@ class piece_turn_visitor
                     prev2 = it2++, the_model.operations[1].seg_id.segment_index++)
             {
                 iterator next2 = next_point(ring2, it2);
+                overlay::retrieve_null_policy<point_type> retrieve_policy2(*next2);
 
                 // TODO: internally get_turn_info calculates robust points.
                 // But they are already calculated.
@@ -253,13 +257,12 @@ class piece_turn_visitor
 #endif
                     > turn_policy;
 
-                overlay::retrieve_null_policy retrieve_policy;
 
-                turn_policy::apply(*prev1, *it1, *next1,
-                                    *prev2, *it2, *next2,
+                turn_policy::apply(*prev1, *it1,
+                                    *prev2, *it2,
                                     the_model,
                                     m_intersection_strategy,
-                                    retrieve_policy, retrieve_policy,
+                                    retrieve_policy1, retrieve_policy2,
                                     m_robust_policy,
                                     std::back_inserter(m_turns));
             }
--- a/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
@@ -52,7 +52,8 @@ struct unique_sub_range_from_piece
         , m_point_retrieved(false)
     {}
 
-    static inline bool is_first() { return false; }
+    static inline bool is_first_segment() { return false; }
+    static inline bool is_last_segment() { return false; }
 
     static inline std::size_t size() { return 3u; }
 
--- a/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
@@ -17,6 +17,7 @@
 #include <boost/core/ignore_unused.hpp>
 #include <boost/range.hpp>
 
+#include <boost/geometry/core/assert.hpp>
 #include <boost/geometry/algorithms/equals.hpp>
 #include <boost/geometry/algorithms/expand.hpp>
 #include <boost/geometry/algorithms/detail/disjoint/box_box.hpp>
@@ -57,6 +58,7 @@ struct unique_sub_range_from_piece
 
     inline point_type const& at(std::size_t index) const
     {
+        BOOST_GEOMETRY_ASSERT(index < size());
         switch (index)
         {
             case 0 : return *m_iterator_at_i;
--- a/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
@@ -34,17 +34,20 @@ namespace boost { namespace geometry
 namespace detail { namespace buffer
 {
 
+// Implements a unique_sub_range for a buffered piece,
+// the range can return subsequent points
+// known as "i", "j" and "k" (and further),  indexed as 0,1,2,3
 template <typename Ring>
 struct unique_sub_range_from_piece
 {
     typedef typename boost::range_iterator<Ring const>::type iterator_type;
     typedef typename geometry::point_type<Ring const>::type point_type;
 
-    unique_sub_range_from_piece(Ring const& ring, iterator_type it0, iterator_type it1)
+    unique_sub_range_from_piece(Ring const& ring,
+                                iterator_type iterator_at_i, iterator_type iterator_at_j)
         : m_ring(ring)
-        , m_point_i(*it0)
-        , m_point_j(*it1)
-        , m_iterator(it1)
+        , m_iterator_at_i(iterator_at_i)
+        , m_iterator_at_j(iterator_at_j)
         , m_point_retrieved(false)
     {}
 
@@ -56,28 +59,26 @@ struct unique_sub_range_from_piece
     {
         switch (index)
         {
-            case 0 : return m_point_i;
-            case 1 : return m_point_j;
-            case 2 : return get_next_point();
-            default : return m_point_i;
+            case 0 : return *m_iterator_at_i;
+            case 1 : return *m_iterator_at_j;
+            case 2 : return get_point_k();
+            default : return *m_iterator_at_i;
         }
     }
 
 private :
 
-    inline point_type const& get_next_point() const
+    inline point_type const& get_point_k() const
     {
         if (! m_point_retrieved)
         {
-            m_point = *next_point(m_iterator);
+            m_iterator_at_k = advance_one(m_iterator_at_j);
             m_point_retrieved = true;
         }
-        return m_point;
+        return *m_iterator_at_k;
     }
 
-private :
-
-    inline void circular_move_along_ring(iterator_type& next) const
+    inline void circular_advance_one(iterator_type& next) const
     {
         ++next;
         if (next == boost::end(m_ring))
@@ -86,26 +87,24 @@ private :
         }
     }
 
-    inline iterator_type next_point(iterator_type it) const
+    inline iterator_type advance_one(iterator_type it) const
     {
         iterator_type result = it;
-        circular_move_along_ring(result);
+        circular_advance_one(result);
 
         // TODO: we could also use piece-boundaries
         // to check if the point equals the last one
         while (geometry::equals(*it, *result))
         {
-            circular_move_along_ring(result);
+            circular_advance_one(result);
         }
         return result;
     }
 
     Ring const& m_ring;
-    point_type m_point_i;
-    point_type m_point_j;
-    iterator_type m_iterator;
-
-    mutable point_type m_point;
+    iterator_type m_iterator_at_i;
+    iterator_type m_iterator_at_j;
+    mutable iterator_type m_iterator_at_k;
     mutable bool m_point_retrieved;
 };
 
@@ -280,9 +279,9 @@ class piece_turn_visitor
                         detail::overlay::assign_null_policy
                     > turn_policy;
 
-                turn_policy::apply(the_model,
+                turn_policy::apply(unique_sub_range1, unique_sub_range2,
+                                   the_model,
                                    m_intersection_strategy,
-                                   unique_sub_range1, unique_sub_range2,
                                    m_robust_policy,
                                    std::back_inserter(m_turns));
             }
--- a/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
@@ -34,35 +34,6 @@ namespace boost { namespace geometry
 namespace detail { namespace buffer
 {
 
-
-#if defined(BOOST_GEOMETRY_BUFFER_USE_SIDE_OF_INTERSECTION)
-struct buffer_assign_turn
-{
-    static bool const include_no_turn = false;
-    static bool const include_degenerate = false;
-    static bool const include_opposite = false;
-
-    template
-    <
-        typename Info,
-        typename Point1,
-        typename Point2,
-        typename IntersectionInfo
-    >
-    static inline void apply(Info& info,
-                             Point1 const& /*p1*/,
-                             Point2 const& /*p2*/,
-                             IntersectionInfo const& iinfo)
-    {
-        info.rob_pi = iinfo.rpi();
-        info.rob_pj = iinfo.rpj();
-        info.rob_qi = iinfo.rqi();
-        info.rob_qj = iinfo.rqj();
-    }
-
-};
-#endif
-
 template <typename Ring>
 struct unique_sub_range_from_piece
 {
@@ -281,6 +252,10 @@ class piece_turn_visitor
             the_model.operations[1].piece_index = piece2.index;
             the_model.operations[1].seg_id = piece2.first_seg_id;
             the_model.operations[1].seg_id.segment_index = index2; // override
+#if defined(BOOST_GEOMETRY_BUFFER_USE_SIDE_OF_INTERSECTION)
+            geometry::recalculate(the_model.rob_pi, *prev1, m_robust_policy);
+            geometry::recalculate(the_model.rob_pj, *it1, m_robust_policy);
+#endif
 
             unique_sub_range_from_piece<ring_type> unique_sub_range1(ring1, prev1, it1);
 
@@ -290,6 +265,10 @@ class piece_turn_visitor
                     prev2 = it2++, the_model.operations[1].seg_id.segment_index++)
             {
                 unique_sub_range_from_piece<ring_type> unique_sub_range2(ring2, prev2, it2);
+#if defined(BOOST_GEOMETRY_BUFFER_USE_SIDE_OF_INTERSECTION)
+                geometry::recalculate(the_model.rob_qi, *prev2, m_robust_policy);
+                geometry::recalculate(the_model.rob_qj, *it2, m_robust_policy);
+#endif
 
                 // TODO: internally get_turn_info calculates robust points.
                 // But they are already calculated.
@@ -298,13 +277,8 @@ class piece_turn_visitor
                 // and iterating in sync with them...
                 typedef detail::overlay::get_turn_info
                     <
-#if defined(BOOST_GEOMETRY_BUFFER_USE_SIDE_OF_INTERSECTION)
-                        buffer_assign_turn
-#else
                         detail::overlay::assign_null_policy
-#endif
                     > turn_policy;
-
 
                 turn_policy::apply(the_model,
                                    m_intersection_strategy,
--- a/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
@@ -63,6 +63,64 @@ struct buffer_assign_turn
 };
 #endif
 
+template <typename Ring>
+struct retrieve_from_piece
+{
+    typedef typename boost::range_iterator<Ring const>::type iterator_type;
+    typedef typename geometry::point_type<Ring const>::type point_type;
+
+    retrieve_from_piece(Ring const& ring, iterator_type start_iterator)
+        : m_ring(ring)
+        , m_start_iterator(start_iterator)
+        , m_point_retrieved(false)
+    {}
+
+    static inline bool is_first() { return false; }
+
+    static inline bool is_last() { return false; }
+
+    inline point_type const& get() const
+    {
+        if (! m_point_retrieved)
+        {
+            m_point = *next_point(m_start_iterator);
+            m_point_retrieved = true;
+        }
+        return m_point;
+    }
+
+private :
+
+    inline void circular_move_along_ring(iterator_type& next) const
+    {
+        ++next;
+        if (next == boost::end(m_ring))
+        {
+            next = boost::begin(m_ring) + 1;
+        }
+    }
+
+    inline iterator_type next_point(iterator_type it) const
+    {
+        iterator_type result = it;
+        circular_move_along_ring(result);
+
+        // TODO: we could also use piece-boundaries
+        // to check if the point equals the last one
+        while (geometry::equals(*it, *result))
+        {
+            circular_move_along_ring(result);
+        }
+        return result;
+    }
+
+    Ring const& m_ring;
+    iterator_type m_start_iterator;
+
+    mutable point_type m_point;
+    mutable bool m_point_retrieved;
+};
+
 template
 <
     typename Pieces,
@@ -102,29 +160,6 @@ class piece_turn_visitor
         return ! m_rings[piece1.first_seg_id.multi_index].has_concave;
     }
 
-    template <typename Range, typename Iterator>
-    inline void move_to_next_point(Range const& range, Iterator& next) const
-    {
-        ++next;
-        if (next == boost::end(range))
-        {
-            next = boost::begin(range) + 1;
-        }
-    }
-
-    template <typename Range, typename Iterator>
-    inline Iterator next_point(Range const& range, Iterator it) const
-    {
-        Iterator result = it;
-        move_to_next_point(range, result);
-        // TODO: we could use either piece-boundaries, or comparison with
-        // robust points, to check if the point equals the last one
-        while(geometry::equals(*it, *result))
-        {
-            move_to_next_point(range, result);
-        }
-        return result;
-    }
 
     template <std::size_t Dimension, typename Iterator, typename Box>
     inline void move_begin_iterator(Iterator& it_begin, Iterator it_beyond,
@@ -231,17 +266,14 @@ class piece_turn_visitor
             the_model.operations[1].seg_id = piece2.first_seg_id;
             the_model.operations[1].seg_id.segment_index = index2; // override
 
-            // TODO: next_point can be moved inside a specified retrieve policy for piece points
-            iterator next1 = next_point(ring1, it1);
-            overlay::retrieve_null_policy<point_type> retrieve_policy1(*next1);
+            retrieve_from_piece<ring_type> retrieve_policy1(ring1, it1);
 
             iterator it2 = it2_first;
             for (iterator prev2 = it2++;
                     it2 != it2_beyond;
                     prev2 = it2++, the_model.operations[1].seg_id.segment_index++)
             {
-                iterator next2 = next_point(ring2, it2);
-                overlay::retrieve_null_policy<point_type> retrieve_policy2(*next2);
+                retrieve_from_piece<ring_type> retrieve_policy2(ring2, it2);
 
                 // TODO: internally get_turn_info calculates robust points.
                 // But they are already calculated.
--- a/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
@@ -64,26 +64,41 @@ struct buffer_assign_turn
 #endif
 
 template <typename Ring>
-struct retrieve_from_piece
+struct unique_sub_range_from_piece
 {
     typedef typename boost::range_iterator<Ring const>::type iterator_type;
     typedef typename geometry::point_type<Ring const>::type point_type;
 
-    retrieve_from_piece(Ring const& ring, iterator_type start_iterator)
+    unique_sub_range_from_piece(Ring const& ring, iterator_type it0, iterator_type it1)
         : m_ring(ring)
-        , m_start_iterator(start_iterator)
+        , m_point_i(*it0)
+        , m_point_j(*it1)
+        , m_iterator(it1)
         , m_point_retrieved(false)
     {}
 
     static inline bool is_first() { return false; }
 
-    static inline bool is_last() { return false; }
+    static inline std::size_t size() { return 3u; }
 
-    inline point_type const& get() const
+    inline point_type const& at(std::size_t index) const
+    {
+        switch (index)
+        {
+            case 0 : return m_point_i;
+            case 1 : return m_point_j;
+            case 2 : return get_next_point();
+            default : return m_point_i;
+        }
+    }
+
+private :
+
+    inline point_type const& get_next_point() const
     {
         if (! m_point_retrieved)
         {
-            m_point = *next_point(m_start_iterator);
+            m_point = *next_point(m_iterator);
             m_point_retrieved = true;
         }
         return m_point;
@@ -115,7 +130,9 @@ private :
     }
 
     Ring const& m_ring;
-    iterator_type m_start_iterator;
+    point_type m_point_i;
+    point_type m_point_j;
+    iterator_type m_iterator;
 
     mutable point_type m_point;
     mutable bool m_point_retrieved;
@@ -265,14 +282,14 @@ class piece_turn_visitor
             the_model.operations[1].seg_id = piece2.first_seg_id;
             the_model.operations[1].seg_id.segment_index = index2; // override
 
-            retrieve_from_piece<ring_type> retrieve_policy1(ring1, it1);
+            unique_sub_range_from_piece<ring_type> unique_sub_range1(ring1, prev1, it1);
 
             iterator it2 = it2_first;
             for (iterator prev2 = it2++;
                     it2 != it2_beyond;
                     prev2 = it2++, the_model.operations[1].seg_id.segment_index++)
             {
-                retrieve_from_piece<ring_type> retrieve_policy2(ring2, it2);
+                unique_sub_range_from_piece<ring_type> unique_sub_range2(ring2, prev2, it2);
 
                 // TODO: internally get_turn_info calculates robust points.
                 // But they are already calculated.
@@ -289,13 +306,11 @@ class piece_turn_visitor
                     > turn_policy;
 
 
-                turn_policy::apply(*prev1, *it1,
-                                    *prev2, *it2,
-                                    the_model,
-                                    m_intersection_strategy,
-                                    retrieve_policy1, retrieve_policy2,
-                                    m_robust_policy,
-                                    std::back_inserter(m_turns));
+                turn_policy::apply(the_model,
+                                   m_intersection_strategy,
+                                   unique_sub_range1, unique_sub_range2,
+                                   m_robust_policy,
+                                   std::back_inserter(m_turns));
             }
         }
     }
--- a/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
@@ -205,7 +205,6 @@ class piece_turn_visitor
         typedef typename boost::range_value<Rings const>::type ring_type;
         typedef typename boost::range_value<Turns const>::type turn_type;
         typedef typename boost::range_iterator<ring_type const>::type iterator;
-        typedef typename point_type<ring_type const>::type point_type;
 
         signed_size_type const piece1_first_index = piece1.first_seg_id.segment_index;
         signed_size_type const piece2_first_index = piece2.first_seg_id.segment_index;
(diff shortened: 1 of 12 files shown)


awulkiew:
Interesting. Regarding future plans, what for will be additional points `pl` and `ql` needed?

The PR looks good to me. I'm only wondering why you've made certain naming decisions:
- The name of the policy. Would it be possible to rename it to know what is retrieved exactly? What is it for on a higher level of abstraction? What is stored and why? AFAIU it's for optional, additional data of an intersection segment: next point and flags for the current segment (is first or last segment). Or are the flags for first/last point? Is that correct? `OptionalPointDataPolicy`, `AdditionalSidePointPolicy`, `OptionalSegmentInfoPolicy`etc.?
- The name of function `get()` may be confusing because it's not clear what is it returning exactly. Furthermore we use `get()` for coordinates. Why not `get_point()` or even explicitly `get_point_k()` (and later `get_point_q()`)?

barendgehrels:
Ah, sorry I meant 1.70
No need now
Thanks

barendgehrels:
Will react tonight on your suggestions

barendgehrels:
Thanks for your comments.

I can make it get_point_k, that is clear (as long as people understand the meaning of "k")...
W.r.t. the name: it retrieves extra information about the segments being intersected. For both segments there is an independent policy (might well be different type, for example for box/poly).

Basic information, per segment, is 
- point i
- point j
They form together the segment. This is not retrieved by the policy (so indeed the policy gets additional information).

The extra information is:
- is_first() -> returns true if i-j is the first segment (only for linear features)
- is_last() -> returns true if i-j is the last segment (only for linear features)
In the last case, there is no point "k"! So maybe (did not think of that earlier) get_point_k should return a pointer (might be nullptr) or an optional (we're not yet dependent on boost optional). Or just as it is now, but its contents should be considered as undefined if is_last == true
- get_point_k() -> returns the first spatially unequal (at least, that should - for some operations that is not yet done) point after j
- get_point_l() -> returns the second spatially unequal point -> and also here it might be (for linear features) we need information about if it does exist, or not.

Maybe it is the best to return it as a pointer? Or return a boolean (and copy the point)? Or an extra method (has_k has_l)? All possible.

Will I rename the policy to retrieve_additional_info_policy ? (you did not suggest this one yet, but I think it is clear). Or optional_info_policy? It is about intersecting segments, that is known by context, and it is not always about side so that does not have to be part of the name I think.

barendgehrels:
Actually - get_point_k() is quite specific, isn't it?
I now consider Point* get_point(int index), where index=2 (for point k) and index=3(for point l). This means index 0 and 1 (for i and j), but they are not used. I think that will be most intuitive. What do you think?

awulkiew:
The names you propose are better then the original for sure but I'm wondering if we could find something even more descriptive.

This is additional info for segment (2 points), or 3-4 non-equal points in a geometry. So a relevant section/fragment/part or a geometry. Thought this is not a representation of the data stored in a geometry because if some of the points are equal some of them may be skipped. So this is actually a cherry-picked/newly-formed fragment of a line. Do you have an idea how something like this could be named?

I like the explicit `get_point_k()` better I think.

How about internally storing pointers to points and return references? This way you wouldn't be forced to store a separate `bool` state, this type would behave like a reference type and `mutable` wouldn't be needed.

Have you considered passing only the policies and storing also points `i` and `j` inside? This way the policy wouldn't be "optional" anymore. It'd become something like `FragmentPolicy`, or `ReferringLinePolicy`.

barendgehrels:
Internally (in get_turns.hpp) we call them nonduplicate points. So then it would be "get_non_duplicate_point_k" - which is quite long... For me it can be omitted there.

I will think about storing a pointer internally - it is then basically up to the policy itself. For example a box is always circular, will never return true there, so there it is not necessary.

Renaming is_last to "has_k" and reversing the meaning is probably convenient. Because then we can add "has_l" for "get_point_l".

I indeed have thought about passing policies with i/j inside. However, that is never necessary, it is really about intersecting a segment, and a segment consists of i/j which are never optional. Therefore I did not do that, I don't want to make too much changes for the sake of the changes themselves now...

awulkiew:
@barendgehrels Of course I agree. ;) Ok, so the policy would represent a segment of potentially newly-created points (intersection point can be one of them) and possibly additional info like flags, next non-equal point in the new geometry and yet another point.

So I have another bunch of possible names describing that this is a new fragment of geometry created by intersecting some geometries: `SubRangePolicy`, `NewSubRingPolicy`, `NewRingPartPolicy`, `IntersectionCurvePolicy`, `IntersectionFragmentPolicy`. Btw, when I'm proposing these names I don't expect you to use them exactly. It's partly a way of expressing abstractions I'm thinking about.

barendgehrels:
(Unique)SubRangePolicy sounds good to me.
Actually I don't understand your remark about "created by intersecting" - it is getting unique points from an existing geometry, and passes it.
They are later intersected indeed, but that is not related at all to this retrieval - do I miss something?

barendgehrels:
Right, that explains indeed some other things I did not understand ;-)
OK, indeed it represents points from only one geometry - so I think referring_linestring or referring_ring (because it is circular) are good descriptions - maybe referring_range because that is more neutral.
I can continue with that, thanks for your input!

vissarion:
This PR looks interesting to me too! 

Regarding naming I had to read the whole thread to understand exactly what is going on. https://github.com/boostorg/geometry/pull/531#issuecomment-442996815 is very descriptive, thanks! 

However, the use of indices `i,j,k,l` to name four successive (modulo duplicates) points took me a while to digest it. Apart from being successive letters in latin alphabet there is no other clue why those points should be successive. Maybe naming them `a,b,c,d,` or `one,two,three,four` or even indexing by `1..4` would be more clear. But this is a matter of taste. 

+1 for the rest.

barendgehrels:
Hi @awulkiew  and @vissarion , it is now revised based on your remarks, and quite nice I think.
So retrieve_policy is now unique_sub_range and has a subset of range concepts (only "at" and "size" replacing get_point_X and has_p). We pass a lot less now...
Curious to your findings.

barendgehrels:
Some more changes, especially removed apply from AssignPolicy because it was not used anymore (buffer did use it until yesterday but that is now solved differently - and will go with rescaling).
I believe it is ready now, running all tests again.

barendgehrels:
All tests pass - if you agree then it can be merged

awulkiew:
Great, I'll review it today. Btw, did you check what's the difference in performance, if any?

barendgehrels:
I'll do a quick check on performance. If any it should be a bit faster because the points known as "k" are no retrieved lazily, so that should save some time. On the other hand some points on buffer are no rescaled non-lazily, but that occurs less (I would think).

barendgehrels:
Performance is increased a bit:

**Union/intersection:**
$ ./recursive_polygons --seed 1234 --level 7 --size 20 --count 100
old: polygons: 25500 type: d time: 6.91922
new: polygons: 25500 type: d time: 6.54316

**Buffer:**
$ ./recursive_polygons_buffer --seed 1234 --level 7 --count 100
old: geometries: 25500 type: d time: 9.22374
new: geometries: 25500 type: d time: 9.13735

include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
```diff
@@ -34,34 +34,80 @@ namespace boost { namespace geometry
...
+    static inline std::size_t size() { return 3u; }
+
+    inline point_type const& at(std::size_t index) const
+    {
+        switch (index)
+        {
+            case 0 : return m_point_i;
+            case 1 : return m_point_j;
+            case 2 : return get_next_point();
+            default : return m_point_i;
```
awulkiew: Why not have an assertion here? Or is it a valid case?

include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
```diff
@@ -34,34 +34,80 @@ namespace boost { namespace geometry
...
+        while (geometry::equals(*it, *result))
+        {
+            circular_move_along_ring(result);
+        }
+        return result;
+    }
+
+    Ring const& m_ring;
+    point_type m_point_i;
+    point_type m_point_j;
```
awulkiew: I'm wondering, why not have iterators here as well to avoid copying? Ar are the points copied because you fear cache misses?

include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
```diff
@@ -34,34 +34,80 @@ namespace boost { namespace geometry
...
+    iterator_type m_iterator;
+
+    mutable point_type m_point;
```
awulkiew: Right, the same here. Is there a reason why simply the iterator isn't set to point to the relevant point in a range and left there instead of copying the point here?

include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
```diff
@@ -246,20 +277,14 @@ class piece_turn_visitor
...
                     > turn_policy;
 
-                turn_policy::apply(*prev1, *it1, *next1,
-                                    *prev2, *it2, *next2,
-                                    false, false, false, false,
-                                    the_model,
-                                    m_intersection_strategy,
-                                    m_robust_policy,
-                                    std::back_inserter(m_turns));
+                turn_policy::apply(the_model,
```
awulkiew: Is there a reason why the order of arguments is changed? Previously the points were passed as the first arguments and after that strategies etc. Not that this is a problem, just pointing it out.

include/boost/geometry/algorithms/detail/overlay/get_intersection_points.hpp
```diff
@@ -44,14 +44,12 @@ template
...
+    template <typename Strategy, typename UniqueSubRange1, typename UniqueSubRange2, typename RobustPolicy, typename OutputIterator>
     static inline OutputIterator apply(
-                Point1 const& pi, Point1 const& pj, Point1 const& /*pk*/,
-                Point2 const& qi, Point2 const& qj, Point2 const& /*qk*/,
-                bool /*is_p_first*/, bool /*is_p_last*/,
-                bool /*is_q_first*/, bool /*is_q_last*/,
                 TurnInfo const& ,
                 Strategy const& strategy,
+                UniqueSubRange1 const& range_p,
+                UniqueSubRange2 const& range_q,
```
awulkiew: Same here. The input ranges are passed after the strategy.

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -525,12 +511,14 @@ struct equal_opposite : public base_turn_handler
...
-        typename Point2,
+        typename UniqueSubRange1,
+        typename UniqueSubRange2,
         typename OutputIterator,
         typename IntersectionInfo
     >
-    static inline void apply(Point1 const& pi, Point2 const& qi,
+    static inline void apply(
+                UniqueSubRange1 const& range_p,
+                UniqueSubRange2 const& range_q,
```
awulkiew: Here they are passed at the beginning of arguments list which is as it was before.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
```diff
@@ -279,37 +317,25 @@ struct get_turn_info_for_endpoint
...
-        BOOST_GEOMETRY_ASSERT(ip_info.is_qj == equals::equals_point_point(qj, inters_pt));
-#endif
-
         // TODO - calculate first/last only if needed
-        bool is_p_first_ip = is_p_first && ip_info.is_pi;
-        bool is_p_last_ip = is_p_last && ip_info.is_pj;
-        bool is_q_first_ip = is_q_first && ip_info.is_qi;
-        bool is_q_last_ip = is_q_last && ip_info.is_qj;
+        bool is_p_first_ip = range_p.is_first() && ip_info.is_pi;
+        bool is_p_last_ip = range_p.size() == 2u && ip_info.is_pj;
```
awulkiew: This condition looks non-intuitively. Is it based on an assumption that for last segment the point `k` is not retrieved? This way here, the caller has to know how the `range_p` and `range_q` below are filled possibly in some other place of the code. Would it be possible to have and explicit function `is_last()` (like `is_first()`) so here it'd be clear what's happening and all details are hidden in the `range_p`?

include/boost/geometry/algorithms/detail/overlay/get_turn_info_helpers.hpp
```diff
@@ -36,183 +37,309 @@ struct turn_operation_linear
...
+
+    //! Get precalculated point
+    Point const& at(std::size_t index) const
+    {
+        switch (index)
+        {
+            case 0 : return m_robust_point_i;
+            case 1 : return m_robust_point_j;
+            case 2 : return get_point_k();
+            default : return m_robust_point_i;
```
awulkiew: Why not an assertion?

include/boost/geometry/algorithms/detail/overlay/get_turn_info_helpers.hpp
```diff
@@ -36,183 +37,309 @@ struct turn_operation_linear
...
+    inline point1_type const& get_pj() const { return m_range_p.at(1); }
+    inline point1_type const& get_pk() const { return m_range_p.at(2); }
+
+    inline point2_type const& get_qi() const { return m_range_q.at(0); }
+    inline point2_type const& get_qj() const { return m_range_q.at(1); }
+    inline point2_type const& get_qk() const { return m_range_q.at(2); }
+
+    SideStrategy m_side_strategy; // NOTE: cannot be const&
+    UniqueSubRange1 const& m_range_p;
+    UniqueSubRange2 const& m_range_q;
```
awulkiew: So here `UniqueSubRange1` and `UniqueSubRange2` are `const&`, yet `SideStrategy ` is not. Are we sure that ranges will live as long as the side_calculator object? Are we sure that they will not be created from temporaries like:

    side_calculator sc(some_range1(), some_range2(), get_side_strategy());

because if they are created like that these references will be dangling.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_helpers.hpp
```diff
@@ -36,183 +37,309 @@ struct turn_operation_linear
...
+    RobustPolicy const& m_robust_policy;
 
-    Pi const& m_pi;
-    Pj const& m_pj;
-    Pk const& m_pk;
-    Qi const& m_qi;
-    Qj const& m_qj;
-    Qk const& m_qk;
+    Point const& m_robust_point_i;
+    Point const& m_robust_point_j;
```
awulkiew: So here robust points are stored as references, are we sure that whatever is passed to the constructor of `robust_subrange_adapter` will live longer than an object of this class?

include/boost/geometry/algorithms/detail/overlay/get_turn_info_helpers.hpp
```diff
@@ -278,34 +404,49 @@ class intersection_info
     // TODO: it's more like is_spike_ip_p
     inline bool is_spike_p() const
     {
+        if (base::m_range_p.size() == 2u)
```
awulkiew: Out of curiosity, in which case something like this can happen? Is this a check for invalid input or a check for last segment as it was before?

include/boost/geometry/algorithms/detail/overlay/get_turn_info_helpers.hpp
```diff
@@ -278,34 +404,49 @@ class intersection_info
...
             }
         }
         
         return false;
     }
 
-    // TODO: it's more like is_spike_ip_q
     inline bool is_spike_q() const
     {
+        if (base::m_range_q.size() == 2u)
```
awulkiew: Right, here it is too. Is the reasoning that if `base::m_range_q.size() == 2u` then there is no `base::rqk()`?

include/boost/geometry/algorithms/detail/overlay/get_turn_info_la.hpp
```diff
@@ -706,35 +663,40 @@ struct get_turn_info_linear_areal
...
+        }
 
-// TODO: is_q_last could probably be replaced by false and removed from parameters
-
-        linear_intersections intersections(pi, qi, inters.result(), is_p_last, is_q_last);
+        linear_intersections intersections(range_p.at(0),
+                                           range_q.at(0),
+                                           inters.result(),
+                                           range_p.size() == 2u,
+                                           range_q.size() == 2u);
```
awulkiew: This is non-intuitive. A person reading it 6 months from now will ask "why 2 exactly"? ;)

include/boost/geometry/algorithms/detail/overlay/get_turn_info_la.hpp
```diff
@@ -827,7 +798,7 @@ struct get_turn_info_linear_areal
 
         // IP on the last point of Linear Geometry
         if ( BOOST_GEOMETRY_CONDITION(EnableLast)
-          && is_p_last
+          && range_p.size() == 2u
```
awulkiew: Here.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_ll.hpp
```diff
@@ -443,28 +403,27 @@ struct get_turn_info_linear_linear
...
-                    if ( is_p_first
-                      && equals::equals_point_point(pi, tp.point) )
+                    if ( range_p.is_first()
+                      && equals::equals_point_point(range_p.at(0), tp.point) )
                     {
                         tp.operations[0].position = position_front;
                     }
-                    else if ( is_p_last
-                           && equals::equals_point_point(pj, tp.point) )
+                    else if ( range_p.size() == 2u
```
awulkiew: Here.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_ll.hpp
```diff
@@ -443,28 +403,27 @@ struct get_turn_info_linear_linear
...
-                    else if ( is_q_first
-                           && equals::equals_point_point(qi, tp.point) )
+                    else if ( range_q.is_first()
+                           && equals::equals_point_point(range_q.at(0), tp.point) )
                     {
                         tp.operations[1].position = position_front;
                     }
-                    else if ( is_q_last
-                           && equals::equals_point_point(qj, tp.point) )
+                    else if ( range_q.size() == 2u
```
awulkiew: Here.

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
```diff
@@ -101,6 +101,105 @@ struct no_interrupt_policy
...
+
+    inline bool is_first() const
+    {
+        return !IsAreal && m_section.is_non_duplicate_first && m_index == m_section.begin_index;
+    }
+
+    inline std::size_t size() const
+    {
+        return IsAreal ? 3
+            : m_section.is_non_duplicate_last && m_index + 1 >= m_section.end_index ? 2 : 3;
```
awulkiew: So this is the cause of the `2` condition above. Right? ;)

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
```diff
@@ -101,6 +101,105 @@ struct no_interrupt_policy
...
+    }
+
+    inline Point const& at(std::size_t index) const
+    {
+        switch (index)
+        {
+            case 0 : return m_previous_point;
+            case 1 : return m_current_point;
+            case 2 : return get_next_point();
+            default : return m_previous_point;
```
awulkiew: Possible assertion.

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
```diff
@@ -101,6 +101,105 @@ struct no_interrupt_policy
...
+        {
+            circular_iterator++;
+            geometry::recalculate(next_robust_point, *circular_iterator, m_robust_policy);
+        }
+    }
+
+    Section const& m_section;
+    signed_size_type m_index;
+    Point const& m_previous_point;
+    Point const& m_current_point;
```
awulkiew: And here points are stored as references. What is the reason behind the inconsistency?

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
```diff
@@ -519,8 +591,9 @@ template
 >
 struct get_turns_cs
 {
-    typedef typename geometry::point_type<Range>::type point_type;
+    typedef typename geometry::point_type<Range>::type range_point_type;
     typedef typename geometry::point_type<Box>::type box_point_type;
+    typedef boost::array<box_point_type, 4> box_array;
```
awulkiew: Boost.Array requires that the stored type has the default constrictor. Just pointing it out, there are probably many default-constructed points in the code. Still AFAIK this requirement is not a part of the Point concept.

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
```diff
@@ -556,22 +689,16 @@ struct get_turns_cs
             return;
         }
 
-        boost::array<box_point_type,4> bp;
-        assign_box_corners_oriented<ReverseBox>(box, bp);
+        box_array box_points;
```
awulkiew: Right, here Boost.Array was used before so this is not a change.

awulkiew:
Thanks! It's great that it's also faster.

In general I'm ok with merging. I see only 1 issue I'd suggest you changed, i.e. `is_last()` instead of `size() == 2`. Maybe also the order of arguments of some functions.

The other things are minor issues and questions. E.g. I don't know if the switch-cases without assertions are needed as they are or not. I'm ok with them and other things like keeping references or copies in various cases if you prefer to keep them as they are. But I'm interested in knowing if there is a key to when points are stored in one or the other way.

barendgehrels:
Thanks for your review! Will continue this weekend.

barendgehrels:
Thanks for your review! Will continue this weekend.

barendgehrels:
Thanks for your review! Will continue this weekend.

barendgehrels:
Oops, pressing button multiple times copies comment...

barendgehrels:
Hi @awulkiew , thanks (again) for your review.
About is_last(), I understand, it makes sense. The only objection I have is that it is a bit ambiguous. Because it does not say which is last (i-j ? j-k ? k-l?). 
I can make it "is_last_segment()", which is maybe a bit more clear especially if we document that unique_sub_range is all about one segment and one or two following points.

About the order: I agree, will change. Cause was actually that I first made it a policy, and only later renamed it but did not move it.
About the assertions: fine for me, will add them. And I will also look at the references.

barendgehrels:
And about is_spike_p / is_spike_q, I did not introduce them, but they all had a condition !is_p_last or !is_q_last before. That made sense because if p or q is last, then it can never be a spike and basically point k should not be accessed. Therefore I moved that condition inside.
So indeed, then there is no base::rqk()

barendgehrels:
Sorry @awulkiew , assertions are currently not possible without a major rewrite. It is because of the get_turn_info_la and get_turn_info_ll (which I did not write, neither of them). They call several strategies (e.g. collinear) of get_turn_info (for areal), which asks for points "k" all the time. And probably there are way more calls to pk, qk or their robust equivalents.
I think we have to keep that now. They did already access these points in previous implementation. Because these points were passed already (it is just the "next" point - and it is valid - but the line ends there, so effectively it is probably the first point). But there was no assertion. I currently cannot rewrite all of these...
But we should certainly revisit this.

barendgehrels:
Right, it is ready again, all tests pass (now in debug mode - to verify assertions)

include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
```diff
@@ -443,14 +413,14 @@ struct get_turn_info_for_endpoint
...
-                    side_calculator_for_endpoint<cs_tag,
-                                    RobustPoint1, RobustPoint2,
-                                    typename IntersectionInfo::side_strategy_type,
-                                    RobustPoint2>
-                        side_calc(ri2, ri1, rj1, ri2, rj2, rk2, inters.get_side_strategy()); // TODO: wrong, ri2/ri1/rj2
+                    // TODO: find out why side is used with respect to non-segments
+                    // (that is: range2.at(0) to range1.at(0), denoted as x)
+                    int const side_pj_q2 = sides.apply(range2.at(1), range2.at(2), range1.at(1));
+                    int const side_pj_x = sides.apply(range2.at(0), range1.at(0), range1.at(1));
+                    int const side_qk_x = sides.apply(range2.at(0), range1.at(0), range2.at(2));
```
awulkiew: It is probably me who wrote it. I don't remember exactly why turns are checked this way, I'd have to analyse the whole file and maybe debug. However I may point you in a right direction right now. This is probably for turns where linear geometries are involved, correct? With linear geometries the turns have slightly different meaning than in areal/areal case. For instance the equality doesn't mean that the whole segment was the same but that the intersection point was equal to one of the points of the geometry. Turns hold additional data indicating whether prior to the intersection point the segments were collinear or are crossing at the intersection point, etc. AFAIR this had something to do with the fact that linear geometries are not ordered like areal and that from various geometrically equal sets of segments, I needed the same set of turns, no matter the direction of the segments. In other words the linear turns are non-ordered like areal, turns. E.g. for areal/areal one relative direction could result in i/i operations but with reverse segments the operations would be x/x. And this is true for all other operations. But linear turns would be the same.

So AFAIR parts of this code are there to transform ordered areal turns into unordered linear turns. Sometimes additional info had to be calculated. I didn't check this for sure but this might have been such case.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
```diff
@@ -497,13 +467,12 @@ struct get_turn_info_for_endpoint
...
-                                    RobustPoint2>
-                        side_calc(ri2, rj1, ri1, ri2, rj2, rk2, inters.get_side_strategy()); // TODO: wrong, ri2/ri1/rj2
-                    
+                    int const side_pi_q2 = sides.apply(range2.at(1), range2.at(2), range1.at(0));
+                    int const side_pi_x = sides.apply(range2.at(0), range1.at(1), range1.at(0));
+                    int const side_qk_x = sides.apply(range2.at(0), range1.at(1), range2.at(2));
+
                     std::pair<operation_type, operation_type>
-                        operations = operations_of_equal(side_calc);
+                        operations = operations_of_equal(side_pi_q2, side_pi_x, side_qk_x);
```
awulkiew: So unless we know what's happening exactly we should probably be careful about changing this code. But if all of the tests pass with this change it should be ok I think.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
```diff
@@ -609,27 +578,23 @@ struct get_turn_info_for_endpoint
...
-        int const side_pk_q2 = side.pk_wrt_q2();
-        int const side_pk_p = side.pk_wrt_p1();
-        int const side_qk_p = side.qk_wrt_p1();
-
         // If pk is collinear with qj-qk, they continue collinearly.
         // This can be on either side of p1 (== q1), or collinear
         // The second condition checks if they do not continue
         // oppositely
-        if ( side_pk_q2 == 0 && side_pk_p == side_qk_p )
+        if ( side_1 == 0 && side_2 == side_3 )
```
awulkiew: Out of curiosity, why did you decide to change the names of variables to less meaningful? Above I see a comment referencing specific points so this condition looks unrelated now.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
awulkiew: Furthermore, this is the code calculating turns for endpoints, i.e. the first or the last point of a linestring. There is no such thing in areal geometries. However in this case we have to know how the other geometry behaves at the endpoint, when one geometry ends or begins. And since the geometry either begins or ends at intersection point maybe this is why "non-segment" points from other geometries are considered.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_la.hpp
```diff
@@ -736,20 +737,12 @@ struct get_turn_info_linear_areal
...
-                            typename IntersectionInfo::cs_tag,
-                            rp1_type, rp2_type,
-                            typename IntersectionInfo::side_strategy_type,
-                            rp2_type
-                        > side_calc(inters.rqi(), inters.rpi(), inters.rpj(),
-                                    inters.rqi(), inters.rqj(), inters.rqk(),
-                                    inters.get_side_strategy());
+                    int const side_pj_q2 = sides.apply(range_q.at(1), range_q.at(2), range_p.at(1));
+                    int const side_pj_x = sides.apply(range_q.at(0), range_p.at(0), range_p.at(1));
+                    int const side_qk_x = sides.apply(range_q.at(0), range_p.at(0), range_q.at(2));
```
awulkiew: Is this new way of expressing points more readable to you (e.g. instead of `i()`, `j()`, `k()`)? For me I don't know, looks confusing. Granted, if someone doesn't know what `i`, `j` or  `k` means it's also confusing but at least one cane distinguish one from the other on first sight. Maybe it's a matter of getting used to it. Is it? How do you feel about it after working with it for a while?

awulkiew:
Looks good, I'm only worried about the replacement of `side_calculator_for_endpoint` with manually calculated sides. AFAIU the whole point of it was to have this code used several times encapsulated as a tool. Why have you decided to remove it and replace with code looking less clear?

barendgehrels:
Oh, Imo it is not less clear. It was using named sides, but substituting them with completely other sides, even from different sources. I was really surprised by that and consider it as wrong.
In the new version you can better see what is happening there, which is indeed still not 100% clear, but imo more clear than just call encapsulated magic in the wrong way. I had to do it in really small steps to avoid breaking anything, took me quite some time but all is equivalent as it was before, and all tests pass of course.

barendgehrels:
Oh, Imo it is not less clear. It was using named sides, but substituting them with completely other sides, even from different sources. I was really surprised by that and consider it as wrong.
In the new version you can better see what is happening there, which is indeed still not 100% clear, but imo more clear than just call encapsulated magic in the wrong way. I had to do it in really small steps to avoid breaking anything, took me quite some time but all is equivalent as it was before, and all tests pass of course.

barendgehrels:
Sorry for duplication again, on Iphone this interface is hardly usable

barendgehrels:
@awulkiew About your question, related to the new method of using things as range_q.at(0):

> Is this new way of expressing points more readable to you (e.g. instead of i(), j(), k())?

I was quite used to i,j and k - but I noticed that @vissarion kind of objected to this way of working, and I think I agree with him. So having range-like methods as at(0) (I also considered just subscripting) make it more clear which point is meant, then just alphabetical order i,j,k,l (l, so L, would have been new, but would have been really confusing with 1).

So to answer your question directly, yes, I think that at(0) etc is more readable to me than .i()

barendgehrels:
About your question

> Out of curiosity, why did you decide to change the names of variables to less meaningful?

Because the names were wrong. It used names like "pk" - while, as you mentioned, there is indeed no "pk" in end-points. It is another point, and which one depends on the situation. It is called 5 or 6 times in different ways, I could not come up with a common name, also because I don't understand all the tiny details of this code.

barendgehrels:
Finally, about your comment:

> So AFAIR parts of this code are there to transform ordered areal turns into unordered linear turns.

I did not realize they were unordered, and I don't think they are. Is that really the case? But for sure, I understand that on the end-points there is different behaviour, and different decisions should be made. What I don't understand (at least not understood) was that side calculator was used, a points were inserted for the k-points, and all looked like there were general calculations happening, but behind the screens there were sides calculated totally different from the normal cases (and called like the normal cases). That was my incentive to change it, besides, of course, the need to pass a range instead of points (which I could not do here) and to access the ranges differently, and pass them differently, as was coming from review remarks...

Also, this code was called from both linear/linear and linear/areal.

Are there bulk tests for these cases, inside or outside (at MySQL side) our codebase? I'm curious if it just always works...

Of course I'm willing to give more meaningful names if I know them, and to understand the code better, and refactor some more things if necessary - but for now I thought this was already quite a step and, as a step itself, consistent.

awulkiew:
We have the tests here:
https://github.com/boostorg/geometry/blob/develop/test/algorithms/overlay/get_turns_linear_linear.cpp
https://github.com/boostorg/geometry/blob/develop/test/algorithms/overlay/get_turns_linear_areal.cpp

So debugging them, stepping inside and seeing in which scenarios which sides are calculated here and which points are taken into account should give us the answer. I'll check it out too. Btw, if you're using Visual Studio you could use the extension I wrote to easily see the geometries, I don't know if you know about it (https://marketplace.visualstudio.com/items?itemName=AdamWulkiewicz.GraphicalDebugging).

awulkiew:
Ok, so this is the code handling a case when the intersection point is at the first or last point of the linestring (geometry P) and the second point of a segment of the other linestring or ring (geometry Q). So as usual we want to check which segment has to be followed when intersection or union is calculated. Because there is no 3rd point of the first geometry (and I probably wanted to use existing utility for this) we have to generate some point and it's sufficient to take the first point of Q.

E.g. in this case: `"LINESTRING(1 0,1 1)", "LINESTRING(0 0,1 0,2 0)"` (https://github.com/boostorg/geometry/blob/develop/test/algorithms/overlay/get_turns_linear_linear.cpp#L40) these are the initial segments (P - 1 red segment, Q - 2 green segments):

image

and these are the **new** segments for P (2 blue) with the added first point from Q and 2 segments of Q (green):

image

So now we can check the sides "the regular way" using the existing tools, i.e. `side_calculator` to check sides and `IntersectionInfo` to check spikes and other stuff.

With that said in L/L case (2 linear geometries) 2 sides checks would be enough: `Pi` WRT Q1 and Q2 (so `side_pk_q2` and `side_pk_p` because this is the same as `side_pk_q` in the original code) followed by spike check to see if the side is opposite. There are additional conditions in `operations_of_equal` because this function is used also explicitly in the L/A (linear and areal) case (e.g. here https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/get_turn_info_la.hpp#L780), so if the segment of linear P is on the left of Q, it's outside Q which means union on P, and if it's right of Q this means intersection on P. So for L/L this additional side calculation could be removed.

awulkiew:
I edited the comment to make it more clear what I'm talking about ;)

I don't suggest that this code should be refactored/cleaned as a part of this PR. I just wanted to help understanding it.

Anyway, I agree with merging it together with possible change of names of variables.

So this would probably be the key:
* is_first
    * side_pk_q2 -> side_pj_q2
    * side_pk_p -> side_pj_q
    * side_qk_p -> side_qk_q
* is_last
    * side_pk_q2 -> side_pi_q2
    * side_pk_p -> side_pi_q
    * side_qk_p -> side_qk_q

awulkiew:
One more clarification, by "unordered" I meant that linear geometries are not CW nor CCW and that there is no interior or exterior on left or right side of linear geometry. That regardless of the relative direction of this geometry the methods and operations are the same. They allow to simply follow the linear geometry and generate intersection or union or stop at `x`. Previously when the regular/areal turns were used for linear geometries different turns were generated based on the side of the geometries, like i/i vs x/x etc. Maybe "unordered" was not a good name, my point was that the operations and methods may mean something different than you're used to. E.g. operation `x` means "the end of linestring" and method `e` may be generated for the first endpoint of linestring because it means "intersection point equal to points of both geometries", not "the segments before arriving at IP are equal" as in A/A case, etc.

barendgehrels:
Thanks a lot for your clarifications! I'll continue later.

barendgehrels:
Thanks again for your comments and clear pictures.

Sure, I know your debugging tools, certainly. But I'm always using QtCreator now - and not Visual Studio. I remember that you did have a tool for that too, but (at that time) it did slow things down, sometimes giving issues, and besides that I usually have the need to add more and more debug information, so therefore I rely on SVG creation and do not visualize geometries themselves in the IDE.

It now becomes clear to me why a point of Q was substituted for P. But I still think it was mainly a trick to be able to use the side calculator. That need is gone now, because (by applying ranges) it became clear that it was messing names up (I don't mean implementation, but just the names, as you confirmed).
If that substitution is not necessary anymore, it is trivial to calculate just the point of p (pj in this case) w.r.t. q1 or q2.

For which you provide the key, as far as I can see, and I could use those names (in the calling functions, so i/o x. But note that they are called as, for example the last one:
`int const side_qk_x = sides.apply(range2.at(0), range1.at(1), range2.at(2));`
If this is side qk_q - why is the second point then coming from range 1? I tried to make it consistent with your naming (so use range2.at(1)) but... that gives me errors. So it is not side_qk_q but the side of qk w.r.t. a segment made from qi to pj. 

Or maybe I did not understand you completely...

awulkiew:
I'm playing with the code in develop and replacing the second parameter of `side_calc` at line 432 (`first1`) from `ri1` to `rj2` and at line 485 (`last1`) from `rj1` to `rj2` doesn't change the results for me. At least all of the set operations tests pass. So in case of the first segment (`first1`) the **P'** is changed from `ri2, ri1, rj1` to `ri2, rj2, rj1`, or in the new notation from `rng2.at(0), rng1.at(0), rng1.at(1)` to `rng2.at(0), rng2.at(1), rng1.at(1)`

Anyway, I also tried to remove `side_calculator` at all and at line 428 put this:

    auto side = inters.get_side_strategy();

    int const side_pj_q2 = side.apply(rj2, rk2, rj1);
    int const side_pj_q1 = side.apply(ri2, rj2, rj1);
    int const side_qk_q1 = side.apply(ri2, rj2, rk2);

    std::pair<operation_type, operation_type>
        operations = operations_of_equal(side_pj_q2, side_pj_q1, side_qk_q1);

and at line 482:

    auto side = inters.get_side_strategy();

    int const side_pi_q2 = side.apply(rj2, rk2, ri1);
    int const side_pi_q1 = side.apply(ri2, rj2, ri1);
    int const side_qk_q1 = side.apply(ri2, rj2, rk2);

    std::pair<operation_type, operation_type>
        operations = operations_of_equal(side_pi_q2, side_pi_q1, side_qk_q1);

of course in `operations_of_equal` take `(int side_px_q2, int side_px_q1, int side_qk_q1)`, where `px` means that it's either `pi` or `pj` so the endpoint of the segment P which is not the intersection point. After this change `get_turns_linear_linear` test passes as well.

awulkiew:
Overlay tests also pass for me. In `get_turns_la` I adapted `operations_of_equal` to new arguments, e.g. like that:

    std::pair<operation_type, operation_type>
        operations = get_info_e::operations_of_equal(side_calc.pk_wrt_q2(), side_calc.pk_wrt_p1(), side_calc.qk_wrt_p1());

I also changed the names of points passed to `side_calculator` when it was possible. Note that at line 793 the intersection point is in the middle of Q1 and is equal to `pi` which means that a new segment `qi`->`pi` has to be considered so in this case sides are calculated WRT this new segment and not Q1. And later the same for last, the intersection point may be in the middle. But this is because `side_calculator` is used, if it were not used and sides calculated manually we could probably instead simply use Q1 because it has the same direction as the segment from `qi` to the intersection point at the Q1.

awulkiew:
Do I understand correctly that you do not get the errors with the code as it is now, correct? Only after making changes? And you suggest I play with it after the merging? If that's the case I agree.

barendgehrels:
Sure, that is what I suggested. So I did not get the errors as it is now.
Thanks for merging! And thanks again for reviewing

awulkiew:
Follow-up: https://github.com/boostorg/geometry/pull/540
It worked for me out of the box. Maybe you had problems because of some typo?
Anyway, let's move the discussion there.

barendgehrels:
All right - I was tired, probably not a typo but tried one of them, not the pair together. So obviously that didn't work... Thanks, will go to that new PR

Issue #515, opened 2018-10-12
Invalid result of sym_difference().


awulkiew:
Detected for several cases of `sym_difference`, e.g. *difference.cpp* test case `intersect_holes_intersect_and_disjoint` used in the example below:

    #include <boost/geometry.hpp>
    #include <iostream>

    int main()
    {
        namespace bg = boost::geometry;
        typedef bg::model::point<double, 2, bg::cs::cartesian> point;
        typedef bg::model::polygon<point> polygon;
        typedef bg::model::multi_polygon<polygon> multi_polygon;

        polygon g1, g2;
        multi_polygon result;

        bg::read_wkt("POLYGON((0 0,0 7,5 7,5 0,0 0),(2 2,3 2,3 3,2 3,2 2),(2 4,3 4,3 5,2 5,2 4))", g1);
        bg::read_wkt("POLYGON((1 1,1 6,6 6,6 1,1 1),(2.5 2.5,3.5 2.5,3.5 3.5,2.5 3.5,2.5 2.5))", g2);

        bg::sym_difference(g1, g2, result);

        std::cout << std::setprecision(10);
        std::cout << bg::wkt(result[3].outer()[4]) << std::endl;
        std::cout << bg::wkt(result[4].outer()[2]) << std::endl;

        std::cout << bg::relation(result[3], result[4]).str() << std::endl;

        return 0;
    }

With develop branch and msvc-14.0 (Visual Studio 2015) x86 Debug mode the output is:

    POINT(3 2.49999965)
    POINT(3.00000035 2.5)
    212101212

![path885](https://user-images.githubusercontent.com/1226951/46887920-4df64400-ce5f-11e8-82ef-b04b5db4a4b7.png)

which means that the interiors of polygons 3 and 4 in the result intersect (internally in `relate` `get_turns` correctly returns 2 crossing turns). Two equal points should be generated as a result of intersection of the holes of the input polygons. Two different points are generated instead.

Side note: ~this is not~ previously this was not detected by `is_valid()` due to a bug in this algorithm incorrectly using `point_on_border()` to check if interiors intersect.

awulkiew:
@barendgehrels

awulkiew:
`difference()` cases currently failing validity test are:
- intersect_holes_intersect_and_disjoint
- intersect_holes_intersect_and_touch
- intersect_holes_intersect
The problematic fragment is the same in all of them.

tinko92:
[...]

Issue #466, opened 2018-03-22
disjoint unexpectedly returns true


csommelet:
In some cases, disjoint function does not return the expected value. See code below for a failing example:

```c++
#include <boost/geometry.hpp>
#include <iostream>

using coordinate_system = boost::geometry::cs::spherical_equatorial<boost::geometry::degree>;
using lonlat = boost::geometry::model::point<double, 2, coordinate_system>;
using box = boost::geometry::model::box<lonlat>;
using polygon = boost::geometry::model::polygon<lonlat>;

int main()
{
    std::cout << "Boost version : " << BOOST_LIB_VERSION << " (" << BOOST_VERSION << ")" << std::endl;

    const std::vector<double> norths{
        48.9021,
        48.90215,
        48.9022,
    };

    for (const auto& north: norths) {
        if (boost::geometry::disjoint(
            box{
                {2.2, 48.88},
                {2.5, 48.9021}
            },
            polygon{{{
                {2.4, north},
                {2.4, 48.89},
                {2.3, 48.89},
                {2.3, north},
            }}}))
            std::cerr << "Unexpected disjoint for north lat : " << std::setprecision(7) << north << std::endl;
    }

    return 0;
}
```
Output:
```
Boost version : 1_66 (106600)
Unexpected disjoint for north lat : 48.90215
```

Reproduced when compiled with Apple Clang:
```
Apple LLVM version 9.0.0 (clang-900.0.39.2)
Target: x86_64-apple-darwin17.4.0
```

Some investigation shows that this happens because the method general_areal::apply receives contradictory information:
- ```disjoint_linear<Geometry1, Geometry2>::apply(geometry1, geometry2, strategy)``` returns true, because, schematically, it considers latitudes 48.90215 and 48.9021 to be close enough to be considered equal
- ```rings_containing(geometry2, geometry1, strategy.template get_point_in_geometry_strategy<Geometry1, Geometry2>())``` returns false because it considers point with latitude 48.90215 to be outside the box

awulkiew:
Thanks! I can reproduce the problem. This is indeed a bug but I'd like to discourage you from checking relations of boxes and polygons or performing any other operation. At least in spherical or geographic.

[...]

PR #533, opened 2018-11-29
Define CS-specific behavior of algorithms only with strategy (part 1).

awulkiew:
In Boost.Geometry if a strategy is passed into the algorithm the coordinate system of geometries are ignored. The CS-specific behavior of a strategy is used. At least this is the case at the "highest-level". Internally still in many places the information about the CS defined by strategy is not propagated, default strategies are created from geometry types and algorithms dispatched by CS in algorithms instead of in strategies. To fix this:
- proper strategies has to be "derived" from "higher-level" strategy and propagated down to each underlying algorithm,
- no default strategy should be created internally,
- to be in line with the library design all CS specific code (dispatched by CS) should be moved to strategies directory.

This PR is the first part of changes addressing the above problems. I decided to propose it because it is big enough already and I want to know if the direction I'm moving in is correct.

In order to find places where change is needed semi-automatically I added new coordinate system `cs::undefined` for which default strategy is not defined nor any algorithm dispatched (besides places where some CS is assumed by default which has to be found manually). Next I modified the following:
- move CS-specific normalization utility to strategies `namespace strategy::normalize`
- move CS-specific parts of `expand` and `envelope` to strategies, replacing algorithm dispatched by CS with proper strategies,
- make `equals`/`disjoint` for P/P clean, i.e. strategy now contains CS-specific code and algorithm calls this strategy (internal utility `equals_point_point` and `disjoint_point_point` requires strategy now). Also create CS-specific ones. Keep the agnostic one using one of the CS specific default strategies based on geometries.
- move CS-specific parts of `disjoint` for Box/Box into strategies, create proper strategy etc.
- move CS-specific parts of `disjoint` for Segment/Box into strategies.
- move CS-specific parts of `within` Point/Box strategy from agnostic strategies to CS-specific, keep agnostic strategy using default CS-specific strategy based on CS defined by geometries.
- `point_on_border` `Midpoint` parameter is removed because it was cartesian-only and only was hiding problems (if there was a problem in algorithm and `Midpoint` used then one could have create rings touching at midpoint and get the same result as in the case of rings touching at endpoint but without `Midpoint`)
- `wkt` now internally creates default `disjoint` strategy, I was reluctant to also add `Strategy` argument to `wkt` because I thought that stream manipulators should have only 1 argument but `dsv` has more. Still passing `disjoint` strategy to `wkt` feels wierd.
- azimuth, normalize, expand, envelope, disjoint, etc. strategies are derived from "higher-level" strategies and passed internally.
- overlapping interiors condition fixed in `is_valid`.
- simplify within concept check parameters

The removal of `Midpoint` in `point_on_border` exposed a bug in `is_valid` erroreously returning a false-positive reported here: https://github.com/boostorg/geometry/issues/515. So at this point several `sym_difference` tests fail, which is correct.

This work is also why I was postponing fixing https://github.com/boostorg/geometry/issues/466. I'm going bottom-up with the changes and only now approaching the level of envelope for Polygon. Fixing this issue requires to have a different algorithm for cartesian and non-cartesian Rings and Single-Geometries contained in Multi-Geometries. So at least parts of the algorithm would have to be moved to strategies.


include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
```diff
@@ -310,23 +317,27 @@ class multi_point_multi_geometry
         }
     };
 
+    template <typename DisjointPointBoxStrategy>
     struct overlaps_box_point
     {
         template <typename Box, typename Point>
         static inline bool apply(Box const& box, Point const& point)
         {
             // The default strategy is enough for Point/Box
```
barendgehrels: Maybe this comment can be removed now (is it still the default strategy? It can now be specified)

include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
```diff
@@ -134,7 +148,8 @@ struct disjoint_segment_box_sphere_or_spheroid
 
         if (lon1 > lon2)
         {
-            swap(lon1, lat1, lon2, lat2);
+            std::swap(lon1, lon2);
+            std::swap(lat1, lat2);
```
barendgehrels: This is more clear indeed

include/boost/geometry/algorithms/detail/envelope/range.hpp
```diff
@@ -63,24 +57,29 @@ struct envelope_range
...
+                    >::apply(mbr, *it, strategy.get_element_expand_strategy());
             }
         }
     }
 
     template <typename Range, typename Box, typename Strategy>
     static inline void apply(Range const& range, Box& mbr, Strategy const& strategy)
     {
-        return apply(boost::begin(range), boost::end(range), mbr, strategy);
+        return apply(Strategy::begin(range), Strategy::end(range), mbr, strategy);
```
barendgehrels: This is surprising. What is this for kind of strategy? So it should implement range like features?

include/boost/geometry/io/wkt/read.hpp
```diff
@@ -306,6 +306,17 @@ struct stateful_range_appender<Geometry, open>
     }
 
 private:
+    static inline bool disjoint(point_type const& p1, point_type const& p2)
+    {
+        // TODO: pass strategy
```
barendgehrels: OK for me, PR is large enough

include/boost/geometry/core/cs.hpp
```diff
@@ -65,13 +65,13 @@ struct coordinate_system_units
...
 template <>
-struct coordinate_system_units<geometry::degree>
+struct define_angular_units<geometry::degree>
 {
     typedef geometry::degree units;
 };
 
 template <>
-struct coordinate_system_units<geometry::radian>
+struct define_angular_units<geometry::radian>
```
barendgehrels: Right, I see. We don't get backward compatibility issues with this?

OK, I see that it is part of core_detail - then it should be no problem

include/boost/geometry/core/cs.hpp
```diff
@@ -107,12 +107,8 @@ known as lat,long or lo,la or phi,lambda
...
 template<typename DegreeOrRadian>
 struct geographic
-{
-    typedef typename core_detail::coordinate_system_units
-        <
-            DegreeOrRadian
-        >::units units;
-};
+    : core_detail::define_angular_units<DegreeOrRadian>
+{};
```
barendgehrels: What is the reason to derive instead of defining it inside?

include/boost/geometry/core/cs.hpp
```diff
@@ -174,12 +162,15 @@ struct spherical_equatorial
...
+
+
+/*!
+\brief Undefined coordinate system
+\ingroup cs
+*/
+struct undefined {};
```
barendgehrels: OK

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
```diff
@@ -327,10 +327,19 @@ private :
...
     typedef typename model::referring_segment<point2_type const> segment2_type;
 
-    template <typename Iterator, typename RangeIterator, typename Section, typename RobustPolicy>
+    template
+    <
+        typename Iterator, typename RangeIterator,
+        typename Section,
+        typename IntersectionStrategy, typename RobustPolicy
+    >
     static inline void advance_to_non_duplicate_next(Iterator& next,
```
barendgehrels: I moved this. So this will give a merge conflict later.

include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
awulkiew: Yes, it should have been removed.

include/boost/geometry/algorithms/detail/envelope/range.hpp
awulkiew: The problem with envelope is that in cartesian we can process points but in spherical and geographic we have to analyse segments becuase geodesic's vertex may have different latitude (higher or lower on northern or southern hemisphere respectively) than the endpoints. So the strategy feeds the algorithm either with points or with segments. Previously the dispatching by CS was done in the algorithm. I decided to do it this way to allow the strategy to define the smallest Cs-specific portions of the algorithm. Otherwise I'd have to implement the whole calculation of envelope of a range as strategy.

include/boost/geometry/io/wkt/read.hpp
awulkiew: Yes, I wanted to postpone it because I didn't want to force the user to pass within point strategy into `read_wkt()`. We should probably have separate wkt strategy for that.

include/boost/geometry/core/cs.hpp
awulkiew: Less lines of code. AFAIU it's a typical way of defining MPL-like meta-functions.

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
awulkiew: No problem, I'll resolve them.

awulkiew:
No problem with merging your PR first.

include/boost/geometry/algorithms/detail/envelope/interface.hpp
```diff
@@ -57,13 +62,15 @@ struct envelope
                              Box& box,
```
vissarion: why do not split here in two structs and dispatch as in area and distance?

include/boost/geometry/algorithms/detail/envelope/range.hpp
```diff
@@ -63,24 +57,29 @@ struct envelope_range
...
+                >::apply(*it, mbr, strategy.get_element_envelope_strategy());
 
             // consider now the remaining elements in the range (if any)
             for (++it; it != last; ++it)
             {
-                dispatch::expand<Box, value_type>::apply(mbr, *it, strategy);
+                dispatch::expand
+                    <
+                        Box, value_type
+                    >::apply(mbr, *it, strategy.get_element_expand_strategy());
```
vissarion: so, `strategy` here is a "envelope_range" strategy that contains an expand strategy for elements. Is this a bit unexpected ? I mean to hide an expand strategy in an envelope strategy.

include/boost/geometry/algorithms/detail/envelope/range.hpp
vissarion: +1

include/boost/geometry/algorithms/detail/equals/point_point.hpp
```diff
@@ -36,17 +33,16 @@ namespace detail { namespace equals
...
     \note To avoid circular references
  */
-template <typename Point1, typename Point2>
-inline bool equals_point_point(Point1 const& point1, Point2 const& point2)
+template <typename Point1, typename Point2, typename Strategy>
+inline bool equals_point_point(Point1 const& point1, Point2 const& point2,
+                               Strategy const& )
 {
-    return ! detail::disjoint::disjoint_point_point(point1, point2);
+    return Strategy::apply(point1, point2);
```
vissarion: Why we need a special strategy for checking if 2 points are equal? Is it the same for all CS? Unless if you consider normalization, but in this case how normalization strategy is passes here? What am I missing...

include/boost/geometry/algorithms/detail/expand/interface.hpp
```diff
@@ -58,14 +60,11 @@ struct expand
                              Geometry const& geometry,
```
vissarion: same here

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
```diff
@@ -482,11 +492,13 @@ class get_turns_generic
 
         typename IntersectionStrategy::envelope_strategy_type const
             envelope_strategy = intersection_strategy.get_envelope_strategy();
```
vissarion: is it needed for intersection strategy to contain both envelope and expand strategies? Why not passing them independently as in the strategies in `disjoint_segment_box`?

include/boost/geometry/algorithms/detail/within/interface.hpp
```diff
@@ -70,13 +70,7 @@ struct within
                              Geometry2 const& geometry2,
```
vissarion: same here. why no dispatching as in area and distance?

test/strategies/segment_envelope.cpp
```diff
@@ -1,6 +1,6 @@
 // Boost.Geometry
```
vissarion: now there is strategy named `envelope_segment` the name of this file should also change

include/boost/geometry/algorithms/area.hpp
```diff
@@ -216,15 +216,20 @@ struct area<MultiGeometry, multi_polygon_tag> : detail::multi_sum
...
+    template <typename Geometry>
     static inline typename area_result<Geometry, Strategy>::type
     apply(Geometry const& geometry, Strategy const& strategy)
     {
         return dispatch::area<Geometry>::apply(geometry, strategy);
     }
+};
 
+template <>
+struct area<default_strategy>
```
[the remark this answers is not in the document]
awulkiew: No, this was done from technical reasons. In order to instantiate `area_result` with correct arguments and because it's a return type of member function I had to do it this way. Otherwise I'd left it as it was.

include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
```diff
@@ -50,21 +53,6 @@ namespace detail { namespace disjoint
 template <typename CS_Tag>
 struct disjoint_segment_box_sphere_or_spheroid
```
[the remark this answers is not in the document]
awulkiew: Yes, technically this should probably be moved to strategies. I either left it here to limit the amount of changes or I simply missed it. It's also probable that this code is still used in some other place too. I'd have to check. If possible I'll move it to strategies.

include/boost/geometry/algorithms/detail/envelope/interface.hpp
awulkiew: Because here it's sufficient to distinguish between strategies on function level, the compiler can pick the correct function automatically. With area and distance the function return type was based on strategy and this caused problems. Even the `Geometry` parameter could probably be moved.

include/boost/geometry/algorithms/detail/envelope/range.hpp
awulkiew: Yes, expand and envelope implementation is tightly bound together. The envelope range has to know how to extract the simple geometries from a range (points in cartesian and segments in spherical/geographic) then how to calculate the bounding boxes from them (e.g. spherical/geographic segment to box) and how to expand by them (e.g. spherical/geographic segment to spherical box and then expansion of spherical box). There are such dependencies in envelope/expand code where one strategy needs lower-level strategies of the other algorithm and this propagates "up". AFAIR I haven't yet refactor the envelope code for Multi-Geometries which adds another layer of dependencies. I left the dependencies as they were before because I wanted to avoid dividing the whole algorithms into strategies one for each kind of Geometry (one for each simple geometry, next for all ranges and next for multi-ranges) but maybe this is how it should be done. But I'd prefer to do it in part 2.

include/boost/geometry/algorithms/detail/equals/point_point.hpp
awulkiew: In general yes, this could be done the same for all coordinate systems. For this we'd have to agree what's the valid range of coordinates, e.g. if `lon_deg = -180+eps/2` is valid, since this would be true: ``bg::math::equals(lon_deg + 360, 180), this coordinate would be ambiguous and cartesian-like comparison would fail. Currently we have this normalization in place which even if the user represents the coordinates ambiguously (`-180` vs `180` or even as something outside valid range) the relation operations still works. In order for this to work we need separate strategies for various CSes even for points.

vissarion: I do not have a strong opinion here, since I cannot predict the user needs, but what about fixing a range e.g. `(-180,180]`, then if the user compare invalid coordinates the result is "unexpected" as with invalid geometries.

vissarion: In any case, this should be done in a new PR.

awulkiew: Yes, the normalization could also be a part of `correct`. However I think that assuming that -180 is correct is common practice and if we changed the existing behavior users could get unexpected results.

awulkiew:
Merged develop with recently merged https://github.com/boostorg/geometry/pull/531 and https://github.com/boostorg/geometry/pull/540.

As a reminder, note that some `sym_difference` tests fails due to this: https://github.com/boostorg/geometry/issues/515

> Additionally, since it seems the right thread to ask: if strategies define the CS-specific computation of an algorithm what is an agnostic strategy? Should just be part of the corresponding algorithm?

@vissarion I'm not sure. Maybe it indeed should simply be a different algorithm with distinguished name, like `bg::algorithm_variant1`, `bg::algorithm_variant2`, `bg::algorithm_variant3` and the default `bg::algorithm`. In that case I'm not sure if the name "Strategy" is a good name since it's more general, maybe it would then be "CSStrategy". But anyway, I'm not sure if we can change that now because of causing problems to the users. Thought If there is only 1 agnostic strategy for algorithm I doubt anyone is using it explicitly. Maybe besides buffer's strategies which are probably used heavily because they are something between Strategies and Parameters. And on top of this we might decide to move to C++11 like some other libraries and this could be a good opportunity to redefine some aspects of the library. IF we want to talk about all of it I suggest to create an Issue.

include/boost/geometry/algorithms/detail/within/interface.hpp
awulkiew: I answered that before but also here, dispatching based on function templates doesn't force us to explicitly state the types. I've done it there only because I was forced to do it, not because I wanted to do it. ;)

test/strategies/segment_envelope.cpp
awulkiew: Will do.

PR #538, opened 2018-12-10
Add workaround for msvc-12 faulty array initialization.

awulkiew:
This is a workaround for the following error:

    boost/geometry/srs/projections/par_data.hpp(92) : error C2536: 'boost::geometry::srs::detail::towgs84<T>::boost::geometry::srs::detail::towgs84<T>::m_data' : cannot specify explicit initializer for arrays

I added `core/config.hpp` file allowing us to add defines conveniently.

### DIFF
--- a/include/boost/geometry/core/config.hpp
+++ b/include/boost/geometry/core/config.hpp
@@ -1,0 +1,20 @@
+// Boost.Geometry
+
+// Copyright (c) 2018 Oracle and/or its affiliates.
+// Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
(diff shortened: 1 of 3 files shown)


include/boost/geometry/core/config.hpp
```diff
@@ -0,0 +1,20 @@
...
+// Copyright (c) 2018 Oracle and/or its affiliates.
+// Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_CORE_CONFIG_HPP
+#define BOOST_GEOMETRY_CORE_CONFIG_HPP
+
```
barendgehrels: Cool, I've added this file too - for removing rescaling.
Your's will be merged earlier so I will adapt/extend it

PR #540, opened 2018-12-15
Make side calculation for linear endpoints clearer.

awulkiew:
It seems that not all sides has to be calculated so this PR also adds an optimization for L/A case.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
@@ -225,6 +225,8 @@ private:
 template <bool EnableFirst, bool EnableLast>
 struct get_turn_info_for_endpoint
 {
+    typedef std::pair<operation_type, operation_type> operations_pair;
+
     BOOST_STATIC_ASSERT(EnableFirst || EnableLast);
 
     template<typename UniqueSubRange1,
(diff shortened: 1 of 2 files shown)


include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
```diff
@@ -416,11 +418,10 @@ struct get_turn_info_for_endpoint
                     // TODO: find out why side is used with respect to non-segments
                     // (that is: range2.at(0) to range1.at(0), denoted as x)
```
barendgehrels: This comment can be removed now

include/boost/geometry/algorithms/detail/overlay/get_turn_info_la.hpp
```diff
@@ -797,12 +802,33 @@ struct get_turn_info_linear_areal
...
+                {
+                    side_pi_y = sides.apply(range_q.at(1), range_q.at(2), range_p.at(0)); // pi wrt q2
+                    side_pi_x = sides.apply(range_q.at(0), range_q.at(1), range_p.at(0)); // pi wrt q1
+                    side_qz_x = sides.apply(range_q.at(0), range_q.at(1), range_q.at(2)); // qk wrt q1   
+                }
+                // 2. ip0 or pj in the middle of q1
+                else
+                {
+                    side_pi_y = sides.apply(range_q.at(0), range_q.at(1), range_p.at(0)); // pi wrt q1
+                    side_pi_x = side_pi_y; // pi wrt q1
```
barendgehrels: Great, so all sides are real sides now, a point w.r.t. an existing segment.
Much better and better documented and named.
I'm glad you worked on this last bit, you know and understand this code much better than I do.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
awulkiew: Done.

PR #544, opened 2019-01-02
Fix Segment View doc: Wrong output

jaredjxyz:
The documentation for "Segment View" is currently incorrect. This is a fix to show the actual output for the given example.


awulkiew:
Thanks!

Issue #542, opened 2018-12-16
Direct Formula returns wrong results


barendgehrels:
Hi @vissarion , @awulkiew ,

When I, for buffer, try to calculate the coordinates of a point, at 180 degrees, there is an error.

See this minimal reproduction scenario:
```
#include <boost/geometry.hpp>

int main(int, char* [])
{
    namespace bg = boost::geometry;
    typedef bg::strategy::andoyer::template direct
        <
            double, true, false, false, false
        > direct_t;

    bg::srs::spheroid<double> spheroid;

    typedef bg::model::point<double, 2, bg::cs::geographic<bg::degree> > point_type;

    point_type point;
    bg::set<0>(point, 4.9);
    bg::set<1>(point, 52.4);

    int const count = 12;
    double const distance = 1.0;
    double const two_pi = bg::math::two_pi<double>();
    double const diff = two_pi / double(count);
    double angle = 0;
    for (std::size_t i = 0; i < count; i++, angle += diff)
    {
        typename direct_t::result_type dir_r
                = direct_t::apply(bg::get_as_radian<0>(point),
                                  bg::get_as_radian<1>(point),
                                  distance, angle, spheroid);
        point_type p;
        bg::set_from_radian<0>(p, dir_r.lon2);
        bg::set_from_radian<1>(p, dir_r.lat2);

        double const pi = bg::math::pi<double>();
        std::cout << std::setprecision(20) << " " << angle * 180.0 / pi
                  << " " << bg::get<0>(p) << " " << bg::get<1>(p) << std::endl;
    }
    return 0;
}
```

Delivering:
```
 0 4.9000000000000003553 52.400008986761207552
 29.999999999999996447 4.9000073460059256902 52.400007782763282194
 59.999999999999992895 4.9000127236545552023 52.400004493379917392
 90 4.9000146920092788605 52.399999999999081979
 119.99999999999998579 4.9000127236519706031 52.399995506618694208
 149.99999999999997158 4.9000073460033402029 52.399992217236267322
 179.99999999999997158 4.9000000000000003553 58.883125680961242665
 209.99999999999997158 4.8999926539966605077 52.399992217236260217
 239.99999999999997158 4.8999872763480309956 52.399995506618694208
 270 4.89998530799072185 52.399999999999081979
 300.00000000000005684 4.8999872763454428437 52.400004493379931603
 330.00000000000005684 4.8999926539940723558 52.400007782763275088
```

[...]

Issue #541, opened 2018-12-16
Geographic Area returns wrong results


barendgehrels:
Hi @vissarion , @awulkiew ,
I'm busy producing buffers in geographical coordinate system around a point. That is work in progress now, and it starts to work. However, when I want to check the area (of the geographic buffer) I get unexpected results.

Below the debug output, containing: 
number of input points, "geographic", number of output points (one more), area-in-geographic, "projected", number of output points (still one more as input), area-in-cartesian. Below that their WKT representations in both geographic and projected (using sterea for the Dutch coordinate system, but that is not relevant for this issue)

Because I specified one meter, one would expect an area of pi. Because the number of points is low, it is less, but in cartesian (I used Boost.Geometry projection/transformation), that all seems allright.

So I have to conclude that the area calculated by boost::geometry::area for a geographic coordinate system is not right... Or maybe I had to specify another strategy, but I debugged the code, and basically it seems to take the right strategies.

You see the geographic area fluctiating from -7106 via --51940 to +13597 etc. In only one case (not below, because from a previous debugging version) I got a value close to a right value...

```
5  geographic  6   -7106.98  projected  6   2.3767
     MULTIPOLYGON(((4.900000014692011 52.40000898675671,4.900013977465128 52.40000276851281,4.900008623854553 52.39999272427848,4.899991352373283 52.39999273484305,4.899986031615026 52.40000278560664,4.900000014692011 52.40000898675671)))
     MULTIPOLYGON(((121816.3935228445 490239.7834838452,121817.3390161667 490239.0853036283,121816.9671865939 490237.9703284562,121815.7918899588 490237.9794161207,121815.437346262 490239.1000077746,121816.3935228445 490239.7834838452)))
7  geographic  8   -51940.44953270301  projected  8   2.735323085114942
     MULTIPOLYGON(((4.900000014692011 52.40000898675671,4.900011495831415 52.40000559612448,4.90001432037284 52.39999799149619,4.900006361382522 52.39999189931258,4.899993612143401 52.39999190711099,4.899985673088598 52.39999800901907,4.899988522489213 52.40000561017674,4.900000014692011 52.40000898675671)))
[...]

PR #546, opened 2019-01-16
Feature/buffer point geographic

barendgehrels:
As discussed, this PR implements a strategy for buffers around points in the Geographic coordinate system.

Besides the strategy, it needed some adaptions in the buffer code.

Included are a unit test and documentation.


barendgehrels:
Is it OK to merge this?

awulkiew:
Sorry that it takes me so long to review it. I'll do this this week.

barendgehrels:
No problem

vissarion:
I am also delayed in this. I'll try to do it the next days (while travelling) otherwise beginning of next week (if this is not an issue).

barendgehrels:
Hi @vissarion , thanks - no problem, no hurry necessary.

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -736,46 +744,47 @@ struct buffered_piece_collection
...
+                        BOOST_GEOMETRY_ASSERT
+                        (
+                            index_in_vector > 0
+                            && index_in_vector < pc.offsetted_count
+                        );
 
-                    pc.robust_ring.insert(boost::begin(pc.robust_ring) + index_in_vector, rit->point);
-                    pc.offsetted_count++;
+                        pc.robust_ring.insert(boost::begin(pc.robust_ring) + index_in_vector, rit->point);
+                        pc.offsetted_count++;
```
awulkiew: tab? ;)

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -736,46 +744,47 @@ struct buffered_piece_collection
...
+                        m_turns[rit->turn_index].operations[rit->operation_index].index_in_robust_ring = index_in_vector + index_offset;
+                    }
                 }
             }
-        }
 
-        BOOST_GEOMETRY_ASSERT(assert_indices_in_robust_rings());
-#endif
+            BOOST_GEOMETRY_ASSERT(assert_indices_in_robust_rings());
+        }
```
awulkiew: tab

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
```diff
@@ -127,85 +122,115 @@ inline bool in_box(Point const& previous,
...
-    }
 
-    // Not monotonic, on left or right side: continue analysing
-    return analyse_continue;
-#endif
-}
+        // Not monotonic, on left or right side: continue analysing
+        return analyse_continue;
+    }
+};
```
awulkiew: Tabs. Do you have an idea why does github display the `;` differently here?

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
```diff
@@ -276,8 +292,51 @@ public :
...
+{
+public :
+    template <typename Turn, typename Piece>
+    static inline analyse_result apply(Turn const& turn, Piece const& piece)
+    {
+        typedef typename Piece::section_type section_type;
+        typedef typename Turn::robust_point_type point_type;
+        typedef typename geometry::coordinate_type<point_type>::type coordinate_type;
+
+        typedef strategy::within::winding<point_type> strategy_type;
```
awulkiew: Actually now there are separate winding strategies for each coordinate system. This one is there for backward compatibility and internally calls one of the CS specific ones based on PointType, see: https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/strategies/agnostic/point_in_poly_winding.hpp#L63. So my guess is that you should get this strategy from the one passed into the buffer.

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
```diff
@@ -697,7 +738,72 @@ class turn_in_piece_visitor
...
+        if (piece.is_convex)
+        {
+            return in_convex_piece(turn, piece);
+        }
+        else
+        {
+
+            // TODO: this point_in_geometry is a performance-bottleneck here and
+            // will be replaced completely by extending analyse_piece functionality
+            return detail::within::point_in_geometry(turn.robust_point, piece.robust_ring);
```
awulkiew: You should pass CS-specific winding strategy extracted from buffer strategy into this algorithm.

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
```diff
@@ -697,7 +738,72 @@ class turn_in_piece_visitor
...
+
+template <>
+struct turn_in_piece<false>
+{
+public :
+
+    template <typename Turn, typename Piece>
+    static inline int apply(Turn const& turn, Piece const& piece)
+    {
+        return detail::within::point_in_geometry(turn.robust_point, piece.robust_ring);
```
awulkiew: Needs explicit winding strategy.

include/boost/geometry/strategies/geographic/buffer_point_circle.hpp
```diff
@@ -0,0 +1,127 @@
...
+\* [link geometry.reference.strategies.strategy_buffer_point_square point_square]
+}
+ */
+template
+<
+    typename FormulaPolicy = strategy::andoyer,
+    typename Spheroid = srs::spheroid<double>,
+    typename CalculationType = void
+>
+class point_circle_geographic
```
awulkiew: The naming is slightly inconsistent with already existing geographic strategies. We have:
- `strategy::distance::geographic`
- `strategy::side::geographic`
- `strategy::intersection::geographic_segments`
- `strategy::distance::geographic_cross_track`

So my guess is that this should be called `geographic_point_circle`.

Side note: I noticed that the naming of strategies is inconsistent with the directories structure. Typically we want to have namespaces aligned with directories. Should there be CS-specific namespaces in `strategy` namespace resembling CS-specific directories in *strategies* directory? For example: 
- `strategy::geographic::distance::point_point`
- `strategy::geographic::distance::point_segment`
- `strategy::geographic::intersection::segments`
- `strategy::geographic::side::point_segment`
- `strategy::cartesian::distance::point_point` <=> `strategy::cartesian::distance::pythagoras`
Just a side question for the future.

test/algorithms/buffer/test_buffer_geo.hpp
```diff
@@ -0,0 +1,83 @@
...
+    // Use the appropriate strategy for geographic points
+    bg::strategy::buffer::point_circle_geographic<> circle_strategy(settings.points_per_circle);
+
+    // Use Thomas strategy to calculate geographic area, because it is
+    // the most precise (unless scale of buffer is only around 1 meter)
+    bg::strategy::area::geographic
+    <
+        bg::strategy::thomas, 5,
+        bg::srs::spheroid<long double>, long double
+    > area_strategy;
```
awulkiew: Just for cross-reference. Issues related to this:
https://github.com/boostorg/geometry/issues/541
https://github.com/boostorg/geometry/issues/542

awulkiew:
Thanks. I'm ok with merging after addressing some minor issues.

include/boost/geometry/strategies/geographic/buffer_point_circle.hpp
vissarion: I agree with the name `geographic_point_circle` and also with the side note. Just to add that we probably need to change also names like pythagoras and andoyer to something to remind that it is Euclidean distance and first order approximation of geographic/ellipsoidal distance.

include/boost/geometry/strategies/geographic/buffer_point_circle.hpp
```diff
@@ -0,0 +1,127 @@
...
+[buffer_point_circle_geographic_output]
+[heading See also]
+\* [link geometry.reference.algorithms.buffer.buffer_7_with_strategies buffer (with strategies)]
+\* [link geometry.reference.strategies.strategy_buffer_point_circle point_circle]
+\* [link geometry.reference.strategies.strategy_buffer_point_square point_square]
+}
+ */
+template
+<
+    typename FormulaPolicy = strategy::andoyer,
```
vissarion: why not `thomas` same as the one used in tests to parametrize `area`?

include/boost/geometry/strategies/geographic/buffer_point_circle.hpp
barendgehrels: We have:

- strategy::buffer::point_circle
- strategy::buffer::point_square
And therefore I selected the fitting:
- strategy::buffer::point_circle_geographic

But yes, it can also be
- strategy::buffer::geographic_point_circle

If you prefer that (and apparently you do) then I will change it.

We can actually not rename existing strategies (pythagoras) because they might be used by users.

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
barendgehrels: Thank you

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
barendgehrels: ; differently might be caused by me, splitting a free function to a member function of a class, and that might also confuse you for the 'tabs'. Because as far as I can see they are all right.

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
barendgehrels: Done, can you verify if this is what you intended?

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
barendgehrels: Also here: done, can you verify if this is what you intended?

include/boost/geometry/strategies/geographic/buffer_point_circle.hpp
barendgehrels: All strategies do it like this (e.g. area, azimuth, densify, etc)
It should not necessarily be aligned with test code, should it?
Of course - we could change all with tomas if you advise that and it is fast enough.
But for now I would keep it consistent with the other strategies

test/algorithms/buffer/buffer_point_geo.cpp
```diff
@@ -0,0 +1,40 @@
+// Boost.Geometry
+// Unit Test
+
+// Copyright (c) 2018-2019 Barend Gehrels, Amsterdam, the Netherlands.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#include <test_buffer_geo.hpp>
```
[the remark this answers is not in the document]
barendgehrels: Yes, we're not so consequent in that indeed.
Fixed all for buffer tests

include/boost/geometry/strategies/geographic/buffer_point_circle.hpp
awulkiew: Right, but we can add another strategy with a different name, deprecate the old one by putting the `#warning` in the strategy header and remove the include from main header `geometry.hpp`.

The above is also true for `strategy::buffer::point_circle` to be consistent it could be renamed to `strategy::buffer::cartesian_point_circle`. But I'd like to propose something different.

Correct me if I'm wrong. In the future other geographic and spherical strategies could be implemented. Furthermore as it is now the interface of `buffer()` would allow passing various-CS strategies at the same time, e.g. one for cartesian, one geographic, yet another spherical, etc. If I'm correct and it's the case I propose to redesign the interface by:
- allowing to pass 1 CS-specific buffer strategy as the last, additional parameter of `buffer()`
- treating current strategies as policies of this strategy
This way the parameters would be consistent by design.

barendgehrels: Yes, indeed, you are right. It currently would allow that (though we could protect that).
So I can do that later - for now I propose to rename it as you both agreed (didn't rename it yet - will do tonight). That is the last point of this review and then I think it can be merged.

include/boost/geometry/strategies/geographic/buffer_point_circle.hpp
vissarion: ok maybe it is not needed to be aligned with the tests, this could be changed together with area default strategy in a separate PR

include/boost/geometry/strategies/geographic/buffer_point_circle.hpp
```diff
@@ -0,0 +1,127 @@
...
+            <
+                calculation_type, true, false, false, false
+            > direct_t;
+
+        calculation_type const two_pi = geometry::math::two_pi<calculation_type>();
+
+        calculation_type const diff = two_pi / calculation_type(m_count);
+        // TODO: after calculation of some angles is corrected,
+        // we can start at 0.0
+        calculation_type angle = 0.001;
```
awulkiew: I wanted to fix it. How can I verify that the angles are correct or incorrect? I didn't find any test case in the test file which was commented out due to this. Do you have in mind this issue: https://github.com/boostorg/geometry/issues/542 or is this a different issue?

I'm asking because I wanted to fix it by making the azimuth passed into the direct formula to be in valid range (-pi, pi], like this in the loop below:

    if (angle > pi)
        angle -= two_pi;

and also to start from 0. I verified that this changes the coordinates but causes more cases to fail. And I don't know if this is because the coordinates are incorrect or that it's inaccurate area's fault.

My understanding is that this `0.001` magic number only hides the problem in the specific several cases in test file. Is that correct? Because in general case if the intermediate results from some formulas are incorrect it doesn't matter where is the starting point since problematic angles can be achieved in various ways (i.e. by also tweaking the point count or changing the starting point).

PR #551, opened 2019-01-30
Feature/remove rescaling preparations

barendgehrels:
Please review this code. It prepares the removal of rescaling, by changes in code necessary in other parts than the intersection itself. These parts are: enrichment, traversal, get turn information.
Also it adds some tests.

There are no regressions.


include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -131,6 +131,17 @@ struct base_turn_handler
...
+    template <typename Point1, typename Point2>
+    static inline typename geometry::coordinate_type<Point1>::type
+            distance_measure(Point1 const& a, Point2 const& b)
+    {
+        // TODO: use comparable distance for point-point instead - but that
+        // causes currently cycling include problems
+        typedef typename geometry::coordinate_type<Point1>::type ctype;
+        ctype const dx = get<0>(a) - get<0>(b);
+        ctype const dy = get<1>(a) - get<1>(b);
+        return dx * dx + dy * dy;
```
awulkiew: Right, there should be similar tool as `equals_point_point` where you only include the algorithm for points. But this tool should take the strategy so it'd be ok to simply call comparable distance strategy here.

This code is cartesian-only, is that correct?

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -642,18 +696,6 @@ struct collinear : public base_turn_handler
...
-    static inline typename geometry::coordinate_type<Point1>::type
-            distance_measure(Point1 const& a, Point2 const& b)
-    {
-        // TODO: use comparable distance for point-point instead - but that
-        // causes currently cycling include problems
-        typedef typename geometry::coordinate_type<Point1>::type ctype;
-        ctype const dx = get<0>(a) - get<0>(b);
-        ctype const dy = get<1>(a) - get<1>(b);
-        return dx * dx + dy * dy;
-    }
```
awulkiew: Right, it was moved from here. I'll update the comment above.

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
```diff
@@ -354,35 +372,71 @@ struct traversal
...
 
-        // Initialize with 0, automatically assigned on first result
+        turn_operation_type const& op0 = turn.operations[0];
+        turn_operation_type const& op1 = turn.operations[1];
+        signed_size_type const next0 = op0.enriched.get_next_turn_index();
+        signed_size_type const next1 = op1.enriched.get_next_turn_index();
+
+        bool possible[2];
+        possible[0] = traverse_possible(next0);
+        possible[1] = traverse_possible(next1);
```
awulkiew: Slightly inconsistent naming. Other variables have indexes at the end of the name instead of being an array. So `possible0`, `possible1`. Or is there a reason for this?

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -195,10 +209,21 @@ struct touch_interior : public base_turn_handler
         }
         else if (side_qi_p == 1 && side_qk_p == 1 && side_qk_q == -1)
         {
+            if (side_pj_q2 == -1)
+            {
             // Q turns right on the left side of P (test "ML3")
-            // Union: take both operation
+                // Union: take both operations
```
awulkiew: tab?

awulkiew:
I'm ok with merging.

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
barendgehrels: Good point. Not really, I will harmonize it.

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
barendgehrels: I upgraded my VM and apparently the settings are not right yet, will fix. Thanks.

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
```diff
@@ -712,24 +766,163 @@ struct traversal
...
+                    if (blti.next_turn_index == lti.next_turn_index
+                            && blti.rank_index == lti.rank_index)
+                    {
+                        return false;
+                    }
+                }
+            }
+        }
 
+        // Traveral can either enter the cluster in the first turn,
```
vissarion: Traversal...

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
barendgehrels: Was not a tab but the rest was missing an indent. Fixed now (will push later).

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
barendgehrels: I answered earlier: "Good point. Not really, I will harmonize it." (but apparently I did something wrong, it was placed under the wrong comment).
Anyway, there was a reason because it is used for iteration below.
But I now harmonized it (using more arrays of 2 and iterations), will push.
Thank you.

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
barendgehrels: Fixed, thnks

barendgehrels:
Thanks for your comments! They are addressed, will merge.

awulkiew:
Something is wrong with this change. It causes assertion failure in all algorithms calling `get_turns`:

    /boost/geometry/algorithms/detail/overlay/get_turns.hpp:149:
    const Point& boost::geometry::detail::get_turns::unique_sub_range_from_section<IsAreal, Section, Point, CircularIterator, IntersectionStrategy, RobustPolicy>::at(std::size_t) const [with bool IsAreal = false; Section = boost::geometry::section<boost::geometry::model::box<boost::geometry::model::point<double, 2ul, boost::geometry::cs::cartesian> >, 2ul>; Point = boost::geometry::model::point<double, 2ul, boost::geometry::cs::cartesian>; CircularIterator = boost::geometry::ever_circling_iterator<__gnu_cxx::__normal_iterator<const boost::geometry::model::point<double, 2ul, boost::geometry::cs::cartesian>*, std::vector<boost::geometry::model::point<double, 2ul, boost::geometry::cs::cartesian>, std::allocator<boost::geometry::model::point<double, 2ul, boost::geometry::cs::cartesian> > > > >; IntersectionStrategy = boost::geometry::strategy::intersection::cartesian_segments<>; RobustPolicy = boost::geometry::detail::no_rescale_policy; std::size_t = long unsigned int]:
    Assertion `index < size()' failed.

barendgehrels:
Hmm, I get cannot see the regression matrix.
However, I get a compilation error by the changes in 846d91cac

barendgehrels:
After fixing that (is submitted), I don't get that assertion. You say "all algorithms", is it literally all?
Can you say me which algorithm fails?

barendgehrels:
Should be fixed, now running regression tests again.
I now realize I did run them but in release mode, so didn't catch these assertions

awulkiew:
Looks good now, thanks: https://circleci.com/gh/boostorg/workflows/geometry

> I don't get that assertion. You say "all algorithms", is it literally all?

Well, I exaggerated ;) All set operations, is_simple and relate. See: https://circleci.com/workflow-run/c23a6228-8e9b-418b-a1d7-2845883f571c

Thanks for fixing so fast!

barendgehrels:
You are welcome ;-)

PR #559, opened 2019-02-23
Enhance/remove no self defines

barendgehrels:
Please review, I would like to merge this.
It removes the BOOST_GEOMETRY_NO_SELF_TURN conditions from the unit tests.
Because self turns are stable, necessary, and unit testing without them is not tested anymore and can be omitted now.,


awulkiew:
Thanks. I'm ok with merging.

PR #561, opened 2019-03-01
Add new algorithm detail::calculate_point_order() with strategies (and optimize remove_spikes()).

awulkiew:
This solves the issue caused by inaccurate area calculation for small polygons with andoyer formula.

For geographic CS instead of area the sum of external angles is calculated.

I tried something new with this algorithm. Instead of being dispatched by geometry tag I dispatched it by strategy's `version_tag`.

There are 2 versions of this algorithm:
- area-based which simply calculates the area and checks its value
- azimuth-based which filters out point duplicates and spikes (it's modified version of `remove_spikes`)

The version of an algorithm is defined by the strategy as a member type `version_tag`. The reason for this is that both algorithms are very different and the alternative would be to implement everything in strategies.

This algorithm works for Rings only (actually Ranges of points). It returns one of three values of type `bg::order_selector`.

As mentioned before geographic, azimuth-based version in modified `remove_spikes`. Instead of calling side strategy (2 azimuth calculated per point) it performs 1 azimuth/reverse-azimuth calculation per segment if needed and stores the result for future use. This means that 2x less azimuths are calculated compared to `remove_spike`. Spike detection is a side effect of the "normal" calculation of external angle for each segment. At a point reverse azimuth of previous segment is subtracted from azimuth of the next segment. Also because azimuths are the only thing that is calculated it's faster than `area` which does other things.

Benchmark results (of area, point_order and remove_spikes as a bonus, plus time taken to calculate azimuth for each segment of a ring):

Circle containing 2000001 points.

| formula | `area` | `remove_spikes` | `calculate_point_order` | azimuth for each segment |
| ----- | ----- | ----- | ----- | ----- |
| andoyer | 1.752 | 1.993 | **1.014** | 0.766 |
| thomas | 2.05 | 2.54 | **1.289** | 1.004 |
| vincenty | 2.723 | 2.858 | **1.646** | 1.184 |

Invalid polygon being one spike `(0 0, 45 45, 0 0, 45 45, ...)` containing 2000001 points.

| formula | `area` | `remove_spikes` | `calculate_point_order` | azimuth for each segment |
| ----- | ----- | ----- | ----- | ----- |
| andoyer | 1.572 | 1.192 | **0.948** | 0.725 |
| thomas | 1.871 | 1.439 | **1.205** | 0.956 |
| vincenty | 3.025 | 2.105 | **1.998** | 1.578 |

Note that `area` is faster than `remove_spikes` in the first case and slower in the second. My guess is that it has something to do with the fact that `remove_spikes` uses a vector of cleared points which is used to store points not being spikes nor duplicates and that in the second case all of the points are spike points so they are being cleared as they go and this temporary vector has at most 4 elements.

Note that in the second case the times of `remove_spikes` and `calculate_point_order` are similar. ~My guess is that it's because here the number of azimuths calculation is not 2x smaller as I stated before because if spike is detected in `calculate_point_order` the azimuths has to be recalculated for both neighboring segments which means that more azimuth calculations is performed here (Looks like 2x as many but I haven't checked)~.

Btw, I think `remove_spikes` could be slightly optimized:
- `cleared` container could contain only iterators to the original points
- the finalizing removal of spikes at the beginning and end of the container could be removed by range of iterators which means that `deque` could be replaced by `vector`.

I've done this in `calculate_point_order` and it seemed to be faster than the original. This is also probably the case why in the second benchmark `calculate_point_order` is still faster than `remove_spikes`.

But to be sure what's happening exactly more investigation would be needed.

EDIT:

I checked the numbers of azimuth calculations and the above remark about 2x as many calculations is not true. These are the numbers of inverse formula calls:

| benchmark # | `area` | `remove_spikes` | `calculate_point_order` |
| ----- | ----- | ----- | ----- |
| 1 | 2000000 | 4000002 | 2000001 |
| 2 | 2000000 | 4000002 | 2000000 |

So the small difference above is not due to the number of azimuths called but something else. Above I also added benchmark for pure azimuths calculation as a base-line. So I don't know right now why does `remove_spike` times fluctuate so much. It's very similar algorithm to `calculate_point_order`. Maybe I'll try to do the same modifications there (mainly replace deque with vector) and see what happens.

### DIFF
--- a/include/boost/geometry/strategies/spherical/point_order.hpp
+++ b/include/boost/geometry/strategies/spherical/point_order.hpp
@@ -11,10 +11,19 @@
 #define BOOST_GEOMETRY_STRATEGIES_SPHERICAL_POINT_ORDER_HPP
 
 
+#include <boost/type_traits/is_same.hpp>
+
+
 #include <boost/geometry/core/tags.hpp>
 
-#include <boost/geometry/strategies/point_order.hpp>
+#include <boost/geometry/formulas/spherical.hpp>
+
 #include <boost/geometry/strategies/spherical/area.hpp>
+#include <boost/geometry/strategies/spherical/point_in_point.hpp>
+#include <boost/geometry/strategies/point_order.hpp>
+
+#include <boost/geometry/util/math.hpp>
+#include <boost/geometry/util/select_calculation_type.hpp>
 
 
 namespace boost { namespace geometry
@@ -22,6 +31,83 @@ namespace boost { namespace geometry
 
 namespace strategy { namespace point_order
 {
+
+//template <typename CalculationType = void>
+//struct spherical
+//{
+//    typedef azimuth_tag version_tag;
+//
+//    template <typename Geometry>
+//    struct result_type
+//    {
+//        typedef typename geometry::select_calculation_type_alt
+//            <
+//                CalculationType, Geometry
+//            >::type type;
+//    };
+//
+//    template <typename Point>
+//    inline bool apply(Point const& p1, Point const& p2,
+//                      typename result_type<Point>::type & azi,
+//                      typename result_type<Point>::type & razi) const
+//    {
+//        typedef typename result_type<Point>::type calc_t;
+//
+//        if (equals_point_point(p1, p2))
+//        {
+//            return false;
+//        }
+//
+//        calc_t lon1 = geometry::get_as_radian<0>(p1);
+//        calc_t lat1 = geometry::get_as_radian<1>(p1);
+//        calc_t lon2 = geometry::get_as_radian<0>(p2);
+//        calc_t lat2 = geometry::get_as_radian<1>(p2);
+//
+//        convert_latitudes<Point>(lat1, lat2);
+//
+//        formula::result_spherical<calc_t>
+//            res = formula::spherical_azimuth<calc_t, true>(lon1, lat1, lon2, lat2);
+//
+//        azi = res.azimuth;
+//        razi = res.reverse_azimuth;
+//
+//        return true;
+//    }
+//
+//    template <typename Point>
+//    inline typename result_type<Point>::type
+//    apply(Point const& /*p0*/, Point const& /*p1*/, Point const& /*p2*/,
+//          typename result_type<Point>::type const& azi1,
+//          typename result_type<Point>::type const& azi2) const
+//    {
+//        // TODO: support poles
+//        return math::longitude_distance_signed<radian>(azi1, azi2);
+//    }
+//
+//private:
+//    template <typename Point>
+//    static bool equals_point_point(Point const& p0, Point const& p1)
+//    {
+//        return strategy::within::spherical_point_point::apply(p0, p1);
+//    }
+//
+//    template <typename Point, typename CalcT>
+//    static void convert_latitudes(CalcT & lat1, CalcT & lat2)
+//    {
+//        static const bool is_polar = boost::is_same
+//            <
+//                typename geometry::cs_tag<Point>::type,
+//                spherical_polar_tag
+//            >::value;
+//
+//        if (BOOST_GEOMETRY_CONDITION(is_polar))
+//        {
+//            CalcT pi_half = math::half_pi<CalcT>();
+//            lat1 = pi_half - lat1;
+//            lat2 = pi_half - lat2;
+//        }
+//    }
+//};
 
 template <typename CalculationType = void>
 struct spherical
@@ -39,11 +125,11 @@ struct default_strategy<spherical_equatorial_tag>
     typedef spherical<> type;
 };
 
-template <>
+/*template <>
 struct default_strategy<spherical_polar_tag>
 {
     typedef spherical<> type;
-};
+};*/
 
 } // namespace services
 
--- a/include/boost/geometry/strategies/spherical/point_order.hpp
+++ b/include/boost/geometry/strategies/spherical/point_order.hpp
@@ -1,0 +1,54 @@
+// Boost.Geometry
+
+// Copyright (c) 2019, Oracle and/or its affiliates.
+
+// Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
+
+// Licensed under the Boost Software License version 1.0.
+// http://www.boost.org/users/license.html
+
+#ifndef BOOST_GEOMETRY_STRATEGIES_SPHERICAL_POINT_ORDER_HPP
+#define BOOST_GEOMETRY_STRATEGIES_SPHERICAL_POINT_ORDER_HPP
+
+
+#include <boost/geometry/core/tags.hpp>
+
+#include <boost/geometry/strategies/point_order.hpp>
+#include <boost/geometry/strategies/spherical/area.hpp>
+
+
+namespace boost { namespace geometry
+{
+
+namespace strategy { namespace point_order
+{
+
+template <typename CalculationType = void>
+struct spherical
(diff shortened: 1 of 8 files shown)


awulkiew:
Inspired by issue found while working on reading shapefiles: https://github.com/boostorg/geometry/pull/554

awulkiew:
I modified the `remove_spikes` algorithm in similar way, i.e. replaced deque with vector and modification of container with iterators tracking front and back. This makes the algorithm ~20% faster in my cases:

Benchmark 1 (as above):

| formula | `area` | `remove_spike` develop | `remove_spike` this PR | `calculate_point_order` | azimuth for each segment |
| ----- | ----- | ----- | ----- | ----- | ----- |
| andoyer | 1.752 | 1.993 | **1.636** | **1.014** | 0.766 |
| thomas | 2.05 | 2.54 | **2.099** | **1.289** | 1.004 |
| vincenty | 2.723 | 2.858 | **2.464** | **1.646** | 1.184 |

Benchmark 2 (as above):

| formula | `area` | `remove_spike` develop | `remove_spike` this PR | `calculate_point_order` | azimuth for each segment |
| ----- | ----- | ----- | ----- | ----- | ----- |
| andoyer | 1.572 | 1.192 | **0.85** | **0.948** | 0.725 |
| thomas | 1.871 | 1.439 | **1.075** | **1.205** | 0.956 |
| vincenty | 3.025 | 2.105 | **1.691** | **1.998** | 1.578 |

Note that in the first benchmark `remove_spike` is slower than `calculate_point_order` but in the second one it's faster. It's probably due to the fact that the latter calculates, stores and checks states of azimuths in case they had to be recalculated and in benchmark 2 they have to be recalculated in all of the cases since the whole geometry is a spike. So this the worst, special case for this algorithm. In "normal" case which I consider to be most frequent one the algorithm is faster.

test/algorithms/detail/Jamfile.v2
```diff
@@ -15,8 +15,9 @@
 
 test-suite boost-geometry-algorithms-detail
     :
-    [ run as_range.cpp  : : : : algorithms_as_range ]
-    [ run partition.cpp : : : : algorithms_partition ]
+    [ run as_range.cpp              : : : : algorithms_as_range ]
+	[ run calculate_point_order.cpp : : : : algorithms_calculate_point_order ]
```
vissarion: tab

vissarion:
Thanks for the extra algorithm. Why `version_tag` is needed? AFAIU, cartesian and spherical use area computations for orientation while geographic the new algorithm based on azimuth computation. Moreover, is it possible that spherical cs could also be benefited by this new algorithm?

awulkiew:
> Why version_tag is needed?

Because otherwise it would be implemented as either of these:
1. 3 CS-dependant strategies containing the whole algorithms.
2. 2 agnostic strategies implementing the CS-agnostic part of the algorithm (one using azimuth and the other area) and 3 CS-dependant strategies. Agnostic strategies would expect certain CS-dependant strategies.

So I decided to try something new. Since the use of a specific CS-specific strategy has sense only in one algorithm or the other this information is used to dispatch the algorithm. I also think this is the correct approach in general since the Strategy in general can change the way how the geometries are processed it should be used together with geometries in dispatching of the algorithm.

Envelope strategies could benefit from it since currently they contain member functions returning either a range of points or range of segments. So with this design this could be moved from strategies to algorithms where it should be.

Furthermore, with this design we could remove the CS-agnostic strategies which AFAIU are algorithms, not strategies. Though in this case it's probably not wise due to backward compatibility reasons.

> Moreover, is it possible that spherical cs could also be benefited by this new algorithm?

Yes. I don't remember the exact reason why I didn't implement it. Probably something along the lines: if it isn't broken, don't fix it. Of course we could consider it.

awulkiew:
> Moreover, is it possible that spherical cs could also be benefited by this new algorithm?

It looks like the strategy based on area is faster: 0.282s and 0.272s vs 0.907s and 0.677s for benchmark 1 and 2 respectively.

awulkiew:
@barendgehrels What do you think about it?

This algorithm could be used to express point-order calculation clearly in algorithms like buffer or in overlay instead of directly calling area for this purpose.

awulkiew:
Today is the last call for 1.71. If we do not merge it then it will be postponed to 1.72.

include/boost/geometry/strategies/spherical/point_order.hpp
```diff
@@ -0,0 +1,140 @@
...
+//            >::value;
+//
+//        if (BOOST_GEOMETRY_CONDITION(is_polar))
+//        {
+//            CalcT pi_half = math::half_pi<CalcT>();
+//            lat1 = pi_half - lat1;
+//            lat2 = pi_half - lat2;
+//        }
+//    }
+//};
```
barendgehrels: Can you remove the commented code, or explain why it is commented?

test/algorithms/detail/calculate_point_order.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+
+inline const char * order_str(bg::order_selector o)
+{
+    if (o == bg::clockwise)
+        return "clockwise";
+    else if (o == bg::counterclockwise)
+        return "counterclockwise";
+    else
+        return "order_undetermined";
```
barendgehrels: Curly braces are missing
(alternatively you can use ternary operations)

test/algorithms/detail/calculate_point_order.cpp
```diff
@@ -0,0 +1,148 @@
...
+   test_one<P>("POLYGON((2 0, 2 1, 2 -1, 2 0, 1 0, 1 -1, 0 -1, 0 1, 1 1, 1 0, 2 0))", bg::clockwise);
+   test_one<P>("POLYGON((2 0, 2 1, 2 -1, 2 0, 1 0, 1 -1, 0 -1, 0 1, 1 1, 1 0))", bg::clockwise);
+   test_one<P>("POLYGON((2 0, 2 1, 3 1, 3 -1, 2 -1, 2 0, 1 0, 1 -1, 0 -1, 0 1, 1 1, 1 0, 2 0))", bg::clockwise);
+   test_one<P>("POLYGON((0 85, 5 85, 5 84, 0 84, 0 85))", bg::clockwise);
+   test_one<P>("POLYGON((0 2, 170 2, 170 0, 0 0, 0 2))", bg::clockwise);
+   test_one<P>("POLYGON((0 2, 170 2, 170 1, 160 1, 10 1, 0 1, 0 2))", bg::clockwise);
+   test_one<P>("POLYGON((0 2, 170 2, 170 -2, 0 -2, 0 2))", bg::clockwise);
+   test_one<P>("POLYGON((5 5, 6 5, 6 6, 6 4, 6 5, 5 5, 5 4, 4 4, 4 6, 5 6, 5 5))", bg::clockwise);
+   test_one<P>("POLYGON((5 5, 6 5, 6 6, 6 4, 6 5, 5 5, 5 6, 4 6, 4 4, 5 4, 5 5))", bg::counterclockwise);
+   test_one<P>("POLYGON((5 5, 6 5, 6 5, 6 6, 6 5, 6 4, 6 5, 6 5, 5 5, 5 4, 4 4, 4 6, 5 6, 5 5))", bg::clockwise);
```
barendgehrels: Indentation is inconsistent

barendgehrels:
Thank you, I'm OK with merging, I've a few remarks but those are minor comments.
Sorry for the delay, I had looked at it earlier but did not finish the review.
Great that the inaccuracy for geographic CS is solved, that will help future developments such as buffer.

PR #563, opened 2019-03-03
Enhance/more remove rescale preparations

barendgehrels:
- Almost all is test-only; in code it only extracts buffer_box to separate headerfile
- Changes BOOST_GEOMETRY_NO_ROBUSTNESS to BOOST_GEOMETRY_USE_RESCALING (with negated meaning). This precedes future versions where rescaling is NOT the default (or, in the end, removed)
- Checks conditionally code w.r.t. runs without rescaling (many is already fixed in earlier commits)
- Simplified conditional testcode
- Adds three testcases from reported issues

Note, this is not yet the last PR, work on intersection strategy is still to come.


include/boost/geometry/algorithms/detail/buffer/buffer_box.hpp
```diff
@@ -0,0 +1,62 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+
+// Copyright (c) 2018 Barend Gehrels, Amsterdam, the Netherlands.
```
awulkiew: 2018-2019 :)

test/algorithms/set_operations/union/union_multi.cpp
```diff
@@ -393,18 +388,10 @@ void test_areal()
...
+        settings.test_validity = BG_IF_RESCALED(false, true);
         test_one<Polygon, MultiPolygon, MultiPolygon>("ticket_9081",
             ticket_9081[0], ticket_9081[1],
-#if defined(BOOST_GEOMETRY_NO_ROBUSTNESS)
-            3,
-#else
-            4,
-#endif
-            0, 31, 0.2187385,
+            BG_IF_RESCALED(4, 3), 0, 31, 0.2187385,
```
awulkiew: So one of these numbers is not correct, right?

barendgehrels: I will merge Wednesday and check this. It might be both right (rescaling might split a polygon which would otherwise be unsplit). The reverse case we also had (should actually be split but is merged in rescaling).
Thank you.

awulkiew: No, I mean in general, in all of these cases where the result depends on rescaling. I'm just curious if you consider both results correct or not. Previously some of the tests were enabled only if the `FAILING_TESTS` flag was enabled, but now this macro is used. So it's not clear which expected result is the correct one.

barendgehrels: But also in general, as far as I'm aware and can think of now, all results are correct (unless marked as _IGNORE when invalid). Otherwise they're excluded from the suite. Or a comment explicitly notes the reason of failure.
I did not remove FAILING_TESTS flags but renamed some of them, and (when denoting invalidity or skipping a testcase) that is handled in the test-header itself now.

Anyway, there are not so much of these new IF_RESCALED flags, I just checked most of them and added some comments. In general I consider the non-rescaled versions (often more splitted) as more correct. We've to check in the next phase what the final result will be.

PR #568, opened 2019-03-12
Add a constructor that takes matrix_type to matrix_transformer<CalculationType, 3, 3>

tinko92:
The example file 06_b_transformation_example contains this useful example for composing matrix_transformers
`matrix_transformer<double, 2, 2> combined(rotate.matrix() * translate.matrix());`.

The analogue of this for matrix_transformer<CalculationType, 3, 3> causes a compiler error because the corresponding constructor is missing in the 3d case. This pull request is designed to fix this by adding the constructor 
`    inline matrix_transformer(matrix_type const& matrix)
        : m_matrix(matrix)
    {}`
to matrix_transformer<CalculationType, 3, 3>.

### DIFF
--- a/include/boost/geometry/strategies/transform/matrix_transformers.hpp
+++ b/include/boost/geometry/strategies/transform/matrix_transformers.hpp
@@ -129,6 +129,10 @@ public :
                     m_2_0, m_2_1, m_2_2)
     {}
 
+    inline matrix_transformer(matrix_type const& matrix)
+        : m_matrix(matrix)
+    {}
+
     inline matrix_transformer()
         : matrix_transformer<CalculationType, 2, 2>()
     {}
--- a/include/boost/geometry/strategies/transform/matrix_transformers.hpp
+++ b/include/boost/geometry/strategies/transform/matrix_transformers.hpp
@@ -157,6 +157,10 @@ public :
         qvm::A<3,0>(m_matrix) = m_3_0; qvm::A<3,1>(m_matrix) = m_3_1; qvm::A<3,2>(m_matrix) = m_3_2; qvm::A<3,3>(m_matrix) = m_3_3;
     }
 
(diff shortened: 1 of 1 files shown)


awulkiew:
I think it's a good idea.

If needed this could even be expanded in the future (additional set of constructors added) to allow any `Matrix` type adapted to QVM Matrix concept.

Btw, 4 commits are not needed for this but with the magic of GitHub I can squash them while merging. Is that ok or would you like to do this by yourself?

include/boost/geometry/strategies/transform/matrix_transformers.hpp
```diff
@@ -129,6 +130,10 @@ public :
                     m_2_0, m_2_1, m_2_2)
     {}
 
+    inline matrix_transformer(matrix_type const& matrix)
+        : matrix_transformer<CalculationType, 2,2>(matrix)
```
vissarion: 2,2 --> 2, 2 (with a space)

barendgehrels:
Milestone is set to 1.70 - but will it be merged? 
1.70 opens for bugfixes today. This is a minor code change, but actually not a bugfix - though you might expect that it behaves the same as its companion, and then it is a bug

awulkiew:
I'd wait and merge it after the release, so target 1.71.

tinko92:
Thank you very much for the quick feedback.

@awulkiew Sorry, for the four commits, I messed it up on the first try and only noticed later, I'll make sure to have cleaner commit histories on my PRs in the future and I'll also look into the magic of Github. 

I also agree that it could be expanded (and I'd have a use case for an expanded version of this in my code), but because these changes would be less trivial, I'll present my case for them in an Issue first.

barendgehrels:
Yep, let's wait then, updated milestone.

awulkiew:
Thanks!

PR #572, opened 2019-03-15
Fix azimuth validity in geographic_point_circle buffer strategy.

awulkiew:
This PR also adds asserts in formulas.

It addresses the issue: https://github.com/boostorg/geometry/issues/542

It also fixed some failing cases for me from algorithms_buffer_point_geo test file (tested only with msvc-14.1).

### DIFF
--- a/include/boost/geometry/formulas/differential_quantities.hpp
+++ b/include/boost/geometry/formulas/differential_quantities.hpp
@@ -11,6 +11,7 @@
 #ifndef BOOST_GEOMETRY_FORMULAS_INVERSE_DIFFERENTIAL_QUANTITIES_HPP
 #define BOOST_GEOMETRY_FORMULAS_INVERSE_DIFFERENTIAL_QUANTITIES_HPP
 
+#include <boost/geometry/core/assert.hpp>
 
 #include <boost/geometry/util/condition.hpp>
 #include <boost/geometry/util/math.hpp>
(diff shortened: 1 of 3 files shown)


include/boost/geometry/formulas/differential_quantities.hpp
```diff
@@ -73,6 +74,9 @@ class differential_quantities
             CT const sig_12 = dlon / one_minus_f;
             if (BOOST_GEOMETRY_CONDITION(EnableReducedLength))
             {
+                CT const pi = math::pi<CT>();
+                BOOST_GEOMETRY_ASSERT(-pi <= azimuth && azimuth <= pi);
```
barendgehrels: You will get a warning for an unused variable pi in release mode. Maybe just use the math::pi<CT>() inside the assertion?

include/boost/geometry/strategies/geographic/buffer_point_circle.hpp
```diff
@@ -96,6 +97,9 @@ public :
 
         for (std::size_t i = 0; i < m_count; i++, angle += diff)
         {
+            if (angle > pi)
+                angle -= two_pi;
```
barendgehrels: please use braces in every if - this is our convention

barendgehrels: Thanks for fixing this!

barendgehrels:
@awulkiew : Thanks for the changes!

PR #573, opened 2019-03-19
interpolate_point_spherical: inline var initialization

Kojoley:
Fixes msvc < 12 failures.


awulkiew:
Thanks! But this member should be removed at all and constant created in each member function using it. This is how it's done in the whole library. Are you willing to give it a go or would you prefer us to do it?

Kojoley:
I have updated the PR. If it is not what you meant I am fine with closing the PR.

awulkiew:
This is it. Thanks!

Issue #560, opened 2019-02-25
karney_direct longitude is inaccurate


kenba:
The karney_direct formula (`formulas/karney_direct.hpp`) is an implementation of Charles Karney’s GeographicLib Geodesic::Direct function. Accuracy is a key feature of the GeographicLib functions, see #449.  

I therefore expect the `karney_direct` formula to produce results: azimuth, latitude and longitude, that are
very close to those produced by the corresponding GeographicLib `Geodesic::Direct` function.

Comparing the output of the `karney_direct` formula with `Geodesic::Direct` over a grid of latitudes and longitudes, the reverse_azimuths and latitudes match the expected values quite closely but the longitudes are less accurate by over 4 orders of magnitude!, see program below.

I believe that the longitude inaccuracy may be caused by a bug in the `evaluate_coeffs_C3` function in the `util/series_expansion.hpp` file: the variable `m` is not being passed to the `math::polyval` function.

The comment above "evaluate_coeffs_C3x" states: "The C++ code ... is generated by the following Maxima script..." Therefore, this issue may be caused by Maxima or the script.

```
#include <boost/geometry/algorithms/detail/azimuth.hpp>
#include <boost/geometry/formulas/karney_direct.hpp>
#include <GeographicLib/Geodesic.hpp>
#define BOOST_TEST_MAIN
#include <boost/test/unit_test.hpp>
#include <iostream>

namespace
{
  constexpr long double wgs84_a{ 6378137.0L };
  constexpr long double wgs84_f{ 1.0L / 298.257223563L };
  constexpr long double one_minus_f{ 1.0L - wgs84_f };
  constexpr long double wgs84_b{ wgs84_a * one_minus_f };

  const boost::geometry::srs::spheroid<long double> BoostWGS84(wgs84_a, wgs84_b);

  // boost karney_direct function class with azimuth output and SeriesOrder = 6
  template <typename T>
  using BoostKarneyDirect_6 = boost::geometry::formula::karney_direct <T, true, true, false, false, 6u>;

  // boost karney_direct function class with azimuth output and SeriesOrder = 8
  template <typename T>
  using BoostKarneyDirect_8 = boost::geometry::formula::karney_direct <T, true, true, false, false, 8u>;

  // boost test BOOST_CHECK_CLOSE macro takes a percentage accuracy parameter
  const auto EPSILON(std::numeric_limits<double>::epsilon());
  const auto CALCULATION_TOLERANCE(100 * EPSILON);
}

[...]

PR #574, opened 2019-03-20
Add tests for series expansion

vissarion:
Add the tests cases from https://github.com/boostorg/geometry/issues/560


vissarion:
Note: I will squash & merge in one commit if accepted.

awulkiew:
I think this test should go into separate file.

PR #576, opened 2019-03-21
Deactivate old tests that fail or not compile in extensions

vissarion:
There are tests in extensions that use old code and does not compile or fail. I propose to deactivate for now and in the future fix them (if  they are relevant) or remove them.


barendgehrels:
OK for me.
I will fix dissolve afterwards.
The offset/selected tests are really old and not touched for years I thinhk.

Issue #579, opened 2019-03-25
Invalid result of disjoint(segment, box) in spherical_equatorial and probably geographic.


awulkiew:
When the segment is going along meridian and is pointing south, e.g.:

    using coordinate_system = boost::geometry::cs::spherical_equatorial<boost::geometry::degree>;
    using lonlat = boost::geometry::model::point<double, 2, coordinate_system>;
    using box = boost::geometry::model::box<lonlat>;
    using segment = boost::geometry::model::segment<lonlat>;

    boost::geometry::disjoint(box{ { 10, 10 },{ 20, 20 } },
                              segment{ { 12, 2 }, { 12, 1 } });

Internally in `disjoint_segment_box_sphere_or_spheroid::apply` envelope of a segment is calculated at line 162 and the result is wrong, i.e. `box{{12, 1}, {12, 90}}`.

AFAIU it's because above `alp1` is calculated and passed into the `envelope_segment_impl` for the segment pointing south (`alp1` is `pi`) and inside `envelope_segment_impl::special_cases` the latitudes are reversed because `lon1 == lon2` at line 264. Now the segment is pointing north and these reversed coordinates are passed into the azimuth strategy to calculate reverse azimuth (`alp2` is `0`). Obviously both azimuths are incompatible.

awulkiew:
Detected while working on PR: https://github.com/boostorg/geometry/issues/466
and using disjoint Segment/Box in implementation of disjoint Areal/Box.

awulkiew:
I added failing cases: https://github.com/boostorg/geometry/commit/30ebfb8cd53ce5041fc55ae5f5a23a0146a5805b

awulkiew:
In `special_cases` the coordinates can be swapped in other cases too, like `is_pole1 && is_pole2` or `lon1 > lon2`. There is no guarantee that the azimuth passed from the outside will be consistent with this function and reverse azimuth calculated afterwards. So I see 3 solutions:

1. Implement some normalization function the same for all spherical and geographic coordinates and call it at the beginning of all non-cartesian calculations in the library, before passing the coordinates to any formula. This is horribly error-prone and would cause trouble.

2. Assume that the azimuth passed from the outside is calculated for the original segment and either recalculate it when the original coordinates are modified or based on the condition modify azimuth to match the swapped coordinates. This would be done more or less half of the time (lon1 > lon2).

[...]

PR #580, opened 2019-03-26
Fix disjoint(segment, box) in spherical_equatorial and geographic CS

vissarion:
Propose a fix for https://github.com/boostorg/geometry/issues/579


include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
```diff
@@ -154,9 +154,6 @@ struct disjoint_segment_box_sphere_or_spheroid
 
         //Compute alp1 outside envelope and pass it to envelope_segment_impl
         //in order for it to be used later in the algorithm
```
awulkiew: This comment is not needed anymore.

PR #582, opened 2019-04-03
Feature/add start turn

barendgehrels:
This PR is another preparation of rescaling:

- adding start turns, which are sometimes necessary in case other turns are missed
- some more decisions based on very small nearly-collineary but not completely situations
- fixing empty output in traversal

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
@@ -14,6 +14,8 @@
 #include <boost/geometry/core/coordinate_type.hpp>
 #include <boost/geometry/util/select_coordinate_type.hpp>
 
+#include <cmath>
+
 namespace boost { namespace geometry
 {
 
@@ -24,10 +26,10 @@ namespace detail
 template <typename T>
 struct distance_measure
 {
-    T value;
+    T measure;
 
     distance_measure()
-        : value(T())
+        : measure(T())
     {}
 
     bool is_small() const { return true; }
@@ -48,7 +50,7 @@ struct distance_measure_floating
     // This is an arbitrary boundary, to enable some behaviour
     // (for example include or exclude turns), which are checked later
     // with other conditions.
-    bool is_small() const { return std::fabs(measure) < 1.0e-3; }
+    bool is_small() const { return std::abs(measure) < 1.0e-3; }
 
     // Returns true if the distance measure is absolutely zero
     bool is_zero() const { return measure == 0.0; }
@@ -76,9 +78,12 @@ struct distance_measure<float>
 namespace detail_dispatch
 {
 
-template <typename CalculationType, typename Tag>
+// TODO: this is effectively a strategy, but for internal usage.
+// It might be moved to the strategies folder.
+
+template <typename CalculationType, typename CsTag>
 struct get_distance_measure
-        : not_implemented<Tag>
+        : not_implemented<CsTag>
 {};
 
 template <typename CalculationType>
@@ -168,6 +173,7 @@ namespace detail
 // 0 (absolutely 0, not even an epsilon) means collinear. Like side,
 // a negative means that p is to the right of p1-p2. And a positive value
 // means that p is to the left of p1-p2.
+
 template <typename cs_tag, typename SegmentPoint, typename Point>
 static distance_measure<typename select_coordinate_type<SegmentPoint, Point>::type>
 get_distance_measure(SegmentPoint const& p1, SegmentPoint const& p2, Point const& p)
--- a/include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
@@ -135,33 +135,8 @@ struct get_distance_measure<CalculationType, spherical_tag>
 };
 
 template <typename CalculationType>
-struct get_distance_measure<CalculationType, spherical_equatorial_tag>
-{
-    typedef detail::distance_measure<CalculationType> result_type;
-
-    template <typename SegmentPoint, typename Point>
-    static result_type apply(SegmentPoint const& , SegmentPoint const& ,
-                             Point const& )
-    {
-        // TODO, optional
-        result_type result;
-        return result;
-    }
-};
-template <typename CalculationType>
 struct get_distance_measure<CalculationType, geographic_tag>
-{
-    typedef detail::distance_measure<CalculationType> result_type;
-
-    template <typename SegmentPoint, typename Point>
-    static result_type apply(SegmentPoint const& , SegmentPoint const& ,
-                             Point const& )
-    {
-        // TODO, optional
-        result_type result;
-        return result;
-    }
-};
+        : get_distance_measure<CalculationType, spherical_tag> {};
 
 
 } // namespace detail_dispatch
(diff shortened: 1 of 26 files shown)


include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+    distance_measure_floating()
+        : measure(T())
+    {}
+
+    // Returns true if the distance measure is small.
+    // This is an arbitrary boundary, to enable some behaviour
+    // (for example include or exclude turns), which are checked later
+    // with other conditions.
+    bool is_small() const { return std::fabs(measure) < 1.0e-3; }
```
vissarion: any clue about this number? does it come from experimentations?

include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
```diff
@@ -0,0 +1,188 @@
...
+{
+    typedef detail::distance_measure<CalculationType> result_type;
+
+    template <typename SegmentPoint, typename Point>
+    static result_type apply(SegmentPoint const& p1, SegmentPoint const& p2,
+                             Point const& p)
+    {
+        typedef CalculationType ct;
+
+        // Construct a line in general form (ax + by + c = 0),
```
vissarion: if this is only for cartesian shouldn't it be in strategies?

include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
awulkiew: Also, probably: `math::abs(measure)`.

barendgehrels: The source is in the comments, right above - it is a more or less arbitrary number.
It is a FP number so fabs should be fine - but I could make it abs too

include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
barendgehrels: I did not make it a real strategy, invokeable by users - it is a detail, so for now I kept it here. Unless you object.

vissarion: Ok, but maybe add a comment or TODO that this should be moved to strategies in the future?

include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
vissarion: Does this mean that changing this number affects the output/correctness or the performance of the algorithm?

include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
awulkiew: > I did not make it a real strategy, invokeable by users

It's not a matter of invoking by the user since it should still be derived from higher-level strategy like intersection strategy in this case.

I guess we could keep it here (at least for now) but could you at least take the coordinate system from strategy instead of geometry? You could add `typedef ... cs_tag` member type to intersection strategies for this purpose.

Actually we could do this with all strategies in order to generate CS-specific parts of the algorithm if needed instead of defining strategies within strategies.

awulkiew: But there is another thing I don't understand. AFAIU this is a utility to make sides calculation more robust, correct? consistent with distance calculation? If that's the case why is this not a part of side strategy? Or does it have a different purpose?

barendgehrels: Working at this:

> I guess we could keep it here (at least for now) but could you at least take the coordinate system from strategy instead of geometry? You could add typedef ... cs_tag member type to intersection strategies for this purpose.

But I'm a puzzled - what value does it add? I assume you find it more configurable because that tag comes from intersection strategy, and not from geometry, and they can be passed independently.

We now use the basic intersection strategy as an umbrella strategy to access all other strategies. So yes, I agree that a cs_tag could be added there. But I think it should be splitted again - one basic intersection strategy (as we had before) and a separate umbrella strategy returning you all the other strategies (area, point-in-geometry, envelope, etc) which all have nothing to do with intersection strategy itself (that, as far as I can see, needs only a side strategy as a dependency).

Anyway, that is a different subject. To make it just a bit more consistent with current implementations I will add that tag - but I'm really hesitant to add an accessor there again for this new strategy.

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -142,6 +143,51 @@ struct base_turn_handler
...
+            > dm_type;
+
+        dm_type const dm = get_distance_measure(range_q.at(index_q - 1),
+            range_q.at(index_q), range_p.at(index_p));
+
+        if (! dm.is_zero())
+        {
+            // Not truely collinear, distinguish for union/intersection
+            // If p goes left (positive), take that for a union
+            ui_else_iu(dm.is_positive(), ti);
```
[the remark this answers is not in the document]
barendgehrels: >     * Indeed, changing the value (a bit) should not matter. I will come back to this.
The value itself is indeed not sensitive, it can be changed a bit without changing behaviour. Of course if you change it completely, things solved by this feature might start to fail.

barendgehrels:
@awulkiew : I applied your remark, defining the cs_tag in the intersect_segment strategies and pass them in get_turn_info to distance_measure. Because it is called from basehandler it has to be passed several times. I now named the strategy **ubrella_strategy** which makes it in my eyes much clearer.

barendgehrels:
OK to merge this?

include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
```diff
@@ -0,0 +1,194 @@
...
+                             Point const& )
+    {
+        // TODO, optional
+        result_type result;
+        return result;
+    }
+};
+
+template <typename CalculationType>
+struct get_distance_measure<CalculationType, spherical_equatorial_tag>
```
awulkiew: This is not really needed since spherical intersection/umbrella strategy defines `spherical_tag`.

awulkiew:
Yes, sorry for late response. I'm ok with merging.

> I now named the strategy ubrella_strategy which makes it in my eyes much clearer.

Sure, it's at the same time umbrella strategy and intersection strategy. In the future we could think about implementing proper umbrella strategies. E.g. in this case all set and relational operations could take one `relate` umbrella strategy, etc.

include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
```diff
@@ -0,0 +1,194 @@
...
+    static result_type apply(SegmentPoint const& , SegmentPoint const& ,
+                             Point const& )
+    {
+        // TODO, optional
+        result_type result;
+        return result;
+    }
+};
+template <typename CalculationType>
+struct get_distance_measure<CalculationType, geographic_tag>
```
awulkiew: This specialization is the same as the one for `spherical_tag` so it could be implemented as:

    template <typename CalculationType>
    struct get_distance_measure<CalculationType, geographic_tag>
        : get_distance_measure<CalculationType, spherical_tag>
    {};

barendgehrels: OK - will do this and merge, thank you

include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
barendgehrels: OK

barendgehrels:
> Sure, it's at the same time umbrella strategy and intersection strategy. In the future we could think about implementing proper umbrella strategies. E.g. in this case all set and relational operations could take one `relate` umbrella strategy, etc.

Right, OK

PR #587, opened 2019-05-01
Bugfix/omit selfturn condition

barendgehrels:
This is a very small PR, changing only one condition (fixing a case without rescaling, all other cases stay fine) and adding a few cases based on precision research/enhancements

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/traversal.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/traversal.hpp
@@ -803,8 +803,7 @@ public :
             {
                 continue;
             }
-            if (is_self_turn<OverlayType>(cluster_turn)
-                || cluster_turn.both(target_operation))
+            if (cluster_turn.both(target_operation))
             {
                 // Not (yet) supported, can be cluster of u/u turns
                 return false;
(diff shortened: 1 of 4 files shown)


barendgehrels:
@awulkiew - how can I trigger all tests on this PR? Is that possible?
Locally I tested all set_operations and buffer, which are fine - but it would be good to test all of it.

awulkiew:
@barendgehrels What do you mean? Do you want to run all of the tests at CircleCI? If that's the case then you'd have to either remove branch filters from CircleCI jobs or add your branch there.

Filters are defined here:
https://github.com/boostorg/geometry/blob/develop/.circleci/config.yml#L487

then they are applied to separate jobs e.g. like that:
https://github.com/boostorg/geometry/blob/develop/.circleci/config.yml#L555

minimal test is run for all branches other than master and develop:
https://github.com/boostorg/geometry/blob/develop/.circleci/config.yml#L631

Have in mind that at CircleCI not all of the tests are executed, e.g. all of the extensions and some index tests are ommited.

Furthermore at CircleCI only one OS and compiler is used. If you want more comprehensive tests it's still better to run locally with several compilers or merge to develop and wait for the regression matrix update.

barendgehrels:
@awulkiew  - thanks for the info! For now I run all the tests locally

barendgehrels:
@vissarion : good points. I have a new robustness test (still to be committed) for removing rescaling, which scans input (adding very small values to some coordinates) and reports if they are failing. Very useful to find border cases.
No, in the final (ideal) case they should all pass, and there will be some thresholds but I try to limit that as much as possible.
Yes, these tests will also pass without the new change, in that sense it is unrelated. But the change is necessary for some other (yet uncommitted) code.

@awulkiew  : all tests pass, I think it can be merged.

Issue #588, opened 2019-05-07
Invalid turns generation (detected as inconsistency in set and relational operations).


awulkiew:
@barendgehrels I've noticed that some operations became inconsistent after recent changes in set operations (with VS2015, msvc-14.0, x86). Consider:
```
red = POLYGON((5 0,7 10,0 15,10 15,15 25,20 15,30 15,22 10,25 0,15 5,5 0))
green = POLYGON((15 0,17 10,10 15,20 15,25 25,30 15,40 15,32 10,35 0,25 5,15 0))
blue = MULTIPOLYGON(((10 15,15 25,20 15,10 15)),((10 15,17 10,15.90909090909091 4.545454545454546,15 5,5 0,7 10,0 15,10 15)),((23.69565217391304 4.347826086956522,25 0,20 2.5,23.69565217391304 4.347826086956522)))
```
![image](https://user-images.githubusercontent.com/1226951/57336100-0ef24d80-7125-11e9-9b72-2c2fd9c4d64e.png)
where:
```
bg::difference(red, green, blue)
```
`blue` should therefore be within `red`:
```
assert(bg::within(blue, red) == true)
```
but it is not. Here it is how it looks like with turns calculated internally in `within` or `relation`:

![image](https://user-images.githubusercontent.com/1226951/57336227-75776b80-7125-11e9-82e7-8b30fe93f071.png)

The lower right `blue` triangle has the following points:
```
{23.695652173913043, 4.3478260869565215}
{25.000000000000000, 0.00000000000000000}
{20.000000000000000, 2.5000000000000000}
{23.695652173913043, 4.3478260869565215}
```
the lower right arm of the `red` star is formed from the points:
```
{22.000000000000000, 10.000000000000000}
{25.000000000000000, 0.00000000000000000}
{15.000000000000000, 5.0000000000000000}
```
and the corresponding turns are:
```
0: touch u/i at {25.000000000000000, 0.00000000000000000}
1: crosses i/u at {20.000001500000678, 2.5000007500003396}
2: crosses u/i at {23.695651652173520, 4.3478258260867602}
```
(note: turns indexes are different than in `within`)

[...]

PR #589, opened 2019-05-15
[TURNS] Disable start-turns, and take the closer point 

barendgehrels:
Closer point to measure distance, which makes that measurement more precise.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/enrich_intersection_points.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/enrich_intersection_points.hpp
@@ -428,7 +428,9 @@ inline void enrich_intersection_points(Turns& turns,
 
     // As long as turn indexes are not used yet, turns might be erased from
     // the vector
-    detail::overlay::erase_colocated_start_turns(turns, geometry1, geometry2);
+    // For now start turns are disabled.
+    // TODO: remove code or fix inconsistencies within validity and relations
+    // detail::overlay::erase_colocated_start_turns(turns, geometry1, geometry2);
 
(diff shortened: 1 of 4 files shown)


awulkiew:
Is this PR needed in your future work on removing rescaling or is the intention fixing the issue https://github.com/boostorg/geometry/issues/588? This PR does not fix it, the reason is something else (see in the issue). If this change is not needed and it interferes with your work I suggest to discard it.

barendgehrels:
@awulkiew  thanks for your investigations.
Though this PR does not fix the issue you reported, I'm still glad you mentioned it because it **does** solve issues in validity, when rescaling is turned off.
Because of that, and because of all issues are now fixed in another way, I currently don't need start-turns and would prefer if they are not necessary. Code wise (less code) and because of what you said, if I disable them in validity, the turns become context-dependent.

So I'm in favour of keeping this PR - but if you want to postpone it, so instead of disabling the code, removing it at all later (or keeping it in the end), that is also fine for me.

awulkiew:
>  but if you want to postpone it

If you prefer this solution than the other I'm ok with that. Less code is better than more code, etc.

vissarion:
Thanks. I am also ok if you want to merge this PR.  

However, I do not see the less code argument. Is this a result of this PR or enables you to remove unnecessary code?

barendgehrels:
About less code: it is indirectly a result of this PR because here we just disable start turns (without throwing away the code) and later I can either throw it away (preferred) or, if unfortunately necessary, re-enable it again but then I've also to distinguish between use cases (overlay-validation) without creating inconsistencies.

Thanks, I will merge it.

PR #591, opened 2019-05-21
Fix call to pj_qsfn in cea

vissarion:
Currently the code does not compile if `T` is different than `double`.


awulkiew:
Thanks!

Issue #592, opened 2019-05-22
Test defines


barendgehrels:
Hi,

I've a small issue, we currently use 5 different defines to enable failing tests:

```
#ifdef BOOST_GEOMETRY_TEST_ENABLE_FAILING
#ifdef BOOST_GEOMETRY_TEST_INCLUDE_FAILING_TESTS
#ifdef BOOST_GEOMETRY_ENABLE_FAILING_TESTS
#ifdef BOOST_GEOMETRY_INCLUDE_FAILING_TESTS
#ifdef GEOMETRY_TEST_INCLUDE_FAILING_TESTS
```

Because for rescaling I'm busy with tests, and also adding/removing these every now and then, I would like to see them harmonized and can do that.

I suppose it should now start with BOOST_GEOMETRY_TEST (because many others do so too, such as BOOST_GEOMETRY_TEST_PERMUTATIONS).

Then **BOOST_GEOMETRY_TEST_INCLUDE_FAILING_TESTS** is the most clear to me. Though a bit long, and TEST is duplicate.

I propose to change all into **BOOST_GEOMETRY_TEST_INCLUDE_FAILURES** which is clear and concise.

Will I change that and include that in my next PR?

awulkiew:
Good idea! 

Regarding the name. "Including" in c++ means something different. ENABLE/DISABLE is clear. Also removing the word would be fine too: BG_TEST_FAILING or BG_TEST_FAILURES.

barendgehrels:
Closing issue, it is merged already a month ago

PR #594, opened 2019-05-29
Test/changes in defines

barendgehrels:
As discussed before in issue #592 - I used Adam's suggestion, BOOST_GEOMETRY_TEST_FAILURES
And there are a few things more in this PR


include/boost/geometry/core/config.hpp
```diff
@@ -17,6 +18,8 @@
 #define BOOST_GEOMETRY_CXX11_ARRAY_UNIFIED_INITIALIZATION
 #endif
 
+#define BOOST_GEOMETRY_USE_KRAMER_RULE
```
mloskot: I wonder, is this a public macro that a user is supposed to (un)define to tweak the behaviour here?

If public, is this documented anywhere?

vissarion: It seems that this macro is not used any more and used once to bypass the validity checks on some test cases in union, but @barendgehrels should be the right person to answer.

barendgehrels: Sorry for the delayed answer.

The macro is transitional, in the time to remove the rescaling. It selects or deselects an algorithm. It is not meant to the public (though they might use it). I will come back to it later.

For now I need it because some testcases depend on its value. Therefore it is part of this PR.

I will write a short comment.

mloskot: @barendgehrels Thanks for the explanation.

PR #605, opened 2019-06-24
Define CS-specific behavior of algorithms only with strategy (part 2).

awulkiew:
Followup of https://github.com/boostorg/geometry/pull/533

The changes:
- rtree may take a strategy as `bgi::parameters<RtreeParams, Strategy>`, not all possible operations are supported, only those required by other algorithms where the rtree is used, for other operations the default strategy is used as it was before. This needs further changes but I think it needs more general changes in strategies, i.e. introduction of "proper" umbrella strategies.
- rescale_policy_type takes optional CSTag (previously CS-related behavior was defined by Geometry type).
- the information about the CS taken from the strategy is propagated internally in various algorithms.
- Point types are now optional in point_in_poly strategies (left as template arguments defaulted to void for backward compatibility).
- getters for CS-specific strategies added to some strategies.


include/boost/geometry/algorithms/detail/is_valid/polygon.hpp
```diff
@@ -453,7 +476,7 @@ class is_valid_polygon
         // compute turns and check if all are acceptable
         debug_phase::apply(3);
 
-        typedef has_valid_self_turns<Polygon> has_valid_turns;
+        typedef has_valid_self_turns<Polygon, typename Strategy::cs_tag> has_valid_turns;
```
vissarion: why not `typedef`ed as above or not `typedef` the above to look alike. Similar cases below. Am I missing something?

awulkiew: No, there is no specific cause. I think I prefer removing `typedef` above because it's used only in one place.

barendgehrels: Thanks! It is huge, and I could not check all details, but what I've seen looks to me. So I'm OK with merging this.

awulkiew: Thanks, I think I'll also add some tests for geometries with undefined CS before merging.

awulkiew:
I addressed the issue raised by @vissarion, resolved the conflict with develop, added tests and fixed one more issue, i.e. passed the strategy into `get_rescale_policy()` which calls `envelope()` and `expand()`. I plan to merge it this weekend.

Issue #596, opened 2019-06-20
boost::geometry::buffer generates (multi)polygon with spike


meastp:
The following test case generates a (multi)polygon with a spike in Boost.Geometry v 1.70

Changing from `boost::geometry::strategy::buffer::join_miter` to `boost::geometry::strategy::buffer::join_round` resolves the issue (no spike). However, as I think no spikes should be produced, I created this issue.

```
#include <iostream>

#include <boost/geometry.hpp>
#include <boost/geometry/geometries/geometries.hpp>

#include <boost/geometry/io/svg/write_svg.hpp>
#include <iostream>
#include <fstream>

int main()
{
	typedef boost::geometry::model::point<double, 2, boost::geometry::cs::cartesian> point_t;
	typedef boost::geometry::model::linestring<point_t> linestring_t;
	linestring_t ls;

	ls.push_back(point_t{ 292979.660, 6688731.370 });
	ls.push_back(point_t{ 292979.600, 6688733.420 });
	ls.push_back(point_t{ 292979.540, 6688735.440 });
	ls.push_back(point_t{ 292979.540, 6688735.500 });

	typedef boost::geometry::model::polygon<point_t> polygon_t;
	typedef boost::geometry::model::multi_polygon<polygon_t> mpolygon_t;
	mpolygon_t _buffer1;

	auto dTol = 0.015;
	const int nNumP_Circle = 12;
	boost::geometry::strategy::buffer::distance_symmetric<double> distance_strategy(dTol);
	boost::geometry::strategy::buffer::join_miter join_strategy;
	boost::geometry::strategy::buffer::end_round end_strategy(nNumP_Circle);
	boost::geometry::strategy::buffer::side_straight side_strategy;
	boost::geometry::strategy::buffer::point_circle point_strategy;

	boost::geometry::buffer(ls, _buffer1, distance_strategy, side_strategy, join_strategy, end_strategy, point_strategy);

	std::ofstream svg("outfile.svg");

	std::ostringstream out;
	out << std::fixed << std::setprecision(6) << boost::geometry::wkt(_buffer1);

	auto str = out.str();
}
```

str is:
```
[...]

PR #606, opened 2019-06-26
Fix of issue #596 - [buffer] fix line/line intersection point calculation for some cases,

barendgehrels:
now using general form (which will be used later more widely). This fixes
some numerical issues.
Including testcase, reported as #596

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/line_line_intersection.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/line_line_intersection.hpp
@@ -1,6 +1,6 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
-// Copyright (c) 2012-2014 Barend Gehrels, Amsterdam, the Netherlands.
+// Copyright (c) 2012-2019 Barend Gehrels, Amsterdam, the Netherlands.
 
 // Use, modification and distribution is subject to the Boost Software License,
 // Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
@@ -10,7 +10,6 @@
 #define BOOST_GEOMETRY_ALGORITHMS_DETAIL_BUFFER_LINE_LINE_INTERSECTION_HPP
 
 
-#include <boost/geometry/arithmetic/determinant.hpp>
 #include <boost/geometry/util/math.hpp>
 #include <boost/geometry/strategies/buffer.hpp>
 #include <boost/geometry/algorithms/detail/buffer/parallel_continue.hpp>
@@ -35,14 +34,19 @@ struct line_line_intersection
     static inline strategy::buffer::join_selector apply(Point const& pi, Point const& pj,
         Point const& qi, Point const& qj, Point& ip)
     {
-        // See http://mathworld.wolfram.com/Line-LineIntersection.html
-        typedef typename coordinate_type<Point>::type coordinate_type;
+        typedef typename coordinate_type<Point>::type ct;
 
-        coordinate_type const denominator
-            = determinant<coordinate_type>(get<0>(pi) - get<0>(pj),
-                get<1>(pi) - get<1>(pj),
-                get<0>(qi) - get<0>(qj),
-                get<1>(qi) - get<1>(qj));
+        // Construct lines in general form (ax + by + c = 0),
+        // (will be replaced by a general_form structure in a next PR)
+        ct const pa = get<1>(pi) - get<1>(pj);
+        ct const pb = get<0>(pj) - get<0>(pi);
+        ct const pc = -pa * get<0>(pi) - pb * get<1>(pi);
+
+        ct const qa = get<1>(qi) - get<1>(qj);
+        ct const qb = get<0>(qj) - get<0>(qi);
+        ct const qc = -qa * get<0>(qi) - qb * get<1>(qi);
+
+        ct const denominator = pb * qa - pa * qb;
 
         // Even if the corner was checked before (so it is convex now), that
         // was done on the original geometry. This function runs on the buffered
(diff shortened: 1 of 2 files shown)


include/boost/geometry/algorithms/detail/buffer/line_line_intersection.hpp
```diff
@@ -51,27 +55,20 @@ struct line_line_intersection
         // to check it again.
 
         // For round joins, it will not be used at all.
-        // For miter joints, there is a miter limit
+        // For miter joins, there is a miter limit
         // If segments are parallel/collinear we must be distinguish two cases:
         // they continue each other, or they form a spike
-        if (math::equals(denominator, coordinate_type()))
+        ct const zero = ct();
```
awulkiew: Why not simply `zero = 0` ?

barendgehrels: Doesn't necessarily work for user defined types.
This is equivalent to current code.
Thanks for your quick approvals.

awulkiew: > Doesn't necessarily work for user defined types.

What exactly do you have in mind? Do you have a specific example?

While it would indeed not work for types to which you couldn't implicitly convert integers, we assign many other numbers in the whole library, like `1`. So the requirement is here already, user-defined types has to allow implicit initialization from integer (and double in some cases).

Another possibility is that a const object cannot be created with a default ctor. But we're not talking about such case (`ct zero;`) but about explicitly assigning an integer or double (`ct zero = 0;`). Which is more readable.

> This is equivalent to current code.

Zero-initialization only works for POD types (https://en.cppreference.com/w/cpp/language/zero_initialization). In case of classes (type with defined constructor) the behavior is defined in the default constructor. It may as well be that the type is in uninitialized state. Why are we assuming it will be initialized to 0? So no, for user defined types this is not equivalent.

barendgehrels: We did it like this from the beginning. For example here:
```promoted_type const zero = promoted_type();``` in side_by_triangle
```T const zero = T();``` in side_of_intersection
and on various other places. Just looked, indeed we don't do it everywhere. So in that sense I could change it back.
Your argument for user-defined-types is a bit convoluted I think. UD types might also not support assignment and then initialization with 0 does not work. And that is (as far as I remember) the reason that we do it like it is (or was).

Yes, we assign many other numbers in the library too, that is true. So (I tested mainly with ttmath) the types will always support assignment (or conversion) from numeric types.

Anyway, it is equivelent to the code which I replaced, I actually meant that by my statement. Not to all current code. I could have left it there (but I find this more readable).

Comment (awulkiew) [reply include/boost/geometry/algorithms/detail/buffer/line_line_intersection.hpp]: > We did it like this from the beginning.

Yes, I know, so my question is: why? What specific case someone wanted to solve with this? Do you remember?

Is it possible that this was about some internally used simple type like `segment_ratio` which we had control on and the person adding it didn't want to implement the assignment from double because it would take a lot of time and didn't want to implement a struct template specializing for this specific type in this specific place and instead someone suggested that zero-initialization would work the same way as long as the default ctor in `ratio` type mimicked it? And since this type would be used only in some places it was needed only in a few strategies?

> UD types might also not support assignment and then initialization with 0 does not work.

Ok, I'll try to be more clear. I wrote about assignment from int/double it in my first point so I'm aware of that. The assignment has to work, this is our requirement. And in addition to this in some places we use zero/value-initialization which may call default ctor working in a different way than assignment. So for UDTs instead of 1 requirement we have 2 requirements:
- assignment from integer/double,
- default-ctor initializing to 0.

> And that is (as far as I remember) the reason that we do it like it is (or was).

I doubt it was for UDTs in general, unless in the very early version of the whole library there wasn't any initialization other than with 0. If you think the initialization is really an issue we should rather have special traits for initialization of UDTs. But before doing that I'd like to see a specific case or a user complain.

> I tested mainly with ttmath

Yes, but this was surely not for `ttmath` because:
- ttmath::Big defines ctors taking int and double (https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/extensions/contrib/ttmath/ttmathbig.h#L3139) and assignment operators (https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/extensions/contrib/ttmath/ttmathbig.h#L3111)
- we have modified version of ttmath anyway, probably for math constants or some operations like `math::sqrt` so we could modify the assignment if it was needed.

But wait a minute, since the assignments are implemented, why do we have the whole library?
Why not simply have traits outside ttmath?

It seems that ttmath is confused about the initialization ot 0 in default constructor:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/extensions/contrib/ttmath/ttmathbig.h#L3617
and it depends on the macro definition!

And in the first version it was not there:
https://github.com/boostorg/geometry/blob/7b1b14a5a073c6dcae9237ac75bc2f1ea59ae168/include/boost/geometry/extensions/contrib/ttmath/ttmathbig.h#L3006

Does someone had to modify the default constructor in order for it to work with the library? So exactly the opposite for what you're saying.

So now I'm confused, why do we zero/value-initialize in the first place if not for ttmath and ttmath was modified to accomodate for it?

Comment (barendgehrels) [reply include/boost/geometry/algorithms/detail/buffer/line_line_intersection.hpp]: Thanks for your explanations.

No, I don't remember exactly the moment when we started it (it was somewhere in the beginning). I think it must still be somewhere in a really old mailing list, but you have to go back to somewhere between 2008 and 2011 I think. It was common practice at that time.

I think I was the one changing ttmath (it is long ago that I touched it or compiled with it).

But it was not only for ttmath - we tested more types. Yes, it is a nuisance indeed to touch external libraries.

Code like this is or was necessary to use UD types, because often you cannot just use the values:
```
CT const delta_sigma = B * sin_sigma * (cos_2sigma_m
  + (B/CT(4)) * ( cos_sigma * (CT(-1) + CT(2)*cos_2sigma_m_sqr)
  - (B/CT(6) * cos_2sigma_m * (CT(-3)+CT(4)*sin_sigma_sqr) 
        * (CT(-3)+CT(4)*cos_2sigma_m_sqr)) )); // (6)
```

Anyway, I'm OK with changing behaviour of course, and ttmath is so long not tested that most code will not work for it anymore (or probably for other UDT)

But for pragmatic reasons maybe we should do that if we go to C++11 (or 14) because it makes no sense to change things now and then again, because in C++11 or higher it is usually done differently.

For now I think my change is acceptable because it was effectively already there in exactly the same way.

awulkiew: I checked the original code (https://www.ttmath.org/) and the macro check in the default ctor is there as well so it was defined by the original author. Still the point remains, it is possible that UDT may initialize variable in the default ctor to a state different than value `0`.

Btw, when you're testing do you have to define `TTMATH_BIG_DEFAULT_CLEAR` otherwise it fails?

E.g. `IsZero()` used in comparison (https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/extensions/contrib/ttmath/ttmathbig.h#L642) only checks the` TTMATH_BIG_ZERO` and in the default ctor if `TTMATH_BIG_DEFAULT_CLEAR` is not defined only `TTMATH_BIG_NAN` flag is set.

awulkiew: > But for pragmatic reasons

Sure.

> For now I think my change is acceptable because it was effectively already there in exactly the same way.

Yes, thanks!

awulkiew: > But it was not only for ttmath - we tested more types

Right, we should probably do this again. At least pick some 2 most frequently used libraries and/or Boost.Multiprecision. @vissarion sometimes uses such library for accuracy testing. But I agree, not now.

vissarion: >For now I think my change is acceptable because it was effectively already there in exactly the same way.

+1

>Right, we should probably do this again. At least pick some 2 most frequently used libraries and/or Boost.Multiprecision. @vissarion sometimes uses such library for accuracy testing. 

I was trying `ttmath` and `gmp` few years ago but with issues, I do not remember the details now but I could compile in most of the cases. Sure it would be useful to try it in the near future!

barendgehrels: Thanks for your thoughts and approvals. It is merged now, will close the issues.

PR #608, opened 2019-07-02
Remove unused parameters

vissarion:
Remove unused parameters in some algorithms to avoid compilation warnings.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
@@ -160,8 +160,7 @@ struct base_turn_handler
             std::size_t index_p, std::size_t index_q,
             TurnInfo& ti)
     {
-        boost::ignore_unused(range_p);
-        boost::ignore_unused(range_q);
+        boost::ignore_unused(range_p, range_q);
         BOOST_GEOMETRY_ASSERT(IndexP + IndexQ == 1);
         BOOST_GEOMETRY_ASSERT(index_p > 0 && index_p <= 2);
         BOOST_GEOMETRY_ASSERT(index_q > 0 && index_q <= 2);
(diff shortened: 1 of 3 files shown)


include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -153,9 +153,18 @@ struct base_turn_handler
...
+    static inline void both_collinear(
+            UniqueSubRange1 const&
+#if ! defined(BOOST_GEOMETRY_USE_RESCALING)
+            range_p
+#endif
+            ,
+            UniqueSubRange2 const&
+#if ! defined(BOOST_GEOMETRY_USE_RESCALING)
+            range_q
+#endif
```
awulkiew: I think I'd prefer passing them to `ignore_unused` to avoid ifdefs distorting reading of the code.

vissarion: Thanks, I agree.

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -153,12 +153,15 @@ struct base_turn_handler
...
+    static inline void both_collinear(
+            UniqueSubRange1 const& range_p,
             UniqueSubRange2 const& range_q,
-            UmbrellaStrategy const& umbrella_strategy,
+            UmbrellaStrategy const&,
             std::size_t index_p, std::size_t index_q,
             TurnInfo& ti)
     {
+        boost::ignore_unused(range_p);
+        boost::ignore_unused(range_q);
```
awulkiew: Btw, you could pass multiple arguments so this could be one line, but it's ok as it is.

awulkiew:
Btw, you could've amended the last commit and then force pushed instead of creating another one. ;)

vissarion:
I know, but my intention is to use github's squash but it didn't worked as expected.

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
barendgehrels: Thank you, I agree with that too.
Furthermore I'm OK, thanks for the PR

PR #609, opened 2019-07-03
Add infinite line (currently in 2D, implemented with the "general form")

barendgehrels:
This pull request adds an **infinite line** in general form. It is not (yet) a concept, but its code was already used in two places, and it will also be used in the rescaling replacement.

A third place (direction_code_ used some other implementation but that is now replaced by this general form too. This fixes a small bug where lines were exactly vertical.

In a later phase we might change it and make it concept based, such that we have the intersection between two infinite lines.

It is Cartesian. In Spherical or Geographical, it might maybe be represented by a Great Circle or Ellipse.

Including unit test.

It also removes some debug code only for unit tests in direction_code (removal was already marked with a TODO)


awulkiew:
My understanding is that this is a utility which will be used only in case of cartesian CS, e.g. in strategies. Should it be a strategy? Are there equivalent concepts in other coordinate systems we'd like to use?

Btw, another place where it could be put is util/math.

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
```diff
@@ -175,6 +176,16 @@ struct buffer_range
         }
     }
 
+    static inline bool similar_direction(output_point_type const& p0,
+            output_point_type const& p1,
+            output_point_type const& p2)
+    {
+        typedef arithmetic::general_form<coordinate_type> gf;
+        gf const p = arithmetic::construct_line<coordinate_type>(p0, p1);
+        gf const q = arithmetic::construct_line<coordinate_type>(p1, p2);
```
awulkiew: Why not simply use a constructor?

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
```diff
@@ -175,6 +176,16 @@ struct buffer_range
         }
     }
 
+    static inline bool similar_direction(output_point_type const& p0,
+            output_point_type const& p1,
+            output_point_type const& p2)
+    {
+        typedef arithmetic::general_form<coordinate_type> gf;
```
awulkiew: I suggest to change the name because in general context it's not clear general form of what is this. I'd rather explicitly name it `line_2d`, `general_2d_line_form` or something similar. Or even more explicitly `cartesian_line_2d`. Because this is only 2d and only cartesian. It's not really that much "general". ;)

include/boost/geometry/arithmetic/general_form.hpp
```diff
@@ -0,0 +1,161 @@
...
+// Structure containing the General Form of a line, a*x + b*y + c == 0
+// Might be conceptized later. Therefore operations are implemented outside
+// the structure itself.
+template <typename GeneralType = double>
+struct general_form
+{
+    general_form()
+        : a(GeneralType())
+        , b(GeneralType())
+        , c(GeneralType())
```
awulkiew: Why not `a(0)`? :) What is our decision about zero-initialization regarding new code? Because we agreed to leave the old code as it is but what about newly added one?

include/boost/geometry/arithmetic/general_form.hpp
```diff
@@ -0,0 +1,161 @@
...
+        , normalized(false)
+    {}
+
+    // Horizontal: a == 0, for example y-3=0, y==3
+    // Vertical: b == 0, for example x-2=0, x==2
+    // Through origin: c == 0
+    GeneralType a;
+    GeneralType b;
+    GeneralType c;
+    bool normalized;
```
awulkiew: Do we only need 2d version?

include/boost/geometry/arithmetic/general_form.hpp
```diff
@@ -0,0 +1,161 @@
...
+inline
+general_form<GeneralType> construct_line(Coordinate const& x1,
+    Coordinate const& y1, Coordinate const& x2, Coordinate const& y2)
+{
+    general_form<GeneralType> result;
+    result.a = y1 - y2;
+    result.b = x2 - x1;
+    result.c = -result.a * x1 - result.b * y1;
+    return result;
+}
```
awulkiew: If we had n-dimensional version this would probably have to be reimplemented.

include/boost/geometry/arithmetic/general_form.hpp
```diff
@@ -0,0 +1,161 @@
...
+    GeneralType a;
+    GeneralType b;
+    GeneralType c;
+    bool normalized;
+};
+
+template <typename GeneralType, typename Coordinate>
+inline
+general_form<GeneralType> construct_line(Coordinate const& x1,
+    Coordinate const& y1, Coordinate const& x2, Coordinate const& y2)
```
awulkiew: Is this the only kind of line which could be constructed from 4 numbers? This could either be the constructor of `general_line` type (or whatever other name it'd have) or this function could take the `general_line` as an argument in case you have more line types that could be constructed in the future.

include/boost/geometry/arithmetic/general_form.hpp
```diff
@@ -0,0 +1,161 @@
...
+                             geometry::get<1, 1>(segment));
+}
+
+
+// Calculates intersection point of two infinite lines.
+// Returns true if the lines intersect.
+// Returns false if lines are parallel (or collinear, possibly opposite)
+template <typename Point, typename GeneralType>
+inline
+bool get_intersection(Point& ip,
```
awulkiew: Typically then something is called `get_XXX` then it returns `XXX`. It is also a mixture of `intersects` and intersection (thought in this case it should take the point as the last output argument). But at the same time it has different arguments than these functions. I think I'd probably call it `intersection()` but put the `Point` as the last argument as it is the case with `bg::intersection()`.

include/boost/geometry/arithmetic/general_form.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+//! Return a distance-side-measure for a point to a line
+//! Point is located left of the line if value is positive,
+//! right of the line is value is negative, and on the line if the value
+//! is exactly zero
+template <typename GeneralType, typename CoordinateType>
+inline
+typename select_most_precise<GeneralType, CoordinateType>::type
+side_value(general_form<GeneralType> const& f,
+    CoordinateType const& x, CoordinateType const& y)
```
awulkiew: Right, so this is another 2d-only function, but here it's ok since it takes `general_form` so it would be possible to implement n-d version by overloading it.

include/boost/geometry/arithmetic/general_form.hpp
```diff
@@ -0,0 +1,161 @@
...
+{
+    // https://en.wikipedia.org/wiki/Distance_from_a_point_to_a_line#Line_defined_by_an_equation
+    // Distance from point to line in general form is given as:
+    // (a * x + b * y + c) / sqrt(a * a + b * b);
+    // In most use cases comparisons are enough, saving the sqrt
+    // and often even the division.
+    // Also, this gives positive values for points left to the line,
+    // and negative values for points right to the line.
+    return f.a * x + f.b * y + f.c;
```
awulkiew: I wonder how good numerically this new solution is. Does it really behave better in various cases than what we have now or does it simply work for some specific cases we test right now but can theoretically make it worse in other cases?

include/boost/geometry/arithmetic/general_form.hpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// Returns true for two lines which are supposed to be (close to) collinear
+// (which is not checked) and have a similar direction
+// (in practice up to 45 degrees, TO BE VERIFIED)
+// true: -----------------> p -----------------> q
+// false: -----------------> p <----------------- q
+template <typename GeneralType>
+inline
+bool similar_direction(const general_form<GeneralType>& p,
```
awulkiew: So the question is, what does "similar" mean? "Maybe up to 45 degrees"? I think the name doesn't express this well enough. Besides does such non-well-defined function is needed? If in some algorithm a check like: "angle between lines < 90 deg or < 45 deg" then this function should be named specifically expressing the desired condition and the condition in the function should check exactly what it is supposed to check.

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
barendgehrels: Hi @awulkiew @vissarion thanks for all your comments.
It is called general because it implements the so-called "general form" (otherwise it would be "generic").
For the rest I come back to all your remarks, hopefully next Sunday

awulkiew: @barendgehrels my point is that it is a "general form" of something specific, i.e. equation of 2d cartesian line. I don't know if "general form" is only used WRT line equations but when I was searching the web I found that this name was always put in quotes so it seems to be rather colloquial term. And I'd assume that many things can have "general form" whatever that may be.

vissarion:
>My understanding is that this is a utility which will be used only in case of cartesian CS, e.g. in strategies. Should it be a strategy? Are there equivalent concepts in other coordinate systems we'd like to use?

I guess this reduces to whether we need rescaling for other coordinate systems, since as far as I understand this utility is for rescaling.

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
barendgehrels: Right, sorry for the slight delay. I'm OK with a name change, and with a move.

Note that this class is more useful than just a utility. It implements an infinite line, like we have a point, segment, linestring, and this is a (non OGC) infinite line. Currently in 2d, yes. Over the years there have been questions about this (calculate intersection between two infinite lines) and this code already implements that.

So in that sense it could be moved to the geometry/geometries folder.

But I did not make it yet as generic as all the other stuff, so it is not yet concept based.

Anyway, this is the reason I kept the class very concise and did not put a normal constructor there.

More answers to come. I will also update the description.

include/boost/geometry/arithmetic/general_form.hpp
barendgehrels: Will update. I wrote this already for the recent discussion.

include/boost/geometry/arithmetic/general_form.hpp
barendgehrels: For now yes. The form which I called "general form" in 3D is a plane. Did not dive into that. But of course infinite lines in 3D do exist.

include/boost/geometry/arithmetic/general_form.hpp
barendgehrels: I don't think that is a problem.

include/boost/geometry/arithmetic/general_form.hpp
barendgehrels: Most code we have is for 2d

include/boost/geometry/arithmetic/general_form.hpp
barendgehrels: I'm OK with changing the name and putting the point as last, sorry about that.

BTW: if the infinite line will indeed be one of our concepts, the general intersection function should be able to find its intersection

include/boost/geometry/arithmetic/general_form.hpp
barendgehrels: It behaves numerically better than the Kramer's rule (that we currently had). Therefore I replaced it in buffer and with that fixed a bug. I will come back later with more info. I'm testing it quite intensively.

include/boost/geometry/arithmetic/general_form.hpp
barendgehrels: The documentation states: 
```
Returns true for two lines which are supposed to be (close to) collinear
(which is not checked) and have a similar direction
```
We need a function which returns true of collinear lines that are opposite, and false if they have the same direction. This simple function does the trick. But what is collinear - that is based on numerical precision. The result should be reliable with all the epsilons we use. So this function does not do any check for collinearity, and does not use any epsilon. Its implementation is straightforward, and (basically as a side effect) it returns true for lines even 30 degrees apart. Which indeed have a `similar` direction (therefore I still like the name, it should not be too precise)

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
barendgehrels: See my other answer, it is concept based. I can rename it to `make` which is more common.

include/boost/geometry/arithmetic/general_form.hpp
awulkiew: > We need a function which returns true of collinear lines that are opposite, and false if they have the same direction. 

So the precondition is that the arguments are two collinear lines (so this is checked before somewhere) and then they either may be going into the same direction or the opposite and you have to check that, correct? If that's correct then this function has strictly defined purpose. Not to check similarity but to check the direction of two collinear lines. So why not name it to something like: `same_collinear_direction` or since in the code we call this being opposite why not simply: `are_opposite` or `check_opposite`?

include/boost/geometry/arithmetic/general_form.hpp
barendgehrels: The calculations between Cramer's rule are actually similar to this code. So it should be numerical equivalent. It is more the way how things are expressed, which is now simpler (and that was actually fixing that bug). Besides that, in a future, this line parameters might be stored in a prepared geometry, improving performance

include/boost/geometry/arithmetic/general_form.hpp
barendgehrels: I don't know why `same_collinear_direction` is more clear than `similar_direction`. With `are_opposite` I have the objection that it does not check collinearity itself. 
But I'm not opposed to renaming ;-)

barendgehrels:
I reworked the code, it is now splitted into an `infinite_line` (class and file), `make` ( to construct it) and `infinite_line_functions` (I did not want to split all these in 4 mini files). I think it is better than it was, thanks for your comments. Please let me know your new opinions.

include/boost/geometry/arithmetic/general_form.hpp
awulkiew: I agree, the names I proposed are also not perfect. If you think of anything better then rename it. If not I'm ok with `similar_direction`.

awulkiew:
Thanks. I'm ok as well.

include/boost/geometry/arithmetic/general_form.hpp
barendgehrels: Thank you, for now I will merge it - maybe we have more ideas later.

PR #613, opened 2019-07-17
Feature/changes for selfturns and tests

barendgehrels:
Some minor changes in code, more changes in some tests


test/algorithms/check_validity.hpp
```diff
@@ -0,0 +1,79 @@
+// Boost.Geometry
+
+// Copyright (c) 2017 Barend Gehrels, Amsterdam, the Netherlands.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_TEST_CHECK_VALIDITY_HPP
+#define BOOST_GEOMETRY_TEST_CHECK_VALIDITY_HPP
```
barendgehrels: Note, this file is moved, it is adapted but not new

barendgehrels:
> Sorry for late review. It looks OK to me. Thanks!

Thank you!

barendgehrels:
@awulkiew is it OK if I merge this? My next PR is dependent on this one.

extensions/test/algorithms/dissolve.cpp
```diff
@@ -652,7 +651,9 @@ void test_all(ut_settings const& settings_for_sensitive_cases)
     TEST_DISSOLVE_WITH(dissolve_reallife, 91756.916526794434, 1, 0, 25,
                        settings_for_sensitive_cases);
 
+#if defined(BOOST_GEOMETRY_TEST_FAILURES)
     TEST_DISSOLVE(gitter_2013_04_a, 3043.9181, 3, 0, 21);
+#endif
```
awulkiew: Is this a regression in dissolve introduced by this PR?

include/boost/geometry/algorithms/detail/is_valid/has_valid_self_turns.hpp
```diff
@@ -87,11 +87,13 @@ class has_valid_self_turns
...
             > interrupt_policy;
 
+        // Calculate self-turns, skipping adjacent segments
         detail::self_get_turn_points::self_turns<false, turn_policy>(geometry,
                                           strategy,
                                           robust_policy,
                                           turns,
-                                          interrupt_policy);
+                                          interrupt_policy,
+                                          0, true);
```
awulkiew: What does this change do? Does this change validity check WRT how it was in the past or is it a change needed because some other change was done in `self_turns`?

awulkiew: According to this comment: 

    // In some cases (dissolve, has_self_intersections)
    // neighbouring segments should be checked
    // (for example to detect spikes properly)

(https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/get_turns.hpp#L393)

Though in `is_valid` spikes are detected separately here:

https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/is_valid/ring.hpp#L210

May these adjacent turns be important for some other reason?

extensions/test/algorithms/dissolve.cpp
barendgehrels: Not that I know of. It did not compile before, so I did not bisect when it started failing

include/boost/geometry/algorithms/detail/is_valid/has_valid_self_turns.hpp
barendgehrels: Thanks, I will verify it.

barendgehrels: It is indeed wrong, I will revert it. Strangely enough the tests in `is_valid` did not catch it, so I will add a testcase there too.

It causes some false positives in my branch so I have to solve that in another way.

Will update this PR today

barendgehrels: Hmm, that was in my derived branch. In the original branch it is all OK, and also well tested.

Indeed, spikes are detected separately and therefore adjacent turns to detect turns in validity checking. So I think we can apply it as is now.

Agree?

awulkiew: Ok, if there are problems in the future we can revert it later.

awulkiew:
I marked it as a bug fix because of dissolve but if you think it's more like an enhancement needed by your future work please change it.

include/boost/geometry/algorithms/detail/is_valid/has_valid_self_turns.hpp
barendgehrels: Thank you, and a bug-label is OK

PR #615, opened 2019-08-21
Stop passing policies to intersection, and pass subrange

barendgehrels:
This PR redefines code around segment_ratios such that it is not necessary anymore to pass them 
to most of the intersection strategies. This is a step forwards.

Also, it passes the subrange to intersection i/o the segment. This can (incidentally) be used to avoid start turns. It is not used yet now but will be in a next PR.

_More in detail, in turn_info_helpers the split is made such that only if robust policy is really used, it is passed to intersection. Cartesian only (automatically, by nature of policy, the rest does not use rescaling). Therefore RobustPolicy is not present anymore in spherical/geographic. Later (but all is transitional to avoid mega-PR's) it will also be omitted from Cartesian._

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
@@ -179,23 +179,22 @@ struct buffered_piece_collection
             typename IntersectionStrategy::cs_tag
         >::type rescale_policy_type;
 
-    typedef typename geometry::segment_ratio_type
+    typedef geometry::segment_ratio
     <
-        point_type,
-        RobustPolicy
-    >::type segment_ratio_type;
+        typename geometry::coordinate_type<robust_point_type>::type
+    > ratio_type;
 
     typedef buffer_turn_info
     <
         point_type,
         robust_point_type,
-        segment_ratio_type
+        ratio_type
     > buffer_turn_info_type;
 
     typedef buffer_turn_operation
     <
         point_type,
-        segment_ratio_type
+        ratio_type
     > buffer_turn_operation_type;
 
     typedef std::vector<buffer_turn_info_type> turn_vector_type;
@@ -206,7 +205,7 @@ struct buffered_piece_collection
         int operation_index;
         robust_point_type point;
         segment_identifier seg_id;
-        segment_ratio_type fraction;
+        ratio_type fraction;
     };
 
     struct piece
--- a/include/boost/geometry/algorithms/detail/disjoint/linear_linear.hpp
+++ b/include/boost/geometry/algorithms/detail/disjoint/linear_linear.hpp
@@ -38,7 +38,6 @@
 
 #include <boost/geometry/policies/disjoint_interrupt_policy.hpp>
 #include <boost/geometry/policies/robustness/no_rescale_policy.hpp>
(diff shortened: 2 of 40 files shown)


awulkiew:
Thanks! Though I'm confused by the change from segments to subranges in intersection strategies. Is it because you plan to pass more than 2 points into intersection strategies later? If that is the case how will the intersection strategy work for 3 points and what would these 3 points be? If it is not the case why did segments replaced?

barendgehrels:
Sure, I plan to use more than 2 points later some in rare cases. This makes the start turns, which were there earlier and quite complex, redundant (so that code, which was not used at this moment, is now  deleted in this PR).
The 3 points are effectively passed already right now, because the subrange is passed. But the third point is not used yet. The third point can make the difference in collinear cases between a collinear situation and a touch (both arcs meet at their endpoints). That difference is otherwise, in Floating Point coordinates, in rare cases, very hard to make without that third point

awulkiew:
> The third point can make the difference in collinear cases between a collinear situation and a touch (both arcs meet at their endpoints). That difference is otherwise, in Floating Point coordinates, in rare cases, very hard to make without that third point

I don't know the algorithm you're planning to implement but shouldn't this be rather a part of the algorithm and not the strategy? How the information about the third point (next in range?) will change the calculation of the intersection point of two segments? Or in other words, is it possible to check some specific conditions before calling the strategy or afterwards in the algorithm? Or is the check CS-dependent?

If this is CS-dependent then my gut feeling is that maybe all of the calculation so intersection+sideness of the next segment(s) should be done once in the intersection strategy for all cases, not only this one? Then you'd move significant part of get_turn_info into the intersection strategies so that'd make it even more complex. But you propose to handle only part of the analysis of next segment in the intersection strategy so this is confusing.

barendgehrels:
The whole idea is to create a "hook"like intersection. So two segments are intersected with two others, and if the intersection is located on both first segments, it is reported. The second segments are only involved as soon as the intersection falls exactly on the midpoint. That way we can guarantee intersection behaviour, and don't need start-turns. We won't miss intersections then (which now might happen, though in really rare cases).

Having said that, that can be build around it and not inside indeed. But the problem is that intersection strategy also calls all policies (how, arrival, sides, etc) and those are dependent on that point. Therefore I think it is better to have them inside. And maybe, indeed, it is wise to create a simplier intersection strategy and have this one CS agnostic (it is already very large anyway - but those other strategies assembled there do belong per CS...)

awulkiew:
I was rather thinking about detecting it before calling the strategy. Is it even possible or is it only possible after the intersection point is calculated, so internally in the strategy or after?

> and have this one CS agnostic

This is only a side note. If something is really CS agnostic (so has to be calculated/called regardless of the strategy) it should rather be moved outside to the algorithm.

I'm not saying to do it here because my gut feeling is that the handling of special cases related to numeric issues is not CS agnostic.

barendgehrels:
No, it is not possible to detect it before. It depends on all the information calculated during segment intersection.

> I'm not saying to do it here because my gut feeling is that the handling of special cases related to numeric issues is not CS agnostic

That makes sense, often it is not.

So do you think this PR is acceptable?

awulkiew:
Yes, I'm ok with it. Thanks!

barendgehrels:
Thank you! Will merge it soon, @vissarion are you also OK?

barendgehrels:
Thanks!

PR #616, opened 2019-08-21
Improvements of srs transformations and shapefile loading.

awulkiew:
transformations:
- Introduced shared_grid concept and separated synchronization primitives from algorithms
- 2 separate shared grids implementations, one depending on Boost.Thread and the other one on STD. Not included automatically, the user picks the one that is needed.
- Algorithms of shared grids moved from different files to the one containing the implementation for regular grids.
- Low level projections are implemented differently, no CRTP, parameters moved outside to wrappers. This allows to share the parameters by multiple projections and to remove some virtual calls from dynamic projection.
- Some memory optimizations:
  - separating proj_params for ellipsoidal and spherical versions in case where only a subset of parameters is used (sinu)
  - sharing the parameters between multiple projections (goode, igh)
  - rearranging parameters WRT allignment

shapefile:
- Added special handling of loading of doubles which should work with the majority of architectures. The purpose is also to see which Boost runners will fail.


awulkiew:
Btw, I added you @mloskot as a reviewer because I'd like to hear you opinion specifically about the automatic handling of endianness of doubles (https://github.com/boostorg/geometry/pull/616/files#diff-bc4368b6a55393894fde6f1848242794).

mloskot:
@awulkiew 
> // It is also not clear why did the generic version above compile in the 
> // past.

Because it has never been used on anything else but either on little or big endian architectures?
Because nobody compiled it on ARMs with half-s?

include/boost/geometry/srs/projections/impl/base_static.hpp
```diff
@@ -45,84 +46,116 @@ struct static_projection_type
...
     inline bool inverse(XY const& xy, LL& lp) const
     {
         try
         {
-            pj_inv(this->m_prj, this->m_par, xy, lp);
+            pj_inv(*this, this->m_par, xy, lp);
             return true;
         }
-        catch(...)
+        catch (...)
```
vissarion: is this change by intention? it seems that `catch(...)` is used more than `catch (...)` in the library

awulkiew: Yes, it seems that it's mixed in the library. In SRS the version with space is used more often. But it doesn't matter because we should probably use `BOOST_CATCH` instead anyway. I propose to change it in a separate PR in the whole library at once.

awulkiew:
@mloskot
There were no test files for shapefiles.
There were no special handling of doubles before this PR.

Boost.Endian was always used for conversion between endians in this code. This library checks the type of variable at compile time and it fails to compile for floating point types. So the original code of shapefiles written before should've failed to compile back then when I was testing it locally as it did when I got back to it recently.

AFAIU the checks were added to Boost.Endian fairly recently:
https://github.com/boostorg/endian/commit/f0f8e398d67adb25314cda47dabb64c964adb005
so the reason is probably that I used older version of this library while testing the code.

mloskot:
@awulkiew I think your explanation makes sense and the Boost.Endian update with new checks may be the reason behind the different compilation results. To be honest, I have not monitored Boost.Endian for a very long time.

include/boost/geometry/extensions/gis/io/shapefile/read.hpp
```diff
@@ -114,8 +114,6 @@ inline void double_endianness_check()
 // TODO: It is not clear that this will work on all machines because some of
 //   them may use mixed endianness (half little-endian, half big-endian) for
 //   doubles or different endianness than the one used for integers.
-//   It is also not clear why did the generic version above compile in the
```
mloskot: Just for the history archive: it's been obsolete as per https://github.com/boostorg/geometry/pull/616#issuecomment-532693484

awulkiew:
@mloskot I also noticed that in WKB code are endian-related defines used that are deprecated now. So I think I'll attempt to unify the endian-related code from WKB and shapefile. Not in this PR. The shapefile change should probably not be a part of this PR in the first place.

@barendgehrels what do you think about the change in projections? In particular about replacing CRTP with wrapper?

mloskot:
@awulkiew I vaguely remember I worked with the Boost.Endian-based code while adding the WKB support. I'm not sure why the git history is lacking of those changes, but here are some archived references:
- [[ggl] [Ticket #2] Implementation of io/wkb module](https://lists.boost.org/geometry/2009/06/0275.php)
- [[ggl] Added boost/ggl/util/endian.hpp](https://lists.boost.org/geometry/2009/06/0272.php)

So, I stole the code from somewhere else, the Spirit and others.

awulkiew:
@mloskot Damn, 10 years ago. Thanks for the info.

barendgehrels:
Thanks for the work! 
I still didn't review it properly, but it is very much appreciated

awulkiew:
@barendgehrels do you plan to review it fully or are you satisfied with the quick one you made?

Issue #590, opened 2019-05-16
Moving to C++11 or C++14


awulkiew:
I'd like to start a discussion about moving to the newer standard. Among other things it would definitely make some of the internals clearer than they are now. What do you think about it?

One solution would be to simply announce that Geometry drops the support for older compilers. We should also decide which C++11 features exactly we'd like to use because the support for different features was added in various versions of different compilers. I prepared a list of the most important features I think we could use and the list of 4 compilers below. I'll update it if someone suggests that additional feature is useful.

The feature I think would be the most useful is `auto` because it'd allow us to get rid of many very long template instantiations, `typedef`s, etc. So I marked the features and minimum versions of compilers based on `auto` feature below (so versions GCC 4.4, Clang 2.9, VS 2010, Intel C++ 11.0 and below). The youngest of these compilers is GCC 4.4 released in March 13, 2012. Of course we can relax it if needed, .e.g Rvalue references would also be useful and they are supported in the same versions besides Intel C++ which needs the next minor version. The next thing is I think lambda expressions but this would require GCC 4.5 and Clang 3.1 which basically means compilers released in and after 2012.

| Feature | GCC | Clang | MSC/MSVC/VS | Intel C++ |
| --------- | --------- | --------- | --------- | --------- |
| Rvalue references                                       | **4.3** | **2.9** | **16.0/10.0/2010** | 11.1 |
| Variadic templates                                      | **4.3** | **2.9** | 18.0/12.0/2013 | 12.1 |
| Initializer lists                                              | **4.4** | 3.1 | 18.0/12.0/2013 | 13.0 |
| **Static assertions**                                    | **4.3** | **2.9** | **16.0/10.0/2010** | **11.0** |
| **auto-typed variables**                             | **4.4** | **2.9** | **16.0/10.0/2010** | **11.0** |
| Lambda expressions                                   | 4.5 | 3.1 | **16.0/10.0/2010** | 12.0 |
| **decltype(expression)**                             | **4.3** | **2.9** | **16.0/10.0/2010** | **11.0** |
| Null pointer constant                                  | 4.6 | **2.9** | **16.0/10.0/2010** | 12.1 |
| Strongly-typed enums                                | **4.4** | **2.9** | 17.0/11.0/2012 | 13.0 |
| constexpr                                                    | 4.6 | 3.1 | 19.0/14.0/2015 | 13.0 |
| Delegating constructors                             | 4.7 | 3.0 | 18.0/12.0/2013 | 14.0 |
[...]

PR #617, opened 2019-09-04
Feature/robust predicates

tinko92:
# Introduction
This is the first (and most simple and self-contained) of three branches that I created during the GSoC 2019 that has now concluded. It provides two strategies for 2d cartesian points:
* a side-strategy that is basically a robust, adaptive alternative to side_by_triangle (more details below). Given p1, p2, p3 it calculates the signed area (times two) of the triangle p1, p2, p3 and (based on the sign) answers the question: Does p3 lie on the left or right side or exactly on the segment p1, p2.
* an in-circle-strategy for 4 points p1, p2, p3, p, that answers the question: Does p lie in the circumcircle of the triangle formed by p1, p2, p3?

The strategy classes come with documentation and tests, currently in the extension-directories.

# Rationale
The new side-strategy can be used where side_by_triangle is used and may help in situations where computations break down because of incorrect results in side_by_triangle.
The in_circle-strategy will be required by algorithms for the construction of Delaunay triangulations such as the one I will propose in a later pull request. It may also be of use in constructing an adaptive, robust side-strategy for circular segments, which were recently proposed by @meastp in the Gitter room and generally useful on its own for users.

# Implementation details
Both strategies are based on evaluating certain determinants (like the existing side_by_triangle), They are robust in the sense that they return correct results even on inputs where direct float arithmetics would fail due to precision issues, such as side of POINT(1, 1) with respect to the segment of POINT( 1e-20, 0) and POINT(1e20, 1e20). They are adaptive in the sense that for inputs where direct float arithmetic can provably produce correct results (like in all non-degenerate inputs), the computation terminates early to save time. This is done using analytically computed error bounds. The algorithm is intended for floating-point parameters but returns correct results for integers that can be translated to the calculation_type without loss of precision. It is not robust with respect to inputs that cause an overflow in the computation of interim results.

All of the math behind this is explained in the papers linked here https://www.cs.cmu.edu/~quake/robust.html. The details are implemented almost directly as explained in the paper. All of the implementation details (the mathematics) are in the file 
include/boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp

The robustness-parameter is my own addition and can be used to limit the maximum precision.  If 0 is provided, the predicates are comparatively fast.

### DIFF
--- a/extensions/test/triangulation/in_circle_robust.cpp
+++ b/extensions/test/triangulation/in_circle_robust.cpp
@@ -1,0 +1,56 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+// Unit Test
+
+// Copyright (c) 2019 Tinko Bartels, Berlin, Germany.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#include <geometry_test_common.hpp>
+
+#include <boost/geometry/geometries/point_xy.hpp>
+#include <boost/geometry/extensions/triangulation/strategies/cartesian/in_circle_robust.hpp>
+
+template <typename P>
+void test_all()
+{
+    typedef bg::strategy::in_circle::in_circle_robust<double, 2> inc2;
+    typedef bg::strategy::in_circle::in_circle_robust<double, 1> inc1;
+    typedef bg::strategy::in_circle::in_circle_robust<double, 0> inc0;
+
+    P col1(0.0, 0.0), col2(1.0, 0.0), col3(0.0, 1.0);
+    P in(0.5,0.5) , on(1.0, 0.0), out(-0.5, -0.5);
+    int in2 = inc2::apply(col1, col2, col3, in);
+    BOOST_CHECK_GT(in2, 0);
+    int in1 = inc1::apply(col1, col2, col3, in);
+    BOOST_CHECK_GT(in1, 0);
+    int in0 = inc0::apply(col1, col2, col3, in);
+    BOOST_CHECK_GT(in0, 0);
+
+    int on2 = inc2::apply(col1, col2, col3, on);
+    BOOST_CHECK_EQUAL(on2, 0);
+    int on1 = inc1::apply(col1, col2, col3, on);
+    BOOST_CHECK_EQUAL(on1, 0);
+    int on0 = inc0::apply(col1, col2, col3, on);
+    BOOST_CHECK_EQUAL(on0, 0);
+
+    int out2 = inc2::apply(col1, col2, col3, out);
+    BOOST_CHECK_GT(0, out2);
+    int out1 = inc1::apply(col1, col2, col3, out);
+    BOOST_CHECK_GT(0, out1);
+    int out0 = inc0::apply(col1, col2, col3, out);
+    BOOST_CHECK_GT(0, out0);
+
+    P hard1(0, 0), hard2(1e20, 0), hard3(0, 1e20);
+    P inhard(0.5, 0.5);
+    int hardr  = inc2::apply(hard1, hard2, hard3, inhard);
+    BOOST_CHECK_GT(hardr, 0);
+}
+
+
+int test_main(int, char* [])
+{
+    test_all<bg::model::d2::point_xy<double> >();
+    return 0;
+}
--- a/extensions/test/triangulation/in_circle_robust.cpp
+++ b/extensions/test/triangulation/in_circle_robust.cpp
@@ -3,6 +3,9 @@
 
 // Copyright (c) 2019 Tinko Bartels, Berlin, Germany.
 
+// Contributed and/or modified by Tinko Bartels,
+//   as part of Google Summer of Code 2019 program.
+
 // Use, modification and distribution is subject to the Boost Software License,
 // Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
 // http://www.boost.org/LICENSE_1_0.txt)
--- a/include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp
+++ b/include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp
@@ -1,0 +1,92 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+
+// Copyright (c) 2019 Tinko Bartels, Berlin, Germany.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_EXTENSIONS_TRIANGULATION_STRATEGIES_CARTESIAN_SIDE_ROBUST_HPP
+#define BOOST_GEOMETRY_EXTENSIONS_TRIANGULATION_STRATEGIES_CARTESIAN_SIDE_ROBUST_HPP
+
+#include <boost/geometry/util/select_most_precise.hpp>
+#include <boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp>
+
+namespace boost { namespace geometry
+{
+
+namespace strategy { namespace side
+{
+
+template
+<
+    typename CalculationType = void,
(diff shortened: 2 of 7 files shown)


extensions/test/triangulation/in_circle_robust.cpp
```diff
@@ -0,0 +1,56 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+// Unit Test
+
+// Copyright (c) 2019 Tinko Bartels, Berlin, Germany.
+
```
vissarion: You could acknowledge GSoC here if you like. See for example https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/formulas/karney_direct.hpp#L5

include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+namespace strategy { namespace side
+{
+
+/*!
+\brief Adaptive precision predicate to check at which side of a segment a point lies:
+    left of segment (>0), right of segment (< 0), on segment (0).
+\ingroup strategies
+\tparam CalculationType \tparam_calculation (numeric_limits<ct>::epsilon() and numeric_limits<ct>::digits must be supported for calculation type ct)
+\tparam robustness Number that determines maximum precision. Values from 0 to 2 may make the calculation terminate faster for inputs that may require higher precision to ensure correctness.
```
vissarion: this is not very clear. Do you want to say values from 0 to 3, where 0 is the fastest and less accurate and 3 the slowest and most accurate ?

include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp
```diff
@@ -0,0 +1,103 @@
...
+    typename CalculationType = void,
+    int robustness = 3
+>
+struct side_robust
+{
+public:
+    //! \brief Computes double the signed area of the CCW triangle p1, p2, p
+    template
+    <
+        typename CoordinateType,
```
vissarion: unused template parameter

include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+#ifndef DOXYGEN_SHOULD_SKIP_THIS
+    template
+    <
+        typename P1,
+        typename P2,
+        typename P
+    >
+    static inline int apply(P1 const& p1, P2 const& p2, P const& p)
```
vissarion: maybe this is too generic. What is the case where `p1` and `p2` are of different types?

include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp
```diff
@@ -0,0 +1,103 @@
...
+    {
+        typedef typename coordinate_type<P1>::type coordinate_type1;
+        typedef typename coordinate_type<P2>::type coordinate_type2;
+        typedef typename coordinate_type<P>::type coordinate_type3;
+
+        typedef typename boost::mpl::if_c
+            <
+                boost::is_void<CalculationType>::type::value,
+                typename select_most_precise
```
vissarion: there is a `select_most_precise` with 3 arguments

include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp
```diff
@@ -0,0 +1,103 @@
...
+        typename P2,
+        typename P
+    >
+    static inline int apply(P1 const& p1, P2 const& p2, P const& p)
+    {
+        typedef typename coordinate_type<P1>::type coordinate_type1;
+        typedef typename coordinate_type<P2>::type coordinate_type2;
+        typedef typename coordinate_type<P>::type coordinate_type3;
+
+        typedef typename boost::mpl::if_c
```
vissarion: this is implemented in utilities as `select_calculation_type_alt`

include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp
```diff
@@ -0,0 +1,103 @@
...
+            >::type coordinate_type;
+        typedef typename select_most_precise
+            <
+                coordinate_type,
+                double
+            >::type promoted_type;
+
+
+        promoted_type sv =
+            side_value<coordinate_type, promoted_type>(p1, p2, p);
```
vissarion: `coordinate_type` here is unused

include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp
tinko92: I understand, I'll reconsider the wording there. It is a little hard to say it in one sentence.
By default (robustness = 3), the strategy will compute up to four successively more precise approximations of the true side_value (with the fourth being guaranteed to be correct) and check for each approximation whether its sign can already be guaranteed using relative error bounds. If that is the case, the remaining approximations (if any) will not be computed and the current approximation will be returned. 

If the robustness parameter is set to 0, it will return the first approximation without any test against the error bounds, if it is set to 1, it will compute the first approximation, test it against the error bound and either return if the precision is sufficient or compute the second approximation. With robustness = 2, up to three approximations will be computed.

include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp
tinko92: I cannot imagine such a case, however, my choice to be that generic was motivated by side_by_triangle having the same interface and I wanted to be as compatible with it as possible.

include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp
```diff
@@ -0,0 +1,94 @@
...
+    left of segment (>0), right of segment (< 0), on segment (0).
+\ingroup strategies
+\tparam CalculationType \tparam_calculation (numeric_limits<ct>::epsilon() and numeric_limits<ct>::digits must be supported for calculation type ct)
+\tparam robustness Number that determines maximum precision. Values from 0 to 2 may make the calculation terminate faster for inputs that may require higher precision to ensure correctness.
+\details This predicate determines at which side of a segment a point lies using an algorithm that is adapted from orient2d as described in "Adaptive Precision Floating-Point Arithmetic and Fast Robust Geometric Predicates" by Jonathan Richard Shewchuk ( https://dl.acm.org/citation.cfm?doid=237218.237337 ). More information and copies of the paper can also be found at https://www.cs.cmu.edu/~quake/robust.html . It is designed to be adaptive in the sense that it should be fast for inputs that lead to correct results with plain float operations but robust for inputs that require higher precision arithmetics.
+ */
+template
+<
+    typename CalculationType = void,
+    int robustness = 3
```
barendgehrels: Tinko, thanks a lot for your contributions!
Can you spell the robustness template parameter as `Robustness` and maybe make it an std::size_t

include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp
barendgehrels: Yes, the same interface is OK.

include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp
vissarion: For me, the technical details are not needed here. Hence, saying something like 0 is the fastest and less accurate and 3 the slowest and most accurate, is fine.

awulkiew:
Thanks for this PR! I'm ok with merging after the issues above are addressed.

I'd like to be explicit regarding one thing though. This PR depends on C++11. After merging it the library will partially depend on newer standard. I'm ok with that since we plan to move to C++11 anyway, but I'd like to get explicit confirmation from you @barendgehrels @vissarion.

vissarion:
@awulkiew I am OK to merge this since it is currently in extensions. Also, since in the discussion here: https://github.com/boostorg/geometry/issues/590 we haven't reached to a conclusion on a plan to move to newer standards, adding new features that depend on newer standards in extensions could be a start.

awulkiew:
Thanks Tinko!

PR #622, opened 2019-10-02
Fix/buffer pass strategies

barendgehrels:
This pull request takes care that strategies used within the buffer algorithm details are properly passed from one subroutine to another.

Besides that, there are some changes in buffer test (which were already failing before this PR)

And some additional changes related to removing the rescaling:
* sectionalize
* an extra condition in get_turn_info
* a minor fix in the new subranges

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
@@ -167,7 +167,7 @@ struct buffered_piece_collection
             robust_point_type
         >::type robust_area_result_type;
 
-    typedef typename strategy::point_in_geometry::services::default_strategy
+    typedef typename IntersectionStrategy::template point_in_geometry_strategy
         <
             robust_point_type,
             robust_ring_type
@@ -348,6 +348,7 @@ struct buffered_piece_collection
     area_strategy_type m_area_strategy;
     envelope_strategy_type m_envelope_strategy;
     expand_strategy_type m_expand_strategy;
+    point_in_geometry_strategy_type m_point_in_geometry_strategy;
 
     robust_area_strategy_type m_robust_area_strategy;
     RobustPolicy const& m_robust_policy;
@@ -367,10 +368,15 @@ struct buffered_piece_collection
         , m_has_deflated(false)
         , m_intersection_strategy(intersection_strategy)
         , m_side_strategy(intersection_strategy.get_side_strategy())
-        , m_area_strategy(intersection_strategy.template get_area_strategy<point_type>())
+        , m_area_strategy(intersection_strategy
+            .template get_area_strategy<point_type>())
         , m_envelope_strategy(intersection_strategy.get_envelope_strategy())
         , m_expand_strategy(intersection_strategy.get_expand_strategy())
-        , m_robust_area_strategy(intersection_strategy.template get_area_strategy<robust_point_type>())
+        , m_point_in_geometry_strategy(intersection_strategy
+            .template get_point_in_geometry_strategy<robust_point_type,
+                        robust_ring_type>())
+        , m_robust_area_strategy(intersection_strategy
+            .template get_area_strategy<robust_point_type>())
         , m_robust_policy(robust_policy)
     {}
 
@@ -658,7 +664,12 @@ struct buffered_piece_collection
                 typename IntersectionStrategy::disjoint_box_box_strategy_type
             > original_ovelaps_box_type;
 
-        turn_in_original_visitor<turn_vector_type> visitor(m_turns);
+        turn_in_original_visitor
+            <
+                turn_vector_type,
+                point_in_geometry_strategy_type
+            > visitor(m_turns, m_point_in_geometry_strategy);
+
         geometry::partition
             <
                 robust_box_type,
@@ -808,14 +819,14 @@ struct buffered_piece_collection
         }
     }
 
-    static inline void determine_properties(piece& pc)
+    inline void determine_properties(piece& pc) const
     {
         pc.is_monotonic_increasing[0] = true;
         pc.is_monotonic_increasing[1] = true;
         pc.is_monotonic_decreasing[0] = true;
         pc.is_monotonic_decreasing[1] = true;
 
-        pc.is_convex = geometry::is_convex(pc.robust_ring);
+        pc.is_convex = geometry::is_convex(pc.robust_ring, m_side_strategy);
 
         if (pc.offsetted_count < 2)
         {
@@ -962,8 +973,11 @@ struct buffered_piece_collection
                 <
                     typename geometry::cs_tag<point_type>::type,
                     turn_vector_type, piece_vector_type,
-                    point_in_geometry_strategy_type
-                > visitor(m_turns, m_pieces, point_in_geometry_strategy_type());
+                    point_in_geometry_strategy_type,
+                    side_strategy_type
+                > visitor(m_turns, m_pieces,
+                          m_point_in_geometry_strategy,
+                          m_side_strategy);
 
             typedef turn_ovelaps_box
                 <
@@ -1421,11 +1435,7 @@ struct buffered_piece_collection
         enrich_intersection_points<false, false, overlay_buffer>(m_turns,
             m_clusters, offsetted_rings, offsetted_rings,
             m_robust_policy,
-            m_intersection_strategy.template get_point_in_geometry_strategy
-                <
-                    buffered_ring<Ring>,
-                    buffered_ring<Ring>
-                >());
+            m_intersection_strategy);
     }
 
     // Discards all rings which do have not-OK intersection points only.
@@ -1476,7 +1486,7 @@ struct buffered_piece_collection
 
             int const geometry_code
                 = detail::within::point_in_geometry(any_point,
-                    original.m_ring, point_in_geometry_strategy_type());
+                    original.m_ring, m_point_in_geometry_strategy);
 
             if (geometry_code == -1)
             {
--- a/include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
+++ b/include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
@@ -33,6 +33,8 @@
 #include <boost/type_traits/is_same.hpp>
 #include <boost/type_traits/is_fundamental.hpp>
 
+#include <boost/geometry/core/config.hpp>
+
 #include <boost/geometry/algorithms/assign.hpp>
 #include <boost/geometry/algorithms/envelope.hpp>
 #include <boost/geometry/algorithms/expand.hpp>
@@ -56,7 +58,11 @@
 #include <boost/geometry/views/reversible_view.hpp>
 #include <boost/geometry/geometries/segment.hpp>
 
+#if defined(BOOST_GEOMETRY_USE_RESCALING)
 #include <boost/geometry/algorithms/detail/expand_by_epsilon.hpp>
+#else
+#include <boost/geometry/algorithms/detail/buffer/buffer_box.hpp>
+#endif
 #include <boost/geometry/strategies/envelope.hpp>
 #include <boost/geometry/strategies/expand.hpp>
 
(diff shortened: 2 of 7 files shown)


include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -808,14 +815,14 @@ struct buffered_piece_collection
         }
     }
 
-    static inline void determine_properties(piece& pc)
+    inline void determine_properties(piece& pc) const
```
vissarion: since you added `const` here you could also added to other non-static function members e.g. the other specialization of `determine_properties`

include/boost/geometry/strategies/cartesian/point_in_poly_winding.hpp
```diff
@@ -54,7 +54,8 @@ template
 <
     typename Point_ = void, // for backward compatibility
     typename PointOfSegment_ = Point_, // for backward compatibility
-    typename CalculationType = void
+    typename CalculationType = void,
+    typename SideStrategyType = side::side_by_triangle<CalculationType>
```
awulkiew: The order of arguments is not correct I think. Is it used anywhere? I cannot find it.

barendgehrels: I don't know why it is not correct - I mean, I can't add the strategy before calculation type without breaking existing functionality. Actually the whole structure is designed for backward compatibility, so maybe we should consider some more refactoring here.

Anyway, I use it in another PR (still to come) so therefore you won't see it here. But it should be possible to use another side strategy here, instead of taking the default.

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
barendgehrels: Thanks, I missed your comments. I can have a look there indeed, better if they are mutually consistent.
And I will ook at the long line.

include/boost/geometry/strategies/cartesian/point_in_poly_winding.hpp
awulkiew: My point was that `CalculationType` typically is the last parameter. And yes this strategy could use some refactoring, i.e. Point types are not needed and the name is bad.

Which other side strategy would you like to use? Maybe instead of adding `SideStrategyType` parameter to this strategy there should be another cartesian point-in-geometry strategy with its own side strategy different than this one?

If it's not used in this PR should it be a part of it?

barendgehrels: Yes but my point was that inserting a parameter breaks existing code using it.
I want to use another side strategy and/or the one Tinko created.
What's the objection to having code more generic than it is now?

awulkiew: In general I think we're too careless with strategies and tend to choose the solution which locally works for a specific use-case but is not consistent with the rest of the library. Now we have many different strategies with various parameters in different orders, various naming schemes etc. Many of them are not documented and it is very hard for the users to know which strategy is expected by certain algorithm taking certain combination of arguments, etc.

This particular strategy was agnostic in the past. It was moved as cartesian with the agnostic one as a wrapper left for backward compatibility. Additional strategies were added for other CSes. Now, a parameters is added for cartesian strategy only. Should the same parameter be added to spherical and geographic point-in-geometry strategies? Should it be used by the agnostic one too? What should the user expect? I understand that it may be easier to just pass a side strategy as template parameter but will it be also easier for the user vs. just passing different point-in-geometry strategy? Does the user even need that? Are strategies part of the public interface and therefore for the users? Or are these internal tools used mainly by us? Or maybe are there two kinds of strategies, some for the users and other for internal use?

E.g. AFAIU you need this for the `buffer` internals, correct? Is this something the user should also be able to pass to an algorithm? E.g. assuming it's for the `buffer`, why not create a base class for internal use taking whatever parameters you like and then leave this strategy for the users as it was? If there is only one side strategy working for `buffer` in certain parts and another side strategy in other parts of the algorithm does it even have sense to allow to pass it by the user?

So I guess I'd be ok with this change because it is not different than what can be found in other strategies right now. But in general I think this should either be avoided or we should think about the interface of the library at this point. So if there is some other solution that you'd also be happy with, like making this cartesian winding strategy an internal/detail strategy which then is used in the `buffer` and as a base class of the strategy for the users, I'd prefer that. Or is the solution you've choosen the best, given the strategies design we have right now?

Side note: Initially I thought that being careful with strategies would be enough, so taking care about naming, adding more parameters etc. but I'm not sure now. I'm starting to lean towards the conclusion that the whole strategies system should be redesigned taking into account all of the complexity we have now and knowledge gathered over several last years. But obviously this is not a task for this PR.

vissarion: +1 for the side note

barendgehrels: @awulkiew : thanks for the note. You have done the most recent work on strategies so you know the current state the best. I was not 100% aware of the history about these winding strategies, it is quite long ago that I had worked on it.

I agree with you that we should redesign the strategies. There are a few requirements (from my point of view):

- all strategies should be flexible (the user, the the developer should be able to pass them)
- mainly because of geographic/spherical: strategies might have a configuration (i.e. they should be passed at runtime)
- strategies can depend on each other (such as here) (so we should first create a dependency graph)
- there will be more

So the change here fullfilled the flexibility, but (of course) not yet the rest indeed.

For my current work on avoiding rescaling (not only buffer, in general) I'm using another side strategy. Therefore I should be able to use such another strategy, and it is really inconvenient that the cartesian winding strategy just takes hardcoded the triangle strategy.

Even if I will not use such another strategy (not sure yet), strategies should support flexible strategies without hacking the code at those places.

Good that you point out that it is now incompatible with the other winding or similar strategies. That was not my intention and I think that is not good. So I will revise that part (either taking the default strategy there, or something else, I will consider various options - keeping the fact that it will not take side_by_triangle there).

It will be just a small change to this PR and I hope to finish it next Friday.

awulkiew: @barendgehrels To be clear. I'm not saying that the spherical and geographic strategy has to take side strategy as template parameter. Though it would certainly be an option. But AFAIK currently none of the algorithms needs that in non-cartesian CS. I was just expressing the problem we're facing here.

But I don't understand something. Why do you plan to use different side strategy? Where in the code the decision about its use will be made? Will it be defined in some strategy? Will it always be used in buffer or will the user be able to choose it? Because if it's the user's decision, if it's important to use the correct side strategy for the task and buffer expects intersection strategy, shouldn't it be passed into the Intersection strategy too? It's the most top-level strategy passed into the buffer after all.

On the other hand if it's defined specifically by you, only for buffer in some parts of the algorithm (with `side_by_triangle` used in other parts) because otherwise this algorithm does not work properly why does it have to be a part of the point-in-geometry strategy's interface if it's not the user who defines it?

Do I understand correctly that two different side strategies are used in buffer in different places? If there will be only one in the end, why not simply change it in point-in-geometry and e.g. enable with `#define`?

The problem with this issue is that if you think that the user has to be able to define side strategy in point-in-geometry strategy then by extension it has to be true for all higher-level strategies containing point-in-geometry strategy, so intersection strategy, point-segment distance strategy and IO strategy. And again by extension, in general case it should probably be possible to define all strategies you need explicitly. For that the whole system would have to be redesigned and the strategies for example all replaced with umbrella strategies.

If it's used only in buffer I suggest to leave point-in-geometry strategies as they are now and implement additional one, with clear interface. E.g.:

    template
    <
        typename SideStrategy, typename CalculateType = void
    >
    struct cartesian_point_in_geometry_strategy_for_buffer
    {
        BOOST_MPL_ASSERT((boost::is_same<typename SideStrategy::cs_tag, cartesian_tag>::value));
    };

Or am I missing something?

I think I do not understand the use-case. Would it make sense to simply temporarily revert this change and propose it later together with another PR, with the code actually using it? So we can understand the problem better and talk about a specific problem then?

barendgehrels: To be clear from my side, I repeat

> So I will revise that part (either taking the default strategy there, or something else, I will consider various options - keeping the fact that it will not take side_by_triangle there).

So yes, I will revert that change. No problem.

> it has to be true for all higher-level strategies containing point-in-geometry strategy, so intersection strategy, point-segment distance strategy...

They don't belong there. Intersection strategy *only* depends on side-strategy and nothing more. Why should a complex algorithm as point-in-geometry be defined inside intersection-strategy, just calculating an intersection between two segments...

But they are currently there indeed, most probably for pragmatic reasons (because that was the strategy originally passed). So I'm totally with you that that part should be redesigned.

Why define another side-strategy? Because there might also appear another intersection strategy. But I come back to that later.

awulkiew: > So yes, I will revert that change. No problem.

Thanks.

> Intersection strategy only depends on side-strategy and nothing more. Why should a complex algorithm as point-in-geometry be defined inside intersection-strategy, just calculating an intersection between two segments...

Sorry for confusion, by point-in-geometry strategy I mean the strategy used by point-in-geometry algorithm which is winding strategy. I started to call it point-in-geometry strategy because this strategy is used both for areal and linear geometries and not always used for calculation of winding number of point in polygon. And this is also how the getter is called in intersection strategy (`get_point_in_geometry_strategy()`).

Or is your point that the intersection strategy should not define any getters at all? And that we should find a different solution for propagating strategies into the internal algorithms, like instead of passing "single-purpose" strategies (like intersection, winding, side, distance, etc.) to pass umbrella strategies gathering other "single-purpose" strategies together? If that's the case then I agree. I'm not sure about the umbrella part (maybe some other mechanism would be better) but for sure I'd prefer removing all of the getters and keep the functionality simple and well defined.

But currently, with getters, if we wanted to not only use different side strategy but also some other like different winding strategy or area strategy then we'd have to pass it into the intersection strategy too (because this strategy is expected by the buffer and has to carry all CS-specific information right now). If we redesigned the strategies, removed the getters and replaced them by something else then indeed the intersection strategy would only depend on side strategy.

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
barendgehrels: Sorry, the other one cannot be made `const` because it modifies `m_pieces`.
And this one should not be `static` anymore because it accesses member strategies.
So current code is OK

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -370,6 +371,7 @@ struct buffered_piece_collection
         , m_area_strategy(intersection_strategy.template get_area_strategy<point_type>())
         , m_envelope_strategy(intersection_strategy.get_envelope_strategy())
         , m_expand_strategy(intersection_strategy.get_expand_strategy())
+        , m_point_in_geometry_strategy(intersection_strategy.template get_point_in_geometry_strategy<robust_point_type, robust_ring_type>())
```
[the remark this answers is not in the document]
barendgehrels: Fixed

include/boost/geometry/strategies/cartesian/point_in_poly_winding.hpp
barendgehrels: Hi @awulkiew : So this part is reverted now.

barendgehrels: > Or is your point that the intersection strategy should not define any getters at all? And that we should find a different solution for propagating strategies into the internal algorithms, like instead of passing "single-purpose" strategies (like intersection, winding, side, distance, etc.) to pass umbrella strategies gathering other "single-purpose" strategies together? If that's the case then I agree.

Yes! Intersection strategy should ideally not define any getters. It depends on side-strategy.
Yes, we should propagate strategies as umbrella strategies, gathering them. But the umbrella is not trivial on itself, because it is hierarchical: side strategy (depends on nothing), intersection (depends on side), point-in-geometry (depends on side), distance (...).

Anyway, that is a different thing now but it should keep our attention.

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
```diff
@@ -775,22 +781,30 @@ struct sectionalize_multi
...
     {
+#if defined(BOOST_GEOMETRY_USE_RESCALING)
         detail::expand_by_epsilon(it->bounding_box);
+
+#else
+        // Expand the box to avoid missing any intersection. The amount is
+        // should be larger than epsilon. About the value itself: the smaller
+        // it is, the higher the risk to miss intersections. The larger it is,
+        // the more comparisons are made. So it should be on the high side.
+        detail::buffer::buffer_box(it->bounding_box, 0.001, it->bounding_box);
```
awulkiew: This is probably also something which in the future should depend on a strategy somehow. As well as the epsilon used in point comparison (equals). But for now this harcoded value is fine.

include/boost/geometry/strategies/cartesian/point_in_poly_winding.hpp
awulkiew: @barendgehrels Ok thanks!

PR #623, opened 2019-10-08
[partition] Fix integral overflow while dividing bounding box in half.

awulkiew:


include/boost/geometry/algorithms/detail/partition.hpp
```diff
@@ -30,15 +32,35 @@ namespace boost { namespace geometry
...
+    }
+};
+
+template <typename T>
+struct divide_interval<T, true>
+{
+    static inline T apply(T const& mi, T const& ma)
+    {
+        // avoid overflow
+        return mi / 2 + ma / 2 + (mi % 2 + ma % 2) / 2;
```
barendgehrels: OK for me. Thanks for the fix.

PR #627, opened 2019-10-16
Implement relops for arguments always resulting in false.

awulkiew:
The purpose is to simplify writing generic code with relational operations. Consider e.g.:

    bg::equals(linestring, polygon);
    bg::touches(point, linestring);
    bg::crosses(point, linestring);
    bg::overlaps(point, linestring);

before: compilation error,
this PR: compiles and always returns `false`.

### DIFF
--- a/include/boost/geometry/algorithms/detail/equals/implementation.hpp
+++ b/include/boost/geometry/algorithms/detail/equals/implementation.hpp
@@ -5,8 +5,8 @@
 // Copyright (c) 2009-2015 Mateusz Loskot, London, UK.
 // Copyright (c) 2014-2015 Adam Wulkiewicz, Lodz, Poland.
 
-// This file was modified by Oracle on 2014, 2015, 2016, 2017, 2018.
-// Modifications copyright (c) 2014-2018 Oracle and/or its affiliates.
+// This file was modified by Oracle on 2014, 2015, 2016, 2017, 2018, 2019.
+// Modifications copyright (c) 2014-2019 Oracle and/or its affiliates.
 
 // Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
 // Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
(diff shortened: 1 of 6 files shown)


include/boost/geometry/algorithms/detail/equals/implementation.hpp
```diff
@@ -275,6 +275,15 @@ struct equals_by_collection_or_relate
     }
 };
 
+struct equals_always_false
+{
+    template <typename Geometry1, typename Geometry2, typename Strategy>
+    static inline bool apply(Geometry1 const& , Geometry2 const& , Strategy const& )
```
mloskot: Nitpicking: `& )` -> `&)` :-)

barendgehrels:
So this is only for geometry-type-combinations which can, by definition, never return `true`?

awulkiew:
@barendgehrels Yes, other cases are already handled. There is no combination which always results in `true`.

include/boost/geometry/algorithms/detail/equals/implementation.hpp
awulkiew: @mloskot Well idk ;) do we have gudelines for that? AFAIU we put space after `&` and before `, ` so I assumed that the missing name is represented by space. Or am I mistaken and all of the spaces should be removed?

barendgehrels:
> 
> 
> @barendgehrels Yes, other cases are already handled. There is no combination which always results in `true`.

Great, thanks

include/boost/geometry/algorithms/detail/equals/implementation.hpp
barendgehrels: AFAIK we don't have a written guideline for that.
I find this readable, but I'm also fine with &)

mloskot: Having nothing constructive to say, I went nit-picking and bike-shedding. Certainly, ignoreable

barendgehrels: Certainly nice to hear from you Mateusz!

mloskot: Barend, I'm happy to be revolving around still

awulkiew: Ok, I saw this in Geometry code but it is possible that it was me who put it in the first place. I don't have an opinion regarding it, but if it's ignorable I'll leave it as it is. ;)

awulkiew:
Thanks for reviews.

PR #632, opened 2019-10-30
Fix/buffer onesided

barendgehrels:
This PR (not related to rescaling / robustness, but found it when working at it) fixes two related things:

- the end-cap was generated wrongly for one-sided buffers
- turn points were classified wrongly for one-sided buffers

### DIFF
--- a/doc/release_notes.qbk
+++ b/doc/release_notes.qbk
@@ -20,6 +20,18 @@
 [section:release_notes Release Notes]
 
 [/=================]
+[heading Boost 1.72]
+[/=================]
+
+[*Improvements]
+
+* Various improvements related to robustness of set and relational operation.
+
+[*Bugfixes]
+
+* Fixed errors in generation of asymmetric one-sided buffer
+
+[/=================]
 [heading Boost 1.71]
 [/=================]
 
--- a/include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
@@ -547,10 +547,6 @@ struct buffered_piece_collection
             {
                 it->location = inside_buffer;
             }
-            if (it->count_on_original_boundary > 0)
-            {
-                it->location = inside_buffer;
-            }
             if (it->count_within_near_offsetted > 0)
             {
                 // Within can have in rare cases a rounding issue. We don't discard this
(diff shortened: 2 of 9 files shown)


include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -547,10 +547,6 @@ struct buffered_piece_collection
             {
                 it->location = inside_buffer;
             }
-            if (it->count_on_original_boundary > 0)
-            {
-                it->location = inside_buffer;
```
barendgehrels: Because this counter is now set in a different way, this condition could go

doc/release_notes.qbk
```diff
@@ -19,6 +19,18 @@
 
 [section:release_notes Release Notes]
 
+[/=================]
+[heading Boost 1.72]
+[/=================]
+
+[*Improvements]
+
+* Various improvements related to robustness of set and relational operation.
```
vissarion: it would be useful to have here the PR links

test/algorithms/buffer/buffer_linestring.cpp
```diff
@@ -325,23 +325,48 @@ void test_all()
...
+    test_one<linestring, polygon>("mysql_25662426a_mostly_right_05", mysql_25662426a, join_round32, end_round32, 14.3419, 0.05, settings, 0.5);
+    test_one<linestring, polygon>("mysql_25662426a_mostly_right_1", mysql_25662426a, join_round32, end_round32, 27.1955, 0.1, settings, 1.0);
+    test_one<linestring, polygon>("mysql_25662426a_mostly_right_2", mysql_25662426a, join_round32, end_round32, 43.1821, 0.2, settings, 2.0);
+    test_one<linestring, polygon>("mysql_25662426a_mostly_right_3", mysql_25662426a, join_round32, end_round32, 54.4337, 0.3, settings, 3.0);
+    test_one<linestring, polygon>("mysql_25662426a_mostly_right_4", mysql_25662426a, join_round32, end_round32, 75.6376, 0.4, settings, 4.0);
+    test_one<linestring, polygon>("mysql_25662426a_mostly_right_5", mysql_25662426a, join_round32, end_round32, 102.6952, 0.5, settings, 5.0);
+    test_one<linestring, polygon>("mysql_25662426a_mostly_right_10", mysql_25662426a, join_round32, end_round32, 260.1850, 1.0, settings, 10.0);
+
     // Left
-    test_one<linestring, polygon>("mysql_25662426a_1", mysql_25662426a, join_round32, end_round32, 54.9018, 1.0, ut_settings(), 0.0);
```
vissarion: are those tests still failing or the new ones are also covering those?

barendgehrels:
Thanks for the quick approvals.

W.r.t. @vissarion 's remark about release notes, that line is copied from earlier releases where the same is done, it is a continuous operation - we might add PR's but I doubt if it is necessary.

PR #637, opened 2019-11-05
Fix MSVC compailation errors in karney_inverse.

awulkiew:
Plus warning fix.

Even though this formula requires C++11 and is no longer included by default (see https://github.com/boostorg/geometry/pull/636) it is still tested (in `test/formulas/inverse.cpp`). Without this fix it does not compile with MSVC even when C++11 is supported.


awulkiew:
Or should we disable testing of this formula as well?

awulkiew:
I decided to put `karney_inverse` cases in a separate file to not loose the ability to run tests of other formulas without C++11 support.

PR #640, opened 2019-11-13
Minor changes

barendgehrels:
This is no functional change (also because it is beta period)

- a reported issue is added to the testsuite
- sort-by-side is refactored just a bit to allow calling it from other situations (the calling itself is not ready yet so not included in this PR).
- Also it updated several tests going right/wrong with different compilerdefine settings
- Finally it reports the type(s) being tested


test/algorithms/buffer/buffer_multi_polygon.cpp
```diff
@@ -413,42 +411,43 @@ void test_all()
...
     test_one<multi_polygon_type, polygon_type>("rt_p11", rt_p11, join_miter, end_flat, 28.7426, 1.0);
     test_one<multi_polygon_type, polygon_type>("rt_p12", rt_p12, join_miter, end_flat, 22.5711, 1.0);
     test_one<multi_polygon_type, polygon_type>("rt_p13", rt_p13, join_miter, end_flat, 19.9142, 1.0);
     test_one<multi_polygon_type, polygon_type>("rt_p14", rt_p14, join_miter, end_flat, 20.8284, 1.0);
     test_one<multi_polygon_type, polygon_type>("rt_p15", rt_p15, join_miter, end_flat, 23.6569, 1.0);
-#if defined(BOOST_GEOMETRY_USE_KRAMER) || defined(BOOST_GEOMETRY_TEST_FAILURES)
+//#if defined(BOOST_GEOMETRY_USE_KRAMER) || defined(BOOST_GEOMETRY_TEST_FAILURES)
     test_one<multi_polygon_type, polygon_type>("rt_p16", rt_p16, join_miter, end_flat, 23.4853, 1.0);
-#endif
+//#endif
```
barendgehrels: Oops, will remove these defines in next commit, it now succeeds

awulkiew:
@barendgehrels Do I understand correctly that you consider this a bugfix and would like to release it with 1.72?

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
```diff
@@ -124,7 +106,7 @@ private :
 
     static const operation_type target_operation = operation_from_overlay<OverlayType>::value;
 
-    typedef typename side_compare<target_operation>::type side_compare_type;
+    typedef typename sort_by_side::side_compare<target_operation>::type side_compare_type;
```
awulkiew: This is not a change request but a question. I'm wondering, since this is the only place where `side_compare` is used, is there a reason why separate struct template was added and not MPL/TT used to define this member type instead? E.g.:

    typedef typename boost::mpl::if_c
        <
            boost::is_same<target_operation, operation_union>::value,
            std::greater<int>,
            std::less<int>
        >::type side_compare_type;

barendgehrels: That would also be possible. This was written years ago when we didn't use is_same so often.
Besides that, now you will get a compilation error if it is not either union or intersection. With the same, you won't get that. So it might be more error-prone (depending what you want, of course)

barendgehrels:
> 
> 
> @barendgehrels Do I understand correctly that you consider this a bugfix and would like to release it with 1.72?

No, this does not need to be in 1.72 (but it might). The only functional change is in the distance measure but that is not relevant for 1.72 itself.

barendgehrels:
It's merged, thanks for your quick reviews

PR #641, opened 2019-11-14
Fix deprecated header for floating_point_comparison

vissarion:
Fixing warning `This header is deprecated. Please use <boost/test/tools/floating_point_comparison.hpp> instead.`


awulkiew:
Thanks!

PR #644, opened 2019-11-18
Move math::difference_angle() because it uses C++11 std::reminder.

awulkiew:
Also add const& to math util arguments.

### DIFF
--- a/include/boost/geometry/formulas/karney_inverse.hpp
+++ b/include/boost/geometry/formulas/karney_inverse.hpp
@@ -42,6 +42,31 @@
 #include <boost/geometry/formulas/result_inverse.hpp>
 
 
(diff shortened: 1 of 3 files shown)


adl1995:
As far as I can see, this function uses the same signs as in GeographicLib: https://sourceforge.net/p/geographiclib/code/ci/release/tree/include/GeographicLib/Math.hpp#l414

awulkiew:
@adl1995 Right, I was looking at the GeographicLib 1.46:
https://sourceforge.net/p/geographiclib/code/ci/f1df2f5da8c05b8ec2fa484894a82861c87b7da3/tree/include/GeographicLib/Math.hpp#l475

Thanks. I'll remove the comment.

PR #645, opened 2019-11-19
fixing `Wunused-parameter` warnings

vissarion:


include/boost/geometry/strategies/cartesian/buffer_end_round.hpp
```diff
@@ -116,6 +116,7 @@ public :
                 DistanceStrategy const& distance,
                 RangeOut& range_out) const
     {
+        boost::ignore_unused(perp_left_point);
```
mloskot: Shouldn't there be explicit `#include <boost/core/ignore_unused.hpp>` for this, to keep the header self-contained, etc.?

vissarion: Yes, this is better. Changed. Thanks!

awulkiew:
Thanks!

include/boost/geometry/strategies/cartesian/buffer_end_round.hpp
barendgehrels: Thanks!

Issue #643, opened 2019-11-18
boost::intersection produces wrong results by tolerance


Legotckoi:
This is related to [#630](https://github.com/boostorg/geometry/issues/630) 
Consider the following:

```
#include <iostream>
#include <boost/geometry.hpp>

using namespace std;
using namespace boost::geometry;
using point2d = model::d2::point_xy<double>;
using polygon = model::polygon<point2d>;
using multiPolygon = model::multi_polygon<polygon>;

void checkIntersection(const std::string& polyString, const std::string& multiPolyString)
{
    polygon poly;
    multiPolygon multiPoly;
    multiPolygon result;

    read_wkt(polyString, poly);
    read_wkt(multiPolyString, multiPoly);

    intersection(multiPoly, poly, result);
    correct(result);

    cout << area(poly) << " " << area(result) << endl;
    cout << result.empty() << " " << intersects(multiPoly, poly) << endl << endl;
}

int main()
{
    polygon poly;
    multiPolygon multiPoly;
    multiPolygon result;

    // this works
    cout << "works" << endl;
    checkIntersection("POLYGON((-0.420826 0.168095,5.29161 5.76881,12.2925 -1.37173,6.58007 -6.97245,-0.420826 0.168095))",
                      "MULTIPOLYGON(((4.75986e-16 -0.261125,0 0,0.723376 0.600649,1.22339 1.19163,1.7234 2.13422,1.7234 -2.0189,4.75986e-16 -0.261125)))");

    // this doesn`t works
    cout << "doesn`t work" << endl;
    checkIntersection("POLYGON((-0.420825839394717959862646239344 0.168094877926996288941552393226,"
                      "5.29161113734201116187705338234 5.76881481261494233336861725547,"
                      "12.2925110557019436896553088445 -1.37173140830596795858298264648,"
                      "6.58007407896521545609402892296 -6.97245134299391278176472042105,"
                      "-0.420825839394717959862646239344 0.168094877926996288941552393226))",
                      "MULTIPOLYGON(((4.75985680877701196159479187986e-16 -0.26112514220036736611874061964,"
                      "0 0,"
                      "0.72337592336357892097709054724 0.600648602980154100450249643472,"
                      "1.223386680467822174023240222 1.19163154396839887638748223253,"
                      "1.72339743757206620422550713556 2.1342216197599452875977021904,"
                      "1.72339742847305732453833115869 -2.01889900623749607433410346857,"
[...]

Issue #630, opened 2019-10-25
boost::intersection produces wrong results


innova-engineering:
This is related to #548 
Consider the following:
```
#include <boost/test/unit_test.hpp>
#include <boost/geometry.hpp>


void testIntersection(const std::string& polyString, const std::string& multiPolyString)
{
    using namespace boost::geometry;
    using point2d = model::d2::point_xy<double>;
    using polygon = model::polygon<point2d, true, false>;
    using multiPolygon = model::multi_polygon<polygon>;

    polygon poly;
    multiPolygon multiPoly;
    multiPolygon result;

    read_wkt(polyString, poly);
    read_wkt(multiPolyString, multiPoly);

    intersection(multiPoly, poly, result);

    BOOST_CHECK_GE(area(poly), area(result));
    BOOST_CHECK_EQUAL(!result.empty(), intersects(multiPoly, poly));
}

BOOST_AUTO_TEST_CASE(Test1)
{
    testIntersection("POLYGON((-0.3 -0.1475,-0.3 +0.1475,+0.3 +0.1475,+0.3 -0.1475,-0.3 -0.1475))", 
                    "MULTIPOLYGON(((-0.605 +0.1575,+0.254777333596 +1.0172773336,+1.53436796127 -0.262313294074,+0.674590627671 -1.12209062767,-0.605 +0.1575)))");
}

BOOST_AUTO_TEST_CASE(Test2)
{
    testIntersection("POLYGON((-0.3 -0.1475,-0.3 +0.1475,+0.3 +0.1475,+0.3 -0.1475,-0.3 -0.1475))", 
                    "MULTIPOLYGON(((-1.215 +0.7675000000000001,-0.4962799075873666 +1.486220092412633,+0.665763075292561 +0.324177109532706,-0.05295701712007228 -0.3945429828799273,-1.215 +0.7675000000000001)))");
}

BOOST_AUTO_TEST_CASE(Test3)
{
    testIntersection("POLYGON((-0.3 -0.1475,-0.3 +0.1475,+0.3 +0.1475,+0.3 -0.1475,-0.3 -0.1475))", 
                    "MULTIPOLYGON(((-0.9099999999999999 +0.4625,-0.1912799075873667 +1.181220092412633,+0.9707630752925609 +0.01917710953270602,+0.2520429828799277 -0.6995429828799273,-0.9099999999999999 +0.4625)))");
}
```

All data represent polygons that actually intersect. 

[...]

PR #648, opened 2019-11-23
[TEST] add testcases for reported issues

barendgehrels:
Adds testcases for issue #643 and, for difference, also for #630 (union/intersection were already there)


barendgehrels:
@awulkiew  thanks for the merge

PR #670, opened 2020-02-29
Symmetrizing read/write wkt

### LINKED ISSUE #654: Proposal: Make read/write WKT more symmetric by adding writing to std::string
AFAIK, the only way to write WKT provided by boost.geometry is a stream manipulator. In many cases it is more convenient to produce a `std::string` representation directly, as it is provided for reading by `read_wkt`. I would like to propose `write_wkt` function operating on a string and geometry.

sudo-panda:
This PR fixes #654 . It adds the following:

- `to_wkt`: This function converts the given geometry to wkt string and returns it instead of modifying a stringstream taken as input like `wkt`.
- `from_wkt`: This function returns the geometry obtained by parsing a wkt string instead of taking in a geometry variable as parameter and storing the geometry in it.
- tests: Tests for both `to_wkt` and `from_wkt` are provided with basically follow the template of already implemented `read_wkt` and `wkt` tests.
- examples and documentation: Example usage of both have been added and documentation has been updated.

This PR is now ready to be reviewed and then merged.

### DIFF
--- a/doc/html/index.html
+++ b/doc/html/index.html
@@ -140,10 +140,14 @@
 <li class="listitem">
         Yaghyavardhan Singh Khangarot (discrete Frechet and Hausdorff distance)
       </li>
+<li class="listitem">
+        Tinko Bartels (Delaunay triangulation, Voronoi diagram, random point generation,
+        ...)
+      </li>
 </ul></div>
 </div>
 <table xmlns:rev="http://www.cs.rpi.edu/~gregod/boost/tools/doc/revision" width="100%"><tr>
-<td align="left"><p><small>Last revised: February 19, 2019 at 14:49:50 GMT</small></p></td>
+<td align="left"><p><small>Last revised: March 01, 2020 at 20:07:18 GMT</small></p></td>
 <td align="right"><div class="copyright-footer"></div></td>
 </tr></table>
 <hr>
--- a/doc/html/index.html
+++ b/doc/html/index.html
@@ -140,14 +140,10 @@
 <li class="listitem">
         Yaghyavardhan Singh Khangarot (discrete Frechet and Hausdorff distance)
       </li>
-<li class="listitem">
-        Tinko Bartels (Delaunay triangulation, Voronoi diagram, random point generation,
-        ...)
-      </li>
 </ul></div>
 </div>
 <table xmlns:rev="http://www.cs.rpi.edu/~gregod/boost/tools/doc/revision" width="100%"><tr>
-<td align="left"><p><small>Last revised: March 01, 2020 at 20:07:18 GMT</small></p></td>
+<td align="left"><p><small>Last revised: February 19, 2019 at 14:49:50 GMT</small></p></td>
 <td align="right"><div class="copyright-footer"></div></td>
 </tr></table>
 <hr>
--- a/doc/reference/io/read_wkt.qbk
+++ b/doc/reference/io/read_wkt.qbk
@@ -19,4 +19,6 @@ is reserved for future usage, which will then have an indication of the used tex
 [read_wkt]
 
 [heading See also]
+* [link geometry.reference.io.wkt.from_wkt From WKT]
+* [link geometry.reference.io.wkt.to_wkt To WKT]
 * [link geometry.reference.io.wkt.wkt WKT streaming manipulator]
--- a/doc/reference/io/to_wkt.qbk
+++ b/doc/reference/io/to_wkt.qbk
@@ -1,0 +1,24 @@
+[/============================================================================
+  Boost.Geometry (aka GGL, Generic Geometry Library)
+
+  Copyright (c) 2009-2014 Barend Gehrels, Amsterdam, the Netherlands.
+
+  Use, modification and distribution is subject to the Boost Software License,
+  Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+  http://www.boost.org/LICENSE_1_0.txt)
+=============================================================================/]
+
+[def __this_function__ to_wkt]
+
+[heading_conformance_ogc __this_function__..AsText]
+[note __this_function__ is not named "AsText" or "as_text" because Boost.Geometry 
+also supports other textformats (svg, dsv)]
+
+[heading Example]
+[to_wkt]
+[to_wkt_output]
+
+[heading See also]
+* [link geometry.reference.io.wkt.read_wkt Read WKT]
+* [link geometry.reference.io.wkt.from_wkt From WKT]
+* [link geometry.reference.io.wkt.wkt WKT streaming manipulator]
--- a/doc/reference/io/to_wkt.qbk
+++ b/doc/reference/io/to_wkt.qbk
@@ -2,6 +2,7 @@
   Boost.Geometry (aka GGL, Generic Geometry Library)
 
   Copyright (c) 2009-2014 Barend Gehrels, Amsterdam, the Netherlands.
+  Copyright (c) 2020 Baidyanath Kundu, Haldia, India.
 
   Use, modification and distribution is subject to the Boost Software License,
   Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
--- a/doc/src/examples/io/from_wkt.cpp
+++ b/doc/src/examples/io/from_wkt.cpp
@@ -18,11 +18,11 @@
 
 int main()
 {
-    typedef boost::geometry::model::d2::point_xy<double> point_type;
-    typedef boost::geometry::model::box<point_type> box_type;
-    typedef boost::geometry::model::segment<point_type> segment_type;
-    typedef boost::geometry::model::polygon<point_type> polygon_type;
-    typedef boost::geometry::model::linestring<point_type> linestring_type;
+    using point_type = boost::geometry::model::d2::point_xy<double>;
+    using box_type = boost::geometry::model::box<point_type>;
+    using segment_type = boost::geometry::model::segment<point_type>;
+    using polygon_type = boost::geometry::model::polygon<point_type>;
+    using linestring_type = boost::geometry::model::linestring<point_type>;
 
     point_type a;
     box_type d;
--- a/doc/src/examples/io/from_wkt.cpp
+++ b/doc/src/examples/io/from_wkt.cpp
@@ -24,19 +24,13 @@ int main()
     using polygon_type = boost::geometry::model::polygon<point_type>;
     using linestring_type = boost::geometry::model::linestring<point_type>;
 
-    point_type a;
-    box_type d;
-    segment_type e;
-    polygon_type c;
-    linestring_type b;
-    a = boost::geometry::from_wkt<point_type>("POINT(1 2)");
-    d = boost::geometry::from_wkt<box_type>("BOX(0 0,3 3)");
-    e = boost::geometry::from_wkt<segment_type>("SEGMENT(1 0,3 4)");
-    c = boost::geometry::from_wkt<polygon_type>("POLYGON((0 0,0 7,4 2,2 0,0 0))");
-    b = boost::geometry::from_wkt<linestring_type>("LINESTRING(0 0,2 2,3 1)");
+    auto const a = boost::geometry::from_wkt<point_type>("POINT(1 2)");
+    auto const d = boost::geometry::from_wkt<box_type>("BOX(0 0,3 3)");
+    auto const e = boost::geometry::from_wkt<segment_type>("SEGMENT(1 0,3 4)");
+    auto const c = boost::geometry::from_wkt<polygon_type>("POLYGON((0 0,0 7,4 2,2 0,0 0))");
+    auto const b = boost::geometry::from_wkt<linestring_type>("LINESTRING(0 0,2 2,3 1)");
 
     return 0;
 }
 
 //]
-
--- a/doc/src/examples/io/from_wkt.cpp
+++ b/doc/src/examples/io/from_wkt.cpp
@@ -2,6 +2,7 @@
 // QuickBook Example
 
 // Copyright (c) 2014 Barend Gehrels, Amsterdam, the Netherlands.
+// Copyright (c) 2020 Baidyanath Kundu, Haldia, India.
 
 // Use, modification and distribution is subject to the Boost Software License,
 // Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
--- a/doc/src/examples/io/from_wkt.cpp
+++ b/doc/src/examples/io/from_wkt.cpp
@@ -1,0 +1,36 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+// QuickBook Example
+
+// Copyright (c) 2014 Barend Gehrels, Amsterdam, the Netherlands.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+//[from_wkt
+//` Shows the usage of from_wkt
+
+#include <boost/geometry.hpp>
+#include <boost/geometry/geometries/point_xy.hpp>
+#include <boost/geometry/geometries/linestring.hpp>
+#include <boost/geometry/geometries/polygon.hpp>
+
+int main()
+{
+    typedef boost::geometry::model::d2::point_xy<double> point_type;
+    typedef boost::geometry::model::linestring<point_type> linestring_type;
+    typedef boost::geometry::model::polygon<point_type> polygon_type;
+    typedef boost::geometry::model::box<point_type> box_type;
+    typedef boost::geometry::model::segment<point_type> segment_type;
+
+    point_type a = boost::geometry::from_wkt<point_type>("POINT(1 2)");
+    linestring_type b = boost::geometry::from_wkt<linestring_type>("LINESTRING(0 0,2 2,3 1)");
+    polygon_type c = boost::geometry::from_wkt<polygon_type>("POLYGON((0 0,0 7,4 2,2 0,0 0))");
+    box_type d = boost::geometry::from_wkt<box_type>("BOX(0 0,3 3)");
+    segment_type e = boost::geometry::from_wkt<segment_type>("SEGMENT(1 0,3 4)");
+
+    return 0;
+}
+
+//]
+
--- a/doc/src/examples/io/from_wkt.cpp
+++ b/doc/src/examples/io/from_wkt.cpp
@@ -18,16 +18,21 @@
 int main()
 {
     typedef boost::geometry::model::d2::point_xy<double> point_type;
-    typedef boost::geometry::model::linestring<point_type> linestring_type;
-    typedef boost::geometry::model::polygon<point_type> polygon_type;
     typedef boost::geometry::model::box<point_type> box_type;
     typedef boost::geometry::model::segment<point_type> segment_type;
+    typedef boost::geometry::model::polygon<point_type> polygon_type;
+    typedef boost::geometry::model::linestring<point_type> linestring_type;
 
-    point_type a = boost::geometry::from_wkt<point_type>("POINT(1 2)");
-    linestring_type b = boost::geometry::from_wkt<linestring_type>("LINESTRING(0 0,2 2,3 1)");
-    polygon_type c = boost::geometry::from_wkt<polygon_type>("POLYGON((0 0,0 7,4 2,2 0,0 0))");
-    box_type d = boost::geometry::from_wkt<box_type>("BOX(0 0,3 3)");
-    segment_type e = boost::geometry::from_wkt<segment_type>("SEGMENT(1 0,3 4)");
+    point_type a;
+    box_type d;
+    segment_type e;
+    polygon_type c;
+    linestring_type b;
+    a = boost::geometry::from_wkt<point_type>("POINT(1 2)");
+    d = boost::geometry::from_wkt<box_type>("BOX(0 0,3 3)");
+    e = boost::geometry::from_wkt<segment_type>("SEGMENT(1 0,3 4)");
+    c = boost::geometry::from_wkt<polygon_type>("POLYGON((0 0,0 7,4 2,2 0,0 0))");
+    b = boost::geometry::from_wkt<linestring_type>("LINESTRING(0 0,2 2,3 1)");
 
     return 0;
 }
--- a/doc/src/examples/io/to_wkt.cpp
+++ b/doc/src/examples/io/to_wkt.cpp
@@ -32,7 +32,7 @@ int main()
     std::cout << boost::geometry::to_wkt(point) << std::endl;
     std::cout << boost::geometry::to_wkt(polygon) << std::endl;
 
-    point_type point_frac = geom::make<point_type>(3.141592654, 2.718281828);
+    point_type point_frac = geom::make<point_type>(3.141592654, 27.18281828);
     geom::model::polygon<point_type> polygon_frac;
     geom::append(geom::exterior_ring(polygon_frac), geom::make<point_type>(0.00000, 0.00000));
     geom::append(geom::exterior_ring(polygon_frac), geom::make<point_type>(0.00000, 4.00001));
@@ -55,7 +55,7 @@ Output:
 [pre
 POINT(3 2)
 POLYGON((0 0,0 4,4 4,4 0,0 0))
-POINT(3.14 2.72)
+POINT(3.14 27.2)
 POLYGON((0 0,0 4,4 4,4 0,0 0))
 ]
 
--- a/doc/src/examples/io/to_wkt.cpp
+++ b/doc/src/examples/io/to_wkt.cpp
@@ -21,7 +21,7 @@ int main()
     namespace geom = boost::geometry;
     typedef geom::model::d2::point_xy<double> point_type;
 
-    point_type point = geom::make<point_type>(3, 6);
+    point_type point = geom::make<point_type>(3, 2);
     geom::model::polygon<point_type> polygon;
     geom::append(geom::exterior_ring(polygon), geom::make<point_type>(0, 0));
     geom::append(geom::exterior_ring(polygon), geom::make<point_type>(0, 4));
@@ -31,6 +31,17 @@ int main()
 
     std::cout << boost::geometry::to_wkt(point) << std::endl;
     std::cout << boost::geometry::to_wkt(polygon) << std::endl;
+
+    point_type point_frac = geom::make<point_type>(3.141592654, 2.718281828);
+    geom::model::polygon<point_type> polygon_frac;
+    geom::append(geom::exterior_ring(polygon_frac), geom::make<point_type>(0.00000, 0.00000));
+    geom::append(geom::exterior_ring(polygon_frac), geom::make<point_type>(0.00000, 4.00001));
+    geom::append(geom::exterior_ring(polygon_frac), geom::make<point_type>(4.00001, 4.00001));
+    geom::append(geom::exterior_ring(polygon_frac), geom::make<point_type>(4.00001, 0.00000));
+    geom::append(geom::exterior_ring(polygon_frac), geom::make<point_type>(0.00000, 0.00000));
+
+    std::cout << boost::geometry::to_wkt(point_frac, 3) << std::endl;
+    std::cout << boost::geometry::to_wkt(polygon_frac, 3) << std::endl;
 
     return 0;
 }
@@ -42,7 +53,9 @@ int main()
 /*`
 Output:
 [pre
-POINT(3 6)
+POINT(3 2)
+POLYGON((0 0,0 4,4 4,4 0,0 0))
+POINT(3.14 2.72)
 POLYGON((0 0,0 4,4 4,4 0,0 0))
 ]
 
--- a/doc/src/examples/io/to_wkt.cpp
+++ b/doc/src/examples/io/to_wkt.cpp
@@ -1,0 +1,51 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+// QuickBook Example
+
+// Copyright (c) 2020 Baidyanath Kundu, Haldia, India.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+//[to_wkt
+//` Shows the usage of to_wkt
+
+#include <iostream>
+
+#include <boost/geometry.hpp>
+#include <boost/geometry/geometries/point_xy.hpp>
+#include <boost/geometry/geometries/polygon.hpp>
+
+int main()
+{
+    namespace geom = boost::geometry;
+    typedef geom::model::d2::point_xy<double> point_type;
+
+    point_type point = geom::make<point_type>(3, 6);
+    geom::model::polygon<point_type> polygon;
+    geom::append(geom::exterior_ring(polygon), geom::make<point_type>(0, 0));
+    geom::append(geom::exterior_ring(polygon), geom::make<point_type>(0, 4));
+    geom::append(geom::exterior_ring(polygon), geom::make<point_type>(4, 4));
+    geom::append(geom::exterior_ring(polygon), geom::make<point_type>(4, 0));
+    geom::append(geom::exterior_ring(polygon), geom::make<point_type>(0, 0));
+
+    std::cout << boost::geometry::to_wkt(point) << std::endl;
+    std::cout << boost::geometry::to_wkt(polygon) << std::endl;
+
+    return 0;
+}
+
+//]
+
+
+//[to_wkt_output
+/*`
+Output:
+[pre
+POINT(3 6)
+POLYGON((0 0,0 4,4 4,4 0,0 0))
+]
+
+
+*/
+//]
--- a/include/boost/geometry/io/wkt/write.hpp
+++ b/include/boost/geometry/io/wkt/write.hpp
@@ -613,7 +613,16 @@ inline wkt_manipulator<Geometry> wkt(Geometry const& geometry)
 \qbk{[include reference/io/to_wkt.qbk]}
 */
 template <typename Geometry>
-inline std::string to_wkt(Geometry const& geometry, int significant_digits = -1)
+inline std::string to_wkt(Geometry const& geometry)
+{
+    concepts::check<Geometry const>();
+    std::string out;
+    dispatch::devarianted_wkt<Geometry>::apply(out, geometry, -1, ! (boost::is_same<typename tag<Geometry>::type, ring_tag>::value));
+    return out;
+}
+
+template <typename Geometry>
+inline std::string to_wkt(Geometry const& geometry, int significant_digits)
 {
     concepts::check<Geometry const>();
     std::string out;
--- a/include/boost/geometry/io/wkt/write.hpp
+++ b/include/boost/geometry/io/wkt/write.hpp
@@ -24,7 +24,6 @@
 
 #include <ostream>
 #include <string>
-#include <limits> 
 
 #include <boost/array.hpp>
 #include <boost/range/begin.hpp>
@@ -67,67 +66,22 @@ namespace boost { namespace geometry
 namespace detail { namespace wkt
 {
 
-template <typename Output>
-struct output_formatter
-{
-    template <typename T>
-    static void append(Output& out, T const& v, int = -1)
-    {
-        out << v;
-    }
-};
-
-template <>
-struct output_formatter<std::string>
-{
-    template <typename T>
-    static void append(std::string& out, T const& val, int significant_digits = -1)
-    {
-        BOOST_STATIC_ASSERT((boost::is_arithmetic<T>::value));
-        std::stringstream ss;
-        if (significant_digits >= 0)
-        {
-            ss.precision(significant_digits);
-        }
-        else if (std::numeric_limits<T>::is_specialized)
-        {
-            ss.precision(std::numeric_limits<T>::digits10);
-        }
-        ss << val;
-        out += ss.str();
-    }
-
-    static void append(std::string& out, char c, int = -1)
-    {
-        out += c;
-    }
-
-    static void append(std::string& out, char const* s, int = -1)
-    {
-        out += s;
-    }
-};
-
 template <typename P, int I, int Count>
 struct stream_coordinate
 {
-    template <typename Output>
-    static inline void apply(Output& out, P const& p, int significant_digits)
+    template <typename Char, typename Traits>
+    static inline void apply(std::basic_ostream<Char, Traits>& os, P const& p)
     {
-        if (I > 0)
-        {
-            output_formatter<Output>::append(out, ' ');
-        }
-        output_formatter<Output>::append(out, get<I>(p), significant_digits);
-        stream_coordinate<P, I + 1, Count>::apply(out, p, significant_digits);
+        os << (I > 0 ? " " : "") << get<I>(p);
+        stream_coordinate<P, I + 1, Count>::apply(os, p);
     }
 };
 
 template <typename P, int Count>
 struct stream_coordinate<P, Count, Count>
 {
-    template <typename Output>
-    static inline void apply(Output&, P const&, int significant_digits)
+    template <typename Char, typename Traits>
+    static inline void apply(std::basic_ostream<Char, Traits>&, P const&)
     {}
 };
 
@@ -163,13 +117,12 @@ struct double_closing_parenthesis
 template <typename Point, typename Policy>
 struct wkt_point
 {
-    template <typename Output>
-    static inline void apply(Output& out, Point const& p, int significant_digits, bool)
+    template <typename Char, typename Traits>
+    static inline void apply(std::basic_ostream<Char, Traits>& os, Point const& p, bool)
     {
-        output_formatter<Output>::append(out, Policy::apply());
-        output_formatter<Output>::append(out, "(");
-        stream_coordinate<Point, 0, dimension<Point>::type::value>::apply(out, p, significant_digits);
-        output_formatter<Output>::append(out, ")");;
+        os << Policy::apply() << "(";
+        stream_coordinate<Point, 0, dimension<Point>::type::value>::apply(os, p);
+        os << ")";
     }
 };
 
@@ -186,9 +139,9 @@ template
 >
 struct wkt_range
 {
-    template <typename Output>
-    static inline void apply(Output& out, Range const& range,
-                int significant_digits, bool force_closure = ForceClosurePossible)
+    template <typename Char, typename Traits>
+    static inline void apply(std::basic_ostream<Char, Traits>& os,
+                Range const& range, bool force_closure = ForceClosurePossible)
     {
         typedef typename boost::range_iterator<Range const>::type iterator_type;
 
@@ -199,7 +152,7 @@ struct wkt_range
 
         bool first = true;
 
-        output_formatter<Output>::append(out, PrefixPolicy::apply());
+        os << PrefixPolicy::apply();
 
         // TODO: check EMPTY here
 
@@ -207,11 +160,8 @@ struct wkt_range
         iterator_type end = boost::end(range);
         for (iterator_type it = begin; it != end; ++it)
         {
-            if (!first)
-            {
-                output_formatter<Output>::append(out, ",");
-            }
-            stream_type::apply(out, *it, significant_digits);
+            os << (first ? "" : ",");
+            stream_type::apply(os, *it);
             first = false;
         }
 
@@ -221,11 +171,11 @@ struct wkt_range
             && boost::size(range) > 1
             && wkt_range::disjoint(*begin, *(end - 1)))
         {
-            output_formatter<Output>::append(out, ",");
-            stream_type::apply(out, *begin, significant_digits);
+            os << ",";
+            stream_type::apply(os, *begin);
         }
 
-        output_formatter<Output>::append(out, SuffixPolicy::apply());
+        os << SuffixPolicy::apply();
     }
 
 
@@ -262,26 +212,26 @@ struct wkt_sequence
 template <typename Polygon, typename PrefixPolicy>
 struct wkt_poly
 {
-    template <typename Output>
-    static inline void apply(Output& out, Polygon const& poly,
-                int significant_digits, bool force_closure)
+    template <typename Char, typename Traits>
+    static inline void apply(std::basic_ostream<Char, Traits>& os,
+                Polygon const& poly, bool force_closure)
     {
         typedef typename ring_type<Polygon const>::type ring;
 
-        output_formatter<Output>::append(out, PrefixPolicy::apply());
+        os << PrefixPolicy::apply();
         // TODO: check EMPTY here
-        output_formatter<Output>::append(out, "(");
-        wkt_sequence<ring>::apply(out, exterior_ring(poly), significant_digits, force_closure);
+        os << "(";
+        wkt_sequence<ring>::apply(os, exterior_ring(poly), force_closure);
 
         typename interior_return_type<Polygon const>::type
             rings = interior_rings(poly);
         for (typename detail::interior_iterator<Polygon const>::type
                 it = boost::begin(rings); it != boost::end(rings); ++it)
         {
-            output_formatter<Output>::append(out, ",");
-            wkt_sequence<ring>::apply(out, *it, significant_digits, force_closure);
+            os << ",";
+            wkt_sequence<ring>::apply(os, *it, force_closure);
         }
-        output_formatter<Output>::append(out, ")");
+        os << ")";
     }
 
 };
@@ -289,13 +239,13 @@ struct wkt_poly
 template <typename Multi, typename StreamPolicy, typename PrefixPolicy>
 struct wkt_multi
 {
-    template <typename Output>
-    static inline void apply(Output& out, Multi const& geometry,
-                int significant_digits, bool force_closure)
+    template <typename Char, typename Traits>
+    static inline void apply(std::basic_ostream<Char, Traits>& os,
+                Multi const& geometry, bool force_closure)
     {
-        output_formatter<Output>::append(out, PrefixPolicy::apply());
+        os << PrefixPolicy::apply();
         // TODO: check EMPTY here
-        output_formatter<Output>::append(out, "(");
+        os << "(";
 
         for (typename boost::range_iterator<Multi const>::type
                     it = boost::begin(geometry);
@@ -304,12 +254,12 @@ struct wkt_multi
         {
             if (it != boost::begin(geometry))
             {
-                output_formatter<Output>::append(out, ",");
+                os << ",";
             }
-            StreamPolicy::apply(out, *it, significant_digits, force_closure);
+            StreamPolicy::apply(os, *it, force_closure);
         }
 
-        output_formatter<Output>::append(out, ")");
+        os << ")";
     }
 };
 
@@ -318,20 +268,20 @@ struct wkt_box
 {
     typedef typename point_type<Box>::type point_type;
 
-    template <typename Output>
-    static inline void apply(Output& os, Box const& box,
-                int significant_digits, bool force_closure)
+    template <typename Char, typename Traits>
+    static inline void apply(std::basic_ostream<Char, Traits>& os,
+                Box const& box, bool force_closure)
     {
         // Convert to a clockwire ring, then stream.
         // Never close it based on last point (box might be empty and
         // that should result in POLYGON((0 0,0 0,0 0,0 0, ...)) )
         if (force_closure)
         {
-            do_apply<model::ring<point_type, true, true> >(os, box, significant_digits);
+            do_apply<model::ring<point_type, true, true> >(os, box);
         }
         else
         {
-            do_apply<model::ring<point_type, true, false> >(os, box, significant_digits);
+            do_apply<model::ring<point_type, true, false> >(os, box);
         }
     }
 
@@ -343,15 +293,15 @@ struct wkt_box
             //assert_dimension<B, 2>();
         }
 
-        template <typename RingType, typename Output>
-        static inline void do_apply(Output& out,
-                    Box const& box, int significant_digits)
+        template <typename RingType, typename Char, typename Traits>
+        static inline void do_apply(std::basic_ostream<Char, Traits>& os,
+                    Box const& box)
         {
             RingType ring;
             geometry::convert(box, ring);
-            output_formatter<Output>::append(out, "POLYGON(");
-            wkt_sequence<RingType, false>::apply(out, ring, significant_digits);
-            output_formatter<Output>::append(out, ")");
+            os << "POLYGON(";
+            wkt_sequence<RingType, false>::apply(os, ring);
+            os << ")";
         }
 
 };
@@ -362,9 +312,9 @@ struct wkt_segment
 {
     typedef typename point_type<Segment>::type point_type;
 
-    template <typename Output>
-    static inline void apply(Output& out, Segment const& segment, 
-                int significant_digits, bool)
+    template <typename Char, typename Traits>
+    static inline void apply(std::basic_ostream<Char, Traits>& os,
+                Segment const& segment, bool)
     {
         // Convert to two points, then stream
         typedef boost::array<point_type, 2> sequence;
@@ -375,8 +325,8 @@ struct wkt_segment
 
         // In Boost.Geometry a segment is represented
         // in WKT-format like (for 2D): LINESTRING(x y,x y)
-        output_formatter<Output>::append(out, "LINESTRING");
-        wkt_sequence<sequence, false>::apply(out, points, significant_digits);
+        os << "LINESTRING";
+        wkt_sequence<sequence, false>::apply(os, points);
     }
 
     private:
@@ -506,45 +456,42 @@ struct wkt<Multi, multi_polygon_tag>
 template <typename Geometry>
 struct devarianted_wkt
 {
-    template <typename Output>
-    static inline void apply(Output& out, Geometry const& geometry,
-                             int significant_digits, bool force_closure)
+    template <typename OutputStream>
+    static inline void apply(OutputStream& os, Geometry const& geometry,
+                             bool force_closure)
     {
-        wkt<Geometry>::apply(out, geometry, significant_digits, force_closure);
+        wkt<Geometry>::apply(os, geometry, force_closure);
     }
 };
 
 template <BOOST_VARIANT_ENUM_PARAMS(typename T)>
 struct devarianted_wkt<variant<BOOST_VARIANT_ENUM_PARAMS(T)> >
 {
-    template <typename Output>
+    template <typename OutputStream>
     struct visitor: static_visitor<void>
     {
-        Output& m_out;
-        int m_significant_digits;
+        OutputStream& m_os;
         bool m_force_closure;
 
-        visitor(Output& out, int significant_digits, bool force_closure)
-            : m_out(out)
-            , m_significant_digits(significant_digits)
+        visitor(OutputStream& os, bool force_closure)
+            : m_os(os)
             , m_force_closure(force_closure)
         {}
 
         template <typename Geometry>
         inline void operator()(Geometry const& geometry) const
         {
-            devarianted_wkt<Geometry>::apply(m_out, geometry, m_significant_digits, m_force_closure);
+            devarianted_wkt<Geometry>::apply(m_os, geometry, m_force_closure);
         }
     };
 
-    template <typename Output>
+    template <typename OutputStream>
     static inline void apply(
-        Output& out,
+        OutputStream& os,
         variant<BOOST_VARIANT_ENUM_PARAMS(T)> const& geometry,
-        int significant_digits,
         bool force_closure)
     {
-        boost::apply_visitor(visitor<Output>(out, significant_digits, force_closure), geometry);
+        boost::apply_visitor(visitor<OutputStream>(os, force_closure), geometry);
     }
 };
 
@@ -572,30 +519,31 @@ public:
 
     // Boost.Geometry, by default, closes polygons explictly, but not rings
     // NOTE: this might change in the future!
-    inline wkt_manipulator(Geometry const& g, bool force_closure = ! is_ring)
+    inline wkt_manipulator(Geometry const& g,
+                           bool force_closure = ! is_ring)
         : m_geometry(g)
         , m_force_closure(force_closure)
     {}
 
     template <typename Char, typename Traits>
     inline friend std::basic_ostream<Char, Traits>& operator<<(
-            std::basic_ostream<Char, Traits>& out,
+            std::basic_ostream<Char, Traits>& os,
             wkt_manipulator const& m)
     {
-        dispatch::devarianted_wkt<Geometry>::apply(out, m.m_geometry, -1, m.m_force_closure);
-        out.flush();
-        return out;
+        dispatch::devarianted_wkt<Geometry>::apply(os, m.m_geometry, m.m_force_closure);
+        os.flush();
+        return os;
     }
 
 private:
     Geometry const& m_geometry;
     bool m_force_closure;
 };
+
 /*!
 \brief Main WKT-streaming function
 \tparam Geometry \tparam_geometry
 \param geometry \param_geometry
-\param significant_digits Specifies the no of significant digits to use in the output wkt
 \ingroup wkt
 \qbk{[include reference/io/wkt.qbk]}
 */
@@ -607,9 +555,8 @@ inline wkt_manipulator<Geometry> wkt(Geometry const& geometry)
     return wkt_manipulator<Geometry>(geometry);
 }
 
-
 /*!
-\brief Main WKT-string formulating function
+\brief WKT-string formulating function
 \tparam Geometry \tparam_geometry
 \param geometry \param_geometry
 \param significant_digits Specifies the no of significant digits to use in the output wkt
@@ -619,19 +566,18 @@ inline wkt_manipulator<Geometry> wkt(Geometry const& geometry)
 template <typename Geometry>
 inline std::string to_wkt(Geometry const& geometry)
 {
-    concepts::check<Geometry const>();
-    std::string out;
-    dispatch::devarianted_wkt<Geometry>::apply(out, geometry, -1, ! (boost::is_same<typename tag<Geometry>::type, ring_tag>::value));
-    return out;
+    std::stringstream ss;
+    ss << boost::geometry::wkt(geometry);
+    return ss.str();
 }
 
 template <typename Geometry>
 inline std::string to_wkt(Geometry const& geometry, int significant_digits)
 {
-    concepts::check<Geometry const>();
-    std::string out;
-    dispatch::devarianted_wkt<Geometry>::apply(out, geometry, significant_digits, ! (boost::is_same<typename tag<Geometry>::type, ring_tag>::value));
-    return out;
+    std::stringstream ss;
+    ss.precision(significant_digits);
+    ss << boost::geometry::wkt(geometry);
+    return ss.str();
 }
 
 #if defined(_MSC_VER)
--- a/include/boost/geometry/io/wkt/write.hpp
+++ b/include/boost/geometry/io/wkt/write.hpp
@@ -596,7 +596,7 @@ private:
 \brief Main WKT-streaming function
 \tparam Geometry \tparam_geometry
 \param geometry \param_geometry
-\param 
+\param significant_digits Specifies the no of significant digits to use in the output wkt
 \ingroup wkt
 \qbk{[include reference/io/wkt.qbk]}
 */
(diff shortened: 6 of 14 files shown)


include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +62,63 @@ namespace boost { namespace geometry
 namespace detail { namespace wkt
 {
 
+template <typename Output>
+struct output_formatter
+{
+    template <typename T>
+    static void add(Output & out, T const& v)
```
mloskot: Perhaps [put](https://en.cppreference.com/w/cpp/io/basic_ostream/put) is a better name, than `add`.
Alternatively, `append` if it should work for both, streams and buffers like `string`. What's your opinion, @awulkiew ?

Nitpick: remove whitespaces before `&`. Try to replicate existing code style.

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +62,63 @@ namespace boost { namespace geometry
...
+        ss << val;
+        out += ss.str();
+    }
+
+    static void add(std::string & out, char c)
+    {
+        out += c;
+    }
+
+    static void add(std::string & out, const char * s)
```
mloskot: Nitpick: `const char *` -> `char const*`

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +62,63 @@ namespace boost { namespace geometry
...
+    {
+        out << v;
+    }
+};
+
+template <>
+struct output_formatter<std::string>
+{
+    template <typename T>
+    static void add(std::string & out, T const& val)
```
mloskot: `std::string` needs to be `std::basic_string`

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +62,63 @@ namespace boost { namespace geometry
...
+    {
+        std::stringstream ss;
```
mloskot: `std::stringstream` needs to be `std::basic_stringstream`

test/io/wkt/wkt.cpp
```diff
@@ -46,6 +46,15 @@ void check_wkt(G const& geometry, std::string const& expected)
                       boost::to_upper_copy(expected));
 }
 
+template <typename G>
```
mloskot: I think you should also update copyright notice, adding yourself.

sudo-panda: What year should I mention in the copyright notice?

mloskot: Year of your contribution.

See also FAQ on maintaining copy notices here https://www.boost.org/users/license.html

include/boost/geometry/io/wkt/write.hpp
awulkiew: @mloskot yes, `append` is a better name, thanks.

include/boost/geometry/io/wkt/write.hpp
awulkiew: Does this rule also apply for pointers, or only references? I have a feeling that this can be confused with `char * const` by a beginner.

include/boost/geometry/io/wkt/write.hpp
awulkiew: And then also take `std::basic_stream` as argument of `to_wkt`.

include/boost/geometry/io/wkt/write.hpp
awulkiew: Also of you change `std::string` to `std::basic_string` then it becomes `CharT const*` or `const CharT *`.

mloskot:
```
T const&
T const*
```

is clearest yet most symmetric syntax possible, also for pointers as even if one writes `const T*`, correct reading is from right to left anyway. I'd actually expect beginners having problems with `const T*` ;)

This was discussed quite some time ago and documented here https://github.com/boostorg/geometry/wiki/Guidelines-for-Developers

doc/html/index.html
```diff
@@ -140,10 +140,14 @@ <h3>
 <li class="listitem">
```
mloskot: Please, remove changes to this file. 
Unfortunately, it gets modified during docs build, see #674, but the changes should be `git restore`-ed before committing.

doc/src/examples/io/from_wkt.cpp
```diff
@@ -0,0 +1,36 @@
...
+    typedef boost::geometry::model::linestring<point_type> linestring_type;
+    typedef boost::geometry::model::polygon<point_type> polygon_type;
+    typedef boost::geometry::model::box<point_type> box_type;
+    typedef boost::geometry::model::segment<point_type> segment_type;
+
+    point_type a = boost::geometry::from_wkt<point_type>("POINT(1 2)");
+    linestring_type b = boost::geometry::from_wkt<linestring_type>("LINESTRING(0 0,2 2,3 1)");
+    polygon_type c = boost::geometry::from_wkt<polygon_type>("POLYGON((0 0,0 7,4 2,2 0,0 0))");
+    box_type d = boost::geometry::from_wkt<box_type>("BOX(0 0,3 3)");
+    segment_type e = boost::geometry::from_wkt<segment_type>("SEGMENT(1 0,3 4)");
```
mloskot: Could you declare the variables first, then assign to avoid the `warning: variable ‘a’ set but not used` flood?

Generally, we don't want to introduce any new warnings. They also flood CI builds logs often causing logs too log  failures or timeouts, and we may start building examples at some point.

doc/reference/io/read_wkt.qbk
```diff
@@ -19,4 +19,6 @@ is reserved for future usage, which will then have an indication of the used tex
 [read_wkt]
 
 [heading See also]
+* [link geometry.reference.io.wkt.from_wkt From WKT]
+* [link geometry.reference.io.wkt.to_wkt To WKT]
```
mloskot: You need to create the two documents you are linking here, which will pull the examples in:

1. `reference/io/from_wkt.qbk`

    ```
    [heading Conformance]
    Other libraries refer to this functionality as [*ST_GeomFromText] or [*STGeomFromText].
    That is not done here because Boost.Geometry support more text formats. The name GeomFromText
    is reserved for future usage, which will then have an indication of the used text format.

    [heading Example]
    [from_wkt]

    [heading See also]
    * [link geometry.reference.io.wkt.to_wkt To WKT]
    * [link geometry.reference.io.wkt.wkt WKT streaming manipulator]
    ```

2. `reference/io/to_wkt.qbk`

    ```
    [heading Conformance]
    Other libraries refer to this functionality as [*ST_AsText].
    That is not done here because Boost.Geometry support more text formats.

    [heading Example]
    [to_wkt]

    [heading See also]
    * [link geometry.reference.io.wkt.from_wkt From WKT]
    * [link geometry.reference.io.wkt.wkt WKT streaming manipulator]
    ```

sudo-panda: I made those files but did not commit them. Sorry about that.

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +63,67 @@ namespace boost { namespace geometry
...
+{
+    template <typename T>
+    static void append(Output& out, T const& v)
+    {
+        out << v;
+    }
+};
+
+template <typename CharT, typename Traits, typename Allocator>
+struct output_formatter<std::basic_string<CharT, Traits, Allocator>>
```
awulkiew: `> >` - we don't require C++11 yet.

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +63,67 @@ namespace boost { namespace geometry
...
+        ss << val;
+        out += ss.str();
+    }
+
+    static void append(std::basic_string<CharT, Traits, Allocator>& out, CharT c)
+    {
+        out += c;
+    }
+
+    static void append(std::basic_string<CharT, Traits, Allocator>& out, const CharT* s)
```
awulkiew: As @mloskot said earlier the type should be `CharT const*`

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +63,67 @@ namespace boost { namespace geometry
...
+    }
+};
+
+template <typename CharT, typename Traits, typename Allocator>
+struct output_formatter<std::basic_string<CharT, Traits, Allocator>>
+{
+    template <typename T>
+    static void append(std::basic_string<CharT, Traits, Allocator>& out, T const& val)
+    {
+        std::basic_stringstream<CharT, Traits, Allocator> ss;
```
awulkiew: Prior to C++11 `basic_stringstream` didn't have `Allocator` parameter [1] so please remove it. Plus it's probably not needed to use a custom allocator here.

[1] https://en.cppreference.com/w/cpp/io/basic_stringstream

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +63,67 @@ namespace boost { namespace geometry
...
+    static void append(std::basic_string<CharT, Traits, Allocator>& out, const CharT* s)
+    {
+        out += s;
+    }
+
+    template <typename AppendCharT, typename AppendTraits, typename AppendAllocator>
+    static void append(
+        std::basic_string<CharT, Traits, Allocator>& out, 
+        std::basic_string<AppendCharT, AppendTraits, AppendAllocator> const& str
+        )
```
awulkiew: I know I had this overload in my example code but is this really needed? Are strings passed into append anywhere?

include/boost/geometry/io/wkt/write.hpp
awulkiew: The problem with this overload and the one above is that they take `CharT` and `CharT const*` which are taken from `std::basic_string`. So if this type is anything else then the first overload of `append()` taking any `T const&` is going to be instantiated. This will happen e.g. when you pass `std::wstring` because this type is derived from `std::basic_string<wchar_t>` [1] so `CharT` will be `wchar_t` but below values of type `char` or `char const*` are passed into `append()`, e.g. `' '` or `"("`.

This is not a problem now, because `to_wkt` always returns `std::string` but if you wanted it to be complete, and e.g. allowed to define the output string type, e.g. like that:

std::wstring wkt = bg::to_wkt<std::wstring>(geom);

then this implementation would be sub-optimal because for all chars and string literals `std::basic_stringstream` would be used anyway.

[1] https://en.cppreference.com/w/cpp/string/basic_string

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +63,67 @@ namespace boost { namespace geometry
...
+};
+
+template <typename CharT, typename Traits, typename Allocator>
+struct output_formatter<std::basic_string<CharT, Traits, Allocator>>
+{
+    template <typename T>
+    static void append(std::basic_string<CharT, Traits, Allocator>& out, T const& val)
+    {
+        std::basic_stringstream<CharT, Traits, Allocator> ss;
+        ss << val;
```
awulkiew: As I mentioned here: https://github.com/boostorg/geometry/issues/654#issuecomment-570940504 I propose to set the precision here to `std::numeric_limits<T>::digits10` if `std::numeric_limits<T>::is_specialized`. @mloskot what do you think?

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +63,67 @@ namespace boost { namespace geometry
...
-    template <typename Char, typename Traits>
-    static inline void apply(std::basic_ostream<Char, Traits>& os, P const& p)
+    template <typename Output>
+    static inline void apply(Output& out, P const& p)
     {
-        os << (I > 0 ? " " : "") << get<I>(p);
-        stream_coordinate<P, I + 1, Count>::apply(os, p);
+        if (I > 0)
+        {
+            output_formatter<Output>::append(out, ' ');
```
awulkiew: So e.g. here `' '` is of type `char` so if the `Output` is `std::basic_string<wchar_t>` then here the `append()` using `std::basic_stringstream` will be instantiated defeating the purpose of having multiple `append()` overloads.

include/boost/geometry/io/wkt/write.hpp
mloskot: I agree with @awulkiew here. I missed the fact all the current WKT IO implementation, by design or not, quite assumes `char`-based output.

Adding support for arbitrary `CharT` is something that could be considered but in separate PR.

Then, with help of [additional layer of indirection](http://www.josuttis.com/tmplbook/), we could specify required characters and WKT tags too using expected types, somethign along these lines:

```
template <typename CharT>
struct output_char_traits;
template <>
struct output_char_traits<char> { static const char space = ' '; };
template <>
struct output_char_traits<wchar_t> { static const wchar_t space = L' '; };

template <typename Output>
void apply(Output& out)
{
    // traits_type common for basic_string and basic_stringstream
    typedef typename Output::traits_type::char_type char_type;
    output_formatter<Output>::append(out, output_char_traits<char_type>::space);
};
```

include/boost/geometry/io/wkt/write.hpp
mloskot: I agree maxing out by `T` should be fine.

We could have something like this to address user requests like https://github.com/boostorg/geometry/issues/654#issuecomment-570835846

```
std::string to_wkt(Geometry const& geometry); // defaults to digits10
std::string to_wkt(Geometry const& geometry, int significant_digits);

File: include/boost/geometry/io/wkt/write.hpp (hunk 14 above)
Comment (mloskot) [reply include/boost/geometry/io/wkt/write.hpp]: Yes, as I also agreed in details in  https://github.com/boostorg/geometry/pull/670#discussion_r387284769

File: include/boost/geometry/io/wkt/write.hpp (hunk 10 above)
Comment (awulkiew) [reply include/boost/geometry/io/wkt/write.hpp]: It is also possible that this could be simplified. I didn't look into it but it's possible that the codes of the basic characters we use (' ', '(', ')', ',', etc.) are the same for all character types (`char`, `wchar_t`, `char16_t`, `char32_t`, etc.). So it's possible that a simple `static_cast` would be enough.

Comment (mloskot) [reply include/boost/geometry/io/wkt/write.hpp]: The cast may not work in a endian-agnostic manner. There is however narrow/widen functions defined and the standard guarantees unambiguous transformation for the [basic charactes  set of 96 characters](https://books.google.pl/books?id=5qybChHSk7QC&pg=PA279&lpg=PA279&dq=char+widen+narrow&source=bl&ots=drhQ3F5n20&sig=ACfU3U0bF4Ro_vwSXVE56F9TGIytpMrWPQ&hl=en&sa=X&ved=2ahUKEwjWiajLsP_nAhXksIsKHd3_DocQ6AEwA3oECAYQAQ#v=onepage&q=char%20widen%20narrow&f=false)

Comment (sudo-panda) [reply include/boost/geometry/io/wkt/write.hpp]: As you have said that the support for CharT should be added in a separate PR, what is to be done in this PR?

Comment (mloskot) [reply include/boost/geometry/io/wkt/write.hpp]: @sudo-panda As you see the review turned into some additional brainstorm, not uncommon. 

I think reverting back to explicit use of narrow character `std::string` and `std::stringstream` is probably the right thing to do, but let's wait for @awulkiew opinion on that.

Extending current all the plain text formatters (WKT and even SVG) to support arbitrary type of character is something desirable, I think, but it is a task separate from this PR.

File: include/boost/geometry/io/wkt/write.hpp [hunk 15]
```diff
@@ -62,22 +63,67 @@ namespace boost { namespace geometry
...
+{
+    template <typename T>
+    static void append(Output& out, T const& v)
+    {
+        out << v;
+    }
+};
+
+template <typename CharT, typename Traits>
+struct output_formatter<std::basic_string<CharT, Traits>>
```
vissarion: `> >` - we don't require `C++11` yet.

awulkiew: `std::basic_string` has 3 template arguments prior to C++11. `std::basic_string` takes `Allocator` [1] while `std::basic_stringstream` does not [2].

[1] https://en.cppreference.com/w/cpp/string/basic_string
[2] https://en.cppreference.com/w/cpp/io/basic_stringstream

sudo-panda: Thankyou for the explanation.

mloskot:
@sudo-panda FYI, next time you update your PR, then documentation and `doc/src/examples` will be built and verified with GitHub Actions.

include/boost/geometry/io/wkt/write.hpp
awulkiew: > I think reverting back to explicit use of narrow character std::string and std::stringstream is probably the right thing to do
@mloskot Yes, I'm ok with that.

include/boost/geometry/io/wkt/write.hpp
sudo-panda: Can you provide any technique to propagate the significant_digits back to the output_formatter?
The ways that I see its possible are 
- Use a tparam or param in each function call from to_wkt till output_formatter
- Use a global variable
Both of these seem to be bad. The latter worse than the former. If you could provide any other way or else I will proceed with the first one.

mloskot: @sudo-panda 
> Use a tparam or param in each function call from to_wkt till output_formatter

IMO, users typically expect it to be specifiable  at run-time, so yes, adding the parameter is the feasible option `output_formatter<T>::append(..., int significant_digits)` .

> Use a global variable

No, we definitely should avoid globals.

> Both of these seem to be bad.

What's bad in an extra parameter?

/cc @vissarion, @awulkiew for insights

sudo-panda: > What's bad in an extra parameter?

It's redundant information for the functions that are between to_wkt and output_formatter. But if that's the only way then I will do it.

mloskot: The existing technique of dispatching, often interleaved (for geometry type, for output type), is based on forwarding of (numerous) parameters which are often ignored at higher levels until they reach place where they find its use. Think of it as functional way.

Alternative would be object-based solution, for example:

```cpp
template <typename Geometry>
inline std::string to_wkt(Geometry const& geometry, int signigicant_digits)
{
    concepts::check<Geometry const>();
    std::string out;
    wkt<dispatch::devarianted_wkt<Geometry>> wkt(out, signigicant_digits);
    wkt.write(out, geometry);
    return out;
}
```

but then `wkt` and others would have to undergo a major re-write.

To me, this is a matter of work-to-feature balance. Adding support for how many decimal digits we output may not necessarily justify a major change to existing code. Major changes have real potential to introduce bugs. Is the feature worth the risk...

sudo-panda: You are right it is not worth the risk and is not how the rest of boost/geometry has been built. I will start work on adding the parameter.

mloskot: If you are not in rush, you may give @awulkiew  and @vissarion some time to respond with their opinions.

sudo-panda: Sure

vissarion: To my point of view, passing an extra parameter seems ok and as @mloskot said it is already performed in the library in many places.

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +64,92 @@ namespace boost { namespace geometry
...
+
+template <>
+struct output_formatter<std::ostream>
+{
+    template <typename T>
+    static void append(std::ostream& out, T const& val, int significant_digits = -1)
+    {
+        if(std::numeric_limits<T>::is_specialized)
+        {
+            if(significant_digits==-1)
```
mloskot: IMO, exact value is not relevant here and the test should be 

```
if (significant_digits < 0)
```

IOW, we want to distinguish specified precision (0, 1, 2, 3...) from unspecified (default).

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +64,92 @@ namespace boost { namespace geometry
...
+        out += s;
+    }
+};
+
+template <>
+struct output_formatter<std::ostream>
+{
+    template <typename T>
+    static void append(std::ostream& out, T const& val, int significant_digits = -1)
+    {
```
mloskot: I think it would be a good idea to add here

```
BOOST_STATIC_ASSERT((boost::is_arithmetic<T>::value));
```

It would be nice if we could use `static_assert` with `std::is_arithmetic<T>` here , but we are not at C++11 yet.

mloskot:
@sudo-panda If you update your PR, you will get #680 change and GitHub Actions will CI check it for you.

sudo-panda:
@mloskot thanks for the guidance and approving the PR.

test/io/wkt/wkt.cpp
```diff
@@ -76,6 +106,44 @@ void test_wkt(std::string const& wkt, std::string const& expected,
...
+void test_wkt_to_from(std::string const& wkt, std::string const& expected,
+              std::size_t n, double len = 0, double ar = 0, double peri = 0)
+{
+    G geometry;
+
+    geometry = bg::from_wkt<G>(wkt);
+
+    /*
+    std::cout << "n=" << bg::num_points(geometry)
```
vissarion: you could use  `#ifdef BOOST_GEOMETRY_TEST_DEBUG` here instead of commenting out

sudo-panda: @vissarion I will do that, thanks for the suggestions.

include/boost/geometry/io/wkt/write.hpp
awulkiew: I'm also ok with an extra parameter.

The only problem I see is that a user may want to pass a set of such parameters, i.e. stream manipulators for coordinates. But this problem is solvable by e.g. allowing to optionally pass a `std::pair` or `std::tuple` of manipulators in the future:

    bg::to_wkt(geometry);
    bg::to_wkt(geometry, int_precision_10);
    bg::to_wkt(geometry, std::make_pair(std::fixed, std::setprecision(10)));

unless you have a better idea of an extensible interface. Just have in mind that in the future we may have to allow passing a strategy into this function as with any other function.

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +64,94 @@ namespace boost { namespace geometry
...
+    static void append(std::string& out, T const& val, int significant_digits = -1)
+    {
+        BOOST_STATIC_ASSERT((boost::is_arithmetic<T>::value));
+        std::stringstream ss;
+        if(std::numeric_limits<T>::is_specialized)
+        {
+            if(significant_digits < 0)
+                ss.precision(std::numeric_limits<T>::digits10);
+            else
+                ss.precision(significant_digits);
```
awulkiew: The precision is set only if `numeric_limits` are specialized for type `T`.

mloskot: @awulkiew Do you mean this?

```cpp
if(significant_digits < 0 && std::numeric_limits<T>::digits10 > 0) // or std::numeric_limits<T>::is_specialized as equivalent
    ss.precision(std::numeric_limits<T>::digits10);
```

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +64,94 @@ namespace boost { namespace geometry
...
+};
+
+template <>
+struct output_formatter<std::ostream>
+{
+    template <typename T>
+    static void append(std::ostream& out, T const& val, int significant_digits = -1)
+    {
+        BOOST_STATIC_ASSERT((boost::is_arithmetic<T>::value));
+        if(std::numeric_limits<T>::is_specialized)
```
awulkiew: Setting precision is only needed for string output isn't it? Why is it here? Did I miss something?

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +64,94 @@ namespace boost { namespace geometry
...
+    }
+
+    static void append(std::string& out, char const* s, int = -1)
+    {
+        out += s;
+    }
+};
+
+template <>
+struct output_formatter<std::ostream>
```
awulkiew: Why is this specialization needed?

include/boost/geometry/io/wkt/write.hpp
mloskot: Why not just let user give own pre-configured stream?
```
to_wkt(std::ostream& os, geometry);
```

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -515,40 +596,60 @@ class wkt_manipulator
...
 \brief Main WKT-streaming function
 \tparam Geometry \tparam_geometry
 \param geometry \param_geometry
+\param significant_digits Specifies the no of significant digits to use in the output wkt
 \ingroup wkt
 \qbk{[include reference/io/wkt.qbk]}
 */
 template <typename Geometry>
-inline wkt_manipulator<Geometry> wkt(Geometry const& geometry)
+inline wkt_manipulator<Geometry> wkt(Geometry const& geometry, int significant_digits = -1)
```
awulkiew: This is not needed here, a user can set precision the old way, using `std::setprecision` for streams.
Am I missing some discussion about it?

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -515,40 +596,60 @@ class wkt_manipulator
...
+/*!
+\brief Main WKT-string formulating function
+\tparam Geometry \tparam_geometry
+\param geometry \param_geometry
+\param significant_digits Specifies the no of significant digits to use in the output wkt
+\ingroup wkt
+\qbk{[include reference/io/to_wkt.qbk]}
+*/
+template <typename Geometry>
+inline std::string to_wkt(Geometry const& geometry, int significant_digits = -1)
```
awulkiew: I don't like this design with a magic number. Shouldn't there be 2 separate overloads of `to_wkt`?
@mloskot @vissarion what is your opinion?

mloskot: The magic of the number is to encode boolean indicator and actual value with single parameter, to distinguish `<0` for not set (use default) or `>=0` for set (use given).

I don't mind separate overloads.

include/boost/geometry/io/wkt/write.hpp
mloskot: No, I don't think you are. I think it slipped through my review unnoticed

include/boost/geometry/io/wkt/write.hpp
sudo-panda: What is the final call then? Should I implement seperate overloads?

include/boost/geometry/io/wkt/write.hpp
sudo-panda: In my opinion if we are providing the option to users to set significant digits in `to_wkt` it would seem symmetric to let users set it in `wkt` too. What do you think @awulkiew and @mloskot?

include/boost/geometry/io/wkt/write.hpp
mloskot: No, you didn't miss anything.

It's just that both, `wkt` stream manipulator and `to_wkt` function use the same formatter implementation, we just got carried away and applied the dedicated argument a bit too far indeed.

include/boost/geometry/io/wkt/write.hpp
mloskot: @sudo-panda Yes, it implicitly seemed logical and that's why it had not been caught..

I don't think it is a major flaw in design as for some  this may be more convenient to stream's `precision` call:

```cpp
std::cout << wkt(g, n)
```
On the other hand, that interface is not as capable as the stream's interface, it does not allow other manipulators. 

Moreover, we want to avoid confusing users allowing them to do

```cpp
std::cout.precision(n1);
std::cout << wkt(g, n2)
```

That would raise unnecessary questions, wouldn't it?

AFAIU @awulkiew, we don't want to compete with `std::ostream` features.

include/boost/geometry/io/wkt/write.hpp
awulkiew: @mloskot I mean that if a user passes `significant_digits` it means he/she wants it to be set in all cases, i.e.:

    if (significant_digits >= 0)
        ss.precision(significant_digits);
    else if (std::numeric_limits<T>::is_specialized)
        ss.precision(std::numeric_limits<T>::digits10);
        
or am I missing something?

mloskot: >  wants it to be set in all cases

...regardless if `std::numeric_limits<T>` is specialized or not.

Yes, correct.

include/boost/geometry/io/wkt/write.hpp
awulkiew: > AFAIU @awulkiew, we don't want to compete with std::ostream features.

@mloskot I think so. It doesn't make sense for me to allow passing the precision here. `bg::wkt` is an IO manipulator similar to `std::setprecision` itself. It has a clear job to do. It is not needed to semantically "merge" both manipulators together especially when we consider that `std::setprecision` is only one of the many that the user can use to tweak the output to his/her needs. See: https://en.cppreference.com/w/cpp/io/manip

IMO the correct usage is:

    std::cout << std::setprecision(n1) << wkt(g);
    // or
    std::cout << std::setprecision(n1) << std::fixed << std::nouppercase << wkt(g);

I rather see the additional parameter of `bg::to_wkt()` as a "convenient workaround", not as an example of "good design" that should be replicated in other places. To the extent that I'd prefer to remove it, only support the default and force the user to use `bg::wkt` if something else was needed. For a long time we didn't even have `to_wkt` because implementing it with `bg::wkt` is trivial if you really need something like that. But I'm ok with this additional parameter for users convenience if you think this would be useful.

mloskot: @awulkiew 
>I rather see the additional parameter of bg::to_wkt() as a "convenient workaround", not as an example of "good design" that should be replicated in other places.

The `to_wkt` is just a simpler function offered for convenience and it has limited capabilities in comparison to the streams. It could be more limited as using default precision or it could be less limited as accepting the extra argument, but nothing else, no other equivalents to the manipulators. I wouldn't call it a bad design, rather a choice, a tradeoff.

> I'm ok with this additional parameter for users convenience if you think this would be useful.

Although the low-level plumbing requires some adjustments to allow the porcelain interface,
I don't think it is a huge interference or it will impose maintenance hassle in future (or does it?). And some users prefer simplified variants of interfaces https://github.com/boostorg/geometry/issues/654#issuecomment-570955873

That's it, there is no deep thoughts behind any of that.

include/boost/geometry/io/wkt/write.hpp
awulkiew: Ok, so `significant_digits` is used here, and in specialization for `std::string`, but not in the default/unspecialized `output_formatter<Output>` (I'm not saying that it should be, I think it should only be used for `std::string`).

AFAIU this specialization for `std::ostream` was added exactly to support `significant_digits` for streams but it will not work for all of them, only for `std::basic_ostream<char>` as long as `std::ostream` is a typedef of `std::basic_ostream<char>` not a separate class because then the default/unspecialized `output_formatter<Output>` would be used.

Is it guaranteed by the standard that `std::ostream` is a typedef and not a separate class?

So anyway, assuming it is, this will be done only for streams using `char`.

I'd just remove the whole specialization together with `significant_digits` from `bg::wkt` stream manipulator. Is my reasoning ok?

include/boost/geometry/io/wkt/write.hpp
awulkiew: I'm ok with leaving it as it is or removing the parameter at all forcing the user to use `bg::wkt` if he/she needs something different than the default.

@barendgehrels do you have an opinion about it?

include/boost/geometry/io/wkt/write.hpp
awulkiew: > I wouldn't call it a bad design, rather a choice, a tradeoff.

Me neither. I don't think it's good. It doesn't mean that it's bad ;). It's just straightforward, ad-hoc solution, not very elegant. We have several functions like that so I'm ok with it. I'm talking about `to_wkt` here. I still think that `wkt` should not have this parameter.

> I don't think it is a huge interference or it will impose maintenance hassle in future (or does it?)

No, I don't think so.

mloskot: > I still think that wkt should not have this parameter.

Yes, I agree. I think it's a good catch you've made. I admit, it's my bad I missed that is unnecessary indeed.

include/boost/geometry/io/wkt/write.hpp
mloskot: > I'd just remove the whole specialization together with significant_digits from bg::wkt stream manipulator. Is my reasoning ok?

@awulkiew Yes, as I explained in https://github.com/boostorg/geometry/pull/670#discussion_r394002323, you've made good catch of this unnecessary change in the `wkt` manipulator.

@sudo-panda I'm sorry I had not caught that earlier myself. I somehow concentrated on the `to_wkt` addition and overlooked how it affects the existing manipulator.

sudo-panda: @mloskot no need to say sorry. I guess sometimes you need someone not involved in the process to look at the bigger picture. I am here to learn no matter how long it takes. 
@awulkiew  thanks for the review. I will push all these changes soon.

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -62,22 +64,94 @@ namespace boost { namespace geometry
...
+template <>
+struct output_formatter<std::ostream>
+{
+    template <typename T>
+    static void append(std::ostream& out, T const& val, int significant_digits = -1)
+    {
+        BOOST_STATIC_ASSERT((boost::is_arithmetic<T>::value));
+        if(std::numeric_limits<T>::is_specialized)
+        {
+            if(significant_digits < 0)
```
vissarion: a space after `if` is needed, please also use brackets for `if...else`

vissarion: there are other cases of missing spaces after `if`, so please address them as well

include/boost/geometry/io/wkt/write.hpp
vissarion: Although, I do not have strong opinion, here is my suggestion in order of preference: different overloads, leaving as is, totally removing the parameter.

sudo-panda:
I had to force push because of wrong commit message. Will be more careful in the future.

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -113,12 +160,13 @@ struct double_closing_parenthesis
 template <typename Point, typename Policy>
 struct wkt_point
 {
-    template <typename Char, typename Traits>
-    static inline void apply(std::basic_ostream<Char, Traits>& os, Point const& p, bool)
+    template <typename Output>
+    static inline void apply(Output& out, Point const& p, int significant_digits, bool)
```
barendgehrels: No, please don't do this.

There is more to stream than significant digits alone.

The original implementation should stay untouched, we should just be able to stream using any stream operator there is.

I thought that a wrapper should be build on top, such that we could stream to a string. But that is not this implementation.

barendgehrels:
To be clear, please don't merge this. It is not good.

include/boost/geometry/io/wkt/write.hpp
mloskot: > No, please don't do this.

Do you mean having the `int significant_digits`  which is ignored in all cases but `std::string`?

sudo-panda: @mloskot that was my first proposal. But after that we agreed that it would be better create the string from the base and not just casting it from the stringstream to a string as mentioned [here](https://github.com/boostorg/geometry/issues/654#issuecomment-592458473). But in my opinion everybody should agree on one after hearing all arguments. So can you explicitly state what is the better approach according to you @barendgehrels ?

doc/reference/io/to_wkt.qbk
```diff
@@ -0,0 +1,24 @@
+[/============================================================================
+  Boost.Geometry (aka GGL, Generic Geometry Library)
+
+  Copyright (c) 2009-2014 Barend Gehrels, Amsterdam, the Netherlands.
```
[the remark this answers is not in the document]
sudo-panda: How much of a file do I need to edit to add my Copyright?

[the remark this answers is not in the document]
mloskot: 3rd in the FAQ here https://www.boost.org/users/license.html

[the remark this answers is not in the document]
sudo-panda: It says it's upto the author but what are your opinions on it?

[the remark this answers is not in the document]
mloskot: There is no universal formula as this is highly case by case specific. I do this:
- I make my own judgement
- I add my copyright (or not)
- I rely on reviewers to object (or not) if my change is eligible for the copyright update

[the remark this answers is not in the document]
sudo-panda: Ok, thanks

include/boost/geometry/io/wkt/write.hpp
awulkiew: @barendgehrels the stream-based implementation should not have been touched, hence my objection above about additional parameters to `bg::wkt()`.

@barendgehrels could you clarify what is the exact problem with this PR?
Is the problem that passing a dummy `significant_digits` into stream-based implementation is not elegant or that passing `significant_digits` into `bg::to_wkt()` at all is not elegant or something else?
Is the argument that there should be more parameters in `bg::to_wkt()` or no parameters at all or that there should be no `bg::to_wkt()`?

Note that if we had this simple version of `bg::to_wkt()` function mentioned by @mloskot then we'd have the same problems. It's only a matter of implementation.

@barendgehrels how about removing `significant_digits` parameter and handle the coordinates always in a default way in `bg::to_wkt()`?

barendgehrels: > 
> 
> @barendgehrels
> 
> > No, please don't do this.
> 
> Do you mean having the `int significant_digits` which is ignored in all cases but `std::string`?
> 
> > I thought that a wrapper should be build on top, such that we could stream to a string.
> 
> Do you mean expected implementation is something like this?
> 
> ```c++
> template <typename Geometry>
> inline std::string to_wkt(Geometry const& geometry, int significant_digits)
> {
>     std::stringstream ss;
>     ss.precision(significant_digits);
>     ss << geometry;
>     return ss.str();
> }
> ```

Yes, I had expected exactly this.

barendgehrels: Thanks for your reactions. The significant_digits not the only issue, the problem is that I don't see a reason to change the implementation at all. If there are good reasons - OK of course. But if significant_digits is the reason, and streaming to a string - that is sooo easy to solve in a different way (see proposal of Mateusz) that I really don't see the reason to add this extra complexity.

barendgehrels: > 
> 
> @mloskot that was my first proposal. But after that we agreed that it would be better create the string from the base and not just casting it from the stringstream to a string as mentioned [here](https://github.com/boostorg/geometry/issues/654#issuecomment-592458473). But in my opinion everybody should agree on one after hearing all arguments. So can you explicitly state what is the better approach according to you @barendgehrels ?

First of all, thanks of course for your efforts! And sorry if my comment came in late an unexpectedly.

To be honest, I missed this suggestion earlier, and I wonder why it would be better. Is the performance better? Both approaches use stringstream internally, so I don't expect significant differences. And adding significant_digits only quite limits it.

OK - so you followed the suggestion and now I call you back... I like this comment of Mateusz there ("My rule of thumb is to always start with option that is the least intrusive to the existing code.") and I wonder why we didn't follow it...

awulkiew: @barendgehrels this was my suggestion based on the fact that I've made a simple benchmark and the result was that `std::string::append()` and `std::string::operator+=` was faster than `std::stringstream::operator<<` (as I mentioned on the Issue page). My test was rather synthetic but now we have an implementation so @sudo-panda could compare the speed of `bg::wkt` and `bg::to_wkt` and make a case that using `std::string` directly is indeed faster, or not.

My additional reasoning was/is as follows:
1. The simple function above is so trivial that the user could implement it using whatever stream manipulators that he/she needs. If we want to add a function like `to_wkt`, it only makes sense for me if our implementation could in some way beat this trivial user-level implementation, e.g. was faster.
2. If we provide a support for some kind of output then we should use the native way of using this kind of output.
3. Adding a layer of abstraction above appending data to streams and strings is of little complexity WRT what we already have in the library.

awulkiew:
@barendgehrels

> Both approaches use stringstream internally, so I don't expect significant differences. 

Only for coordinates. For characters like `','` and native strings like `"MULTIPOLYGON("` `std::stringstream` is not used but the characters are directly appended to `std::string`. And nothing stands in the way of detecting native coordinate types and using e.g. C++11 `std::to_string` for them. I proposed to use `std::stringstream` only in case some user had his own user-defined coordinate type and had `operator<<` defined for this coordinate type, then it would be supported out of the box.

Note that the above means that we should probably get rid of `significant_digits` parameter even for `std::string` because some output methods may not support it like e.g. Streams do. E.g. standard functions does not take this parameter as well as `boost::lexical_cast`.

> And adding significant_digits only quite limits it.

This is only an unelegant implementation where this variable is passed for both Strings and Streams but is ignored by the latter. See the default one for stream:
https://github.com/boostorg/geometry/blob/3c62182c82e294d3446173187126ffb0ef6cf513/include/boost/geometry/io/wkt/write.hpp#L73

The more "proper" way would be to e.g. pass `OutputPolicy` into `devarianted_wkt` and this policy would either be empty for Streams or contained whatever data is needed for Strings. But as I mentioned above I think we should consider simply removing this parameter altogether because we may decide to use something different for coordinates in case of Strings.

include/boost/geometry/io/wkt/write.hpp
sudo-panda: So which design are we going with?

barendgehrels: > 
> 
> So which design are we going with?

@awulkiew  thanks for your explanation. I remember now you mentioned this before, but I didn't realize it would have these consequences.

@sudo-panda could you compare the speed between the two implementations as @awulkiew  suggested?

Given all discussions, if you are all in favour of this new implementation (but there were still some discussions about the significant digits), I'm not against and you can go ahead.

sudo-panda: Sure. I would love to benchmark them.

awulkiew: @sudo-panda could you also benchmark alternative implementations of appending coordinates?

If you want your benchmark to be quite full you could consider checking at least the following:
```
std::stringstream ss;
ss.precision(std::numeric_limits<T>::digits10);
ss << val;
out += ss.str();
```
and
```
std::stringstream ss;
// without precision
ss << val;
out += ss.str();
```
and
```
out += boost::lexical_cast<std::string>(val);
```
and
```
out += std::to_string(val);
```

All of the above for:
1. geometry containing only coordinates having little number of significant digits, e.g. 0, 1, etc.
2. geometry containing only coordinates having many significant digits, e.g. 0.333333333333, 0.666666666667, 1/3, 2/3, sqrt(2), pi, etc.

All of the above for geometries having less coordinates/points and more in order to check different coordinate/character ratios in WKT strings, so e.g. to have several MULTIPOINTs containing 1, 10, 100 and 1000 points (at least 2 for 1 and 100?).

So at least 16 combinations.

sudo-panda: ### Result:

I made a heat map on plot.ly. You can access it [here](https://plot.ly/~sudo-panda/3/).

Method Used | No of Points | Time Taken
-------------------|------------------|----------------
stringstream w precision, < sig dgts | 1 points| 0
stringstream w precision, > sig dgts | 1 points| 3.2
stringstream w precision, < sig dgts | 10 points| 26.86
stringstream w precision, > sig dgts | 10 points| 26.16
stringstream w precision, < sig dgts | 100 points| 249.81
stringstream w precision, > sig dgts | 100 points| 265.87
stringstream w precision, < sig dgts | 1000 points| 2496.11
stringstream w precision, > sig dgts | 1000 points| 2616.56
stringstream w precision, < sig dgts | 10000 points| 24895.1
stringstream w precision, > sig dgts | 10000 points| 26237.3
stringstream w/o precision, < sig dgts | 1 points| 0
stringstream w/o precision, > sig dgts | 1 points| 3.37
stringstream w/o precision, < sig dgts | 10 points| 66.75
stringstream w/o precision, > sig dgts | 10 points| 59.92
stringstream w/o precision, < sig dgts | 100 points| 273.41
stringstream w/o precision, > sig dgts | 100 points| 219.3
stringstream w/o precision, < sig dgts | 1000 points| 2625.64
stringstream w/o precision, > sig dgts | 1000 points| 2816.33
stringstream w/o precision, < sig dgts | 10000 points| 21972.1
stringstream w/o precision, > sig dgts | 10000 points| 17633.8
lexical cast, < sig dgts | 1 points| 0
lexical cast, > sig dgts | 1 points| 1.73
lexical cast, < sig dgts | 10 points| 15
lexical cast, > sig dgts | 10 points| 64.22
lexical cast, < sig dgts | 100 points| 600.01
lexical cast, > sig dgts | 100 points| 379.06
lexical cast, < sig dgts | 1000 points| 2525.91
lexical cast, > sig dgts | 1000 points| 2541.71
lexical cast, < sig dgts | 10000 points| 19197.2
lexical cast, > sig dgts | 10000 points| 19841.5
to_string, < sig dgts | 1 points| 0
to_string, > sig dgts | 1 points| 1.02
to_string, < sig dgts | 10 points| 11.17
to_string, > sig dgts | 10 points| 9.03
to_string, < sig dgts | 100 points| 129.17
to_string, > sig dgts | 100 points| 125.14
to_string, < sig dgts | 1000 points| 1065.07
to_string, > sig dgts | 1000 points| 973.36
to_string, < sig dgts | 10000 points| 9447.4
to_string, > sig dgts | 10000 points| 8865.27



> ***N.B.:*** Values for `< sig dgts` with just *one point* was giving time taken values as *9.41413e+11* so i set them to *0*.

The code used to benchmark the methods :
``` C++
#include <iostream>
#include <ctime>
#include <cstdlib>
#include <cmath>
#include <chrono> 
#include <boost/geometry.hpp>
#include <boost/geometry/geometries/multi_point.hpp>

int impr_to_wkt(int no_of_points)
{
    namespace geom = boost::geometry;
    typedef geom::model::d2::point_xy<double> point_type;

    geom::model::multi_point<point_type> multi_pt;
    for(int i = 0; i < no_of_points; i++)
    {
        srand( unsigned(i*time(NULL)));
        geom::append(multi_pt, geom::make<point_type>(rand() % 1000 / 100.0, rand() % 1000 / 100.0));
    }
        

    std::string out;
    std::chrono::_V2::system_clock::time_point start = std::chrono::high_resolution_clock::now();
    out = geom::to_wkt(multi_pt);
    std::chrono::_V2::system_clock::time_point stop = std::chrono::high_resolution_clock::now();
    std::cerr<<out.length()<<"\t";

    return std::chrono::duration_cast<std::chrono::microseconds>(stop - start).count();
}

int prec_to_wkt(int no_of_points)
{
    namespace geom = boost::geometry;
    typedef geom::model::d2::point_xy<double> point_type;

    geom::model::multi_point<point_type> multi_pt;
    for(int i = 0; i < no_of_points; i++)
    {
        srand( unsigned(i*time(NULL)));
        geom::append(multi_pt, geom::make<point_type>((double)rand()/(RAND_MAX), (double)rand()/(RAND_MAX)));
    }
    
    std::string out;
    std::chrono::_V2::system_clock::time_point start = std::chrono::high_resolution_clock::now();
    out = geom::to_wkt(multi_pt);
    std::chrono::_V2::system_clock::time_point stop = std::chrono::high_resolution_clock::now();
    std::cerr<<out.length()<<"\t";

    return std::chrono::duration_cast<std::chrono::microseconds>(stop - start).count();
}

int main()
{
    for(int i = 0; i < 5; i++)
    {
        long long sum;
        for(int j = 0 ; j<100; j++)
        {
            sum += impr_to_wkt(pow(10, i));
        }
        std::cout<<"stringstream w precision, "<<" < sig dgts, "<<(int) pow(10, i)<<" points, "<<sum/100.0<<std::endl;
        sum = 0;
        for(int j = 0 ; j<100; j++)
        {
            sum += prec_to_wkt(pow(10, i));
        }
        std::cout<<"stringstream w precision, "<<" > sig dgts, "<<(int) pow(10, i)<<" points, "<<sum/100.0<<std::endl;
    }
}
```

### Inference:
Time taken by `to_string` is the least followed by `lexical_cast`. The `stringstream` method is the slowest When the precision is set, it makes it slower.

awulkiew: @sudo-panda thanks! I don't see the benchmark for `bg::wkt()`. Have you tested it too?

Do I see correctly that stringstream with precision set to digits10 is faster than the one without it? That's wierd.

awulkiew: What compiler version and options have you used?

sudo-panda: > I don't see the benchmark for `bg::wkt()`. Have you tested it too?

I haven't tested it.

> Do I see correctly that stringstream with precision set to digits10 is faster than the one without it? That's wierd.

No `stringstream` with precision is slower than the one without as I said:

>Time taken by `to_string` is the least followed by `lexical_cast`. The `stringstream` method is the slowest When the precision is set, it makes it slower.

It can also be seen in the heat map provided.

> What compiler version and options have you used?

I have used the default `b2` configuration which uses `gcc-7` in my case. As for the options can you tell me how can I find that?

mloskot: @sudo-panda
>  As for the options can you tell me how can I find that?

You definitally want to benchmark **optimised** build, not debug (which is default), that is:

```
b2 variant=release address-model=32
b2 variant=release address-model=64
```

More options at https://boostorg.github.io/build/manual/develop/index.html#bbv2.overview.invocation.properties

awulkiew: @sudo-panda

> No stringstream with precision is slower

Only for geometries with 10000 points and 100 > sig dgts.

> I have used the default b2 configuration

Ok, so the results are irrelevant for now.

I suggest to NOT use b2 at all and to build your program directly with a compiler and pass -O2 or -O3 option explicitly.

mloskot: > I suggest to NOT use b2 at all and to build your program directly with a compiler and pass -O2 or -O3 option explicitly.

I'm not advocating for b2, but for completeness of the picture, it is also allows that:

```
b2 variant=release address-model=64 optimization=space
b2 variant=release address-model=64 optimization=speed
b2 variant=release address-model=64 optimization=full
```

If you want to see actual compiler command line invoked, just add `-d2 as in `b2 -d2 variant=release ....`

mloskot: I have added `wkt` and `to_wkt` operations to my string benchmark in https://github.com/mloskot/string_benchmark/blob/master/benchmark_boost_geometry_wkt.cpp, where

- `std::to_string` is the baseline
- problem space means number of points (pairs of X/Y)

### GCC 9

CSV files with results: https://github.com/mloskot/string_benchmark/blob/master/results/gcc9_geometry_wkt.csv

image

### MSVC++ 14.25 (VS2019)

CSV files with results: https://github.com/mloskot/string_benchmark/blob/master/results/vs2019_geometry_wkt.csv

image

------

I picked up @awulkiew's theme on speeding things up because I think speed is important here. Adam's suggestion to re-work the internals opened possibility to completely avoid using streams for `to_wkt` and use of `std:;string` only, even with size `reserve`-ed heuristically based on number of points and dimension of geometry to further tweak performance (along with measuring it, of course). I hope this helps.

sudo-panda: What is the difference in `wkt` and `stringstream` experiments?

mloskot: @sudo-panda
> What is the difference in `wkt` and `stringstream` experiments?

Check the very _trivial_ code of [benchmark_boost_geometry_wkt.cpp](https://github.com/mloskot/string_benchmark/blob/master/benchmark_boost_geometry_wkt.cpp), you will see immediately. The `to_string`, `stringstream` and `lexical_cast` are generic cases outputting **N** points then `wkt` and `to_wkt` do similar for `multi_point` of **N** points, where  **N** is the **Problem Space** number of points (pairs of X/Y). When you compare results, compare test cases of the the same problem spaces.

sudo-panda: Ok I got confused thinking that you were using them in the `output_formatter` rather than benchmarking then standalone.

mloskot: No, I did not apply any modifications to the Boost.Geometry (code of this PR) at all.

I just wrote quick benchmark which just compares

1. 3 x basic general-purpose methods to output list of `double`-s to string methods: `to_string`, `stringstream` and `lexical_cast`
2. 2 x Boost.Geometry methods: `wkt` and `to_wkt`

I think of the 1. as reference results and I picked `to_string` as the single baseline to contrast all other methods against..
I think of the 2. as results of interest for further optimisation.

So, next, the benchmark could be completed with the Geometry's methods further tweaked, optimised, etc.

sudo-panda: Sorry for the late reply.

I tried celero for the first time and the tool is really great although I don't know what to make of the varied results I got.

I have uploaded my code to my repo: https://github.com/sudo-panda/bg_wkt_benchmark

The compiler I have used is GNU C++ 7.5.0.

I have also tried both O2 and O3 and an unoptimized version(as stated in celero docs)

You can find the results of the respective operations in the results directory of my repo. I have also formatted it and sorted it in a google sheet and you can find that here: https://github.com/sudo-panda/bg_wkt_benchmarkhttps://docs.google.com/spreadsheets/d/1AYqr0Unob5Ee6_-6ThEg3Z3iROzOuE01tmLQ46_JJis/edit?usp=sharing

mloskot: @sudo-pand I didjust quick look, but I'm not sure.
I would suggest you to let Celero do the statistics, remove `30` and stick `0`, then bump the `1` to more meaningful size.  You will also notice that I deliberately use `0` for number of iterations in `ExperimentValues`. Another thing is that I'm not sure about Celero's statistics when you force baseline to no-op. If you want to control benchmark with hard-wired runs, iterations, etc. I'd suggest you to use https://github.com/google/benchmark or https://github.com/nickbruun/hayai instead. Benchmarking needs time. My variant is simple and I think it gives quite clear results, but it does not include all test cases we may be interested in.

sudo-panda: I made the changes you suggested but I still can't find the method that is clearly better than the others. The results are a bit mixed. The changes are uploaded to the develop branch. 

https://github.com/sudo-panda/bg_wkt_benchmark/tree/develop

I also updated the google sheet with new results:

https://docs.google.com/spreadsheets/d/1AYqr0Unob5Ee6_-6ThEg3Z3iROzOuE01tmLQ46_JJis/edit?usp=sharing

mloskot:
(I'm moving the thread https://github.com/boostorg/geometry/pull/670#discussion_r401816305 to main stream of comments as I need more horizontal space :))

### Benchmark update

@sudo-panda I played a bit more with benchmarking to better understand where your implementation sits versus the stream manipulator `wkt` and `to_wkt` as the basic wrapper.

First, I defined `to_wkt` as the basic wrapper according to what we discussed with @barendgehrels in https://github.com/boostorg/geometry/pull/670#discussion_r394707569

```cpp
template <typename Geometry>
inline std::string to_wkt_wrapper(Geometry const& geometry)
{
    std::stringstream ss;
    ss << boost::geometry::wkt(geometry);
    return ss.str();
}
template <typename Geometry>
inline std::string to_wkt_wrapper(Geometry const& geometry, int significant_digits)
{
    std::stringstream ss;
    ss.precision(significant_digits);
    ss << boost::geometry::wkt(geometry);
    return ss.str();
}
```

then I added that to my [benchmark_boost_geometry_wkt.cpp](https://github.com/mloskot/string_benchmark/blob/master/benchmark_boost_geometry_wkt.cpp) as `to_wkt_wrap`.

Here is the result:

```
|     Group      |   Experiment    |   Prob. Space   |     Samples     |   Iterations    |    Baseline     |  us/Iteration   | Iterations/sec  |
|:--------------:|:---------------:|:---------------:|:---------------:|:---------------:|:---------------:|:---------------:|:---------------:|
|wkt             | to_string       |               1 |              30 |          262144 |         1.00000 |         0.90276 |      1107714.67 |
|wkt             | to_string       |             256 |              30 |              16 |         1.00000 |       245.25000 |         4077.47 |
|wkt             | to_string       |             512 |              30 |               8 |         1.00000 |       490.37500 |         2039.26 |
|wkt             | to_string       |            1024 |              30 |               4 |         1.00000 |       982.00000 |         1018.33 |
|wkt             | stringstream    |               1 |              30 |          131072 |         1.66540 |         1.50346 |       665134.15 |
|wkt             | stringstream    |             256 |              30 |              16 |         1.26988 |       311.43750 |         3210.92 |
|wkt             | stringstream    |             512 |              30 |               8 |         1.24369 |       609.87500 |         1639.68 |
|wkt             | stringstream    |            1024 |              30 |               4 |         1.25305 |      1230.50000 |          812.68 |
|wkt             | lexical_cast    |               1 |              30 |          131072 |         1.76430 |         1.59274 |       627847.71 |
|wkt             | lexical_cast    |             256 |              30 |              16 |         1.40979 |       345.75000 |         2892.26 |
|wkt             | lexical_cast    |             512 |              30 |               8 |         1.40836 |       690.62500 |         1447.96 |
|wkt             | lexical_cast    |            1024 |              30 |               4 |         1.40860 |      1383.25000 |          722.94 |
|wkt             | wkt             |               1 |              30 |          131072 |         1.99188 |         1.79819 |       556115.61 |
|wkt             | wkt             |             256 |              30 |              16 |         1.49847 |       367.50000 |         2721.09 |
|wkt             | wkt             |             512 |              30 |               8 |         1.41601 |       694.37500 |         1440.14 |
|wkt             | wkt             |            1024 |              30 |               4 |         1.40657 |      1381.25000 |          723.98 |
|wkt             | to_wkt_wrap     |               1 |              30 |          131072 |         1.96920 |         1.77772 |       562519.04 |
|wkt             | to_wkt_wrap     |             256 |              30 |              16 |         1.42635 |       349.81250 |         2858.67 |
|wkt             | to_wkt_wrap     |             512 |              30 |               8 |         1.43946 |       705.87500 |         1416.68 |
|wkt             | to_wkt_wrap     |            1024 |              30 |               4 |         1.39282 |      1367.75000 |          731.13 |
|wkt             | to_wkt_wrap_3di |               1 |              30 |          131072 |         1.90817 |         1.72262 |       580511.72 |
|wkt             | to_wkt_wrap_3di |             256 |              30 |              16 |         1.33384 |       327.12500 |         3056.94 |
|wkt             | to_wkt_wrap_3di |             512 |              30 |               8 |         1.32858 |       651.50000 |         1534.92 |
|wkt             | to_wkt_wrap_3di |            1024 |              30 |               4 |         1.32612 |      1302.25000 |          767.90 |
|wkt             | to_wkt_wrap_9di |               1 |              30 |          131072 |         3.01452 |         2.72139 |       367459.31 |
|wkt             | to_wkt_wrap_9di |             256 |              30 |              16 |         1.56600 |       384.06250 |         2603.74 |
|wkt             | to_wkt_wrap_9di |             512 |              30 |               8 |         1.53887 |       754.62500 |         1325.16 |
|wkt             | to_wkt_wrap_9di |            1024 |              30 |               4 |         1.56263 |      1534.50000 |          651.68 |
|wkt             | to_wkt          |               1 |              30 |           65536 |         2.93074 |         2.64575 |       377964.38 |
|wkt             | to_wkt          |             256 |              30 |              16 |         2.43552 |       597.31250 |         1674.17 |
|wkt             | to_wkt          |             512 |              30 |               8 |         2.44099 |      1197.00000 |          835.42 |
|wkt             | to_wkt          |            1024 |              30 |               4 |         2.41802 |      2374.50000 |          421.14 |
|wkt             | to_wkt_3dig     |               1 |              30 |          131072 |         1.99573 |         1.80167 |       555041.75 |
|wkt             | to_wkt_3dig     |             256 |              30 |              16 |         1.84174 |       451.68750 |         2213.92 |
|wkt             | to_wkt_3dig     |             512 |              30 |               8 |         1.83533 |       900.00000 |         1111.11 |
|wkt             | to_wkt_3dig     |            1024 |              30 |               4 |         1.83656 |      1803.50000 |          554.48 |
|wkt             | to_wkt_9dig     |               1 |              30 |          131072 |         2.87525 |         2.59566 |       385258.86 |
|wkt             | to_wkt_9dig     |             256 |              30 |              16 |         2.00739 |       492.31250 |         2031.23 |
|wkt             | to_wkt_9dig     |             512 |              30 |               8 |         2.00408 |       982.75000 |         1017.55 |
|wkt             | to_wkt_9dig     |            1024 |              30 |               4 |         2.00178 |      1965.75000 |          508.71 |
```

Next, I tried a basic optimisation: do not use `std::stringstream` if (theoretically) not necessary. Here is my version of `output_formatter<std::string>::append<T>`:

```cpp
template <typename T>
static void append(std::string& out, T const& val, int significant_digits = -1)
{
    BOOST_STATIC_ASSERT((boost::is_arithmetic<T>::value));
    if (significant_digits >= 0)
    {
        std::stringstream ss;
        ss.precision(significant_digits);
        ss << val;
        out += ss.str();
    }
    else
    {
        out += std::to_string(val);
    }
}
```

Now, look at how this speeds up your `to_wkt` for the case of the default number of digits against the `to_wkt_wrap` from the results above as well as against `to_wkt_3dig` and `to_wkt_9dig` which go via `std::stringstream`:

```
|wkt             | to_wkt          |               1 |              30 |          131072 |         1.77079 |         1.64599 |       607537.67 |
|wkt             | to_wkt          |             256 |              30 |              16 |         1.04082 |       255.00000 |         3921.57 |
|wkt             | to_wkt          |             512 |              30 |               8 |         1.03724 |       508.25000 |         1967.54 |
|wkt             | to_wkt          |            1024 |              30 |               4 |         1.03931 |      1018.00000 |          982.32 |
|wkt             | to_wkt_3dig     |               1 |              30 |          131072 |         1.94660 |         1.80941 |       552666.31 |
|wkt             | to_wkt_3dig     |             256 |              30 |              16 |         1.85459 |       454.37500 |         2200.83 |
|wkt             | to_wkt_3dig     |             512 |              30 |               8 |         1.92270 |       942.12500 |         1061.43 |
|wkt             | to_wkt_3dig     |            1024 |              30 |               4 |         1.91756 |      1878.25000 |          532.41 |
|wkt             | to_wkt_9dig     |               1 |              30 |          131072 |         2.82303 |         2.62407 |       381087.51 |
|wkt             | to_wkt_9dig     |             256 |              30 |              16 |         2.02883 |       497.06250 |         2011.82 |
|wkt             | to_wkt_9dig     |             512 |              30 |               8 |         2.01913 |       989.37500 |         1010.74 |
|wkt             | to_wkt_9dig     |            1024 |              30 |               4 |         2.01174 |      1970.50000 |          507.49 |
```

### Observations

1. Your current `to_wkt` makes lots of calls to `std::stringstream::str()`, while the `to_wkt_wrap` makes single call to `str()` at the end, that makes the former much slower.
2. The modified `to_wkt_` using `to_string` is fast becasue `to_string` calls `sprintf` (or equivalent) and not `strignstream`.

Regarding the 2., on MSVC++, it calls `scprintf` to get the required length, then constructs `string s(length, '\0')` as buffer, then `sprintf(&s[0], len + 1, "%f", value)` to convert `value` into string buffer `s`.

### Conclusion

If `output_formatter<std::string>` is implemented in terms of `sprintf` it will offer best performance which is also ~50% better performance than `to_wkt_wrap` implemented as thin wrapper on top of the `wkt` stream manipulator.

Folks, please, let me know if my reasoning is correct or not.

include/boost/geometry/io/wkt/write.hpp
mloskot: @sudo-panda I posted benchmarks update in the main thread of comments here https://github.com/boostorg/geometry/pull/670#issuecomment-620277337

sudo-panda:
@mloskot for now I am making `to_wkt` as the wrapper as it seems faster. As for the design of `output_formatter ` I will put another PR if other reviewers agree to follow through.

And thanks for putting in so much effort. :)

mloskot:
@sudo-panda 
> for now I am making to_wkt into a wrapper as it seems faster. 

Yes, I personally think, a solution that is not slower than the stream manipulator is good enough.

Optimising with `sprintf`-like way to avoid `stringstream` would require more work.

sudo-panda:
@barendgehrels , @awulkiew could you please review it so that it can be merged.

mloskot (approved):
AFAICT, this has now become a small addition that is not intrusive to the existing WKT I/O interface and operations, so I think it can be merged.

/cc @barendgehrels @awulkiew @vissarion @sudo-panda

barendgehrels:
> 
> 
> AFAICT, this has now become a small addition that is not intrusive to the existing WKT I/O interface and operations, so I think it can be merged.
> 
> /cc @barendgehrels @awulkiew @vissarion @sudo-panda

Yes, it's not intrusive so that is a good thing.

But I'm wondering what it adds now. On top of of `std::cout << bg::wkt(x);` which was there for 10 years, a stream operator as it should be, with all its flexibility in terms of precision, fixed/exponential, width, etc, we get a bg::to_wkt(x,3) now which works indirectly (using a stringstream) and only supports precision. What does this add to the library? Why having two alternatives for the same functionality? Isn't that confusing?

And if not, shouldn't we build that then too for `dsv`, for example? (In my opinion not).

With respect to the `from` variant I: having a x = from_wkt<Type>(s) might be useful for some users, it can be written on one line instead of two (though the documentation doesn't provide this). But also here the question: is that worth to offer an additional variant? You can say: yes, because we also have that for `envelope`, for example. But there (and for some other functions) we use the convention `return_envelope` . So that would mean this should be called `return_wkt`. That would be acceptable for me.

sudo-panda:
> So that would mean this should be called `return_wkt`. That would be acceptable for me.

@barendgehrels Is that the only change required for this PR to be approved by you?

barendgehrels (commented):
> 
> 
> > So that would mean this should be called `return_wkt`. That would be acceptable for me.
> 
> @barendgehrels Is that the only change required for this PR to be approved by you?

Note, I scratched that comment right after. So `from_wkt` is still acceptable, `return_wkt` would be weird because it doesn't return the wkt itself (`return_envelope` for example does). Sorry for the confusion.

The PR is acceptable for me if `to_wkt` is removed, unless other reviewers find it a useful addition.

doc/src/examples/io/to_wkt.cpp
```diff
@@ -0,0 +1,64 @@
...
+    geom::model::polygon<point_type> polygon_frac;
+    geom::append(geom::exterior_ring(polygon_frac), geom::make<point_type>(0.00000, 0.00000));
+    geom::append(geom::exterior_ring(polygon_frac), geom::make<point_type>(0.00000, 4.00001));
+    geom::append(geom::exterior_ring(polygon_frac), geom::make<point_type>(4.00001, 4.00001));
+    geom::append(geom::exterior_ring(polygon_frac), geom::make<point_type>(4.00001, 0.00000));
+    geom::append(geom::exterior_ring(polygon_frac), geom::make<point_type>(0.00000, 0.00000));
+
+    std::cout << boost::geometry::to_wkt(point_frac, 3) << std::endl;
+    std::cout << boost::geometry::to_wkt(polygon_frac, 3) << std::endl;
+
```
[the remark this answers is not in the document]
sudo-panda: I am waiting for comments from other reviewers before I change anything w.r.t. `to_wkt`.

vissarion (commented):
I am also OK with merging most of this PR but (as @barendgehrels ) I do not understand exactly what  `to_wkt` adds on top of the stream operator.

awulkiew:
@barendgehrels @vissarion This PR implements this issue: https://github.com/boostorg/geometry/issues/654

vissarion:
> @barendgehrels @vissarion This PR implements this issue: #654

Right, thanks! So it seems there is user need so I will not go against. I am OK with this PR.

mloskot:
@vissarion, @barendgehrels
> 
> Right, thanks! So it seems there is user need so I will not go against. I am OK with this PR.

Yes, there seem to be a need indeed.

Although the lengthy discussion we've had on this PR may given an impression it's a big feature, it is not.
It just a little thing that smoothens up that one particular rough edge of the WKT I/O interface making it just a bit more expressive and fluent - every little helps :-)

barendgehrels:
> 
> 
> @vissarion, @barendgehrels
> 
> > Right, thanks! So it seems there is user need so I will not go against. I am OK with this PR.
> 
> Yes, there seem to be a need indeed.
> 
> Although the lengthy discussion we've had on this PR may given an impression it's a big feature, it is not.
> It just a little thing that smoothens up a particular rough edge of the WKT I/O interface making it just a bit more expressive and fluent. Apparently, for users, every little helps :-)

Thanks for the link. Yes, I remember now and I was OK back then, so I'm still OK now. I will approve.

barendgehrels (approved):

awulkiew (approved):

awulkiew:
I appologize it took so long. Thanks for this PR!

PR #675, opened 2020-03-03
Adding missing unit tests for geometry models

### LINKED ISSUE #673: Add unit tests for geometry models
This task is about adding new tests for basic classes like geometry models (e.g. `point`, `point_xy`, etc.) to exercise their basic operations.

Although we can be fairly sure most, if not all, are covered by the extensive tests of algorithms, there is a value of having separate unit tests for each of the models. Such tests clearly show (document!) what operations are actually tested. Not something that can be easily displayed from all the complex tests, unless one knows what to `grep` through the code for.

A good place to start is to look at and follow the unit test for `box` in [test/geometries/box.cpp](https://github.com/boostorg/geometry/blob/develop/test/geometries/box.cpp).

### References

Covering basic classes like geometry models with unit tests have been discussed here
https://github.com/boostorg/geometry/pull/671#issuecomment-593110240

digu-007:
This PR is meant to fix #673. I have added unit tests for all the geometry models. Test includes working of default and copy constructors, assignment operator and custom test using `std::initializer_list`. I have covered tests for all the coordinate systems. I have tested all the files locally using GCC 7.4.

### DIFF
--- a/test/geometries/linestring.cpp
+++ b/test/geometries/linestring.cpp
@@ -50,14 +50,29 @@ void check_linestring(L& to_check, T x, T y, T z)
 }
 
 template <typename P>
-void test_construction()
+void test_default_constructor()
 {
     bg::model::linestring<P> l1(create_linestring<P>());
     check_linestring(l1, 1, 2, 3);
 }
 
 template <typename P>
-void test_compilation()
+void test_copy_constructor()
+{
+    bg::model::linestring<P> l1 = create_linestring<P>();
+    check_linestring(l1, 1, 2, 3);
+}
+
+template <typename P>
+void test_copy_assignment()
+{
+    bg::model::linestring<P> l1(create_linestring<P>()), l2;
+    l2 = l1;
+    check_linestring(l2, 1, 2, 3);
+}
+
+template <typename P>
+void test_concept()
 {   
     typedef bg::model::linestring<P> L;
 
@@ -72,8 +87,10 @@ void test_compilation()
 template <typename P>
 void test_all()
 {   
-    test_construction<P>();
-    test_compilation<P>();
+    test_default_constructor<P>();
+    test_copy_constructor<P>();
+    test_copy_assignment<P>();
+    test_concept<P>();
 }
 
 template <typename P>
--- a/test/geometries/linestring.cpp
+++ b/test/geometries/linestring.cpp
@@ -3,9 +3,6 @@
 
 // Copyright (c) 2020 Digvijay Janartha, Hamirpur, India.
 
-// Parts of Boost.Geometry are redesigned from Geodan's Geographic Library
-// (geolib/GGL), copyright (c) 1995-2010 Geodan, Amsterdam, the Netherlands.
-
 // Use, modification and distribution is subject to the Boost Software License,
 // Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
 // http://www.boost.org/LICENSE_1_0.txt)
@@ -109,14 +106,24 @@ void test_custom()
     test_custom_linestring<P>(IL);
 }
 
+template <typename CS>
+void test_cs()
+{
+    test_all<bg::model::point<int, 3, CS> >();
+    test_all<bg::model::point<float, 3, CS> >();
+    test_all<bg::model::point<double, 3, CS> >();
+
+    test_custom<bg::model::point<double, 2, CS> >();
+}
+
 
 int test_main(int, char* [])
 {   
-    test_all<bg::model::point<int, 3, bg::cs::cartesian> >();
-    test_all<bg::model::point<float, 3, bg::cs::cartesian> >();
-    test_all<bg::model::point<double, 3, bg::cs::cartesian> >();
+    test_cs<bg::cs::cartesian>();
+    test_cs<bg::cs::spherical<bg::degree> >();
+    test_cs<bg::cs::spherical_equatorial<bg::degree> >();
+    test_cs<bg::cs::geographic<bg::degree> >();
 
-    test_custom<bg::model::point<double, 2, bg::cs::cartesian> >();
     test_custom<bg::model::d2::point_xy<double> >();
 
     return 0;
--- a/test/geometries/linestring.cpp
+++ b/test/geometries/linestring.cpp
@@ -1,0 +1,106 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+// Unit Test
+
+// Copyright (c) 2020 Digvijay Janartha, Hamirpur, India.
+
+// Parts of Boost.Geometry are redesigned from Geodan's Geographic Library
+// (geolib/GGL), copyright (c) 1995-2010 Geodan, Amsterdam, the Netherlands.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#include <iostream>
+
+#include <geometry_test_common.hpp>
+
+#include <boost/core/ignore_unused.hpp>
+#include <boost/geometry/algorithms/make.hpp>
+#include <boost/geometry/algorithms/append.hpp>
+#include <boost/geometry/geometries/point.hpp>
+#include <boost/geometry/geometries/point_xy.hpp>
+#include <boost/geometry/geometries/linestring.hpp>
+#include <boost/geometry/geometries/concepts/linestring_concept.hpp>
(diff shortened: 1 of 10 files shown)


digu-007:
I added unit test for two more models `point` and `linestring`.

Also, in [this](https://www.boost.org/doc/libs/1_72_0/libs/geometry/doc/html/geometry/reference/models/model_d2_point_xy.html) page same note is written twice, this is not deliberate right?

test/geometries/linestring.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+#include <boost/geometry/geometries/concepts/linestring_concept.hpp>
+
+
+#include <boost/geometry/geometries/adapted/c_array.hpp>
+#include <boost/geometry/geometries/adapted/boost_tuple.hpp>
+
+#include <boost/geometry/io/dsv/write.hpp>
+
+#include <test_common/test_point.hpp>
```
mloskot: That is too much of blank lines  :)

test/geometries/point.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <boost/geometry/geometries/concepts/point_concept.hpp>
+
+#include <boost/geometry/algorithms/make.hpp>
+
+
+#include <boost/geometry/geometries/point.hpp>
+#include <test_common/test_point.hpp>
+
+
+template <typename T>
```
mloskot: Please, remove the blanks

test/geometries/linestring.cpp
digu-007: I thought it was a convention to leave a line if the header is from different directory, like in [segment.cpp](https://github.com/boostorg/geometry/blob/develop/test/geometries/segment.cpp), but I'll fix this :)

mloskot: Not really, let's say it's unintentional legacy and any code should be crafted more carefully, compact, readable. See [#include directives shall be ordered...](https://github.com/boostorg/geometry/wiki/Guidelines-for-Developers#code-structure) guidelines.

digu-007:
@mloskot I removed the extra lines and now I am using constructor for `linestring` and `initializer_list` for custom test. I have added more functions for better clarity of code.

test/geometries/linestring.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+template <typename P>
+void test_construction()
+{
+    bg::model::linestring<P> l1(create_linestring<P>());
+    check_linestring(l1, 1, 2, 3);
+}
+
+template <typename P>
+void test_compilation()
```
mloskot: Should read `test_concept`

test/geometries/linestring.cpp
```diff
@@ -0,0 +1,106 @@
...
+template <typename L, typename T>
+void check_linestring(L& to_check, T x, T y, T z)
+{
+    BOOST_CHECK_EQUAL(bg::get<0>(to_check[0]), x);
+    BOOST_CHECK_EQUAL(bg::get<1>(to_check[0]), y);
+    BOOST_CHECK_EQUAL(bg::get<2>(to_check[0]), z);
+}
+
+template <typename P>
+void test_construction()
```
mloskot: This in fact only tests default construction. Since we are at the unit test, I was wondering about such test cases

- `test_default_constructor`
- `test_copy_constructor`
- `test_copy_assignment`

digu-007: > what's your opinion

Since it is unit tests, I will add `copy_constructor` and `copy_assignment` as well. I completely agree with you. Just one thing, adding so many assertions would make testing slow right? But not much I guess, what's the approximate complexity of these functions we use for testing in boost?

mloskot: > adding so many assertions would make testing slow right?

In our case the compilation time, due to code traits unrelated to use of assertions (e.g. meta-programming, header-only, etc.), is much much longer than run time of any of our tests, even if you stick hundreds of assertions.

> But not much I guess, what's the approximate complexity of these functions we use for testing in boost?

If you have `a == b`, complexity depends on type of the operands. If `a` and `b` are `int` it will be very different than if they are `std::vector<int>`, etc.

Performance is not our concern here, but correctness.

digu-007: > Performance is not our concern here, but correctness.

Yes, I agree. I'll add other functions as well in the next commit.

test/geometries/multi_linestring.cpp
```diff
@@ -0,0 +1,147 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+// Unit Test
+
+// Copyright (c) 2020 Digvijay Janartha, Hamirpur, India.
+
+// Parts of Boost.Geometry are redesigned from Geodan's Geographic Library
```
vissarion: Is this needed for completely new files?

test/geometries/multi_linestring.cpp
```diff
@@ -0,0 +1,147 @@
...
+    bg::model::linestring<P> l1;
+    bg::append(l1, IL1);
+    bg::append(l1, IL2);
+    test_custom_multi_linestring<P>(l1);
+}
+
+
+int test_main(int, char* [])
+{   
+    test_all<bg::model::point<int, 2, bg::cs::cartesian> >();
```
vissarion: I think it make sense to also test geometry models with the other coordinate systems

test/geometries/multi_linestring.cpp
digu-007: You mean line 6, shall I remove this in all the files?

mloskot: Good questions. I think the canonical form of the notice is just

```
//
// Copyright (c) 2020 Digvijay Janartha, Hamirpur, India.
//
```

The other copyrights are added as necessary by authors of further modifications, that is according to [How should Boost programmers maintain the copyright messages?](https://www.boost.org/users/license.html).

digu-007: Ok I'll change to this form in the next commit, alongside tests for other cs as well :)

test/geometries/linestring.cpp
```diff
@@ -0,0 +1,130 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+// Unit Test
+
+// Copyright (c) 2020 Digvijay Janartha, Hamirpur, India.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#include <iostream>
```
mloskot: Here and in other `.cpp` files missing is

```
#ifdef BOOST_NO_CXX11_HDR_INITIALIZER_LIST
#include <initializer_list>
#endif
```

test/geometries/linestring.cpp
```diff
@@ -0,0 +1,130 @@
...
+    out << bg::dsv(l1);
+    BOOST_CHECK_EQUAL(out.str(), "((1, 2), (2, 3), (3, 4))");
+}
+
+template <typename P>
+void test_custom()
+{   
+    std::initializer_list<P> IL = {P(1, 2), P(2, 3), P(3, 4)};
+    test_custom_linestring<P>(IL);
+}
```
mloskot: These test cases, and in the other `.cpp` files, require C++11 and must be wrapped with `BOOST_NO_CXX11_HDR_INITIALIZER_LIST`, see details at https://www.boost.org/libs/config/doc/html/boost_config/boost_macro_reference.html

test/geometries/linestring.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include <boost/geometry/io/dsv/write.hpp>
+
+#include <test_common/test_point.hpp>
+
+BOOST_GEOMETRY_REGISTER_C_ARRAY_CS(cs::cartesian)
+BOOST_GEOMETRY_REGISTER_BOOST_TUPLE_CS(cs::cartesian)
+
+#ifdef BOOST_NO_CXX11_HDR_INITIALIZER_LIST
+#include <initializer_list>
+#endif//BOOST_NO_CXX11_HDR_INITIALIZER_LIST
```
mloskot: I think you moved the include because you noticed `BOOST_NO_CXX11_HDR_INITIALIZER_LIST` is not defined.
Yes, in order to get those macros defined, `#include <boost/config.hpp>` must be included first (and it is included by some of the `#include <boost/*>` headers.

digu-007: > I think you moved the include because you noticed `BOOST_NO_CXX11_HDR_INITIALIZER_LIST` is not defined.

Yes, I assumed so.

digu-007:
@mloskot @vissarion Thanks for all the suggestions and merging :)

PR #679, opened 2020-03-06
Rename Jamfile.v2 to Jamfile

mloskot:
There is no reason to maintain the legacy `.v2` file extension, which is unnecessary clutter, may be confusing. The canonical name is `Jamfile`.


mloskot:
Thanks @awulkiew 
I think we all agree on such simplification, so I will merge it now.

PR #685, opened 2020-03-21
Add missing headers to precise_math.hpp

tinko92:
I noticed these two missing includes when I tried to use the header on its own.

### DIFF
--- a/extensions/test/headers/Jamfile
+++ b/extensions/test/headers/Jamfile
@@ -48,3 +48,4 @@ alias ext_algebra : [ generate_self_contained_headers extensions/algebra ] ;
 alias ext_algorithms : [ generate_self_contained_headers extensions/algorithms ] ;
 alias ext_index : [ generate_self_contained_headers extensions/index ] ;
 alias ext_gis : [ generate_self_contained_headers extensions/gis ] ;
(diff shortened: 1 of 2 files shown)


mloskot:
@tinko92 Could you try adding `alias ext_triangulation` and see if this catches the missing headers for you? If it does, then this will also serve as re-runnable confirmation of the fix.

https://github.com/boostorg/geometry/blob/6af84a3a4994ac127c5bfe76799e45f0f83d662c/extensions/test/headers/Jamfile#L47-L50

tinko92:
It does. Thanks, I was unaware of this testing facility. It also catches other things that I missed in my tests, I will rebase in a couple of minutes.

tinko92:
The PR now includes the line proposed for the extensions/test/headers/Jamfile and another missing header that was detected through this.

I am not familiar with the other extensions so I am hesitant to open an issue but I noticed that running b2 (in my working copy) in the headers directory produced 17 more failures in total, in extensions_gis, extensions_algorithms and extensions_index. If these are not known false positives, then maybe this is an issue.

vissarion:
First, I am ok with this PR. Regarding failures in extensions I also have 17. I guess they are related to work in progress or outdated code. @barendgehrels @awulkiew should know more.

PR #687, opened 2020-03-25
Fix dangling reference in distance

awulkiew:


include/boost/geometry/algorithms/detail/distance/multipoint_to_geometry.hpp
```diff
@@ -1,6 +1,6 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
-// Copyright (c) 2014, 2019, Oracle and/or its affiliates.
+// Copyright (c) 2014, 2019, 2020 Oracle and/or its affiliates.
```
mloskot: I'm curious: why not `2014-2020`?

awulkiew: We have this policy about copyrights to put all of the dates in the message. Or at least I think we have. However I choose to drop it as soon as the line becomes too long. So I guess I may not wait until then and change it now if you like.

mloskot: Yes, I've been interpreting the policy in such way that it considers `1,2,3` and  `1-3` as equivalents.

PR #689, opened 2020-03-27
Deprecate C++03 in Boost 1.73 as preparation for C+14 switch

mloskot:
Following discussion in #590, we are identifying support for C++03
as a candidate for removal from future releases of Boost.Geometry.

Issue [deprecation](http://eel.is/c++draft/depr#2) warning during compilation in C++03 conformance mode
Users can define BOOST_GEOMETRY_DISABLE_DEPRECATED_03_WARNING to disable it.

### Tasklist

- [x] Decide between switch to C++11 or to C++14 - C++14 it is 
- [x] Decide which Boost version, 1.74 or 1.75 or later, will require the later C++ version - 1.75 it is

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # ![Boost.Geometry](doc/other/logo/logo_bkg.png)
 
+> **CAUTION**: Boost.Geometry in Boost 1.73 deprecates support for the C++03 and will require C++14 from Boost 1.75 onwards (see issue [#590](https://github.com/boostorg/geometry/issues/590)).
+
 Boost.Geometry, part of collection of the [Boost C++ Libraries](http://github.com/boostorg), defines concepts, primitives and algorithms for solving geometry problems.
 
 [![Licence](https://img.shields.io/badge/license-boost-4480cc.png)](http://www.boost.org/LICENSE_1_0.txt)
--- a/doc/compiling.qbk
+++ b/doc/compiling.qbk
@@ -15,6 +15,8 @@
 [def __msvc__ MSVC]
 [def __stlport__ [@http://sourceforge.net/projects/stlport STLport]]
 
+[caution Boost.Geometry in Boost 1.73 deprecates support for the C++03 and will require C++14 from Boost 1.75 onwards]
+
 __boost_geometry__ is a headers-only library. Users only need to include the 
 library headers in their programs in order to be able to access definitions 
 and algorithms provided by the __boost_geometry__ library. No linking against 
--- a/include/boost/geometry/geometry.hpp
+++ b/include/boost/geometry/geometry.hpp
@@ -20,6 +20,15 @@
 #ifndef BOOST_GEOMETRY_GEOMETRY_HPP
 #define BOOST_GEOMETRY_GEOMETRY_HPP
 
+#include <boost/config.hpp>
+#include <boost/config/pragma_message.hpp>
+#if defined(BOOST_NO_CXX11_VARIADIC_TEMPLATES) || defined(BOOST_NO_CXX11_RVALUE_REFERENCES) || defined(BOOST_NO_CXX11_HDR_MEMORY)
+#if !defined(BOOST_GEOMETRY_DISABLE_DEPRECATED_03_WARNING)
+BOOST_PRAGMA_MESSAGE("CAUTION: Boost.Geometry in Boost 1.73 deprecates support for the C++03 and will require C++14 from Boost 1.75 onwards.")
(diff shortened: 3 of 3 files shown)


Mike-Devel:
@mloskot : The checks don't reflect the message: You are checking for c++11 stuff, but state that c++14 will be required. 

wouldnt something like `#if (__cplusplus < 201402L) && (_MSVC_LANG < 201402L) `be more appropriate.

mloskot:
@Mike-Devel Yes, you are correct. Initially, before we settled at C++14, we only needed to distinguish between C++03 and later.

Why not  Boost macros? For example, `BOOST_NO_CXX14_CONSTEXPR`,, `BOOST_NO_CXX14_DECLTYPE_AUTO`, `BOOST_NO_CXX14_RETURN_TYPE_DEDUCTION`

Mike-Devel:
> Why not Boost macros? 

Because I didn't know them offhand ;). 

On the other hand  - unless you already know exactly what features you want to use - doesn't it make more sense to simply test for the c++ language version instead for individual features?

awulkiew:
I'm leaning towards using Boost.Config macros. Even checking single `BOOST_NO_CXX14_CONSTEXPR` would probably be enough because this is one of the features that was added last to MSVC AFAIR.

The test based on `__cplusplus` and `_MSVC_LANG` would require more care and testing to make it right for all supported compilers/preprocessors. I'd assume that some compilers could complain that `_MSVC_LANG` is not defined. At least on MSVC `__cplusplus` is meaningless so it wouldn't surprise me that there is a different compiler which doesn't define `_MSVC_LANG` and also meaningless `__cplusplus`. Not to mention that earlier versions of MSVC doesn't define `_MSVC_LANG`. I also don't know what will be the values with Clang on Windows which partially emulates MSVC. Etc. Etc.

On the other hand we can assume that Boost.Config works.

Furthermore, this is not the only way of informing the users about the transition. In addition to this annoying macro we will also announce on the mailing lists, in the docs, release notes, slack channels, gitter, etc.

mloskot:
I agree. I also prefer the Boost Config macros. Moreover, even if `_MSVC_LANG` is defined regardlessoof CL.EXE switches, `_MSVC_LANG < 201402` is never true.

It's good to provoke comments even on  seemingly trivial issues :)

mloskot:
@awulkiew I updated #696 to check for `BOOST_NO_CXX14_CONSTEXPR`. That should be enough approx. for the purpose of the deprecation warning.

Mike-Devel:
> Moreover, even if _MSVC_LANG is defined regardlessoof CL.EXE switches, _MSVC_LANG < 201402 is never true.

That is true, but the point is that it directly expresses what you say you want to check for (c++14 support) and older versions of msvc that don't support c++14 don't define that macro, so the expression evaluates to true and you get the desired warning .

Again, if you know what you want from c++14, then checking for features is a good approach,  but checking for seemingly arbitrary macros that you expect are (not) defined on compilers you plan to support seems backwards to me.

Not that I have any stake in this - I don't use boost geometry.

mloskot:
Although I understand your point, it's valid, for this particular case of deprecation warning, any missing C++14 feature will serve fine.

Once migration starts, finer feature-wise control will be needed. Then, it may also turn out that VS 2015 with incomplete C++14 support, which AFAIK does not define the the lang macro, may work fine  compiling the library.

Mike-Devel:
It is definitely defined in VS2015 Update 3 (probably not RTM). 

But it really doesn't matter. Sorry for arguing about it.

mloskot:
@Mike-Devel You pointed out a valid issue, so the arguing is not all that pointless :)

Mike-Devel:
I meant arguing about checking `__cplusplus`/`_MSVC_LANG` vs `BOOST_NO_CXX14_CONSTEXPR` ;)

mloskot:
Yes, I know what you mean :)  It still provoked some extra considerations which is good.

Mike-Devel:
@awulkiew , @mloskot 
> Furthermore, this is not the only way of informing the users about the transition. In addition to this annoying macro we will also announce on the mailing lists, in the docs, release notes, slack channels, gitter, etc.

Speaking of which, I don't see release notes for boost geometry (and boost math for that matter) at https://github.com/boostorg/website/blob/master/feed/history/boost_1_73_0.qbk

mloskot:
@Mike-Devel I've just submitted the release notes https://github.com/boostorg/website/pull/499

mloskot:
@awulkiew I also added in https://github.com/boostorg/geometry/commit/56a9f79036dc3bce8dcd0483cfa728a196997f81 the deprecation notice to `doc/release_notes.qbk`. It may be too late to merge the develop to master though.

awulkiew:
@mloskot Master is still opened so it's not too late. I'll do it today.

PR #692, opened 2020-04-01
[buffer] revise turn_in_piece visitor

barendgehrels:
- avoid side_of_intersection (necessary for rescale removal)
- use new abstraction piece_border to verify point-in-piece (necessary for rescale removal
- this replaces the robust_ring
- and much more

Please start the review. I want to add unit tests next week, for the new class `piece_border` and the new strategy `turn_in_ring_winding`. The last one is more specific than other strategies and maybe I should move it to the buffer folder. Also it is a bit (but by far not as much as replaced implementation) dependent on Cartesian.

It is larger than I initially thought but it is quite a cleanup.
(Edited): performance is just slightly better, measured on `recursive_polygons_buffer.cpp`:
560 s (new) vs 600 s (old) and it makes slightly less errors (there are still errors, to be looked at later when rescaling is gone, this test is quite merciless)


include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -364,124 +349,169 @@ struct buffered_piece_collection
         , m_point_in_geometry_strategy(intersection_strategy
             .template get_point_in_geometry_strategy<robust_point_type,
                         robust_ring_type>())
-        , m_robust_area_strategy(intersection_strategy
-            .template get_area_strategy<robust_point_type>())
         , m_robust_policy(robust_policy)
     {}
 
+    template <typename Turn>
+    std::string sidx(Turn const& turn)
```
barendgehrels, line 356: Oops, I still have to remove this, will do

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
```diff
@@ -1,831 +1,46 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
-// Copyright (c) 2012-2014 Barend Gehrels, Amsterdam, the Netherlands.
-// Copyright (c) 2017 Adam Wulkiewicz, Lodz, Poland.
-
-// This file was modified by Oracle on 2016, 2018.
-// Modifications copyright (c) 2016-2018 Oracle and/or its affiliates.
-// Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
+// Copyright (c) 2020 Barend Gehrels, Amsterdam, the Netherlands.
```
barendgehrels, line 3: This file is for 99% replaced (I started with a new file).
So I removed the Oracle copyright, but if you want to keep it, OK for me.

barendgehrels:
It 's still work in progress, I added a unit test and I've to investigate the failures (5)

barendgehrels:
@awulkiew  and @vissarion  are you OK to merge this?

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -256,6 +237,7 @@ struct buffered_piece_collection
 
     struct original_ring
     {
+        typedef geometry::model::box<robust_point_type> robust_box_type;
```
barendgehrels, line 240: In the next (smaller) step I will update point-in-original-ring with non rescaled types

barendgehrels:
> 
> 
> It is ok for me. Sorry for the late review I though it was a PR in progress.

It was in progress earlier indeed. Thanks for reviewing!

awulkiew:
@barendgehrels  Sorry for a late response. I'll review it today.

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
```diff
@@ -157,54 +151,116 @@ struct buffer_turn_info
...
+};
+
+struct piece_get_box
+{
+    template <typename Box, typename Piece>
+    static inline void apply(Box& total, Piece const& piece)
+    {
+        if (piece.m_piece_border.m_has_envelope)
+        {
+            geometry::expand(total, piece.m_piece_border.m_envelope);
```
awulkiew, line 207: ExpandBoxBox strategy is needed here.

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
```diff
@@ -157,54 +151,116 @@ struct buffer_turn_info
...
-        // Sort them descending
-        return sl == sr
-            ? left.fraction < right.fraction
-            : sl < sr;
+struct turn_get_box
+{
+    template <typename Box, typename Turn>
+    static inline void apply(Box& total, Turn const& turn)
+    {
+        geometry::expand(total, turn.point);
```
awulkiew, line 242: ExpandBoxPoint strategy is needed here.

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -131,12 +131,8 @@ struct buffered_piece_collection
 
     // Robust ring/polygon type, always clockwise
     typedef geometry::model::ring<robust_point_type> robust_ring_type;
-    typedef geometry::model::box<robust_point_type> robust_box_type;
 
-    typedef typename default_comparable_distance_result
-        <
-            robust_point_type
-        >::type robust_comparable_radius_type;
+    typedef geometry::model::box<point_type> normal_box_type;
```
awulkiew, line 135: Maybe simply `box_type`? The robust ones will go anyway.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
```diff
@@ -0,0 +1,509 @@
...
+
+    // Only used for debugging (SVG)
+    Ring get_full_ring() const
+    {
+        Ring result;
+        if (! valid())
+        {
+            return result;
+        }
+        std::copy(m_ring->begin() + m_begin,
```
awulkiew, line 160: This code should be enabled with an `#ifdef` because right now it requires `<iostream>` to be included which should not be the case.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
```diff
@@ -0,0 +1,509 @@
...
+        // to stop further processing.
+        return strategy.apply(point, p1, p2, dm, place_on_ring, state);
+    }
+
+    template <typename It, typename Box>
+    void expand_envelope(Box& envelope, It begin, It end) const
+    {
+        for (It it = begin; it != end; ++it)
+        {
+            geometry::expand(envelope, *it);
```
awulkiew, line 361: This needs a strategy.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
```diff
@@ -0,0 +1,509 @@
...
+    template <typename Box>
+    bool calculate_envelope(Box& envelope) const
+    {
+        geometry::assign_inverse(envelope);
+        if (! valid())
+        {
+            return false;
+        }
+        expand_envelope(envelope, m_ring->begin() + m_begin, m_ring->begin() + m_end);
+        expand_envelope(envelope, m_originals.begin(), m_originals.begin() + m_original_size);
```
awulkiew, line 223: They need a strategy.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
```diff
@@ -0,0 +1,509 @@
...
+                  std::back_inserter(result));
+        // Add the closing point
+        result.push_back(*(m_ring->begin() + m_begin));
+
+        return result;
+    }
+
+    void get_properties_of_border(bool is_point_buffer, Point const& center)
+    {
+        m_has_envelope = calculate_envelope(m_envelope);
```
awulkiew, line 174: This needs a strategy.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
```diff
@@ -0,0 +1,509 @@
...
+        if (m_has_envelope)
+        {
+            // Take roundings into account, enlarge box
+            detail::buffer::buffer_box(m_envelope, 1.0e-3, m_envelope);
```
awulkiew, line 178: What is the source of the `1.0e-3` number? Shouldn't it depend on the machine epsilon and the coordinates themselves? For very small coordinates this number if too big, it's not a problem though. However for very big coordinates this number will be too small.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
```diff
@@ -0,0 +1,509 @@
...
+        }
+    }
+
+    void set_offsetted(Ring const& ring, std::size_t begin, std::size_t end)
+    {
+        BOOST_GEOMETRY_ASSERT(begin <= end);
+        BOOST_GEOMETRY_ASSERT(begin < boost::size(ring));
+        BOOST_GEOMETRY_ASSERT(end <= boost::size(ring));
+
+        m_ring = &ring;
```
awulkiew, line 203: Can `Ring` be a user-defined type or is it our type and we can be sure that the `operator&` is not defined? Because otherwise `boost::addressof` should be used here.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
```diff
@@ -0,0 +1,509 @@
...
+    {
+        // A step between original/offsetted ring is always convex
+        // (unless the join strategy generates points left of it -
+        //  future: convexity might be added to the buffer-join-strategy)
+        // Therefore, if the state count > 0, it means the point is left of it,
+        // and because it is convex, we can stop
+
+        typedef typename geometry::coordinate_type<Point>::type coordinate_type;
+        typedef geometry::detail::distance_measure<coordinate_type> dm_type;
+        dm_type const dm = geometry::detail::get_distance_measure<cartesian_tag>(point, p1, p2);
```
awulkiew, line 344: Shouldn't this tag be taken from the Strategy?

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
barendgehrels: Thanks, will fix (this probably was already the case earlier, but will fix anyway)

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
barendgehrels: I renamed it in the next PR (which is nearly ready) because indeed the robust one is gone then.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
barendgehrels: It's actually the same as we did in the sectionalize.
Good point - I will take over what we did there.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
barendgehrels: No, the ring is an offsetted ring and defined by ourselves.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
barendgehrels: ah, yes, of course. Will fix.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
```diff
@@ -0,0 +1,509 @@
...
+        , m_is_monotonic_increasing(false)
+        , m_is_monotonic_decreasing(false)
+        , m_min_comparable_radius(0)
+        , m_max_comparable_radius(0)
+    {
+    }
+
+    //! Returns true if there is a ring, the ring has one or more points,
+    //! and the original has one or two points
+    bool valid() const
```
awulkiew, line 147: Could this function have a different name? Validity in our library has a different meaning. E.g. something directly referring to what this function return, like `is_non_empty_ring_set()` or something. What do you think?

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
```diff
@@ -0,0 +1,509 @@
...
+        {
+            return false;
+        }
+
+        // Walk over the different parts of the ring, in clockwise order
+        // For performance reasons: start with the helper part (one segment)
+        // then the original part (one segment, if any), then the other helper
+        // part (one segment), and only then the offsetted part
+        // (probably more segments, check monotonicity)
+        geometry::strategy::buffer::turn_in_ring_winding<coordinate_type> tir;
```
awulkiew, line 244: Right, so this is a cartesian strategy right? I assume it is and in the future it should be derived from the strategy passed by the user because in general should depend on the CS. So in geographic CS the ellipsoid should be propagated here. Could we at for now implement it in such a way that when someone tries to pass a non-cartesian geometry it does not compile? E.g. add static assert checking cs_tag == cartesian_tag or something. This way it will be easier to find it in the future. It might be done here or in the strategy itself.

awulkiew: Actually, the strategy passed by the user should be checked to be cartesian strategy, not the geometry. The intersection strategy defines cs_tag AFAIR so this should be easy.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
```diff
@@ -0,0 +1,509 @@
...
+
+        bool first = true;
+
+        // An offsetted point-buffer ring around a point is supposed to be closed,
+        // therefore walking from start to end is fine.
+        It it = begin;
+        for (It previous = it++; it != end; ++previous, ++it)
+        {
+            segment_type const s(*previous, *it);
+            radius_type const d = geometry::comparable_distance(center, s);
```
awulkiew, line 487: This needs a strategy. But there is a problem with it because it cannot be derived from the intersection strategy since the intersection strategy is lower-level than the distance point/segment strategy I think, so there would be a cyclic-dependency. I know that because recently I looked into this while checking whether or not it is possible to use `buffer` with `cs_undefined` CS and it is not possible right now because internally, at the top-most level the default intersection strategy is created and internally the default `comparable_distance` is called. But anyway, the solution I came up with for this cyclic-dependency is to implement a "proper" CS-specific buffer strategy providing both CS-specific intersection and comparable_distance strategies. So this would be a greater change outside of the scope of this PR I think. Therefore I propose to leave this at is is for now and later I'll propose my solution for this issue (I'd like to propose some more changes too along with it). Do you agree?

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
awulkiew: I have no opinion. I'm not a legal expert. :)
I'd rather worry about files to which you copied the code from this one because I see a lot of changes done by me regarding the propagation of strategies and one change done by Menelaos in the past so technically if you copied any part affected by us into a different file this file should have all of the copyrights too I think. But this is only my guess.

awulkiew: When I view this version of the file and click "Blame" I still see some changes made by me and Menelaos so even though the majority of the code was removed our changes are still there. This one even adds my copyrights: https://github.com/boostorg/geometry/commit/fd72a1aa16a71ac465db334890964a3799bf5686
Not sure why, though, maybe I changed something during weekend ;)

include/boost/geometry/strategies/cartesian/turn_in_ring_winding.hpp
```diff
@@ -0,0 +1,211 @@
...
+    //
+
+
+public:
+
+    struct counter
+    {
+        inline counter()
+            : m_count(0)
+            , m_min_distance((std::numeric_limits<CalculationType>::max)())
```
awulkiew, line 64: Can `CalculationType` be user-defined? For user defined types max will be initialized by `T()`. Can this be a problem?

barendgehrels:
> 
> 
> Ok, done. I'm of course ok with this change after addressing the issues I raised. Alternatively you can postpone it or leave it to me because I plan to prepare buffer() to work with cs_undefined geometries anyway and this will require propagating the information about the CS everywhere where it's relevant.

Hi @awulkiew  thanks for your thorough review. I will address most of the issues you raise on Wednesday, but will probably leave some of the CS related to you then, thanks.

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
barendgehrels: > When I view this version of the file and click "Blame" I still see some changes made by me and Menelaos, see:

No problem, I'll put it back.

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
barendgehrels: Not supported... There is something needed in the handling of expand with a strategy and a point. I leave this to you to avoid expanding the scope of this PR

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
barendgehrels: -> ring_or_original_empty() (with reversed meaning and now private)
:heavy_check_mark:

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
barendgehrels: I leave this for now, first it was already targetted to cartesian (I made it less dependent by removing side_by_intersection and other specific things, but this PR does not have the purpose to make it completely cs agnostic). Second, buffer of geographic points is already supported so I should not blindly check the cs_tag.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
barendgehrels: We discussed a while ago: I'm quite against adding sub-strategies to the `IntersectionStrategy`. There should be a proper `UbrellaStrategy` and this is *not* the IntersectionStrategy. Because why? It doesn't use most of the strategies defined there (I think it uses only `side` and nothing more). It's just the wrong place. Why should the intersection of two segments define a `point_in_polygon`? I just don't get that.

Well, I understand how it happened - there needed to be a place and this strategy was passed everywhere - so it was added there. But it is inconvenient and should be refactored.

So yes, if this could be fixed by moving all these sub-strategies to a proper umbrella strategy, that would be very welcome.

And that umbrella strategy can be passed to buffer indeed, and there will be no cyclic dependencies.

include/boost/geometry/strategies/cartesian/turn_in_ring_winding.hpp
barendgehrels: Isn't this a problem then everywhere where we use `numeric_limits`?
But anyway, here it was easy to get rid of it, done
:heavy_check_mark:

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
barendgehrels: The expand strategies do not have the Earth model anyway (is there ever a reason to do that?).
So I can define and use them locally here, they don't need to be passed from the IntersectionStrategy.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
barendgehrels: Done :heavy_check_mark:

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
barendgehrels: It's a local function, but the final expand gets a strategy now

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
barendgehrels: Same here

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
awulkiew: > The expand strategies do not have the Earth model anyway

That's because they are the same as the spherical ones and the unit sphere is enough as a model. Still it's different than cartesian.

barendgehrels: > 
> 
> > The expand strategies do not have the Earth model anyway
> 
> That's because they are the same as the spherical ones and the unit sphere is enough as a model. Still it's different than cartesian.

Yep - so I think what I now submitted (and merged) is fine?

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
awulkiew: > this PR does not have the purpose to make it completely cs agnostic

Right.

> Second, buffer of geographic points is already supported so I should not blindly check the cs_tag.

Ok, fixing it now would probably be outside of this PR.

Also my thoughts regarding this issue. Internally `turn_in_ring_winding` relies on cartesian `model::infinite_line` (unless I'm missing something). It is therefore a cartesian strategy. Using it in spherical and geographic buffer is technically wrong (unless you're certain that the distances passed by the user are small which we can't guarantee). Checking it wouldn't be blind, it'd make sure that the user doesn't get the incorrect results thinking they are correct. I'd rather argue that using it in the first place for all CSes was blindly assuming things. If it is used for geographic points then it means it's a bug. It should probably have been modified along with introducing support for geographic points. My guess is that it was not because the compiler didin't argue about it, if the CS was checked here the compiler would point to the place that is wrong in the code.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
awulkiew: A disclaimer: I agree that the `IntersectionStrategy` is not the place for all of the getters that currently are there. Below I describe the rationale behind the current state, not claiming that it should stay as it is because I think it should be changed in general. I'm not sure that the `UbrellaStrategy` is the best solution though.

> There should be a proper UbrellaStrategy and this is not the IntersectionStrategy

That would be a breaking change. Note that the `IntersectionStrategy` in this case is at the same time the "original" strategy and also the umbrella strategy containing all strategies needed in algorithms which the `IntersectionStrategy` is passed into since the inclusion of Boost.Geometry into Boost. So it is both kinds of strategy right now, simply because otherwise it'd break the existing code.

> Why should the intersection of two segments define a point_in_polygon?

As above. Because of the design of the library and the desire to not introduce breaking changes.

Consider e.g. set operations, let's say intersection of polygons. The strategy which was always required since the inclusion of Boost.Geometry was the `IntersectionStrategy`. In addition to this the library requires to pass one strategy for an algorithm and it also guarantees that the CS of the strategy "overwrites" the CS of the geometry. So in order to not break the users code we have to require passing the same strategy as usual, in this case the `IntersectionStrategy` and this strategy has to be the umbrella strategy as well because of all of the other strategies needed deeper in the algorithms.

It would be possible to ensure backward compatibility with "proper" `UmbrellaStrategy` though. We'd have to create the `UmbrellaStrategy` from another "original" strategy (e.g. umbrella strategy for set operations created from classically passed intersection strategy). Note however that this means that all strategies in the umbrella strategy would be created based on the original intersection strategy. And this means that the umbrella strategy would be not needed since the library would create all other strategies automatically anyway.

There is also an extensibility issue. What if we introduced a "proper" `UbrellaStrategy` and after a while another lower-level sub-strategy was needed internally in the algorithm? Would we require to pass another sub-strategy into the `UbrellaStrategy` changing the interface and therefore breaking user's code or would we create it automatically based on other strategies that were in the `UmbrellaStrategy` before the change? If the answer is the latter then I'd argue that the `UbrellaStrategy` is not needed in the first place since it should be possible to create any sub-strategy from any set of other sub-strategies.

Btw, I'd be ok with redesigning the strategies part to something different even if this would mean breaking changes. The `UbrellaStrategy` is the most straightforward thing to do. It would cause the problems I described above and probably other which we don't take into account right now. Right now, the `IntersectionStrategy` is partially the `UmbrellaStrategy` so we'd have similar problems with the "proper one". I'm not even sure if we'd solve the cyclic dependency across sub-strategies this way.

Note that all of the problems I mentioned are derived from the original design of the library so I'd consider to do something more radical like e.g. replacing the strategies mechanism with something else instead of adding another layer of strategies for the user to use. I'm talking about the interface here.

Of course I may be mistaken with all of the above so if you have a proposal we can create an Issue and talk about it.

include/boost/geometry/strategies/cartesian/turn_in_ring_winding.hpp
awulkiew: > Isn't this a problem then everywhere where we use `numeric_limits`?
Yes.

awulkiew:
Ok, thanks for the PR! I raised some additional issues but they are more general and not really related with this PR specifically.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
barendgehrels: Thanks! I will reread this later and react on it, probably this weekend.

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
awulkiew: Almost ;) The CS should be taken from the strategy passed by the user instead of a geometry.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
barendgehrels: Any buffer around a single point will not call this code.
Only when multi-points (or linear/poly geometries) are buffered, then this code would be called.
But if we make a compile-time check here, the geo-point-buffer won't compile. That is the reason I cannot do it now. It was already like this before (but then not `infinite_line` (indeed) but `side_by_intersection` and other things which were cartesian related)

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
barendgehrels: See my message in the other PR - the passed strategy doesn't compile for points, because it's defined as a segment-specific strategy (for boxes it does compile but still wrong because a box is not a segment)

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
awulkiew: Ok, got it, no problem. As I said, fixing it is outside of the scope of this PR.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
awulkiew: @barendgehrels After giving it more thought I think we could indeed try with "proper" `UmbrellaStrategy` at least for one function and see how it works out to avoid over-engineering things. I think we could workaround the problems I mentioned above and the result would still be the code which is more clear than it is now (the strategies in particular which I agree should have well-defined purpose).

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
awulkiew: I'm ok with leaving it as it is. I'll go back into it later if you don't want to deal with it. But I'll describe what we're dealing with here for informative purposes.

I'll write how this should work in general. At the top-most level the user passes a strategy (doesn't matter which one). This strategy at the top-most level can also be created based on the geometry if the user doesn't pass it explicitly. Then the strategy is passed below into the internals of an algorithm. This strategy defines the CS of all of the operations below (or defines all of the operations themselves as sub-strategies, it doesn't matter). What matters is that the information about the CS is passed by the user with the top-most level strategy.

So here `expand` is called. If the user's intention (by proxy) is to perform expansion of a box with some other geometry in a specific CS the following strategies should be used:
* cartesian
  * point - `strategy::expand::cartesian_point`
  * box - `strategy::expand::cartesian_box`
  * segment - `strategy::expand::cartesian_segment`
* spherical
  * point - `strategy::expand::spherical_point`
  * box - `strategy::expand::spherical_box`
  * segment - `strategy::expand::spherical_segment`
* geographic
  * point - `strategy::expand::spherical_point`
  * box - `strategy::expand::spherical_box`
  * segment - `strategy::expand::geographic_segment`

In this case we're talking about `expand(box, point)` so the strategy is either `strategy::expand::cartesian_point` or `strategy::expand::spherical_point`. And at the top-most level of `buffer` the strategy created is the `IntersectionStrategy`. So one of the two `expand` strategies has to be derived from `IntersectionStrategy`. At least this is how it is right now, if we change it in the future it might be different (e.g. it might be derived from an `UmbrellaStrategy`). But right now, the `IntersectionStrategy` directly only defines Box (`expand_box_strategy_type`) and Segment (`expand_strategy_type`) expand strategies. So you have several options here:
1. Take it from one of the underlying strategies, e.g.:
```
IntersectionStrategy::template point_in_geometry_strategy<BoxPoint, TurnPoint>::expand_point_strategy_type
```
2. Create default one based on `IntersectionStrategy::cs_tag`:
```
strategy::expand::services::default_strategy<point_tag, IntersectionStrategy::cs_tag>::type
```
This will probably not compile right now for spherical because for spherical `IntersectionStrategy` the `cs_tag` is `spherical_tag` and `default_strategy` is not specialized for this tag so with this you'd have to add specialization here:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/strategies/spherical/expand_point.hpp#L205-L221
3. Create it based on `IntersectionStrategy::cs_tag` without using `default_strategy` with, e.g.: 
```
typedef typename boost::if_c
    <
        boost::is_same<typename IntersectionStrategy::cs_tag, cartesian_tag>::value,
        strategy::expand::cartesian_point,
        strategy::expand::spherical_point,
    >::type expand_point_strategy_type;

// you could add some check for safety or not:

static const bool cs_check =
    boost::is_same<typename IntersectionStrategy::cs_tag, cartesian_tag>::value
    || boost::is_same<typename IntersectionStrategy::cs_tag, spherical_tag>::value
    || boost::is_same<typename IntersectionStrategy::cs_tag, geographic_tag>::value;
BOOST_MPL_ASSERT(cs_check);
```

awulkiew:
Now when `include/boost/geometry/strategies/cartesian/side_of_intersection.hpp` is removed do we still need the test case `test/strategies/side_of_intersection.cpp`? I commented it out in the Jamfile for now without removing it in case you wanted to still use it for something else.

include/boost/geometry/algorithms/detail/buffer/buffer_policies.hpp
barendgehrels: Thank you.

I think you're missing one important point I make. **It does not compile.** And not because the CS is wrong, it does not compile because the expand strategy, defined in intersection strategy, is defined for `segment` and not on `box` or `point`.
Therefore **it does not compile if you expand with a point.**

And this is an important flaw of how the strategy is passed. We might solve it by adding two other strategies to intersection_strategy. But I'm really opposed to that idea.

I still don't know why expand strategy should be passed at all. For other strategies, I understand the rationale (distance, area, point-in-polygon, side, etc). But this one is just an implementation detail, different for different cs, true, but not user-dependent or earth-model dependent.

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
barendgehrels: I really welcome this idea!

Will we create an issue for it - to get it in our regular process?

barendgehrels:
> 
> 
> Now when `include/boost/geometry/strategies/cartesian/side_of_intersection.hpp` is removed do we still need the test case `test/strategies/side_of_intersection.cpp`? I commented it out in the Jamfile for now without removing it in case you wanted to still use it for something else.

Thanks I will remove it in next PR. Don't need it again.

PR #695, opened 2020-04-13
Support tupled-output in difference, sym_difference and union_.

awulkiew:
This is a followup PR after https://github.com/boostorg/geometry/pull/650

It adds support for tupled-output in the rest of set operations. The interface is the same for all of them and allows the user to use the same output type. Here is an example:
```
polygon po1, po2;
bg::read_wkt("POLYGON((0 0, 0 5, 5 5, 5 0, 0 0))", po1);
bg::read_wkt("POLYGON((0 0, 0 1, 6 1, 5 2, 5 5, 5 6, 4 5, 4 7, 7 7, 7 0, 0 0))", po2);
bg::correct(po1);
bg::correct(po2);

std::tuple<multi_point, multi_linestring, multi_polygon> d, i, sd, u;
bg::difference(po1, po2, d);
bg::intersection(po1, po2, i);
bg::sym_difference(po1, po2, sd);
bg::union_(po1, po2, u);

auto print_result = [&](char op, auto const& res) {
    std::cout << bg::wkt(po1) << ' ' << op << ' ' << bg::wkt(po2) << '\n'
                << bg::wkt(std::get<0>(res)) << '\n'
                << bg::wkt(std::get<1>(res)) << '\n'
                << bg::wkt(std::get<2>(res)) << '\n'
                << std::endl;
};

print_result('-', d);
print_result('*', i);
print_result('^', sd);
print_result('+', u);
```
image

The output is:
```
POLYGON((0 0,0 5,5 5,5 0,0 0)) - POLYGON((0 0,0 1,6 1,5 2,5 5,5 6,4 5,4 7,7 7,7 0,0 0))
MULTIPOINT()
MULTILINESTRING()
MULTIPOLYGON(((0 1,0 5,5 5,5 1,0 1)))
```
image

```
POLYGON((0 0,0 5,5 5,5 0,0 0)) * POLYGON((0 0,0 1,6 1,5 2,5 5,5 6,4 5,4 7,7 7,7 0,0 0))
MULTIPOINT((4 5))
MULTILINESTRING((5 5,5 2))
MULTIPOLYGON(((0 1,5 1,5 0,0 0,0 1)))
```
image

```
POLYGON((0 0,0 5,5 5,5 0,0 0)) ^ POLYGON((0 0,0 1,6 1,5 2,5 5,5 6,4 5,4 7,7 7,7 0,0 0))
MULTIPOINT()
MULTILINESTRING()
MULTIPOLYGON(((4 5,4 7,7 7,7 0,5 0,5 1,0 1,0 5,4 5),(5 5,5 6,4 5,5 5),(5 2,5 1,6 1,5 2)))
```
image

```
POLYGON((0 0,0 5,5 5,5 0,0 0)) + POLYGON((0 0,0 1,6 1,5 2,5 5,5 6,4 5,4 7,7 7,7 0,0 0))
MULTIPOINT()
MULTILINESTRING()
MULTIPOLYGON(((0 5,4 5,4 7,7 7,7 0,0 0,0 5),(5 1,6 1,5 2,5 1),(5 5,5 6,4 5,5 5)))
```
image

Note that in this case the set operations other than intersection can produce only multi polygon but this is not the case when the input has different topological dimention. E.g. when a linestring and a polygon is passed into union the result will be a polygon and the difference of the linestring and the polygon. So the use of tupled output will have more sense.

### DIFF
--- a/include/boost/geometry/algorithms/detail/tupled_output.hpp
+++ b/include/boost/geometry/algorithms/detail/tupled_output.hpp
@@ -9,13 +9,17 @@
 #ifndef BOOST_GEOMETRY_ALGORITHMS_DETAIL_TUPLED_OUTPUT_HPP
 #define BOOST_GEOMETRY_ALGORITHMS_DETAIL_TUPLED_OUTPUT_HPP
 
+#include <boost/geometry/algorithms/convert.hpp>
 #include <boost/geometry/core/config.hpp>
 #include <boost/geometry/core/tag.hpp>
+#include <boost/geometry/core/tag_cast.hpp>
 #include <boost/geometry/core/tags.hpp>
+#include <boost/geometry/geometries/concepts/check.hpp>
 #include <boost/geometry/util/range.hpp>
 #include <boost/geometry/util/tuples.hpp>
 
 #include <boost/mpl/and.hpp>
+#include <boost/mpl/if.hpp>
 #include <boost/range/value_type.hpp>
 #include <boost/type_traits/detail/yes_no_type.hpp>
 #include <boost/type_traits/integral_constant.hpp>
@@ -53,7 +57,7 @@ struct is_multi_geometry
 
 // true for point, linestring or polygon
 template <typename T>
-struct is_multi_geometry_value
+struct is_multi_geometry_element
     : boost::integral_constant
         <
             bool,
@@ -86,23 +90,6 @@ struct is_range
 {};
 
 
-// geometry tag of Rng value_type
-template <typename Rng>
-struct range_value_tag
-    : geometry::tag<typename boost::range_value<Rng>::type>
-{};
-
-// true if geometry tag of Rng is the same as Tag
-template <typename Rng, typename Tag>
-struct is_range_value_tag_same_as
-    : boost::is_same
-        <
-            typename range_value_tag<Rng>::type,
-            Tag
-        >
-{};
-
-
 template <typename T, bool IsRange = is_range<T>::value>
 struct is_tupled_output_element_base
     : boost::integral_constant<bool, false>
@@ -114,12 +101,13 @@ struct is_tupled_output_element_base<T, true>
         <
             bool,
             (is_multi_geometry<T>::value
-            ||
-            ((! is_geometry<T>::value)
-                &&
-                ((is_range_value_tag_same_as<T, point_tag>::value)
-                || (is_range_value_tag_same_as<T, linestring_tag>::value)
-                || (is_range_value_tag_same_as<T, polygon_tag>::value))))
+                ||
+                ((! is_geometry<T>::value)
+                    &&
+                    is_multi_geometry_element
+                        <
+                            typename boost::range_value<T>::type
+                        >::value))
         >
 {};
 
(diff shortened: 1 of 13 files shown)


include/boost/geometry/algorithms/detail/tupled_output.hpp
```diff
@@ -527,6 +502,234 @@ struct output_geometry_access<GeometryOut, Tag, DefaultTag, DefaultTag>
...
+    typename Geometry1, typename Geometry2, typename TupledOut,
+    typename Tag1
+>
+struct expect_output<Geometry1, Geometry2, TupledOut, Tag1, void, void>
+    : expect_output_assert<Geometry1, Geometry2, TupledOut, Tag1>
+{};
+
+
+template <typename CastedTag>
+struct casted_tag_to_single_tag;
```
barendgehrels, line 670: Names of casts are typically suffixed with `cast`. Can you maybe rename this to `to_single_tag_cast` or `single_tag_cast` ?

include/boost/geometry/algorithms/union.hpp
```diff
@@ -209,7 +347,8 @@ inline OutputIterator union_insert(Geometry1 const& geometry1,
 {
     concepts::check<Geometry1 const>();
     concepts::check<Geometry2 const>();
-    concepts::check<GeometryOut>();
+    //concepts::check<GeometryOut>();
```
barendgehrels, line 350: This line is replaced and I think you can delete it?

include/boost/geometry/algorithms/union.hpp
```diff
@@ -139,14 +113,13 @@ struct union_insert
 template
 <
     typename Linear1, typename Linear2, typename LineStringOut,
-    bool Reverse1, bool Reverse2, bool ReverseOut
+    typename TagIn1, typename TagIn2
```
barendgehrels, line 116: Where are the Reverse parameters now? (I can probably look that up but in a review screen that is not so clear)

include/boost/geometry/algorithms/union.hpp
```diff
@@ -175,6 +147,172 @@ struct union_insert
...
+                                       OutputIterator out,
+                                       Strategy const& strategy)
+    {
+        access1::get(out) = geometry::dispatch::intersection_insert
+            <
+                Geometry1, Geometry2,
+                typename access1::type,
+                overlay_difference,
+                geometry::detail::overlay::do_reverse<geometry::point_order<Geometry1>::value>::value,
+                geometry::detail::overlay::do_reverse<geometry::point_order<Geometry2>::value, true>::value
```
barendgehrels, line 228: oh, probably this is the place where `Reverse` is done now

include/boost/geometry/algorithms/detail/tupled_output.hpp
awulkiew: This is not a cast. It's a metafunction translating/mapping from `CastedTag` (dispatch template argument which can be pointlike_tag, linear_tag, areal_tag) to tag of a "single" geometry which is an element of multi-geometry (point_tag, linestring_tag, polygon_tag respectively).

include/boost/geometry/algorithms/union.hpp
awulkiew: Done.

include/boost/geometry/algorithms/detail/tupled_output.hpp
awulkiew: I renamed it to `single_tag_from_base_tag` to avoid "cast" word. Is that name better?

PR #696, opened 2020-04-14
Correct #689 with check for C++14 support instead of C++11 features

mloskot:
Refines #689

------

Thanks to @Mike-Devel for pointing that out in https://github.com/boostorg/geometry/pull/689#issuecomment-613523835

### DIFF
--- a/include/boost/geometry/geometry.hpp
+++ b/include/boost/geometry/geometry.hpp
@@ -20,11 +20,11 @@
 #ifndef BOOST_GEOMETRY_GEOMETRY_HPP
 #define BOOST_GEOMETRY_GEOMETRY_HPP
 
-#include <boost/config.hpp>
-#include <boost/config/pragma_message.hpp>
-#if defined(BOOST_NO_CXX11_VARIADIC_TEMPLATES) || defined(BOOST_NO_CXX11_RVALUE_REFERENCES) || defined(BOOST_NO_CXX11_HDR_MEMORY)
 #if !defined(BOOST_GEOMETRY_DISABLE_DEPRECATED_03_WARNING)
-BOOST_PRAGMA_MESSAGE("CAUTION: Boost.Geometry in Boost 1.73 deprecates support for the C++03 and will require C++14 from Boost 1.75 onwards.")
+#include <boost/config.hpp>
+#if defined(BOOST_NO_CXX14_CONSTEXPR) || (__cplusplus < 201402L)
+#include <boost/config/pragma_message.hpp>
+BOOST_PRAGMA_MESSAGE("CAUTION: Boost.Geometry in Boost 1.73 deprecates support for C++03 and will require C++14 from Boost 1.75 onwards.")
 BOOST_PRAGMA_MESSAGE("CAUTION: Define BOOST_GEOMETRY_DISABLE_DEPRECATED_03_WARNING to suppress this message.")
 #endif
 #endif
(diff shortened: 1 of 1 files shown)


mloskot:
@awulkiew Do you have any comments to this or can we merge?

include/boost/geometry/geometry.hpp
```diff
@@ -20,11 +20,11 @@
 #ifndef BOOST_GEOMETRY_GEOMETRY_HPP
 #define BOOST_GEOMETRY_GEOMETRY_HPP
 
+#if !defined(BOOST_GEOMETRY_DISABLE_DEPRECATED_03_WARNING)
 #include <boost/config.hpp>
+#if defined(BOOST_NO_CXX14_CONSTEXPR) || (__cplusplus < 201402L)
```
Mike-Devel, line 25: This will evaluate to true on all versions of MSVC by default (you have to set a special compile switch for MSVC to define a "orrect" value for `__cplusplus`). I thought the conclusion in the other thread was to just check for `BOOST_NO_CXX14_CONSTEXPR`?

mloskot: I meant to use `_MSVC_LANG` in stead of  `__cplusplus`. I'll remove that sub-condition.

> I thought the conclusion in the other thread was to just check for BOOST_NO_CXX14_CONSTEXPR?

Yes, that was the plan. I'll update it.

include/boost/geometry/geometry.hpp
```diff
@@ -20,11 +20,11 @@
...
 #define BOOST_GEOMETRY_GEOMETRY_HPP
 
+#if !defined(BOOST_GEOMETRY_DISABLE_DEPRECATED_03_WARNING)
 #include <boost/config.hpp>
+#if defined(BOOST_NO_CXX14_CONSTEXPR)
 #include <boost/config/pragma_message.hpp>
-#if defined(BOOST_NO_CXX11_VARIADIC_TEMPLATES) || defined(BOOST_NO_CXX11_RVALUE_REFERENCES) || defined(BOOST_NO_CXX11_HDR_MEMORY)
-#if !defined(BOOST_GEOMETRY_DISABLE_DEPRECATED_03_WARNING)
-BOOST_PRAGMA_MESSAGE("CAUTION: Boost.Geometry in Boost 1.73 deprecates support for the C++03 and will require C++14 from Boost 1.75 onwards.")
+BOOST_PRAGMA_MESSAGE("CAUTION: Boost.Geometry in Boost 1.73 deprecates support for C++03 and will require C++14 from Boost 1.75 onwards.")
```
barendgehrels, line 27: It is possible to use Boost.Geometry without including `geometry.hpp` - we do that often ourselves in unit tests. And in the examples.
But in practice it is probably nearly always included like this.

So I'm OK with this place - but we might also place it in `boost/geometry/core/config.hpp` (including `boost/config.hpp` already)

barendgehrels:
Thanks, I'm basically OK

include/boost/geometry/geometry.hpp
awulkiew: That's a good idea, though `core/config.hpp` is included only by a handful of files basically related to `get_turns` and tuples handling. So a different core file could be used, e.g. `core/tags.hpp` which at first glance (https://github.com/awulkiew/cxx-dependency-visualizer) is one of the most included bottom-level headers.

mloskot: I think we can keep it in `geometry.hpp`. There will be also docs, README, and other means to announce.

mloskot:
Remaining question is if/when to schedule develop to master merge for Boost 1.73.
How that is going to be handled?

Issue #657, opened 2020-01-10
Incorrect results when transforming from 4326 to 3785/3857


grokse:
The results look correct when I use a proj4 string to create my transformation, but if I use boost::geometry::srs::epsg it's about ~25km south of what I expect. 

```c++
#include <boost/geometry.hpp>
#include <boost/geometry/srs/epsg.hpp>
#include <boost/geometry/srs/transformation.hpp>
#include <iostream>

int main() {
  using point_geo = bg::model::point<double, 2, bg::cs::geographic<bg::degree>>;
  using point_car = bg::model::point<double, 2, bg::cs::cartesian>;

  point_geo original = point_geo(-114.7399212, 36.0160698);
  point_car epsg_out;
  point_car proj_out;

  bg::srs::transformation<> epsg_tr{bg::srs::epsg(4326),
                                                       bg::srs::epsg(3785)};
  bg::srs::transformation<> proj_tr{
      bg::srs::epsg(4326),
      bg::srs::proj4(
          "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 "
          "+y_0=0 +k=1.0 +units=m +nadgrids=@null +wktext  +no_defs")};

  epsg_tr.forward(original, epsg_out);
  proj_tr.forward(original, proj_out);
  std::cout << std::setprecision(12) << "epsg output: " << bg::get<0>(epsg_out) << " "
            << bg::get<1>(epsg_out) << std::endl;

  std::cout << std::setprecision(12) << "proj output: " << bg::get<0>(proj_out) << " "
            << bg::get<1>(proj_out) << std::endl;

  return 0;
}
```

epsg output: -12772789.6016 4277696.96863
proj output: -12772789.6016 4302832.77709

awulkiew:
Thanks. It seems that https://epsg.io/3785 and https://spatialreference.org/ref/epsg/popular-visualisation-crs-mercator/ list different proj4 string for the projection. The difference is that the latter defines towgs84 parameter and the former doesn't. So the latter takes the coordinates as if they were on the ellipsoid and the former converts between WGS84 ellipsoid and sphere before applying mercator projection.

epsg.io defines `TOWGS84[0,0,0,0,0,0,0]` in WKT but not in proj4 string.

It seems Proj took the string from epsg.io as well: https://github.com/OSGeo/PROJ/blob/5.2/nad/epsg#L4605

What makes you think that one is correct and the other is not?

[...]

PR #697, opened 2020-04-16
Fix EPSG 3785 definitions.

awulkiew:
Fix for https://github.com/boostorg/geometry/issues/657


awulkiew:
Just doing my job good sir! ;)

PR #700, opened 2020-04-22
Add missing cases for default strategies in distance algorithm

vissarion:
This PR fixes an issue on default strategy selection for distance algorithm and affects all coordinate systems. 

The default strategy for the cases `polygonal/box` and `linear/box` should be a segment/box strategy and not a point/segment one.

### DIFF
--- a/include/boost/geometry/algorithms/detail/distance/default_strategies.hpp
+++ b/include/boost/geometry/algorithms/detail/distance/default_strategies.hpp
@@ -5,9 +5,10 @@
 // Copyright (c) 2009-2014 Mateusz Loskot, London, UK.
 // Copyright (c) 2013-2014 Adam Wulkiewicz, Lodz, Poland.
 
-// This file was modified by Oracle on 2014.
-// Modifications copyright (c) 2014, Oracle and/or its affiliates.
+// This file was modified by Oracle on 2014, 2020.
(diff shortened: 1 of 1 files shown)


include/boost/geometry/algorithms/detail/distance/default_strategies.hpp
```diff
@@ -110,26 +111,45 @@ struct default_strategy<Box1, Box2, box_tag, box_tag, false>
...
+
 template <typename Linear, typename Box>
 struct default_strategy<Linear, Box, segment_tag, box_tag, false>
-    : strategy::distance::services::default_strategy
-          <
-              segment_tag, box_tag,
-              typename point_type<Linear>::type,
-              typename point_type<Box>::type
-          >
+    : default_strategy_polygonal_or_linear<Linear, Box>
```
awulkiew, line 126: Doesn't it cause some cycle of base classes?

`default_strategy<segment, box>` is derived from `default_strategy_polygonal_or_linear` which is derived from `default_strategy<segment, box>` etc.

My gut feeling tells me that this particular specialization should define a strategy directly with `typedef`. Or am I missing something?

awulkiew: What I'm missing is that this `default_strategy` is in the `namespace detail::distance`.

Issue #699, opened 2020-04-19
Verifying that cartesian side_robust works well in all cases


tinko92:
I would like to submit a PR in the future to move side_robust ( https://github.com/boostorg/geometry/blob/e1521a2735f21f319b7f00929b78dc6f9a3dfdaf/include/boost/geometry/extensions/triangulation/strategies/cartesian/side_robust.hpp ) from extensions into the library proper. It is a robust cartesian side strategy with tuneable robustness/performance. 

In #690 it was mentioned by @barendgehrels that it has not been verified whether it works well for all cases. Because of that, I thought it might be sensible to submit an issue first, so that this matter can be discussed/possibly resolved.

I thought about how to verify correctness and robustness and came up with the following test:

https://gist.github.com/tinko92/c6d8b862b8e4b3188b0eea41cfdfb17d

The gist of the gist is that the results of side_robust for points with
- uniformly distributed double coordinates
- coordinates with random magnitude
- random coordinates on a grid, randomly perturbed by rotation and scale transformations
are internally consistent and consistent with the results of Exact_predicates_exact_constructions_kernel from CGAL. So I am reasonably certain that the implementation is correct.

What is not yet covered by side_robust, though, are underflow and overflow.

barendgehrels:
Hi @tinko92 , thanks for this, it is very interesting.
Can you also compare the (speed) performance? Side is called a lot in our code, so we have to know how much slower (or faster, but robust is usually a bit slower) it is...

tinko92:
After some further analysis, I was able to improve the performance of side_robust for non-degenerate input (see PR #701 ). For the sake of brevity, I will only include results for uniformly distributed points and purposefully degenerate points and only for the computer with the Xeon CPU:

uniformly distributed points
GCC 9.3: 93ms
Clang 10: 65ms

purposefully degenerate points
GCC 9.3: 813ms
Clang 10: 980ms

Correctness is unaffected. The results for points with random exponents and for points on a perturbed grid are similar to uniformly distributed points. With these changes, side_robust is now more than twice as fast as side_by_triangle for inputs that are not intentionally created in a way that triggers the most expensive code path.

I believe the only question that is really matters when it comes to the performance penalty for robustness is what the execution path looks like for non-degenerate input: 

[...]

PR #701, opened 2020-04-22
Performance improvements for side_robust

tinko92:
Hello,

this PR contains some changes for implementation details of side_robust that (on my machine, for the benchmarks that can be found in https://github.com/boostorg/geometry/issues/699#issuecomment-617775379 ) lead to a speedup of ~x1.5 with GCC and ~x4 with Clang in the execution of side_robust::apply for completely random input and a more moderate speedup for degenerate input. Correctness has been verified using the tests linked in the opening post found in https://github.com/boostorg/geometry/issues/699 .

### DIFF
--- a/include/boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp
+++ b/include/boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp
@@ -88,42 +88,6 @@ inline std::array<RealNumber, 2> two_diff(RealNumber const a,
     return {{ x, y }};
 }
 
-// constexpr power-method, helper for splitter
-template
-<
-    typename RealNumber
->
-constexpr RealNumber int_pow(RealNumber const base,
-                             int exp,
-                             RealNumber out = 1.0)
-{
-    return exp < 1 ? out :
-        int_pow<RealNumber>(base*base, exp/2, (exp % 2) ? out*base : out);
-}
-
-// consexpr method to compute 2^s + 1 as in Theorem 17, page 18
-template
-<
-    typename RealNumber
->
-constexpr RealNumber splitter()
-{
-    return int_pow<RealNumber>(2.0,
-            (std::numeric_limits<RealNumber>::digits + 1) / 2) + 1;
-}
-
-// see theorem 17, page 18
-template
-<
-    typename RealNumber
->
-inline std::array<RealNumber, 2> split(RealNumber const a) {
-    RealNumber c = splitter<RealNumber>() * a;
-    RealNumber a_big = c - a;
-    RealNumber a_hi = c - a_big;
-    return {{ a_hi, a - a_hi }};
-}
-
 // see theorem 18, page 19
 template
 <
@@ -133,12 +97,7 @@ inline RealNumber two_product_tail(RealNumber const a,
                                    RealNumber const b,
                                    RealNumber const x)
 {
-    std::array<RealNumber, 2> a_expansion = split(a);
-    std::array<RealNumber, 2> b_expansion = split(b);
-    RealNumber err1 = x - (a_expansion[0] * b_expansion[0]);
-    RealNumber err2 = err1 - (a_expansion[1] * b_expansion[0]);
-    RealNumber err3 = err2 - (a_expansion[0] * b_expansion[1]);
-    return (a_expansion[1] * b_expansion[1]) - err3;
+    return std::fma(a, b, -x);
 }
 
 // see theorem 18, page 19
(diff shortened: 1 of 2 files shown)


include/boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp
```diff
@@ -376,6 +314,47 @@ inline RealNumber orient2d(std::array<RealNumber, 2> const& p1,
...
+template
+<
+    typename RealNumber,
+    std::size_t Robustness = 3
+>
+inline RealNumber orient2d(std::array<RealNumber, 2> const& p1,
+                           std::array<RealNumber, 2> const& p2,
+                           std::array<RealNumber, 2> const& p3)
+{
+    if(Robustness == 0) {
```
barendgehrels, line 327: @tinko92 thanks for your benchmark and immediate actions on it!
I'll just write down some minor things, including stylistic details that we're used to do differently.

This `{` should go to next line

include/boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp
```diff
@@ -376,6 +314,47 @@ inline RealNumber orient2d(std::array<RealNumber, 2> const& p1,
...
+        return (p1[0]-p3[0])*(p2[1]-p3[1]) - (p1[1]-p3[1])*(p2[0] - p3[0]);
```
barendgehrels, line 328: Can you add spaces here between all mathematical operators? `(p1[0] - p3[0]) * `

include/boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp
```diff
@@ -376,6 +314,47 @@ inline RealNumber orient2d(std::array<RealNumber, 2> const& p1,
...
+    }
+    std::array<RealNumber, 2> t1, t2, t3, t4;
```
barendgehrels, line 330: Using this means that we can merge it only in `1.75` as C++03 is then gone.
Alternatively you can use `boost::array` - but I don't know if more of your code is dependent on C++11

include/boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp
```diff
@@ -376,6 +314,47 @@ inline RealNumber orient2d(std::array<RealNumber, 2> const& p1,
...
+    t3[0] = p1[1] - p3[1];
+    t4[0] = p2[0] - p3[0];
+    std::array<RealNumber, 2> t5_01, t6_01;
+    t5_01[0] = t1[0] * t2[0];
+    t6_01[0] = t3[0] * t4[0];
+    RealNumber det = t5_01[0] - t6_01[0];
+    RealNumber const magnitude = std::abs(t5_01[0]) + std::abs(t6_01[0]);
+
+    // see p.39, mind the different definition of epsilon for error bound
+    RealNumber const A_relative_bound =
```
barendgehrels, line 342: We usually don't use capitals in variable names (unless there are good reasons to do so - and maybe you have them, if they are literally on the pages you refer to)

tinko92:
Thank you for the review, I will try to address all those issues by tomorrow evening.

include/boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp
tinko92: That is indeed the reason. All capitals in variable names in this file where chosen to mimic the notation of Shewchuk's paper as closely as possible.

include/boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp
tinko92: Would that establish an interface or would later breaking that interface (to use std::array again) be ok because it is in the detail namespace?

I would be fine if it lives in develop/extensions until 1.75 next year. I could also leave it this way and revisit the option to change it to `boost::array` only, if the inclusion of side_robust into the master branch will be considered before 1.74.

include/boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp
```diff
@@ -133,12 +97,7 @@ inline RealNumber two_product_tail(RealNumber const a,
                                    RealNumber const b,
                                    RealNumber const x)
 {
-    std::array<RealNumber, 2> a_expansion = split(a);
-    std::array<RealNumber, 2> b_expansion = split(b);
-    RealNumber err1 = x - (a_expansion[0] * b_expansion[0]);
-    RealNumber err2 = err1 - (a_expansion[1] * b_expansion[0]);
-    RealNumber err3 = err2 - (a_expansion[0] * b_expansion[1]);
-    return (a_expansion[1] * b_expansion[1]) - err3;
+    return std::fma(a, b, -x);
```
awulkiew, line 100: This also requires C++11 so either we have to wait til 1.75 or enable it conditionally based on the compiler.

include/boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp
awulkiew: Would it make sense to replace `array` with your own clearly named type? It's a 2d vector correct? If that's the case then why not call it like that (e.g. `vec2d<>`) and have `x` and `y` members or use `boost::model::point_xy<>`?

awulkiew:
Regarding the C++11 requirement. It would still be fine to merge it not only into the extensions but also in the official part of Geometry and even prior to 1.75 as long as this file is not included by any other file in the library that the user may be including right now. We already has C++11 in one of the Karney geodesic formulas released. It is simply not included by any other file.

include/boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp
tinko92: Thanks for the suggestion with regard to the interface, it is true that the values taken there correspond to points, so having points as parameters would be sensible. The matter of the remaining std::arrays would no longer be a matter of the interface of the function.

I care a lot about the performance of those functions, though. Judging from the source code, I guess that point_xy and boost::array are zero-overhead like std::array but I'll verify that in the next couple of days via benchmarks/looking at the assembly and then I'll commit the necessary changes.

tinko92: @awulkiew I replaced the arrays for p1, p2, p3 with an own type (named vec2d). All other arrays in the function are expansions of single values and not 2d or nd vectors, so I did not touch those. Out of your two suggestions (own type and point_xy), I opted for my own type because I noticed that point_xy is not included almost nowhere (except in geometry/strategies/geographic/distance.hpp ), so I didn't want to create a header that depends on it.

I'm fine with your general idea about not including side_robust anywhere by default before 1.75, So I did not change anything with regard to std::array and std::fma.

Issue #703, opened 2020-04-23
read_wkt() too strict on whitespace?


dkondor:
This might be more of a feature request than a real issue. Any case, I'm deciding to report it here in case someone else runs into the same problem.

Currently, read_wkt() throws an exception if the input contains a tab or newline character anywhere. This can be a hassle if the WKT is coming from a source which includes any of these.

I understand this might be an intentional design decision. At the same time, not being an expert in the field, I did not easily find information about what kind of whitespace is allowed in WKT format.

Would it be possible for read_wkt() to be more relaxed in this regard? (Either by default or with an optional parameter)

If not, it would be really helpful if the documentation explicitly stated that only spaces are allowed. Also, the exception texts are hard to interpret, e.g. from the exception texts, it is hard to understand that the problem is with the wrong type of whitespace included in the string. It would be really helpful if the exception texts could mention that the problematic character is a tab or newline.

I'm attaching a short example code with a few failing cases.
[wkt_test.txt](https://github.com/boostorg/geometry/files/4520544/wkt_test.txt)

awulkiew:
> I understand this might be an intentional design decision.

Yes, AFAIU the standard allows only space, see:
https://www.ogc.org/standards/sfa
page 54
```
<space>= " " // unicode "U+0020" (space)
```

Allowing tabs seems to be reasonable but newlines not so much. And there is also `\r` as well as all other unicode spaces. And this would be checked for all users, not only the ones who deal with non-conformant inputs. It could be an option, but this would make the interface unnecessarily more complex, even though slightly.

Therefore currently I'm leaning towards leaving it as it is and forcing the user to prepare the correct input before passing it into `read_wkt()`, e.g. with Boost.StringAlgorithm's `replace()` or even better `erase()` to replace/remove all whitespaces.

What is your opinion @mloskot @barendgehrels @vissarion ?

awulkiew:
> I understand this might be an intentional design decision.

Yes, AFAIU the standard allows only space, see:
https://www.ogc.org/standards/sfa
page 54
```
<space>= " " // unicode "U+0020" (space)
```

Allowing tabs seems to be reasonable but newlines not so much. And there is also `\r` as well as all other unicode spaces. And this would be checked for all users, not only the ones who deal with non-conformant inputs. It could be an option, but this would make the interface unnecessarily more complex, even though slightly.

[...]

PR #705, opened 2020-04-28
Added note to documentation about handling of whitespace

dkondor:
Explicitly note that only space is allowed. Should clarify issue #703

### DIFF
--- a/include/boost/geometry/io/wkt/read.hpp
+++ b/include/boost/geometry/io/wkt/read.hpp
@@ -903,6 +903,12 @@ struct read_wkt<segment_tag, Segment>
 \tparam Geometry \tparam_geometry
 \param wkt string containing \ref WKT
 \param geometry \param_geometry output geometry
+
+Note that the wkt string is interpreted strictly according to the
+[standard](https://www.ogc.org/standards/sfa); notably, only spaces are
+allowed as whitespace, newline and tab characters will result in an
+error.
+
 \ingroup wkt
 \qbk{[include reference/io/read_wkt.qbk]}
 */


awulkiew:
Thanks for this PR @dkondor ! There are problems with it though.

Your comment is probably not going to be visible in the documentation. Have you [built the docs](https://github.com/boostorg/geometry/wiki/Generating-and-Improving-Documentation) and verified that it works as intended?

I think you have 4 options here:

1. Add it in the qbk file which is included below.
https://github.com/boostorg/geometry/blob/develop/doc/reference/io/read_wkt.qbk
It is included under *Parameters* section in the reference:
https://www.boost.org/doc/libs/1_73_0/libs/geometry/doc/html/geometry/reference/io/wkt/read_wkt.html
with the below
```
\qbk{[include reference/io/read_wkt.qbk]}
```

2. Add your comment in the `\qbk{...}` section which currently only includes the additional qbk file
```
\qbk{
Note that the wkt string ...

[include reference/io/read_wkt.qbk]
}
```

3. Add `\details` doxygen section and put the information there, like it is the case e.g. in `area`:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/area.hpp#L300
which is used to build this:
https://www.boost.org/doc/libs/1_73_0/libs/geometry/doc/html/geometry/reference/algorithms/area/area_1.html
Though in this case I'm not sure what to do with the link (see below). It is possible that enclosing it with `\qbk{...}` could work but I'm not sure. You should test it.

4. Add it in the documentation directory above
https://github.com/boostorg/geometry/blob/develop/doc/reference/io/wkt_format.qbk
which is used to build this page with general info:
https://www.boost.org/doc/libs/1_73_0/libs/geometry/doc/html/geometry/reference/io/wkt.html

----

Furthermore we use QuickBook so the markup language is different than GitHub's. Links are defined as follows:
```
[@https://www.ogc.org/standards/sfa standard]
```

see [QuickBook documentation](https://www.boost.org/doc/libs/1_73_0/doc/html/quickbook/syntax/phrase.html#quickbook.syntax.phrase.links).

----

Last but not least, I'm not sure where is the best place for this info. Without giving it much thought I'd say that either the `\details` section (3.) or the directory above with the general info about the standard (4.). What is your opinion @dkondor @mloskot @vissarion @barendgehrels?

vissarion:
>Last but not least, I'm not sure where is the best place for this info. Without giving it much thought I'd say that either the \details section (3.) or the directory above with the general info about the standard (4.). What is your opinion @dkondor @mloskot @vissarion @barendgehrels?

Adding this to  the general info about the standard (4.) makes more sense to me.

barendgehrels:
Replaced by #1106

PR #706, opened 2020-04-28
Tweaked exception text in case problem is with invalid whitespace

dkondor:
Clarify exception text in case the parsing error might be the result of the string including tab ('\t') or newline ('\r' or '\n') character. Suggestion for #703 

In my opinion, current exception text can be hard to interpret in this case. Since looking at the output, all whitespace looks similar, it might not be obvious which character is problematic. E.g.
```
Too many tokens at '
' in 'LINESTRING(1.0 2.0, 2.0 1.0)
'
```

or especially:
```
bad lexical cast: source type value could not be interpreted as target at '
' in 'LINESTRING(1.0 2.0, 
 2.0 1.0)'
```
since this looks like a regular line break in my terminal.

### DIFF
--- a/include/boost/geometry/io/wkt/read.hpp
+++ b/include/boost/geometry/io/wkt/read.hpp
@@ -82,6 +82,12 @@ struct read_wkt_exception : public geometry::exception
             source += "'";
         }
         complete = message + source + " in '" + wkt.substr(0, 100) + "'";
+        // add extra diagnostic message if the problematic token included
+        // common invalid whitespace characters to help users
+        if (it->find_first_of("\t\r\n") != std::string::npos)
+        {
+			complete += "\n(note: newline and tab characters are not allowed in WKT)";
+		}
     }
 
     read_wkt_exception(std::string const& msg, std::string const& wkt)


include/boost/geometry/io/wkt/read.hpp
```diff
@@ -82,6 +82,12 @@ struct read_wkt_exception : public geometry::exception
             source += "'";
         }
         complete = message + source + " in '" + wkt.substr(0, 100) + "'";
+        // add extra diagnostic message if the problematic token included
+        // common invalid whitespace characters to help users
+        if (it->find_first_of("\t\r\n") != std::string::npos)
+        {
+			complete += "\n(note: newline and tab characters are not allowed in WKT)";
+		}
```
awulkiew, line 90: We don't use tabs. :)

mloskot: https://github.com/boostorg/geometry/wiki/Guidelines-for-Developers

include/boost/geometry/io/wkt/read.hpp
```diff
@@ -82,6 +82,12 @@ struct read_wkt_exception : public geometry::exception
             source += "'";
         }
         complete = message + source + " in '" + wkt.substr(0, 100) + "'";
+        // add extra diagnostic message if the problematic token included
+        // common invalid whitespace characters to help users
+        if (it->find_first_of("\t\r\n") != std::string::npos)
+        {
+			complete += "\n(note: newline and tab characters are not allowed in WKT)";
```
awulkiew, line 89: Should we add line break here instead of simply appending the note to the message? We con't know exactly how this message is going to be used by the user. It not necessarily is going to be displayed using cout/cerr.

include/boost/geometry/io/wkt/read.hpp
```diff
@@ -82,6 +82,12 @@ struct read_wkt_exception : public geometry::exception
             source += "'";
         }
         complete = message + source + " in '" + wkt.substr(0, 100) + "'";
+        // add extra diagnostic message if the problematic token included
+        // common invalid whitespace characters to help users
+        if (it->find_first_of("\t\r\n") != std::string::npos)
```
awulkiew, line 87: I'm ok with this approach but I'd like to present different ones and get your feedback. What do you think about the following alternatives:

1. Check wider range of invalid characters because this might potentially help other users, not only the ones dealing with whitespaces and add this specific character to the additional info note below.

2. Escape special characters in `wkt.substr(0, 100)` in the message instead of adding the note, e.g. put tab as `\t`.

barendgehrels:
It's now supported.
https://github.com/boostorg/geometry/commit/f92671b933e585ecba3c1b02c830f6e90e813da9
Closing.

PR #708, opened 2020-04-29
[buffer] remove rescaling from the turn-in-original check

barendgehrels:
This is the follow up, it is shorter

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
@@ -123,16 +123,11 @@ struct buffered_piece_collection
 {
     typedef typename geometry::point_type<Ring>::type point_type;
     typedef typename geometry::coordinate_type<Ring>::type coordinate_type;
-    typedef typename geometry::robust_point_type
-    <
-        point_type,
-        RobustPolicy
-    >::type robust_point_type;
 
     // Robust ring/polygon type, always clockwise
-    typedef geometry::model::ring<robust_point_type> robust_ring_type;
+    typedef geometry::model::ring<point_type> clockwise_ring_type;
 
-    typedef geometry::model::box<point_type> normal_box_type;
+    typedef geometry::model::box<point_type> box_type;
 
     typedef typename IntersectionStrategy::side_strategy_type side_strategy_type;
     typedef typename IntersectionStrategy::envelope_strategy_type envelope_strategy_type;
@@ -150,32 +145,21 @@ struct buffered_piece_collection
 
     typedef typename IntersectionStrategy::template point_in_geometry_strategy
         <
-            robust_point_type,
-            robust_ring_type
+            point_type,
+            clockwise_ring_type
         >::type point_in_geometry_strategy_type;
 
-    typedef typename geometry::rescale_policy_type
-        <
-            typename geometry::point_type<Ring>::type,
-            typename IntersectionStrategy::cs_tag
-        >::type rescale_policy_type;
-
-    typedef geometry::segment_ratio
-    <
-        typename geometry::coordinate_type<robust_point_type>::type
-    > ratio_type;
 
     typedef buffer_turn_info
     <
         point_type,
-        robust_point_type,
-        ratio_type
+        typename segment_ratio_type<point_type, RobustPolicy>::type
     > buffer_turn_info_type;
 
     typedef buffer_turn_operation
     <
         point_type,
-        ratio_type
+        typename segment_ratio_type<point_type, RobustPolicy>::type
     > buffer_turn_operation_type;
 
     typedef std::vector<buffer_turn_info_type> turn_vector_type;
@@ -237,8 +221,7 @@ struct buffered_piece_collection
 
     struct original_ring
     {
-        typedef geometry::model::box<robust_point_type> robust_box_type;
-        typedef geometry::sections<robust_box_type, 1> sections_type;
+        typedef geometry::sections<box_type, 1> sections_type;
 
         // Creates an empty instance
         inline original_ring()
@@ -246,7 +229,7 @@ struct buffered_piece_collection
             , m_has_interiors(false)
         {}
 
-        inline original_ring(robust_ring_type const& ring,
+        inline original_ring(clockwise_ring_type const& ring,
                 bool is_interior, bool has_interiors,
                 envelope_strategy_type const& envelope_strategy,
                 expand_strategy_type const& expand_strategy)
@@ -266,8 +249,8 @@ struct buffered_piece_collection
                     envelope_strategy, expand_strategy);
         }
 
-        robust_ring_type m_ring;
-        robust_box_type m_box;
+        clockwise_ring_type m_ring;
+        box_type m_box;
         sections_type m_sections;
 
         bool m_is_interior;
@@ -293,7 +276,7 @@ struct buffered_piece_collection
 
     // Specificly for offsetted rings around points
     // but also for large joins with many points
-    typedef geometry::sections<normal_box_type, 2> sections_type;
+    typedef geometry::sections<box_type, 2> sections_type;
     sections_type monotonic_sections;
 
     // Define the clusters, mapping cluster_id -> turns
@@ -329,8 +312,7 @@ struct buffered_piece_collection
         , m_envelope_strategy(intersection_strategy.get_envelope_strategy())
         , m_expand_strategy(intersection_strategy.get_expand_strategy())
         , m_point_in_geometry_strategy(intersection_strategy
-            .template get_point_in_geometry_strategy<robust_point_type,
-                        robust_ring_type>())
+            .template get_point_in_geometry_strategy<point_type, clockwise_ring_type>())
         , m_robust_policy(robust_policy)
     {}
 
@@ -496,7 +478,7 @@ struct buffered_piece_collection
 
         geometry::partition
             <
-                normal_box_type,
+                box_type,
                 include_turn_policy,
                 detail::partition::include_all_policy
             >::apply(m_turns, original_rings, visitor,
@@ -533,9 +515,6 @@ struct buffered_piece_collection
             boost::begin(m_turns); it != boost::end(m_turns); ++it, ++index)
         {
             buffer_turn_info_type& turn = *it;
-
-            // Calculate rescaled turn points
-            geometry::recalculate(turn.robust_point, turn.point, m_robust_policy);
 
             // Update member used
             turn.turn_index = index;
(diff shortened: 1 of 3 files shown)


include/boost/geometry/algorithms/detail/buffer/turn_in_original_visitor.hpp
```diff
@@ -35,7 +35,12 @@ struct original_get_box
     template <typename Box, typename Original>
     static inline void apply(Box& total, Original const& original)
     {
-        geometry::expand(total, original.m_box);
+        typedef typename strategy::expand::services::default_strategy
+            <
+                box_tag, typename cs_tag<Box>::type
+            >::type expand_strategy_type;
```
awulkiew, line 41: The CS should depend on the strategy passed by the user instead of being derived from the Box which AFAIU is based on the Point type of the geometry.

barendgehrels: The strategy passed (currently `IntersectionStrategy`) defines a segment strategy, and not a box or point strategy. So it is better to define it locally.

Besides that, I thought more about this and I don't see any point for a user-defined expand strategy... They are not earth-model dependant, the only special thing is that the dateline and the poles should be handled correctly. Do I miss something?

awulkiew: > The strategy passed (currently IntersectionStrategy) defines a segment strategy, and not a box or point strategy.

It also defines the desired CS of the operation passed by the user. If you pass a strategy explicitly it defines how the operations are done. Geometries are used only at the highest level of an algorithm to get the default main strategy in order to pass it further. Internally the CS is defined always by the strategies.

> So it is better to define it locally. 

It's easier for sure. But it's wrong according to the design rationale of the library.

> I don't see any point for a user-defined expand strategy

Box, expanding, envelope, etc. are different in cartesian vs spherical/geographic. Just like any other CS-specific operation. Call e.g. `expand(BOX(170 0 179 1), POINT(-170 0))` and you'll see the difference which would be `BOX(-170 0 179 1)` in cartesian vs `BOX(170 0 190 1) in spherical/geographic`. Similar is in the case of a box expanded with another box. And I even not mentioned an area near poles where things are even wierder.

With that said this will work, the boxes will simply be bigger than they could be. However we considered passing cartesian geometries and geographic strategy. What if you do the opposite? You have geographic geometries and you pass cartesian strategy. Now this expand will take the winding of the CS at the antimeridian and the poles into account when the user wanted to calculate the cartesian intersection. Example:
1. User stores geographic geometries using `cs::geographic` containing e.g. `POINT(0, 50)`
2. User performs a map projection and stores the result using the same point type, now this point is e.g. `POINT(433000 526440)`
3. User calculates set operation or buffer passing the cartesian strategy explicitly (assuming that it is possible to pass a strategy into `buffer`).

At 3, internally at the line we're talking about here the CS is taken from the geometry and the geographic expand is calculated e.g. for this point `POINT(433000 526440)`. Technically this point is invalid geographical point. In practice `expand` will try to normalize them somehow. The result of this `expand` is going to be some arbitrary junk.

awulkiew:
I raised an issue regarding the strategies. However AFAIU this is how the `buffer` worked before this PR was proposed and the purpose of this PR is not to add propagation of strategies but to remove rescaling. This is why I'm ok with merging it as it is.

include/boost/geometry/algorithms/detail/buffer/turn_in_original_visitor.hpp
```diff
@@ -226,13 +231,13 @@ class turn_in_original_visitor
             return true;
         }
 
-        if (geometry::disjoint(turn.robust_point, original.m_box))
+        if (geometry::disjoint(turn.point, original.m_box))
```
awulkiew, line 234: Just for completness, here the strategy is needed as well. I'm ok with merging it though.

include/boost/geometry/algorithms/detail/buffer/turn_in_original_visitor.hpp
barendgehrels: Thanks for your answer. But this was not exactly my point. I know that expand should be different for cartesian and for spherical/geographic. But the CS is already part of the geometry, as you know. And that is exactly how I am deriving it.

I don't get the point about passing a cartesian geometry and a geographic strategy. Shouldn't that be avoided? But the way I implement it here there is no risk for that.

So my question is again: what is the need for a user-specific expand strategy? Will there ever be an Earth model relevant there?

awulkiew: > I don't get the point about passing a cartesian geometry and a geographic strategy.

I always thought this was the intention. For example it was always possible to call:
```
distance(cart_pt1, cart_pt2, haversine<>());
```
and the result was correct and defined by the strategy.

It was also always possible to call `get_as_radian` with cartesian points.

> Shouldn't that be avoided?

The code compiles, the library doesn't complain about it, there is no static assert, there is no explicit information in the documentation about it AFAICS. You say "avoided" as if it was "it sometimes works and sometimes doesn't", it isn't a particularly firm and well-defined position. :)

I'm surprised that this issue is raised only now. After these pull requests adding the propagation of strategies and `cs::undefined` were merged:
https://github.com/boostorg/geometry/pull/533
https://github.com/boostorg/geometry/pull/605
So yes, it is possible to create geometries with undefined CS and then the user has to pass the strategy and the information about the CS is propagated downwards to all algorithms. As it is possible to not pass a strategy and use the default one taken from the geometry. So it works both ways where the strategy has the priority.

> But the way I implement it here there is no risk for that.

You mean this code will only be called in one CS?

> So my question is again: what is the need for a user-specific expand strategy? Will there ever be an Earth model relevant there?

It doesn't matter if we consider the earth model or the strategy containing it. The earth model also has the information about the CS. My reasoning is based on the fact that if the user passes a strategy, this strategy defines a CS/model and that all of the strategies used in an algorithm has to be consistent with each other.

At this point we have a few options:
0. continue without expressing explicitly what the behavior is when CSes of geometries and strategies are mixed
1. enforce a match of the CS between geometries and strategies with a static assert which I think is a breaking change at this point
2. only use the CS information of the geometry, impossible since strategies can be passed
3. only use the CS information of the strategy, impossible as well because of the default behavior is defined this way
4. allow both with one preferred behavior, define all CS-related information in the strategy with the geometry defining the default behavior.

I thought that 4. was always the intention but we were not pursuing it well enough and were conveniently avoiding it.

Btw, if the propagation of the model is the only problem here, why do we even need an umbrella strategy? If the CS is defined by the geometry and the only purpose of the strategy passed by the user is to define the model then we could simply take this model and create all default strategies below from it. But the model also defines the CS. We'd still have the same problem here, i.e. if the model was the ellipsoid should we use cartesian expand or geographic/spherical expand?

My reasoning is that either we define everything CS-related in a strategy or neither of it. Otherwise it's confusing which information is taken from where.

barendgehrels: Thanks, I'll come back to this later.

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -293,7 +276,7 @@ struct buffered_piece_collection
 
     // Specificly for offsetted rings around points
     // but also for large joins with many points
-    typedef geometry::sections<normal_box_type, 2> sections_type;
+    typedef geometry::sections<box_type, 2> sections_type;
```
awulkiew, line 279: @barendgehrels I stumbled upon a problem regarding this change. The sections and their boxes are used in partition:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp#L592
and then deeper here:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp#L224-L225
and then here:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp#L161-L164
then here:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/sections/section_functions.hpp#L136
which as you can see from the parameters relies on the fact that both points and boxes are robust.
And finally here:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/sections/section_functions.hpp#L136
which causes robust point be compared with non-robust boxes.

Now, I didn't check how sections/boxes are created. If boxes are created from robust points and then simply the value is stored using `double` then I guess we can loose precision when integer values are compared with doubles in which case the bounding box may not cover the robust point (maybe?). If they are created from original points then comparing them with robust points doesn't make any sense because robust coordinates are rescaled.

From the values I see:
```
point            | {6.3920000000000003, 2.5440000000000000}    | boost::geometry::model::point<double,2,boost::geometry::cs::cartesian>
robust_point     | {152743, -763011}                           | boost::geometry::model::point<__int64,2,boost::geometry::cs::cartesian>
other_robust_box | {{-428079.00000000012, -554791.00000000012},
                    {428086.00000000012, -554790.99999999988}} | boost::geometry::model::box<boost::geometry::model::point<double,2,boost::geometry::cs::cartesian> >
point_robust_box | {{-360956.00000000006, -763011.00000000012},
                    {152743.00000000003, -78078.999999999985}} | boost::geometry::model::box<boost::geometry::model::point<double,2,boost::geometry::cs::cartesian> >
```
I think it is the former, so normal point is passed here, sections use the same coordinate type as the normal point (double) but they hold the rescaled values. If that's the case then changing `sections_type` to use `box_type` instead of `robust_box_type` was wrong in the worst case and useless in the best case because values are still rescaled anyway.

This is non-detectable for native integer robust types but if you wanted to use MultiPrecision integral type as robust type then the less-comparison between MultiPrecision integral and `double` does not exist. I have a branch where this is done if you want to play with it.

What do you think we should do here?

All of this can be reproduced e.g. with `buffer_multi_polygon.cpp` test case.

P.S. I also don't understand why does buffer deal with sections manually like that. Partition should do the space partitioning and call visitor accordingly. An algorithm should not interfere with that. But that's a different story.

barendgehrels: Hi Adam, thanks, I think I recently saw this too and fixed it in my next PR, which is not yet ready.
I’ll come back to it next Wednesday.

barendgehrels: I can reproduce it, but it's a different thing than I thought. I'm working on it.

PR #711, opened 2020-05-06
[buffer] Fix direction code for some new cases in buffer

barendgehrels:
The similar_direction function could be erroneous for corner cases (related to accuracy),
refactored it out.

This also makes it a bit less dependent on infinite_line (cartesian).

In the test `recursive_polygons_buffer.cpp` with options `--count 10000 --level 6 --form triangle` 
the error count reduces from ~120 to ~15 (of 10000 cases), so a considerable reduction in wrong buffering. So many these wrong results were apparently caused by this error. Four of these cases, previously going wrong, are added to the unit test.


include/boost/geometry/arithmetic/infinite_line_functions.hpp
```diff
@@ -76,19 +76,6 @@ side_value(model::infinite_line<Type> const& line, Point const& p)
...
-// false: -----------------> p <----------------- q
-template <typename Type>
-inline
-bool similar_direction(const model::infinite_line<Type>& p,
-                       const model::infinite_line<Type>& q)
-{
-    return p.a * q.a >= 0 && p.b * q.b >= 0;
-}
-
 template <typename Type>
```
barendgehrels, line 79: What could go wrong was that (fr example) `p.b` was `-1.00e-16` and `q.b` was `+1.00e-16` and therefore their product was negative, while the segments obviously had the same direction...

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
```diff
@@ -177,14 +167,14 @@ struct buffer_range
...
+    static inline bool same_direction(output_point_type const& p0,
             output_point_type const& p1,
             output_point_type const& p2)
     {
-        typedef model::infinite_line<coordinate_type> line_type;
-        line_type const p = detail::make::make_infinite_line<coordinate_type>(p0, p1);
-        line_type const q = detail::make::make_infinite_line<coordinate_type>(p1, p2);
-        return arithmetic::similar_direction(p, q);
+        typedef typename cs_tag<output_point_type>::type cs_tag;
+        return direction_code<cs_tag>(p0, p1, p2) == 1;
```
awulkiew, line 177: In geographic this'd need the earth model right?

barendgehrels: For this one I'm less sure (I mean: less than expand).
But I guess it's not necessary here either.

The spherical/geographical implementation is not made by me, and I didn't concentrate on it today, I just saw it was there and calling it makes the functionality less Cartesian dependent (than calling infinite_line).

awulkiew: Ah right, this test has to be done after the collinearity check so we know the segments are collinear already and on the same ellipsoid.

barendgehrels:
I assume it's OK to merge, it small. If there still is anything I can address it later

awulkiew:
@barendgehrels This change causes some problems:
https://circleci.com/api/v1.1/project/github/boostorg/geometry/8170/output/102/0?file=true

barendgehrels:
> 
> 
> @barendgehrels This change causes some problems:
> https://circleci.com/api/v1.1/project/github/boostorg/geometry/8170/output/102/0?file=true

Thanks I'll have a look on Wednesday

PR #712, opened 2020-05-13
[overlay] Fix possible instabilities in sort_by_side

barendgehrels:
* caused a crash for non-rescaled code (currently protected by TEST_FAILURE)
* it made the `std::sort`, where it was used, instable and output rubbish (alternatively we could use `stable_sort`)
* plus some minor code refactoring

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
@@ -186,7 +186,14 @@ struct less_by_side
             return on_same(first, second);
         }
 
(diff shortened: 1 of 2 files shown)


include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
```diff
@@ -186,7 +186,14 @@ struct less_by_side
             return on_same(first, second);
         }
 
-        int const side_first_wrt_second = -side_second_wrt_first;
+        int const side_first_wrt_second = m_strategy.apply(m_turn_point, second.point, first.point);
```
barendgehrels, line 189: This block is the fix itself

Issue #718, opened 2020-05-26
Managing memory when R-Tree bulk loading


Caian:
Greetings, I am using R-Tree + shared memory file to store / query a dataset of around 400 million 4D points and I'm relying on bulk loading to favor construction time. But unfortunately I'm projecting that the main memory consumption (currently at 27GB) will be too large for some larger cases.

The consumption above does not happen when calling `insert(begin, end)` from an empty tree, which led me to believe that it is related to the bulk loading process.

If storage is not a limiting factor, is there a way to make the shared memory file larger and use the allocator passed to the constructor for the intermediate results as well?

Thank you,

Caian

awulkiew:
Hi, yes bulk loading algorithm creates a temporary vector for all elements and currently the standard allocator is used for that so elements are stored in-memory which indeed can cause problems if many elements are in the range:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/index/detail/rtree/pack_create.hpp#L177

So the answer is: currently it is not possible, at least not officially.

A workaround could be to modify your local version of Boost at the line I mentioned above passing e.g. the same allocator that was passed into the R-tree. This would however pollute your shared memory with the temporary objects. So something like this (untested):
```
typedef typename boost::container::allocator_traits
        <
            typename allocators_typa::allocator_type
        >::template rebind_alloc<entry_type> temp_allocator_type;
temp_allocator_type temp_allocator(allocators.allocator());
std::vector<entry_type, temp_allocator_type> entries(temp_allocator);
```

I agree that it should be possible to specify how the temporaries are created. It's not clear to me what would be the beast way of doing it. Using the same allocator doesn't seem to be correct approach, plus it could break some existing programs relying on a specific size of shared file, because the memory consumption would be bigger with temporaries stored together with the rtree data.

Another possibilities are e.g.:
- passing special Boost.Geometry-specific kind of compound_allocator, allocator_set, storage_definition, etc. into the rtree, optionally instead of allocator
- passing additional allocator for temporary data into the rtree constructor

So I have to think about it and choose what's best. I also have to look at the newer standards because maybe there is a preferred way of solving this problem. If you have some ideas I'm open to suggestion.

Caian:
Thank you @awulkiew. I stumbled upon this vector and tried to use the shmem allocator, but I think `std::vector`, but it caused problems with `offset_ptr`. 

[...]

PR #720, opened 2020-06-02
Add custom pack allocator to r-tree constructor

Caian:
As discussed in #718, it's not currently possible to specify an allocator for the packing process used during the construction of the r-rtee. This is desirable in some situations when dealing with constrained memory systems.

This PR adds overloads for both `bgi::detail::index::rtree::pack` and `bgi::rtree` classes to support the aforementioned allocator without interfering with the existing interfaces.

### DIFF
--- a/include/boost/geometry/index/rtree.hpp
+++ b/include/boost/geometry/index/rtree.hpp
@@ -439,8 +439,8 @@ public:
     \param parameters        The parameters object.
     \param getter            The function object extracting Indexable from Value.
     \param equal             The function object comparing Values.
-    \param tree_allocator    The allocator object for persistent data in the tree..
-    \param pack_allocator    The temporary allocator object used when packing
+    \param allocator         The allocator object for persistent data in the tree.
+    \param temp_allocator    The temporary allocator object used when packing
 
     \par Throws
     \li If allocator copy constructor throws.
@@ -452,11 +452,11 @@ public:
                  parameters_type const& parameters,
                  indexable_getter const& getter,
                  value_equal const& equal,
-                 allocator_type const& tree_allocator,
-                 PackAlloc const& pack_allocator)
-        : m_members(getter, equal, parameters, tree_allocator)
+                 allocator_type const& allocator,
+                 PackAlloc const& temp_allocator)
+        : m_members(getter, equal, parameters, allocator)
     {
-        pack_construct(first, last, pack_allocator);
+        pack_construct(first, last, temp_allocator);
     }
 
     /*!
@@ -468,8 +468,8 @@ public:
     \param parameters        The parameters object.
     \param getter            The function object extracting Indexable from Value.
     \param equal             The function object comparing Values.
-    \param tree_allocator    The allocator object for persistent data in the tree.
-    \param pack_allocator    The temporary allocator object used when packing.
+    \param allocator         The allocator object for persistent data in the tree.
+    \param temp_allocator    The temporary allocator object used when packing.
 
     \par Throws
     \li If allocator copy constructor throws.
@@ -481,11 +481,11 @@ public:
                           parameters_type const& parameters,
                           indexable_getter const& getter,
                           value_equal const& equal,
-                          allocator_type const& tree_allocator,
-                          PackAlloc const& pack_allocator)
-        : m_members(getter, equal, parameters, tree_allocator)
+                          allocator_type const& allocator,
+                          PackAlloc const& temp_allocator)
+        : m_members(getter, equal, parameters, allocator)
     {
-        pack_construct(::boost::begin(rng), ::boost::end(rng), pack_allocator);
+        pack_construct(::boost::begin(rng), ::boost::end(rng), temp_allocator);
     }
 
     /*!
@@ -493,9 +493,9 @@ public:
 
     The tree is created using packing algorithm and a temporary packing allocator.
 
-    \param first             The beginning of the range of Values.
-    \param last              The end of the range of Values.
-    \param tree_allocator    The allocator object for persistent data in the tree.
+    \param first        The beginning of the range of Values.
+    \param last         The end of the range of Values.
+    \param allocator    The allocator object for persistent data in the tree.
 
     \par Throws
     \li If allocator copy constructor throws.
@@ -504,8 +504,8 @@ public:
     */
     template<typename Iterator>
     inline rtree(Iterator first, Iterator last,
-                 allocator_type const& tree_allocator)
-        : m_members(indexable_getter(), value_equal(), parameters_type(), tree_allocator)
+                 allocator_type const& allocator)
+        : m_members(indexable_getter(), value_equal(), parameters_type(), allocator)
     {
         pack_construct(first, last, boost::container::new_allocator<void>());
     }
@@ -515,8 +515,8 @@ public:
 
     The tree is created using packing algorithm and a temporary packing allocator.
 
-    \param rng               The range of Values.
-    \param tree_allocator    The allocator object for persistent data in the tree.
+    \param rng          The range of Values.
+    \param allocator    The allocator object for persistent data in the tree.
 
     \par Throws
     \li If allocator copy constructor throws.
@@ -525,8 +525,8 @@ public:
     */
     template<typename Range>
     inline explicit rtree(Range const& rng,
-                          allocator_type const& tree_allocator)
-        : m_members(indexable_getter(), value_equal(), parameters_type(), tree_allocator)
+                          allocator_type const& allocator)
+        : m_members(indexable_getter(), value_equal(), parameters_type(), allocator)
     {
         pack_construct(::boost::begin(rng), ::boost::end(rng), boost::container::new_allocator<void>());
     }
@@ -538,8 +538,8 @@ public:
 
     \param first             The beginning of the range of Values.
     \param last              The end of the range of Values.
-    \param tree_allocator    The allocator object for persistent data in the tree.
-    \param pack_allocator    The temporary allocator object used when packing.
+    \param allocator         The allocator object for persistent data in the tree.
+    \param temp_allocator    The temporary allocator object used when packing.
 
     \par Throws
     \li If allocator copy constructor throws.
@@ -548,11 +548,11 @@ public:
     */
     template<typename Iterator, typename PackAlloc>
     inline rtree(Iterator first, Iterator last,
-                 allocator_type const& tree_allocator,
-                 PackAlloc const& pack_allocator)
-        : m_members(indexable_getter(), value_equal(), parameters_type(), tree_allocator)
+                 allocator_type const& allocator,
+                 PackAlloc const& temp_allocator)
+        : m_members(indexable_getter(), value_equal(), parameters_type(), allocator)
     {
-        pack_construct(first, last, pack_allocator);
+        pack_construct(first, last, temp_allocator);
     }
 
     /*!
@@ -561,8 +561,8 @@ public:
     The tree is created using packing algorithm and a temporary packing allocator.
 
     \param rng               The range of Values.
-    \param tree_allocator    The allocator object for persistent data in the tree.
-    \param pack_allocator    The temporary allocator object used when packing.
+    \param allocator         The allocator object for persistent data in the tree.
+    \param temp_allocator    The temporary allocator object used when packing.
 
     \par Throws
     \li If allocator copy constructor throws.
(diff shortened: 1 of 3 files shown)


awulkiew:
Thanks for this PR!

There is one thing I'd like to discuss though, i.e. the number and the order of constructor's arguments.

You placed the temporary allocator just after the range but the tree allocator at the end. I wonder why? My guess would be that if the user wants to pass a custom allocator then he'd either pass the TreeAllocator only and default-construct the TemporaryAllocator or pass both of them. Does your use case require to pass only the TemporaryAllocator but to use the standard allocator as the TreeAllocator?

Back to the subject. I think that the TemporaryAllocator should not be passed right after the range. It should rather be passed at the end. But I know this may be inconvenient, so we could do something different, i.e. mimic the C++14 interface of `std::set` where it is possible to pass only the `Allocator` and `Compare` is default-constructed (see: https://en.cppreference.com/w/cpp/container/set/set). In the case of the R-tree all three `Parameters`, `IndexableGetter` and `EqualTo` could be ommitted and default-constructed, allowing to pass the Allocator and optionally TempAllocator. So:

Right now these ctors are defined (simplified code):
```
rtree(Par = Par(), InGet = InGet(), EqTo = EqTo())
rtree(Par, InGet, EqTo, All)
rtree(It, It, Par = Par(), InGet= InGet(), EqTo = EqTo(), All = All())
rtree(Rng, Par = Par(), InGet= InGet(), EqTo = EqTo(), All = All())
```

So I propose to add these:
```
rtree(It, It, Par, InGet, EqTo, All, TempAll)
rtree(Rng, Par, InGet, EqTo, All, TempAll)
rtree(It, It, All)
rtree(Rng, All)
rtree(It, It, All, TempAll)
rtree(Rng, All, TempAll)
```

What do you think?

Caian:
Thank you for the feedback @awulkiew,

> Thanks for this PR!
> 
> There is one thing I'd like to discuss though, i.e. the number and the order of constructor's arguments.
> 
> You placed the temporary allocator just after the range but the tree allocator at the end. I wonder why? My guess would be that if the user wants to pass a custom allocator then he'd either pass the TreeAllocator only and default-construct the TemporaryAllocator or pass both of them. Does your use case require to pass only the TemporaryAllocator but to use the standard allocator as the TreeAllocator?

You are right, I either specify both allocators or just the tree allocator. 

I tried to keep the interface as closest as possible to the original overload without reordering arguments. I would put the temporary allocator after the tree allocator for consistency, but C++ would not allow template type inference from the default value of `pack_allocator`.

> Back to the subject. I think that the TemporaryAllocator should not be passed right after the range. It should rather be passed at the end. But I know this may be inconvenient, so we could do something different, i.e. mimic the C++14 interface of `std::set` where it is possible to pass only the `Allocator` and `Compare` is default-constructed (see: https://en.cppreference.com/w/cpp/container/set/set). In the case of the R-tree all three `Parameters`, `IndexableGetter` and `EqualTo` could be ommitted and default-constructed, allowing to pass the Allocator and optionally TempAllocator. So:
> 
> Right now these ctors are defined (simplified code):
> 
> ```
> rtree(Par = Par(), InGet = InGet(), EqTo = EqTo())
> rtree(Par, InGet, EqTo, All)
> rtree(It, It, Par = Par(), InGet= InGet(), EqTo = EqTo(), All = All())
> rtree(Rng, Par = Par(), InGet= InGet(), EqTo = EqTo(), All = All())
> ```
> 
> So I propose to add these:
> 
> ```
> rtree(It, It, Par, InGet, EqTo, All, TempAll)
> rtree(Rng, Par, InGet, EqTo, All, TempAll)
> rtree(It, It, All)
> rtree(Rng, All)
> rtree(It, It, All, TempAll)
> rtree(Rng, All, TempAll)
> ```
> 
> What do you think?

Sound good, no arguments reordered.

Caian:
Constructors modified. Here's a proposal for testing:

```c++
// Adapted from:
// https://www.boost.org/doc/libs/1_73_0/libs/geometry/doc/html/geometry/spatial_indexes/rtree_examples/index_stored_in_shared_memory_using_boost_interprocess.html
//
// TODO other authors?
// Copyright (c) 2011-2014 Adam Wulkiewicz.
// Copyright (c) 2020 Caian Benedicto, Campinas, Brazil.

#include <boost/geometry.hpp>
#include <boost/geometry/geometries/point.hpp>
#include <boost/geometry/geometries/box.hpp>

#include <boost/geometry/index/rtree.hpp>

#include <boost/foreach.hpp>

#include <boost/interprocess/managed_mapped_file.hpp>
#include <boost/interprocess/allocators/allocator.hpp>

#include <vector>
#include <string>

typedef boost::geometry::model::point<float, 2, boost::geometry::cs::cartesian> P;
typedef boost::geometry::model::box<P> B;

template <class Tree>
void query_test(const Tree* tree)
{
    namespace bgi = bg::index;

    std::vector<B> expected;
    expected.push_back(B(P(45, 45), P(45.5, 45.5)));
    expected.push_back(B(P(46, 46), P(46.5, 46.5)));
    expected.push_back(B(P(47, 47), P(47.5, 47.5)));
    expected.push_back(B(P(48, 48), P(48.5, 48.5)));
    expected.push_back(B(P(49, 49), P(49.5, 49.5)));
    expected.push_back(B(P(50, 50), P(50.5, 50.5)));
    expected.push_back(B(P(51, 51), P(51.5, 51.5)));
    expected.push_back(B(P(52, 52), P(52.5, 52.5)));
    expected.push_back(B(P(53, 53), P(53.5, 53.5)));
    expected.push_back(B(P(54, 54), P(54.5, 54.5)));
    expected.push_back(B(P(55, 55), P(55.5, 55.5)));

    std::vector<B> result; // Yes, could have been done with iterators...
    unsigned k = tree->query(bgi::intersects(B(P(45, 45), P(55, 55))),
        std::back_inserter(result));

    // TODO ASSERT
    if (k != 11) throw 0;

    BOOST_FOREACH(B const& b, result)
    {
        // TODO ASSERT
        // TODO box comparison?

        bool found = false;

        BOOST_FOREACH(B const& e, expected)
        {
            if (bg::equals(b, e)) { found = true; break; }
        }

        if (!found) throw 0;
    }
}

template <class Par, class I, class E>
void test_tree()
{
    using namespace boost::interprocess;

    namespace bgi = bg::index;

    typedef allocator<B, managed_mapped_file::segment_manager> Alloc;
    typedef bgi::rtree<B, Par, I, E, Alloc> Rtree;

    const char* tree_filename = "RRreeFile";
    const size_t tree_filesize = 10*1024*1024; // 10MB should be more than enough

    const char* pack_filename = "CacheFile";
    const size_t pack_filesize = 10*1024*1024; // 10MB should be more than enough

    // Create the tree

    {
        file_mapping::remove(tree_filename);
        managed_mapped_file tree_segment(create_only, tree_filename, tree_filesize);

        file_mapping::remove(pack_filename);
        managed_mapped_file pack_segment(create_only, pack_filename, pack_filesize);

        std::vector<B> objects;

        for ( float i = 0 ; i < 100 ; i += 1 )
            objects.push_back(B(P(i, i), P(i+0.5f, i+0.5f)));

        Rtree* tree;

        // rtree(It, It, Par, InGet, EqTo, All, TempAll)

        tree = tree_segment.construct<Rtree>("Rtree1")(
            objects.begin(), objects.end(),
            Par(), I(), E(),
            Alloc(tree_segment.get_segment_manager()),
            Alloc(pack_segment.get_segment_manager()));

        query_test(tree);

        // rtree(Rng, Par, InGet, EqTo, All, TempAll)

        tree = tree_segment.construct<Rtree>("Rtree2")(
            objects,
            Par(), I(), E(),
            Alloc(tree_segment.get_segment_manager()),
            Alloc(pack_segment.get_segment_manager()));

        query_test(tree);

        // rtree(It, It, All)

        tree = tree_segment.construct<Rtree>("Rtree3")(
            objects.begin(), objects.end(),
            Alloc(tree_segment.get_segment_manager()));

        query_test(tree);

        // rtree(Rng, All)

        tree = tree_segment.construct<Rtree>("Rtree4")(
            objects,
            Alloc(tree_segment.get_segment_manager()));

        query_test(tree);

        // rtree(It, It, All, TempAll)

        tree = tree_segment.construct<Rtree>("Rtree5")(
            objects.begin(), objects.end(),
            Alloc(tree_segment.get_segment_manager()),
            Alloc(pack_segment.get_segment_manager()));

        query_test(tree);

        // rtree(Rng, All, TempAll)

        tree = tree_segment.construct<Rtree>("Rtree6")(
            objects,
            Alloc(tree_segment.get_segment_manager()),
            Alloc(pack_segment.get_segment_manager()));

        query_test(tree);
    }

    {
        managed_mapped_file segment(open_only, tree_filename);
        Rtree* tree = segment.find<Rtree>("Rtree1").first;
        query_test(tree);
        segment.destroy<Rtree>("Rtree1");
    }

    {
        managed_mapped_file segment(open_only, tree_filename);
        Rtree* tree = segment.find<Rtree>("Rtree2").first;
        query_test(tree);
        segment.destroy<Rtree>("Rtree2");
    }

    {
        managed_mapped_file segment(open_only, tree_filename);
        Rtree* tree = segment.find<Rtree>("Rtree3").first;
        query_test(tree);
        segment.destroy<Rtree>("Rtree3");
    }

    {
        managed_mapped_file segment(open_only, tree_filename);
        Rtree* tree = segment.find<Rtree>("Rtree4").first;
        query_test(tree);
        segment.destroy<Rtree>("Rtree4");
    }

    {
        managed_mapped_file segment(open_only, tree_filename);
        Rtree* tree = segment.find<Rtree>("Rtree5").first;
        query_test(tree);
        segment.destroy<Rtree>("Rtree5");
    }

    {
        managed_mapped_file segment(open_only, tree_filename);
        Rtree* tree = segment.find<Rtree>("Rtree6").first;
        query_test(tree);
        segment.destroy<Rtree>("Rtree6");
    }

    file_mapping::remove(tree_filename);
    file_mapping::remove(pack_filename);
};

int main(int argc, char *argv[])
{
    namespace bgi = bg::index;

    test_tree<
        bgi::linear<32, 8>,
        bgi::indexable<B>,
        bgi::equal_to<B>
    >();

    return 0;
}
```

awulkiew:
Thanks! I'll review it tomorrow.

Regarding the testing. Here are the tests for Rtree+interprocess:
https://github.com/boostorg/geometry/tree/develop/index/test/rtree/interprocess
These constructors could either be added there somewhere or a separate file could be created. The latter is probably simpler. Have in mind we use Boost.Test.

include/boost/geometry/index/rtree.hpp
```diff
@@ -431,13 +426,156 @@ class rtree
...
+
+    The tree is created using packing algorithm and a temporary packing allocator.
+
+    \param first             The beginning of the range of Values.
+    \param last              The end of the range of Values.
+    \param parameters        The parameters object.
+    \param getter            The function object extracting Indexable from Value.
+    \param equal             The function object comparing Values.
+    \param tree_allocator    The allocator object for persistent data in the tree..
+    \param pack_allocator    The temporary allocator object used when packing
```
awulkiew, line 443: I think for consistency with other constructors I'd prefer keeping the old name `allocator` instead of `tree_allocator`. I think it's clear enough.

What do you think about changing the name of `pack_allocator` to `temp_allocator`? I think it would be more clear what is happening there, i.e. it's an allocator used in construction of temporary data.

awulkiew:
Looks good. I only had one remark regarding naming.
Btw, if you like please add yourself to the list of contributors: https://github.com/boostorg/geometry/blob/develop/doc/geometry.qbk#L109

include/boost/geometry/index/rtree.hpp
Caian: I agree.

Caian:
> Looks good. I only had one remark regarding naming.
> Btw, if you like please add yourself to the list of contributors: https://github.com/boostorg/geometry/blob/develop/doc/geometry.qbk#L109

I'll be glad. Should I add copyrights to lines 8 and 17 as well?

awulkiew:
> Should I add copyrights to lines 8 and 17 as well?
No, AFAIU the authors section is reserved for maintainers. Regarding the line 8, that would technically be needed. So to avoid further confusion I'll do it after the PR is merged.

Caian:
Updated

awulkiew:
Would you like me to merge this now or do you plan to add a test file for this PR?

Caian:
I prefer to add it to a next PR.

awulkiew:
Ok, thanks!

PR #723, opened 2020-06-10
[test] update testsuite

barendgehrels:
- some cases are fixed in the meantime, this is updated
- in `difference` there might be slivers which are still OK, therefore an `count_set` is used there
- when `BOOST_GEOMETRY_TEST_FAILURES` is defined, it also writes expected failures - this makes it much easier to detect improvements (or regressions)
- this also shows insight in the current state, which is more or less in par with the rescaling - this means we might switch in next release (and one or two releases later remove the rescaling code)


barendgehrels:
Merged, assuming no objections, and next PR is to built on top of this one.
If there are still considerations, let me know.

PR #724, opened 2020-06-10
[intersection] refactor segment-intersection class

barendgehrels:
This is to avoid using both non-rescaled and rescaled coordinates in one method.

It is completely independent of the other PR created today, and should result in no failures


awulkiew:
Should I mark these changes related to rescaling as enhancements or bugfixes?

barendgehrels:
@awulkiew it's not a bugfix, and indeed it's related to rescaling. So yes, in detail it's an enhancement

PR #730, opened 2020-06-19
Fix MSVC warning C4127: conditional expression is constant

### LINKED ISSUE #727: MSVC Warning: error C4127: conditional expression is constant
Compiling with boost 1.73.0, MSVC 2017 Version 15.8.0 with compiler options /W4 and /std:c++latest, I get "conditional expression is constant" in two places:
```
\boost_1_73_0\boost\geometry\algorithms\detail\overlay\follow.hpp(491): error C4127: conditional expression is constant
\boost_1_73_0\boost\geometry\algorithms\detail\overlay\follow.hpp(491): note: consider using 'if constexpr' statement instead
\boost_1_73_0\boost\geometry\algorithms\detail\overlay\intersection_insert.hpp(390): note: see reference to function template instantiation 'OutputIterator boost::geometry::detail::overlay::follow<GeometryOut,LineString,Areal,boost::geometry::overlay_difference,false,false>::apply<std::deque<turn_info,std::allocator<_Ty>>,OutputIterator,RobustPolicy,Strategy>(const LineString &,const Polygon &,boost::geometry::detail::overlay::operation_type,Turns &,const RobustPolicy &,OutputIterator,const Strategy &)' being compiled
```
and
```
\boost_1_73_0\boost\geometry\algorithms\detail\overlay\follow.hpp(523): error C4127: conditional expression is constant
\boost_1_73_0\boost\geometry\algorithms\detail\overlay\follow.hpp(523): note: consider using 'if constexpr' statement instead
```

mloskot:
Fixes #727 

-----

MWE that triggers the warning

```cpp
#include <deque>
#include <boost/geometry.hpp>
int main()
{
    using line_t = bg::model::linestring<bg::model::d2::point_xy<double>>;
    line_t green, blue;

    bg::read_wkt("LINESTRING(0 0, 2 0)", green);
    bg::read_wkt("LINESTRING(-1 -1, -1 2)", blue);

    std::deque<line_t> output;
    bg::union_(green, blue, output);
    bg::intersection(green, blue, output);
}
```

------

It's worth to be aware usage of the macro

https://github.com/boostorg/geometry/blob/cff2ee9d6c30afba4fe6cf64b3d65080372f8a5d/include/boost/geometry/util/condition.hpp#L35

may trigger the MSVC [code analysis warning C6319](https://docs.microsoft.com/en-us/cpp/code-quality/c6319).

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/follow.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/follow.hpp
@@ -30,6 +30,8 @@
 #include <boost/geometry/algorithms/detail/relate/turns.hpp>
 #include <boost/geometry/algorithms/detail/tupled_output.hpp>
 
+#include <boost/geometry/util/condition.hpp>
+
 namespace boost { namespace geometry
 {
 
@@ -487,7 +489,7 @@ public :
                     strategy, robust_policy,
                     linear::get(out));
             }
-            else if (FollowIsolatedPoints
+            else if (BOOST_GEOMETRY_CONDITION(FollowIsolatedPoints)
                   && following::is_touching(*it, *iit, entered))
             {
                 debug_traverse(*it, *iit, "-> Isolated point");
@@ -519,7 +521,7 @@ public :
         {
             *linear::get(out)++ = current_piece;
         }
-        else if (FollowIsolatedPoints
+        else if (BOOST_GEOMETRY_CONDITION(FollowIsolatedPoints)
               && current_piece_size == 1)
         {
             action::template isolated_point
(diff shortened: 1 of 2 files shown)


include/boost/geometry/algorithms/detail/overlay/follow.hpp
```diff
@@ -487,8 +487,7 @@ public :
                     strategy, robust_policy,
                     linear::get(out));
             }
-            else if (FollowIsolatedPoints
-                  && following::is_touching(*it, *iit, entered))
+            else if (BOOST_GEOMETRY_CONDITION(FollowIsolatedPoints && following::is_touching(*it, *iit, entered)))
```
awulkiew, line 490: Is there a reason why did you decide to put it in one line instead of keeping the old layout? I.e.:
```
else if (BOOST_GEOMETRY_CONDITION(FollowIsolatedPoints)
      && following::is_touching(*it, *iit, entered))
```

mloskot: None. Perhaps just a legacy and paranoid habit to fit a macro invocation in a single line.

I've reverted the formatting.

mloskot:
Good catch, thanks! Fixed.

PR #732, opened 2020-06-24
Fix/traversal cluster exits

barendgehrels:
* Fix some cases in buffer which went wrong without rescaling
* Therefore extracted cluster_exits functionality (was already there, but in traversal. Now in external file)
* Clean up some defines in robustness test and write test configuration


include/boost/geometry/algorithms/detail/overlay/traversal.hpp
```diff
@@ -741,158 +727,6 @@ public :
...
-                return rp.rank;
-            }
-        }
-        return -1;
-    }
-
-    // Function checks simple cases, such as a cluster with two turns,
-    // arriving at the first turn, first turn points to second turn,
-    // second turn points further.
-    inline bool select_turn_from_cluster_linked(signed_size_type& turn_index,
```
barendgehrels, line 763: this functionality is now moved to `cluster_exit`, a struct is made, and there is one additional condition (the fix)

test/algorithms/buffer/buffer_multi_polygon.cpp
```diff
@@ -551,7 +543,7 @@ int test_main(int, char* [])
 #endif
     
 #if defined(BOOST_GEOMETRY_TEST_FAILURES)
-    BoostGeometryWriteExpectedFailures(1, 9);
+    BoostGeometryWriteExpectedFailures(1, 5);
```
barendgehrels, line 546: It effectively fixes 2 cases

PR #733, opened 2020-07-08
Add example showing how std::array<T, N> can model the Point concept

BenFrantzDale:
This is the documentation I found to be missing. I posted something like it at https://stackoverflow.com/a/62777496/874660 but switched this example to `std::array<T,N>` to be more standard.

### DIFF
--- a/doc/concept/point.qbk
+++ b/doc/concept/point.qbk
@@ -33,6 +33,31 @@ The Point Concept is defined as following:
 	* `get` to get a coordinate value
 	* `set` to set a coordinate value (this one is not checked for ConstPoint)
 	
+[heading Example]
+
+While you can `#include boost/geometry/geometries/adapted/std_array.hpp` to use `std::array<T, D>` to model the Point concept,
+the following code spells it out in detail, as you might use for your own point types:
+```
+namespace boost { namespace geometry { namespace traits
+{
+
+template <typename T, std::size_t D> struct tag<std::array<T, D>> { using type = point_tag; };
+template <typename T, std::size_t D> struct dimension<std::array<T, D>> : boost::mpl::int_<D> {};
+template <typename T, std::size_t D> struct coordinate_type<std::array<T, D>> { using type = T; };
+template <typename T, std::size_t D> struct coordinate_system<std::array<T, D>> { using type = boost::geometry::cs::cartesian; };
+
+template <typename T, std::size_t D, std::size_t Index>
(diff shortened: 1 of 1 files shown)


awulkiew:
I think it's a good idea to provide an example in the documentation.

This example is written in newer standard than the one we currently support. We're moving to the newer one in Boost ver 1.75. So if you prefer it to be included in 1.74 the example should be rewritten. Otherwise we should wait with merging.

Regarding the example at SO, you wrote that the example is in C++17. Why is that? I can only see C++11. Which feature am I missing?

BenFrantzDale:
> This example is written in newer standard than the one we currently support. We're moving to the newer one in Boost ver 1.75. So if you prefer it to be included in 1.74 the example should be rewritten. Otherwise we should wait with merging.

My day job is all C++17 so I'm rusty on what showed up when, but `namespace boost::geometry::traits {` vs `namespace boost { namespace geometry { namespace traits {` is post-C++11. (Looks like it's C++17 https://en.cppreference.com/w/cpp/language/namespace) That may be the only thing. I just rewrote it with the expanded C++11-compatible namespaces.

Also, I'm not familiar with `.qbk` files. I assumed triple back-tick is a code block.

awulkiew:
Ah right, nested namespaces. I didn't even look at them. Thanks for the explanation!

Btw, you could put them in one line. We use this pattern in the library:
```
namespace boost { namespace geometry { namespace traits
{
// ...
}}} // namespace boost::geometry::traits
```
If you prefer it like it is now I'm also fine with that.

Ok, so since you decided to use C++11 I understand that we're waiting for 1.75.

BenFrantzDale:
I’m fine with whatever style you like. Just let me know and I’ll change it.

awulkiew:
I'd prefer all of the namespaces in one line (as I show above) but let's wait with the change for other opinions.

doc/concept/point.qbk
```diff
@@ -33,6 +33,30 @@ The Point Concept is defined as following:
 	* `get` to get a coordinate value
 	* `set` to set a coordinate value (this one is not checked for ConstPoint)
 	
+[heading Example]
+
+While you can `#include boost/geometry/geometries/adapted/std_array.hpp` to use `std::array<T, D>` to model the Point concept, the following code spells it out in detail, as you ight use for your own point types:
```
vissarion, line 38: `ight` -> `might`

PR #736, opened 2020-07-15
Fix errors in some floating point cases

barendgehrels:
* in intersection a check is made for overlap (solving some errors mainly in small scale AND nearly collinear situations)
* in touch a special case is added where intersection reports touching segments, but they are just not touching
* in touch some sides were reported as 0 while, in FP, they are distinguishable and that can be used to fix some errors

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
@@ -369,15 +369,6 @@ struct touch : public base_turn_handler
         return side1 == side2 && ! opposite(side1, turn);
     }
 
-    /*static inline void block_second(bool block, TurnInfo& ti)
-    {
-        if (block)
-        {
-            ti.operations[1].operation = operation_blocked;
-        }
-    }*/
-
-
     template
     <
         typename UniqueSubRange1,
@@ -457,7 +448,6 @@ struct touch : public base_turn_handler
                     {
                         ti.operations[1].operation = operation_blocked;
                     }
-                    //block_second(block_q, ti);
                     return;
                 }
 
@@ -483,7 +473,6 @@ struct touch : public base_turn_handler
                     {
                         ti.touch_only = true;
                     }
-                    //block_second(block_q, ti);
                     return;
                 }
             }
@@ -608,16 +597,16 @@ struct equal : public base_turn_handler
 
             typedef detail::distance_measure<coordinate_type> dm_type;
 
-            dm_type const dm_qk_p
+            dm_type const dm_pk_q2
                = get_distance_measure(range_q.at(1), range_q.at(2), range_p.at(2));
-            dm_type const dm_pk_q
+            dm_type const dm_qk_p2
                = get_distance_measure(range_p.at(1), range_p.at(2), range_q.at(2));
 
-            if (dm_pk_q.measure != dm_qk_p.measure)
+            if (dm_qk_p2.measure != dm_pk_q2.measure)
             {
                 // A (possibly very small) difference is detected, which
                 // can be used to distinguish between union/intersection
-                ui_else_iu(dm_pk_q.measure < dm_qk_p.measure, ti);
+                ui_else_iu(dm_qk_p2.measure < dm_pk_q2.measure, ti);
                 return;
             }
         }
--- a/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
@@ -111,6 +111,60 @@ struct base_turn_handler
         both(ti, condition ? operation_union : operation_intersection);
     }
 
+
+#if ! defined(BOOST_GEOMETRY_USE_RESCALING)
+    template
+    <
+        typename UniqueSubRange1,
+        typename UniqueSubRange2
+    >
+    static inline int side_with_distance_measure(UniqueSubRange1 const& range_p,
+            UniqueSubRange2 const& range_q,
+            int range_index, int point_index)
+    {
+        if (range_index >= 1 && range_p.is_last_segment())
+        {
+            return 0;
+        }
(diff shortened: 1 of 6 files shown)


include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -111,6 +111,60 @@ struct base_turn_handler
...
+
+        typedef typename select_coordinate_type
+                <
+                    typename UniqueSubRange1::point_type,
+                    typename UniqueSubRange2::point_type
+                    >::type coordinate_type;
+
+        typedef detail::distance_measure<coordinate_type> dm_type;
+
+        dm_type const dm = get_distance_measure(range_p.at(range_index), range_p.at(range_index + 1), range_q.at(point_index));
```
awulkiew, line 142: What are these ranges? I don't remember. Shouldn't the type of the index rather be some kind of unsigned integral type like `size_t` or ideally defined by the `UniqueSubRange` type as e.g. `size_type`?

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -369,14 +423,80 @@ struct touch : public base_turn_handler
...
+        // |  \qj       In this case it should block P and
+        // |  ^------   set Q to Union
+        // >----->P     qj is LEFT of P1 and pi is LEFT of Q2
+        //              (the other way round is also possible)
+
+        typedef typename select_coordinate_type
+                <
+                    typename UniqueSubRange1::point_type,
+                    typename UniqueSubRange2::point_type
+                    >::type coordinate_type;
```
awulkiew, line 462: Too much indentation. ;)

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
barendgehrels: We pass small ranges to be able to know the next points (point after two intersecting segments).
I will look what we normally do, assuming indeed we can pass size_t

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
barendgehrels: will fix, thanks & thanks for your review

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
awulkiew: So the size of the range is 1, 2 or 3. Ok, bigger type is not needed then.

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -111,6 +111,60 @@ struct base_turn_handler
...
 
+
+#if ! defined(BOOST_GEOMETRY_USE_RESCALING)
+    template
+    <
+        typename UniqueSubRange1,
+        typename UniqueSubRange2
+    >
+    static inline int side_with_distance_measure(UniqueSubRange1 const& range_p,
+            UniqueSubRange2 const& range_q,
```
vissarion, line 122: is indentation right here?

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -111,6 +111,60 @@ struct base_turn_handler
...
+        if (point_index >= 2 && range_q.is_last_segment())
+        {
+            return 0;
+        }
+
+        typedef typename select_coordinate_type
+                <
+                    typename UniqueSubRange1::point_type,
+                    typename UniqueSubRange2::point_type
+                    >::type coordinate_type;
```
vissarion, line 138: Too much indentation also here

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -369,14 +423,80 @@ struct touch : public base_turn_handler
...
 
-    /*static inline void block_second(bool block, TurnInfo& ti)
+#if ! defined(BOOST_GEOMETRY_USE_RESCALING)
+    template
+    <
+        typename UniqueSubRange1,
+        typename UniqueSubRange2
+    >
+    static inline bool handle_imperfect_touch(UniqueSubRange1 const& range_p,
+            UniqueSubRange2 const& range_q, TurnInfo& ti)
```
vissarion, line 433: indentation issue?

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
barendgehrels: Indeed, OK, thanks

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
barendgehrels: Well, this is how we (at least I) did it from the beginning - an indentation of 8. However, currently IDE's usually automatically align it under the next bracket and some features (`clang_format`) or people also do this. So now we've a mixed approach.

IMO both are not perfect... For the first method it is currently unusual, I think. For the second method, with long names and modifiers, it goes really to the right and needs more lines.

I keep this for now, if you've strong opinions, just comment.

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
barendgehrels: Thanks, indeed. Fixed now, all in this file (there were some other ones too somehow)

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
barendgehrels: all fixed, amended and pushed

PR #738, opened 2020-07-21
[formulas] Normalize resulting longitudes in direct formulas.

awulkiew:
Currently when the point is near antimeridian and the azimuth is pointing out to the other side of it the longitude of the resulting point is out of range, e.g.: `point{179.5, 0}`, `azimuth12{90}` and `distance{111680}` results in longitude around `180.5`.


vissarion:
Why this is a problem? Is it written somewhere that formulas guarantee (-180,180) range or this normalization could be done (if needed) by a higher level algorithm (that uses the direct formulas)?

awulkiew:
Yes, technically that's correct. But you'd have to do the normalization in all strategies and algorithms using these formulas and remember that they do not return coordinates in correct range. This is error-prone. I'd rather have formulas generating correct coordinates from the start. Besides coordinates are already normalized in Karney's formula.

include/boost/geometry/formulas/karney_direct.hpp
```diff
@@ -217,6 +217,14 @@ class karney_direct
             math::normalize_longitude<degree, CT>(lon12);
 
             result.lon2 = lon1 + lon12;
+
+            // For longitudes close to the antimeridian the result can be out
+            // of range. Therefore normalize.
+            // In other formulas this has to be done at the end because
+            // otherwise because differential quantities are calculated
```
vissarion, line 224: second "because" not needed

include/boost/geometry/formulas/spherical.hpp
```diff
@@ -267,6 +267,14 @@ inline result_direct<CT> spherical_direct(CT const& lon1,
...
         result.lon2 = lon1 + lon2 - omg1;
         result.lat2 = lat2;
+
+        // For longitudes close to the antimeridian the result can be out
+        // of range. Therefore normalize.
+        CT const pi = math::pi<CT>();
+        if (result.lon2 > pi)
+            result.lon2 -= 2 * pi;
+        else if (result.lon2 <= -pi)
+            result.lon2 += 2 * pi;
```
vissarion, line 277: why not using `math::normalize_longitude` as in `karney_direct`?

awulkiew: I was wondering about it but decided to not use this one because it is more lightweight. `math::normalize_longitude` is able to normalize any coordinates no matter how far the value is from the correct range. Here we know that the result will be only slightly off.

awulkiew: Yes, that's a good idea. I'll do it.

awulkiew:
> OK, I agree, but since we support this it is useful to state it somewhere (probably in documentation) that the results of the algorithms for non-cartesian CSs are normalized in (-180,180). Also, I am not sure that this holds for all algorithms.

AFAIK they should be. Theoretically we also require the input coordinates in this range but in practice we normalize input coordinates before processing them. The only exception is `envelope()` which can return max longitude greater than 180 deg because max has to be greater than min.

And yes, it should probably be clearly stated in the documentation. A good place would probably be the documentation for `bg::cs::geographic` and `bg::cs::cartesian`. Or there could be a new/separate section describing the difference between coordinate systems, etc. because the users are sometimes confused. We could do this along with introducing the new umbrella strategies which would be consistent accross coordinate systems.

include/boost/geometry/formulas/spherical.hpp
awulkiew: Done. Please let me know whether or not you're fine with the functions.

vissarion: Thanks, it's fine!

PR #739, opened 2020-07-28
 Automatically generated, fast floating-point predicates

tinko92:
This pull request contains the first part of my GSoC 2020 project. This code is still work in progress and there might be changes to the interface before the end of GSoC.

The project aims to implement a number of facilities that allow easy creation of fast and robust geometric floating-point predicates, such as, for example, orientation tests (Boost.Geometry currently has a 2d orientation test in the side-strategy). As of writing this, the pull request contains templates to 

- automatically derive error bounds for the approximate evaluation of such expressions
- build static, almost static and semi-static filters from these expressions

I intend to add
- generation of functions for the exact evaluation of arbitrary arithmetic floating-point expressions containing +, - and * using expansion arithmetic
- improvements for the error bound derivation
- an alternative method for error bound derivation
- some additional filters
- automatic detection of reusable interim results between filter stages

The usage of the various templates for the creation of geometric predicates is illustrated in the tests and the example.

This pull request depends on Boost.MP11, which depends on C++11 but has no additional dependencies. The code itself depends on C++14 because of certain constexpr features.

### DIFF
--- a/include/boost/geometry/extensions/generic_robust_predicates/strategies/cartesian/detail/expressions.hpp
+++ b/include/boost/geometry/extensions/generic_robust_predicates/strategies/cartesian/detail/expressions.hpp
@@ -1,0 +1,103 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+
+// Copyright (c) 2020 Tinko Bartels, Berlin, Germany.
+
+// Contributed and/or modified by Tinko Bartels,
+//   as part of Google Summer of Code 2020 program.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_EXTENSIONS_GENERIC_ROBUST_PREDICATES_STRATEGIES_CARTESIAN_DETAIL_EXPRESSIONS_HPP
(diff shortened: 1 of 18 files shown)


include/boost/geometry/extensions/generic_robust_predicates/strategies/cartesian/detail/expressions.hpp
```diff
@@ -0,0 +1,103 @@
...
+    typename A21, typename A22
+>
+using det2x2 = difference
+    <
+        product<A11, A22>,
+        product<A12, A21>
+    >;
+
+using orient2d = det2x2
+        <
```
vissarion, line 35: why 8 spaces here vs 4 above?

tinko92:
I have made the following changes to the PR:

- The template-based error bound calculation has been replaced by a constexpr based implementation. This improves compile-time, template instantiation depth and readability.
- I have tried to address all feedback (spaces after if, unused parameters, position of "const" in types"
- Staged predicates have been added.

The PR now includes a new test-case in https://github.com/BoostGSoC20/geometry/blob/feature/floating_point_filters/extensions/test/generic_robust_predicates/staged.cpp . It illustrates the intended use of the code in this PR. 

For a given geometric predicate (in the example the side-predicate), an expression type needs to be constructed. Using the expression type and a floating-point calculation type (like double), filters can be constructed from the stage_* files. These filters can be combined into a staged_predicate for usage, for example, in a robust side strategy or any other strategy that can be computed as the sign of a polynomial of floating-point coordinates.

The remaining headers are not intended to be used directly and can be considered implementation details.

Issue #716, opened 2020-05-20
Ring iterators required to return references to points, otherwise dangling references happen


autc04:
According to my reading of the documentation, iterators for the vertices of an adapted ring type are allowed to return any type convertible to `T` from their `operator*()`.

However, linear_segment_or_box.hpp, line 101 constructs a `referring_segment` from dereferenced iterators into a ring:

```
                range_segment rng_segment(*it0, *it1);
```

where `range_segment` is defined as:
```
 typedef typename geometry::model::referring_segment
            <
                point_type const
            > range_segment;
```

If the return value of `*it0` and `*it1` is anything but a `point_type&` or a `const point_type&`, this silently constructs a dangling reference to a temporary, leading to hard-to-debug failures.

The documentation for the ring concept refers to the Boost.Range Random Access Range concept, which says that the iterators must be [Random Access Traversal Iteratos](https://www.boost.org/doc/libs/1_72_0/libs/iterator/doc/new-iter-concepts.html#random-access-traversal-iterators-lib-random-access-traversal-iterators).  There, the only requirement on the return type of `*a` is that it is 
"Convertible to T".

Requiring the return type to be (const) `point_type&`, besides being currently undocumented, precludes writing adapters for types where the points are not stored explicitly.

awulkiew:
Yes, I think it's a bug in the library. Thanks for reporting!

awulkiew:
@autc04 I have a patch for this issue:
https://github.com/boostorg/geometry/pull/740
Could you check if it works for you?

PR #740, opened 2020-08-02
Fix dangling references for referring_segment and proxy references returned by iterators.

awulkiew:
A fix for https://github.com/boostorg/geometry/issues/716

It also removes some unneeded typedefs.

### DIFF
--- a/include/boost/geometry/algorithms/for_each.hpp
+++ b/include/boost/geometry/algorithms/for_each.hpp
@@ -5,10 +5,11 @@
 // Copyright (c) 2009-2014 Mateusz Loskot, London, UK.
 // Copyright (c) 2014 Adam Wulkiewicz, Lodz, Poland.
 
-// This file was modified by Oracle on 2014.
-// Modifications copyright (c) 2014, Oracle and/or its affiliates.
+// This file was modified by Oracle on 2014-2020.
+// Modifications copyright (c) 2014-2020, Oracle and/or its affiliates.
 
 // Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
+// Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
 
 // Parts of Boost.Geometry are redesigned from Geodan's Geographic Library
 // (geolib/GGL), copyright (c) 1995-2010 Geodan, Amsterdam, the Netherlands.
@@ -25,6 +26,7 @@
 
 #include <boost/range.hpp>
 #include <boost/type_traits/is_const.hpp>
+#include <boost/type_traits/is_same.hpp>
 #include <boost/type_traits/remove_reference.hpp>
 
 #include <boost/geometry/algorithms/detail/interior_iterator.hpp>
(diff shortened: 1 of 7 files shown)


include/boost/geometry/algorithms/for_each.hpp
```diff
@@ -94,31 +95,88 @@ struct fe_range_per_point
...
+                <
+                    typename add_const_if_c
+                        <
+                            is_const<Range>::value,
+                            typename point_type<Range>::type
+                        >::type,
+                    typename boost::range_reference<Range>::type,
+                    is_const<Range>::value
+                >::apply(previous, it, f);
             previous = it++;
```
barendgehrels, line 180: This is quite a complex solution to add a const, or not...
Isn't a simpler type possible? 
And in C++11/14 (with auto) maybe?

Just questions, I'm fine anyway.

awulkiew: I can't think of anything else besides encapsulating it inside `fe_range_per_segment_call_f` so at least it wouldn't be required to write it twice, here and below.

At least as long as we want to preserve the current behavior. This is an implementation of `for_each_segment` and calling this for mutable ranges probably has no sense because there are no separate segments in them so if a user modified a point of a mutable segment then another segment would be modified indirectly. So a simple and reasonable fix would be to remove all of this `is_const` logic at all and always require the Range to be `const`. In this case also the `fe_range_per_segment_call_f` would be not needed at all.

Side info: `for_each_segment ` is not used anywhere in the library. At least for now. We use `segment_iterator` instead. But when we move to C++14 I think I'd prefer to use it (or rather implement a similar algorithm `for_each_segment_if` allowing to break the iteration) together with generic lambdas instead of `segment_iterator` because it's a lot lighter.

@mloskot do you have some other idea?

barendgehrels: Thanks for your explanation. Indeed - using `for_each_segment` doesn't really make sense. But removing it is a breaking change.

mloskot: Sorry, no better ideas. If nothing requires passing non-const, then narrowing the type qualification down to single possibility is certainly a good idea, and a  simplification.

awulkiew: @barendgehrels ok, I simplified it preserving backward compatibility.

barendgehrels: Cool! Thank you!

include/boost/geometry/algorithms/for_each.hpp
```diff
@@ -94,31 +96,76 @@ struct fe_range_per_point
...
 
+template
+<
+    typename Range,
+    typename Point = typename add_const_if_c
+        <
+            boost::is_const<Range>::value,
+            typename point_type<Range>::type
+        >::type,
+    bool UseRefernces =
```
barendgehrels, line 107: Small typo, missing `e`

include/boost/geometry/algorithms/for_each.hpp
```diff
@@ -94,31 +96,76 @@ struct fe_range_per_point
...
+struct fe_range_per_segment_call_f<Range, Point, false>
+{
+    template <typename Iterator, typename Functor>
+    static inline void apply(Iterator it0, Iterator it1, Functor& f)
+    {
+        // Non-true mutable references returned by iterators
+        // temporary points has to be created and assigned afterwards.
+        Point p0 = *it0;
+        Point p1 = *it1;
+        model::referring_segment<Point> s(p0, p1);
```
barendgehrels, line 142: Clear, it's fine to me. I kind of see the problem. Is it ever used like this? But anyway, it supports what it did before. Thank you.

(You can change `has` to `have`)

awulkiew: I think using `for_each_segment` for mutable ranges doesn't make sense but since they were supported I decided to keep that behavior even for proxy references. We could talk about dropping it, in fact I have a next PR nearly ready related to for_each so we could talk about it there because this PR is fixing a bug.

include/boost/geometry/algorithms/for_each.hpp
awulkiew: Done.

PR #743, opened 2020-08-10
Refactor for_each_x and implement new algorithms.

awulkiew:
I was playing with for_each family of functions and C++14. With the newer standard we can greatly simplify some parts of the code applying some Actor/Policy to Points/Segments/Ranges or checking some predicate for them.
Let me know what you think.

Changes summary:

`for_each_range` now expects functors instead of structs with `apply()` member function.

`for_each_segment` now calls functor also for degenerated ranges for consistency with `segment_iterator`.

New algorithms based on `std::all_of`, `std::any_of` and `std::none_of`
- `all_points_of`, `any_point_of`, `none_point_of`
- `all_segments_of`, `any_segment_of`, `none_segment_of`
- `all_ranges_of`, `any_range_of`, `none_range_of`

New algorithms used in several places in the library.

### DIFF
--- a/.github/workflows/minimal-clang.yml
+++ b/.github/workflows/minimal-clang.yml
@@ -32,28 +32,28 @@ jobs:
 
         include:
           - b2_toolset: clang-3.9
-            b2_cxxstd: 03,11
+            b2_cxxstd: 14
             version: "3.9"
           - b2_toolset: clang-4.0
-            b2_cxxstd: 03,11
+            b2_cxxstd: 14
             version: "4.0"
           - b2_toolset: clang-5.0
-            b2_cxxstd: 03,11,14
+            b2_cxxstd: 14
             version: "5.0"
           - b2_toolset: clang-6.0
-            b2_cxxstd: 03,11,14
+            b2_cxxstd: 14
             version: "6.0"
           - b2_toolset: clang-7
-            b2_cxxstd: 03,11,14,17
+            b2_cxxstd: 14,17
             version: "7"
           - b2_toolset: clang-8
-            b2_cxxstd: 03,11,14,17
+            b2_cxxstd: 14,17
             version: "8"
           - b2_toolset: clang-9
-            b2_cxxstd: 03,11,14,17,2a
+            b2_cxxstd: 14,17,2a
             version: "9"
           - b2_toolset: clang-10
-            b2_cxxstd: 03,11,14,17,2a
+            b2_cxxstd: 14,17,2a
             version: "10"
 
     steps:
--- a/include/boost/geometry/algorithms/detail/disjoint/areal_areal.hpp
+++ b/include/boost/geometry/algorithms/detail/disjoint/areal_areal.hpp
@@ -5,8 +5,8 @@
 // Copyright (c) 2009-2014 Mateusz Loskot, London, UK.
 // Copyright (c) 2013-2014 Adam Wulkiewicz, Lodz, Poland.
 
-// This file was modified by Oracle on 2013-2019.
-// Modifications copyright (c) 2013-2019, Oracle and/or its affiliates.
+// This file was modified by Oracle on 2013-2020.
+// Modifications copyright (c) 2013-2020, Oracle and/or its affiliates.
 
 // Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
 // Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
@@ -30,7 +30,7 @@
 #include <boost/geometry/algorithms/detail/disjoint/linear_linear.hpp>
 #include <boost/geometry/algorithms/detail/disjoint/segment_box.hpp>
 
-#include <boost/geometry/iterators/segment_iterator.hpp>
+#include <boost/geometry/algorithms/for_each.hpp>
 
 
 namespace boost { namespace geometry
@@ -51,46 +51,23 @@ inline bool point_on_border_covered_by(Geometry1 const& geometry1,
         && geometry::covered_by(pt, geometry2, strategy);
 }
 
-/*!
-\tparam Strategy point_in_geometry strategy
-*/
-template<typename Geometry, typename Strategy>
-struct check_each_ring_for_within
-{
-    bool not_disjoint;
-    Geometry const& m_geometry;
-    Strategy const& m_strategy;
-
-    inline check_each_ring_for_within(Geometry const& g,
-                                      Strategy const& strategy)
-        : not_disjoint(false)
-        , m_geometry(g)
-        , m_strategy(strategy)
-    {}
-
-    template <typename Range>
-    inline void apply(Range const& range)
-    {
-        not_disjoint = not_disjoint
-                    || point_on_border_covered_by(range, m_geometry, m_strategy);
-    }
-};
-
 
 /*!
 \tparam Strategy point_in_geometry strategy
 */
-template <typename FirstGeometry, typename SecondGeometry, typename Strategy>
-inline bool rings_containing(FirstGeometry const& geometry1,
-                             SecondGeometry const& geometry2,
+template <typename Geometry1, typename Geometry2, typename Strategy>
+inline bool rings_containing(Geometry1 const& geometry1,
+                             Geometry2 const& geometry2,
                              Strategy const& strategy)
 {
-    check_each_ring_for_within
-        <
-            FirstGeometry, Strategy
-        > checker(geometry1, strategy);
-    geometry::detail::for_each_range(geometry2, checker);
-    return checker.not_disjoint;
+    // TODO: This will be removed when IntersectionStrategy is replaced with
+    //       UmbrellaStrategy
+    auto const pgs = strategy.template get_point_in_geometry_strategy<Geometry2, Geometry1>();
+
+    return geometry::detail::any_range_of(geometry2, [&](auto const& range)
+    {
+        return point_on_border_covered_by(range, geometry1, pgs);
+    });
 }
 
 
@@ -117,10 +94,8 @@ struct areal_areal
         // We check that using a point on the border (external boundary),
         // and see if that is contained in the other geometry. And vice versa.
 
-        if ( rings_containing(geometry1, geometry2,
-                              strategy.template get_point_in_geometry_strategy<Geometry2, Geometry1>())
-          || rings_containing(geometry2, geometry1,
-                              strategy.template get_point_in_geometry_strategy<Geometry1, Geometry2>()) )
+        if ( rings_containing(geometry1, geometry2, strategy)
+          || rings_containing(geometry2, geometry1, strategy) )
         {
             return false;
         }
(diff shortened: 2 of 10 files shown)


.github/workflows/minimal-clang.yml
```diff
@@ -32,28 +32,36 @@ jobs:
 
         include:
           - b2_toolset: clang-3.9
-            b2_cxxstd: 03,11
+#            b2_cxxstd: 03,11
+            b2_cxxstd: 14
```
mloskot, line 36: I'd simply remove the commented lines to keep the YAML-s compact, easier to maintain.

include/boost/geometry/algorithms/detail/disjoint/areal_areal.hpp
```diff
@@ -141,10 +116,11 @@ struct areal_box
                              Box const& box,
                              Strategy const& strategy)
     {
-        if ( ! for_each_segment(geometry::segments_begin(areal),
-                                geometry::segments_end(areal),
-                                box,
-                                strategy.get_disjoint_segment_box_strategy()) )
+        auto ds = strategy.get_disjoint_segment_box_strategy();
+        if (! geometry::all_segments_of(areal, [&](auto const& s)
```
mloskot, line 120: Nitpicks:
- could be `auto const ds` as it's passed to `apply` via reference to `const`, AFAIR.
- capture `[&]` could be `[&box, &ds]` for clarity and, theoretically as I have not checeked, optimisation

include/boost/geometry/algorithms/detail/for_each_range.hpp
```diff
@@ -181,14 +229,55 @@ struct for_each_range<Geometry, Actor, multi_polygon_tag>
...
+template <typename Geometry, typename UnaryPredicate>
+inline bool any_range_of(Geometry const& geometry, UnaryPredicate p)
 {
-    dispatch::for_each_range
-        <
-            Geometry const,
-            Actor
-        >::apply(geometry, actor);
+    return ! dispatch::for_each_range<Geometry const>::apply(geometry,
+                [&](auto&& range)
```
mloskot, line 248: The `[&]` captures only `p`, doesn't it? If so, could be limited too.

include/boost/geometry/algorithms/detail/touches/implementation.hpp
```diff
@@ -199,51 +199,30 @@ struct areal_interrupt_policy
...
-        geometry::point_on_border(p, range);
-        if ( !has_within && geometry::within(p, m_geometry, m_strategy) )
-        {
-            has_within = true;
-        }
-    }
-};
+    typename geometry::point_type<Geometry1>::type pt;
+    return geometry::point_on_border(pt, geometry1)
+        && geometry::within(pt, geometry2, strategy);
```
mloskot, line 209: A theoretical question, does anyone know if optimisers can handle this old-fashioned piece better?

```cpp
if  (geometry::point_on_border(pt, geometry1)) // pt is output
     return geometry::within(pt, geometry2, strategy); // pt is input
else
    return false;
```

I remember old compilers/optimisers tripping on some uses of ternary operator :)

include/boost/geometry/algorithms/for_each.hpp
```diff
@@ -54,209 +56,310 @@ namespace detail { namespace for_each
...
-        model::referring_segment<Point> s(p0, p1);
-        f(s);
+        typedef typename fe_point_type<Range>::type point_type;
+        point_type p0 = *it0;
+        point_type p1 = *it1;
+        model::referring_segment<point_type> s(p0, p1);
+        bool result = f(s);
         *it0 = p0;
         *it1 = p1;
+        return result;
```
mloskot, line 236: There is a reason why `f(s)` is called (and returned boolean cached) before `p0` and `p1` are accessed, right?

Shouldn't the assignments be skipped all together if the `f` fails?

```cpp
if (!f(s))
    return false;

*it0 = p0;
*it1 = p1;
return true;
```

include/boost/geometry/algorithms/detail/disjoint/areal_areal.hpp
awulkiew: I'll add `const`.

I used `&` to have the line short as possible. Do you think lambdas automatically store references to all local variables?

mloskot: The lazy `[&]` captures everything by reference and the compiler should be clever enough to optimise away anything that is not explicitly referenced inside the lambda.

I aimed to provoke some thoughts on which one is clearer for a reader, but yeah, shorter seems better.

include/boost/geometry/algorithms/for_each.hpp
awulkiew: I agree with you in general however my intention was to generate exactly the same outcome as in case of the version for lvalue mutable iterators. The caller don't know whether or not the points are going to be modified inside `f`. The points can be modified and the functor can still return `false`. The final assignment would reflect that, i.e. this code would work exactly how it works for real references.

An example would be an algorithm transforming N points of a geometry or something and stopping as soon as they are transformed. I don't know if this algorithm makes sense but I'm just showing a scenario when the outcomes would be different.

Or am I missing something?

mloskot: > The points can be modified and the functor can still return false.

It is an important clarification.
It makes the difference to the approach which I assumed, i.e. in-out arguments remain untouched on failure.

My questions have been answered, thanks.

include/boost/geometry/algorithms/detail/touches/implementation.hpp
awulkiew: I don't know, so the version with ternary operator would be yet another one:
```
return geometry::point_on_border(pt, geometry1)
     ? geometry::within(pt, geometry2, strategy)
     : false;
```

I always assumed the compiler can do whatever it wants in any case and if we wanted to be sure we would have to check each version of each one of them (a good example is `switch` and which `case` is checked first, etc.). So I tend to use logical `&&` to check conditions one by one because it is the shortest.

mloskot: > I tend to use logical && to check conditions one by one because it is the shortest.

Yes, makes sense, shorter seems better again.
I'm too lazy to check the Compiler Explorer output right now.

>  version with ternary operator would be yet another one

Generally, I try to avoid the ternary operator. I'd rather use if-else instead.

include/boost/geometry/algorithms/detail/disjoint/areal_areal.hpp
awulkiew: I don't have an opinion about it actually. If the compiler captures everything then maybe it'd be a good idea to explicitly limit it and not use the default captures at all. This would be a premature optimization though. This needs more research.

mloskot: I'm sure things are optimised away in this case.
My suggestion was of the same category as declare/include what you use, declare variables in as narrow scope as possible.
Anyway, thanks for comments, it's not an issue that needs any fixing for me.

.github/workflows/minimal-clang.yml
awulkiew: Done.

awulkiew:
Thanks for the review @mloskot .

I was searching for a candidate for refactoring using these functions. `length()` seems to be the most straightforward one. For all linear geometries it could be implemented as something like this:
```
for_each_segment(linear, [&](auto const& s){
    strategy.apply(make_indexed_view<0>(s), make_indexed_view<1>(s));
});
```

We could also have a version of `for_each_segment` (with a different name) taking a functor taking 2 points instead of a segment which would make this even simpler:
```
for_each_something(linear, [&](auto const& p0, auto const& p1){
    strategy.apply(p0, p1);
});
```

`transform()` could nearly be implemented with `all_points_of()` in its entirety but the problem is that it resizes the geometries in the process so we'd need a different algorithm for that. An algorithm which maybe could also be used for `simplify()`.

include/boost/geometry/algorithms/detail/for_each_range.hpp
```diff
@@ -38,47 +48,78 @@ namespace detail { namespace for_each
...
 struct fe_range_multi
 {
-    static inline void apply(Multi & multi, Actor & actor)
+    template <typename Functor>
+    static inline bool apply(Multi& multi, Functor&& f)
     {
-        for ( typename boost::range_iterator<Multi>::type
-                it = boost::begin(multi); it != boost::end(multi); ++it)
+        auto const end = boost::end(multi);
+        for (auto it = boost::begin(multi); it != end; ++it)
```
barendgehrels, line 115: Since you're reworking this, it's nicer to use a range based loop here
`for (auto& geometry : multi)` {... `apply(geometry, f)` }

include/boost/geometry/strategies/agnostic/hull_graham_andrew.hpp
```diff
@@ -116,48 +100,35 @@ struct get_extremes
...
-    inline void apply(InputRange const& range)
+    geometry::detail::for_each_range(geometry, [&](auto const& range)
     {
-        typedef SideStrategy side;
-
         // Put points in one of the two output sequences
-        for (RangeIterator it = boost::begin(range);
-            it != boost::end(range);
-            ++it)
+        for (auto it = boost::begin(range); it != boost::end(range); ++it)
```
barendgehrels, line 127: Also here a range based loop instead?

barendgehrels (approved):
Thanks, looks good.

include/boost/geometry/algorithms/detail/for_each_range.hpp
awulkiew: This is probably not possible. It's because we require `MultiGeometry` to be adapted to Boost.Range concept. So it's possible to adapt it and at the same time avoid implementing `begin` and `end` members or free functions that could be found via ADL. Consider e.g. `std::pair<T*, T*>` which is adapted to Boost.Range concept but can't be used in range-based for loop.

Unless I'm missing something and e.g. Boost.Range defines some free functions enabled conditionally for adapted types. I quickly tried to find something like this but was unable to do it. Plus I'm not sure if they would implement something like this because it's not clear to me that this would not cause any problems.

include/boost/geometry/strategies/agnostic/hull_graham_andrew.hpp
awulkiew: Same here, we require `Range` adapted to Boost.Range concept which is not required to work with range-based for loops.

include/boost/geometry/algorithms/detail/for_each_range.hpp
awulkiew: Actually we could replace it with something like:
```
for (auto& geometry : make_range(multi))
```
where
```
template <typename Range>
inline auto make_range(Range & rng)
{
    return boost::make_iterator_range(boost::begin(rng), boost::end(rng));
}
```

PR #744, opened 2020-08-12
[get_turn_info] Remove the recently added is_ii

barendgehrels:
I removed the is_ii check, which caused all the mentioned problems, and the `side_of_range` too (I made a local copy because I want to reuse it later).
To my surprise the case which it was supposed to fix, was fixed anyway, probably by later changes in the same branch. There is one regression in `difference` (without rescaling) but that can/should be solved differently.

These are technical details - I'm still thinking that that situation should be "ii" (`intersection/intersection`) i/o "u/i" (`union/intersection`), but it always worked that way, so I should not change that now.

I tested all the tests and they are all fine (for testing everything, I had to change my environment, because of disk usage problems, but that is fine now).


barendgehrels:
> 
> 
> Thanks!
> So the rest of the change was just refactoring, correct?

Not really ("handle as touch"), but it is only called if rescaling is turned off. So low risk on breaking things currently.

Thanks for your quick review.

PR #747, opened 2020-08-26
Add optional traits::make and use it in cross_product() and make().

awulkiew:
This change addresses the backward compatibility issue raised here: https://github.com/boostorg/geometry/pull/745

With the new optional trait it is possible to define how a geometry like Point, Box or Segment is constructed.

`bg::make()` for 2d and 3d points is `constexpr` now if the trait is specialized.


include/boost/geometry/geometries/point_xyz.hpp
```diff
@@ -124,6 +124,21 @@ struct access<model::d3::point_xyz<CoordinateType, CoordinateSystem>, Dimension>
...
 };
 
+template<typename CoordinateType, typename CoordinateSystem>
+struct make<model::d3::point_xyz<CoordinateType, CoordinateSystem> >
+{
+    typedef model::d3::point_xyz<CoordinateType, CoordinateSystem> point_type;
+
+    static const bool is_specialized = true;
+
+    static constexpr point_type apply(CoordinateType const& v0,
```
vissarion, line 134: do you need 3 coordinates here?

include/boost/geometry/geometries/point_xyz.hpp
```diff
@@ -124,6 +124,21 @@ struct access<model::d3::point_xyz<CoordinateType, CoordinateSystem>, Dimension>
...
+        CoordinateType const& v1)
```
vissarion, line 135: indentation issue

vissarion (approved):
Looks OK to me. Thanks!
As an addition, it could be useful to add some basic unit tests.

include/boost/geometry/geometries/point_xyz.hpp
awulkiew: Yes, I'll fix it.

awulkiew:
> it could be useful to add some basic unit tests.

Right.

awulkiew:
What do you think about moving `point_xy` and `point_xyz` from `model::d2` and `model::d3` namespace into `model` namespace? Of course leaving the alias for backward compatibility for now. These additional namespaces seems to be not needed. Or is there a reason why they are there?

vissarion:
> What do you think about moving `point_xy` and `point_xyz` from `model::d2` and `model::d3` namespace into `model` namespace? Of course leaving the alias for backward compatibility for now. These additional namespaces seems to be not needed. 

I am ok with that.

>Or is there a reason why they are there?

@barendgehrels should know

include/boost/geometry/geometries/point_xyz.hpp
awulkiew: done

awulkiew:
@barendgehrels are you ok with this change?

barendgehrels:
> 
> 
> > What do you think about moving `point_xy` and `point_xyz` from `model::d2` and `model::d3` namespace into `model` namespace? Of course leaving the alias for backward compatibility for now. These additional namespaces seems to be not needed.
> 
> I am ok with that.
> 
> > Or is there a reason why they are there?
> 
> @barendgehrels should know

I'm OK with the move. But it is not backwards compatible, users might use it. We should add it into the release notes then. And not for 1.75 then...

barendgehrels (approved):
I'm OK, sorry for the delay.

awulkiew:
@barendgehrels Thanks!

> I'm OK with the move. But it is not backwards compatible, users might use it. We should add it into the release notes then. And not for 1.75 then...

Yes, it's too late for that. I think we should add aliases in `d2` and `d3` namespaces and this way this change would be backward compatible.

PR #748, opened 2020-08-29
Remove direct usage of Boost.MPL and Boost.TypeTraits from the library.

awulkiew:
Changes:
- new files:
   - `util/type_traits_std.hpp` - type traits not available in C++14 dependent only on standard headers, e.g. C++17 `bool_constant`, C++20 `remove_cvref`, non-standard `transcribe_const`, etc.
   - `util/type_traits.hpp` - geometry-related type traits dependent on tags, e.g. `is_linear`, `enable_if_point`, etc.
   - `util/sequence.hpp` - tools for manipulating `std::integer_sequence` and non-standard `type_sequence` (replacing `mpl::vector_c` and `mpl::vector` respectively), the interface is mimicking the one of `std::tuple` metafunctions, e.g. `sequence_element` replaces `mpl::at_c`
- `add_const_if_c` moved to util/type_traits_std.hpp, all usage removed or replaced with newly added `transcribe_const`
- `util::bare_type` moved to util/type_traits_std.hpp, all usage replaced with newly added `remove_cptrref`
- `is_range` moved to `util/range.hpp`
- added new tags:
   - `polylinear_tag` - range-based linear geometry (Ls, MLs)
   - `polysegmental_tag` - range-based linear or areal geometry (Ls, MLs, Ring, Poly, MPoly)
- added variadic templates in utilities like `select_most_precise`, `select_calculation_type_alt`
- `select_calculation_type` used instead of `select_most_precise` in `buffer::geographic_point_circle` to correctly handle `CalculationType`
- Boost.TypeTraits and Boost.MPL replaced in the majority of the library, what's left is:
   - `MPL_ASSERT` - I'd like to propose a separate PR for that
   - `Boost.Fusion` adaptation
   - in deprecated headers not included automatically like `util/compress_variant.hpp`
   - in `util/parameter_type_of.hpp` because Boost.FunctionTypes is used there for now
- `geometry::dimension<>` is derived from `std::integral_constant<std::size_t, trait::dimension<Geometry>::value>`, rationale:
    - to guarantee the type (standard uses std::size_t for such things, e.g. `std::array`'s size, `std::tuple_size<>::value`)
    - we already use `std::size_t` across the library for Dimensions
    - to allow users to define `trait::dimension` however they like as long as there is `value` member
    - this is a potentially breaking change
- `distance_result` for variants is implemented differently, it no longer can be a variant type, it is always the most precise numeric type for any combination of variant geometries and strategy (so result of a strategy), rationale:
   - simplicity
   - in order to get variant result the input types had to be specific which makes it unlikely, i.e. variant types had to hold geometries using different coordinate types and at least one of these types had to be more precise than double (so e.g. user-defined or `long double` in a compiler actually providing this as separate type to `double`), because in distance strategies all fundamental result types are promoted to `double`
   - still this is a potentially breaking change


include/boost/geometry/util/type_traits.hpp
```diff
@@ -0,0 +1,323 @@
...
+
+template <typename T>
+struct is_multi_linestring
+    : std::is_same<multi_linestring_tag, typename tag<T>::type>
+{};
+
+template <typename T>
+struct is_polylinear
+    : std::is_base_of<polylinear_tag, typename tag<T>::type>
+{};
```
awulkiew, line 132: I decided to use *polylinear* name for linear geometries which are not single segment. Similar to *polygonal* for areal geometries that are not a box. If you have a different idea feel free to propose a change.

include/boost/geometry/util/type_traits.hpp
```diff
@@ -0,0 +1,323 @@
...
+template <typename T>
+struct is_areal
+    : std::is_base_of<areal_tag, typename tag<T>::type>
+{};
+
+
+template <typename T>
+struct is_polysegmental
+    : bool_constant<is_polylinear<T>::value || is_polygonal<T>::value>
+{};
```
awulkiew, line 174: Here is another group of geometries, i.e. geometries based on ranges of segments which are not a single segment, both linear and areal. I named it `polysegmental` feel free to suggest a different name.

include/boost/geometry/util/type_traits.hpp
```diff
@@ -0,0 +1,323 @@
...
+template <typename T>
+struct is_multi
+    : std::is_base_of<multi_tag, typename tag<T>::type>
+{};
+
+
+template <typename T>
+struct is_multi_element
+    : bool_constant<is_point<T>::value || is_linestring<T>::value || is_polygon<T>::value>
+{};
```
awulkiew, line 186: Here is another group, single geometries which can be elements of a multi geometry. I didn't have an idea about a name so the one I used is more straightforward than the other ones I mentioned. I'm open to suggestion.

mloskot (approved):
I have not reviewed the code changes, but I like the general direction (removal of legacy libraries).
So, thank you @awulkiew and I give 👍  assuming tests and examples compile and pass, examples compile.

awulkiew:
I decided to move type_traits and sequence utilities from `detail` namespace to `util` namespace because:
- it's shorter
- it's easier to use in deeper `detail` namespaces like `strategies::detail`, e.g. `util::enable_if_linear` vs `geometry::detail::enable_if_linear`
Please let me know if you're ok with it.

awulkiew:
@mloskot I'll put it this way: the number of passing tests and compiling examples remains the same. ;)

awulkiew:
I'll wait with merging this for @barendgehrels merging https://github.com/boostorg/geometry/pull/752.

mloskot:
> I decided to move type_traits and sequence utilities from `detail` namespace to `util` namespace because
> it's shorter
> it's easier to use

I'd just point out that semantic of the `detail` namespace is "not a public interface, can change/deprecate/remove w/o notice".
The proposed simplification breaks that notion and it may hit back with users complaining about when things have been removed from the `util` what I see likely.

awulkiew:
@mloskot yes, I'm aware of that. I find it unlikely that these type traits will be modified in the future. The only candidate I can think of are metafunctions for sequences (`merge`, `combine`, etc.) which could have names changed to names consistent with other utils like `sequence_element`, `sequence_size`, etc. (named after metafunctions for `std::tuple`). So they could be named `sequence_merge`, `sequence_combine` or `merge_sequences`, `combine_sequences`, etc. Other than that I think the interface is ok.
With that said I wouldn't hesitate much if we wanted to do any change in the future because these tools are undocumented and AFAIU anything that is undocumented may be changed as well. I think it was already done in the past with algorithms in `boost::geometry` namespace. An example can be `sectionalize` which is in `algorithms/detail` directory but the function is in the `boost::geometry` namespace and at some point an additional `DimensionVector` template argument was added to it. I'm not endorsing this practice in general. I'm only pointing out that the code that is not documented is probably rarely used.

barendgehrels (approved):
I glanced through the changes, and it looks good to me, thanks!
But it's too much to review it in detail.

barendgehrels:
> I decided to move type_traits and sequence utilities from `detail` namespace to `util` namespace because:
> 
>     * it's shorter
> 
>     * it's easier to use in deeper `detail` namespaces like `strategies::detail`, e.g. `util::enable_if_linear` vs `geometry::detail::enable_if_linear`
>       Please let me know if you're ok with it.

Fine for me.

include/boost/geometry/util/type_traits.hpp
barendgehrels: Fine for me, it's clear.

include/boost/geometry/util/type_traits.hpp
barendgehrels: What is the difference with `linestring` ? Those are also linear geometries with not a single segment. But I assume you mean something else?

awulkiew: Yes, polylinear means linestring or multi_linestring. I added the tag in case someone wanted to use `tag_cast` instead and because we already have `polygonal_tag`.

include/boost/geometry/util/type_traits.hpp
vissarion: The names are fine by me too. I do not know if they make a lot of geometric sense but they are consistent and clear. So I have no objection.

PR #749, opened 2020-08-31
Area strategy for more accurate computations in cartesian CS

vissarion:
This PR uses precise mathematical formulas to compute more accurate areas in cartesian CS. The inaccuracy arise from errors in floating point summation. See also https://svn.boost.org/trac10/ticket/11928.

A test case is added for illustration. This approach can be used in other places in the library where inaccurate summation can occur. A similar approach is used in `GeographicLib` cf. https://geographiclib.sourceforge.io/html/Accumulator_8hpp_source.html


include/boost/geometry/extensions/triangulation/strategies/cartesian/in_circle_robust.hpp
```diff
@@ -12,7 +12,7 @@
 #ifndef BOOST_GEOMETRY_EXTENSIONS_TRIANGULATION_STRATEGIES_CARTESIAN_IN_CIRCLE_ROBUST_HPP
 #define BOOST_GEOMETRY_EXTENSIONS_TRIANGULATION_STRATEGIES_CARTESIAN_IN_CIRCLE_ROBUST_HPP
 
-#include<boost/geometry/extensions/triangulation/strategies/cartesian/detail/precise_math.hpp>
+#include<boost/geometry/util/precise_math.hpp>
```
awulkiew, line 15: Or would `formulas` be a better place?

awulkiew:
Looks good, thanks!

What is the difference in performance of both strategies?
I'm asking because if it's close to the original strategy we could consider keeping only one.

awulkiew (approved):

include/boost/geometry/extensions/triangulation/strategies/cartesian/in_circle_robust.hpp
awulkiew: Together with `util/series_expansion.hpp` (probably in a different PR).

include/boost/geometry/strategy/cartesian/precise_area.hpp
```diff
@@ -0,0 +1,117 @@
...
+namespace boost { namespace geometry
+{
+
+namespace strategy { namespace area
+{
+
+/*!
+\brief Cartesian area calculation
+\ingroup strategies
+\details Calculates cartesian area using the trapezoidal rule and precise summation
```
tinko92, line 32: A minor suggestion: The "precise summation" here kind of implies it already because summation precision is no issue with integral types, but maybe it would be helpful for some users who are less familiar with this to explicitly state that this is only intended to be used for floating-point CalculationType. Because, if I understand area result_type correctly, it will only promote to double if the user did not specify something.

Or maybe there could be a static_assert of std::is_floating_point<return_type> in the strategy (it could also be argued to have it in the precise_math header, but I think from the users perspective, getting that error for the strategy might be more helpful).

awulkiew: Yes, if the user passes CalculationType we believe in his intentions and use this type, no matter what it is. So I have no opinion regarding this suggestion.

Have in mind that a user can pass his own, precise numeric type. If this method is good for such types then the test should rather be `(! std::is_integral<return_type>::value)`.

vissarion: I clarify it more in the `\details` for the documentation. Since this is supposed to be for advanced users I do not think we need an assert or return an error. Are you OK with this @tinko92 ?

include/boost/geometry/extensions/triangulation/strategies/cartesian/in_circle_robust.hpp
vissarion: I put it `util` since it contains basic operations (summation, multiplication etc) just like `math.hpp`, formulas are more "geometric" but I am open to move it in a new PR together with other functions if we are convinced by the concept of formulas vs. util.

vissarion:
> Looks good, thanks!
> 
> What is the difference in performance of both strategies?
> I'm asking because if it's close to the original strategy we could consider keeping only one.

The original strategy is performing 3 operations to compute the determinant and 1 sum for each segment, while the new adds 6 more operations (instead of one to compute a more accurate sum). A similar implementation in geographiclib https://geographiclib.sourceforge.io/html/classGeographicLib_1_1Accumulator.html is reported to be more than 10 times slower. To conclude I think both methods are useful. We have to understand how often this accurate strategy will be needed and then maybe make it default.

include/boost/geometry/strategy/cartesian/precise_area.hpp
tinko92: Yes, I think this is OK.

vissarion:
@barendgehrels do you agree to merge this to develop to go to 1.75 ?

include/boost/geometry/strategy/cartesian/precise_area.hpp
```diff
@@ -0,0 +1,118 @@
...
+    static inline void apply(PointOfSegment const& p1,
+                             PointOfSegment const& p2,
+                             state<Geometry>& st)
+    {
+        typedef typename state<Geometry>::return_type return_type;
+
+        auto det = (return_type(get<0>(p1)) + return_type(get<0>(p2)))
+                 * (return_type(get<1>(p1)) - return_type(get<1>(p2)));
+
+        auto res = boost::geometry::detail::precise_math::two_sum(st.sum1, det);
```
barendgehrels, line 95: Minor, these two parts could be const.

test/algorithms/area/area.cpp
```diff
@@ -175,6 +177,34 @@ void test_large_integers()
...
+                           0.20000000000000001 0.20000000000000004,\
+                           0.79999999999999993 0.80000000000000004,\
+                           1.267650600228229e30 1.2676506002282291e30,\
+                           0.10000000000000001 0.10000000000000001))", poly);
+
+    double inaccurate = bg::area(poly);
+    double accurate = bg::area(poly, precise_cartesian());
+
+    BOOST_CHECK_CLOSE(inaccurate, 0, 0.0001);
+    BOOST_CHECK_CLOSE(accurate, -0.315, 0.0001);
```
barendgehrels, line 205: Quite a difference. Is the polygon indeed anti clockwise?

test/algorithms/area/area.cpp
```diff
@@ -175,6 +177,34 @@ void test_large_integers()
...
 
+struct precise_cartesian : bg::strategies::detail::cartesian_base
+{
+    template <typename Geometry>
+    static auto area(Geometry const&)
+    {
+        return bg::strategy::area::precise_cartesian<>();
+    }
+};
+
```
barendgehrels, line 188: Maybe you can add one `simplex` case, just the area of a triangle or square, such that we can also judge the basic functionality.
But besides that I'm OK, thanks.

barendgehrels (approved):

test/algorithms/area/area.cpp
vissarion: It depends on how you perform the computation. With double arithmetic the area algorithm computes 4 determinants with the following values
```
-0.030000000000000012768
-0.5999999999999999778
-1606938044258989383512154151116236941093472087848332596084736
1606938044258989383512154151116236941093472087848332596084736
```
which then has to sum up and divide by 2. The implementation before this PR computes `0` while after this PR  `-0.315` which is clearly more accurate. 

Now computing the determinants accurately is a different story, not related to this PR. Note that doing this computation with boost multi-precision the sign of the area is positive so the polygon is CW oriented. I added another example where both precise area an multi-precision returns positive sign while the old area strategy returns `0` to emphasize the usefulness of performing a sum more accurately.

test/algorithms/area/area.cpp
vissarion: good idea, done

test/algorithms/area/area.cpp
```diff
@@ -175,6 +178,70 @@ void test_large_integers()
...
+            bg::cs::cartesian
+        > mp_point_type;
+
+    auto const poly0_string = "POLYGON((0 0,0 1,1 0,0 0))";
+
+    bg::model::polygon<point_type> poly0;
+    bg::read_wkt(poly0_string, poly0);
+
+    BOOST_CHECK_CLOSE(bg::area(poly0), 0.5, 0.0001);
+    BOOST_CHECK_CLOSE(bg::area(poly0, precise_cartesian()), 0.5, 0.0001);
```
barendgehrels, line 206: Perfect, thank you

test/algorithms/area/area.cpp
barendgehrels: I see. Good, thanks.

PR #752, opened 2020-09-09
[buffer] Several changes necessary to handle different types

barendgehrels:
This should (at least partly) fix the recent observation of @awulkiew , and contains parts of my work I did before to support types other than double (without rescaling)

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
@@ -149,44 +149,6 @@ class piece_turn_visitor
         return ! m_rings[piece1.first_seg_id.multi_index].has_concave;
     }
 
-
-    template <std::size_t Dimension, typename Iterator, typename Box>
-    inline void move_begin_iterator(Iterator& it_begin, Iterator it_beyond,
-                                    signed_size_type& index, int dir,
-                                    Box const& this_bounding_box,
-                                    Box const& other_bounding_box)
-    {
-        for(; it_begin != it_beyond
-                && it_begin + 1 != it_beyond
-                && detail::section::preceding<Dimension>(dir, *(it_begin + 1),
-                                                         this_bounding_box,
-                                                         other_bounding_box,
-                                                         m_robust_policy);
-            ++it_begin, index++)
-        {}
-    }
-
-    template <std::size_t Dimension, typename Iterator, typename Box>
-    inline void move_end_iterator(Iterator it_begin, Iterator& it_beyond,
-                                  int dir, Box const& this_bounding_box,
-                                  Box const& other_bounding_box)
-    {
-        while (it_beyond != it_begin
-            && it_beyond - 1 != it_begin
-            && it_beyond - 2 != it_begin)
-        {
-            if (detail::section::exceeding<Dimension>(dir, *(it_beyond - 2),
-                        this_bounding_box, other_bounding_box, m_robust_policy))
-            {
-                --it_beyond;
-            }
-            else
-            {
-                return;
-            }
-        }
-    }
-
     template <typename Piece, typename Section>
     inline void calculate_turns(Piece const& piece1, Piece const& piece2,
         Section const& section1, Section const& section2)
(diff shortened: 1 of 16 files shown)


include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
```diff
@@ -149,44 +149,6 @@ class piece_turn_visitor
...
-            {
-                --it_beyond;
-            }
-            else
-            {
-                return;
-            }
-        }
-    }
-
```
barendgehrels, line 189: @awulkiew : I don't remember why this was there, indeed. It is around 7 years old. But I disabled it, and it (now) runs all tests without problems, so I removed it. I have to come back to it, because a similar thing is also in the non-buffered turns.

This most probably fixes the type conflict for you. I can compile with Boost MultiPrecision, and it runs but still gives many errors. So I will continue later with this. The Union (for example) runs fine with Boost MP, so indeed it has to be something with buffer.

include/boost/geometry/algorithms/detail/buffer/line_line_intersection.hpp
```diff
@@ -39,12 +40,7 @@ struct line_line_intersection
 
         // The input lines are not parallel, they intersect, because
         // their join type is checked before.
-        Point ip;
-        bool const intersecting = arithmetic::intersection_point(p, q, ip);
-        BOOST_GEOMETRY_ASSERT(intersecting);
-        boost::ignore_unused(intersecting);
-
-        return ip;
+        return arithmetic::intersection_point(p, q, ip);
```
barendgehrels, line 43: For some types, with extreme coordinates, they don't intersect.

include/boost/geometry/algorithms/detail/sections/section_functions.hpp
```diff
@@ -131,7 +132,14 @@ static inline bool preceding(int dir,
                              RobustBox const& other_robust_box,
                              RobustPolicy const& robust_policy)
 {
-    typename geometry::robust_point_type<Point, RobustPolicy>::type robust_point;
+    typedef typename geometry::robust_point_type<Point, RobustPolicy>::type robust_point_type;
+    BOOST_STATIC_ASSERT((boost::is_same
+                         <
+                             typename geometry::coordinate_type<robust_point_type>::type,
+                             typename geometry::coordinate_type<RobustBox>::type
+                         >::value));
```
barendgehrels, line 140: This should make applying wrong types impossible

include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
vissarion: I think in general it could be very useful if we add minimal tests for all algorithms with some other type than `double` or `float`. Boost MP is a good choice I think.

barendgehrels: Yes, I think we can clean up old code (ttmath, GMP and some other types which were all not part of boost, but we used to verify that behaviour) and add MP there instead.
For union it works quite well (I think perfect, but some expectations need a small update). For buffer there is still a problem.

awulkiew: It would make the testing more complete for sure. Though I had a different scenario in mind when I wrote the comment for the commit. In my branch I use MP types instead of robust/rescaled/integral coordinates to avoid overflow. So the original coordinate type is e.g. `double`. Still in both cases less-comparison would fail to compile. The original MP coordinate type would also cause error in rescaling function because AFAIR there is no implicit conversion from MP to fundamental types.

include/boost/geometry/algorithms/detail/sections/section_functions.hpp
awulkiew: We can use `std::is_same` now. ;) But none of the included files probably include `<type_traits>`, so let's leave it as it is until my PR is merged and I'll change this afterwards.

include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
awulkiew: @barendgehrels It looks like some kind of optimization searching for subranges in both rings corresponding to bounding boxes of sections in order to ignore the rest of the rings. So my guess is that you could notice a difference in performance in some cases. Still this was probably a premature optimization so I'm ok with removing it.

awulkiew (approved):
Thanks for the fix!

vissarion (approved):

include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
barendgehrels: I'll test more thoroughly coming week.

PR #754, opened 2020-09-23
Fix/buffer mp

barendgehrels:
This reverts the removal of monotonic sections in previous PR, because it is indeed for optimization as Adam mentioned. Difference in performance is around 20%

Instead it fixes the type used for partitioning, and extends the type checking to various places where partitioning is used in buffer (and in assign_parents).

For the type checking I added a file (and it uses `static_assert` and `std::is_same` now :-) )

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
@@ -149,6 +149,43 @@ class piece_turn_visitor
         return ! m_rings[piece1.first_seg_id.multi_index].has_concave;
     }
 
+    template <std::size_t Dimension, typename Iterator, typename Box>
+    inline void move_begin_iterator(Iterator& it_begin, Iterator it_beyond,
+                                    signed_size_type& index, int dir,
+                                    Box const& this_bounding_box,
+                                    Box const& other_bounding_box)
+    {
+        for(; it_begin != it_beyond
+                && it_begin + 1 != it_beyond
+                && detail::section::preceding<Dimension>(dir, *(it_begin + 1),
+                                                         this_bounding_box,
+                                                         other_bounding_box,
+                                                         m_robust_policy);
+            ++it_begin, index++)
+        {}
+    }
+
+    template <std::size_t Dimension, typename Iterator, typename Box>
+    inline void move_end_iterator(Iterator it_begin, Iterator& it_beyond,
+                                  int dir, Box const& this_bounding_box,
+                                  Box const& other_bounding_box)
+    {
+        while (it_beyond != it_begin
+            && it_beyond - 1 != it_begin
+            && it_beyond - 2 != it_begin)
+        {
+            if (detail::section::exceeding<Dimension>(dir, *(it_beyond - 2),
+                        this_bounding_box, other_bounding_box, m_robust_policy))
+            {
+                --it_beyond;
+            }
+            else
+            {
+                return;
+            }
+        }
+    }
+
     template <typename Piece, typename Section>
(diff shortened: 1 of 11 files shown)


vissarion (approved):
That looks ok to me.

include/boost/geometry/core/check_for_same_type.hpp
```diff
@@ -0,0 +1,44 @@
...
+    template <typename Geometry1, typename Geometry2>
+    static inline void apply(Geometry1 const& , Geometry2 const& )
+    {
+      static_assert(std::is_same
+          <
+              typename geometry::coordinate_type<Geometry1>::type,
+              typename geometry::coordinate_type<Geometry2>::type
+          >::value, "Coordinate types in geometries should be the same");
+    }
+};
```
awulkiew, line 37: 3 problems with that:
- this tool static asserts for the same `coordinate_type` not the same type
- we already have similar tools for checking assertions about dimensions placed in a file where `geometry::dimension` is defined
- these tools are free functions

So I think this should be moved to `core/coordinate_type.hpp`, be called `assert_same_coordinate_type` or `assert_coordinate_type_equal` etc. and be implemented as a free function just like dimension utils.

barendgehrels: Sounds good, thank you. I'll update the PR.

barendgehrels: It's updated

include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
```diff
@@ -181,10 +218,31 @@ class piece_turn_visitor
...
+
+        signed_size_type index2 = sec2_first_index;
+        move_begin_iterator<0>(it2_first, it2_beyond, index2,
+                    section2.directions[0], section2.bounding_box, section1.bounding_box);
+        move_end_iterator<0>(it2_first, it2_beyond,
+                    section2.directions[0], section2.bounding_box, section1.bounding_box);
+        move_begin_iterator<1>(it2_first, it2_beyond, index2,
+                    section2.directions[1], section2.bounding_box, section1.bounding_box);
+        move_end_iterator<1>(it2_first, it2_beyond,
+                    section2.directions[1], section2.bounding_box, section1.bounding_box);
```
awulkiew, line 240: So this is needed after all? What is the reason of reverting the last change, performance?

include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
```diff
@@ -35,51 +35,33 @@ struct distance_measure
...
-    bool is_zero() const { return measure == 0.0; }
+    bool is_zero() const
+    {
+      return ! is_positive() && ! is_negative();
+    }
 
     // Returns true if the distance measure is positive. Distance measure
     // algorithm returns positive value if it is located on the left side.
-    bool is_positive() const { return measure > 0.0; }
+    bool is_positive() const { return measure > zero(); }
```
awulkiew, line 58: Why not simply `T(0)` which would be shorter, more clear IMO and not require an additional function?

awulkiew (approved):

include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
barendgehrels: Yes, see description. It was ~20% slower in the bulk performance test.

include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp
barendgehrels: It's updated.

PR #756, opened 2020-09-25
Non cartesian and robust convex hulls

vissarion:
This PR adds support for non-cartecian coordinate systems to convex hull algorithm, namely spherical and geographic. For the cartesian case it adds and tests a robust strategy implemented in https://github.com/boostorg/geometry/pull/617

### Details

* the graham-andrew algorithm implementation (a.k.a. Andrew's monotone chain convex hull algorithm) is moved to `algorithms` directory from agnostic strategies,
* new umbrella strategies are added, that basically call the side strategy for each coordinate system,
* robust strategies for cartesian CS are supported. 

### Robustness

The file `convex_hull_robust.cpp` in `test/convex_hull` directory demonstrates what can go wrong with non robust strategies in convex hull computation (this could be extended to other geometric algorithms as well). In the examples, the resulting convex hull could (a) miss a point, (b) be slightly non-convex, (c) could result in a polygon with wrong area sign or (d) has an arbitrarily large error in the resulting area. More information about robustness here: https://people.mpi-inf.mpg.de/~mehlhorn/ftp/classroomExamplesNonrobustness.pdf 
Also as noted in https://github.com/boostorg/geometry/issues/699 the robust side predicate seems to be faster than the non-robust one (`side_by_triangle`) and could be useful to have a fast but non-robust side predicate.


include/boost/geometry/strategies/convex_hull/cartesian.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace strategies { namespace convex_hull
+{
+
+template <typename CalculationType = void>
+class cartesian : strategies::detail::cartesian_base
+{
+public:
+    template <typename Geometry>
+    static auto side(Geometry const&)
```
awulkiew, line 29: Side strategy may be called for points of the same geometry (e.g. in `is_simple`) or for points of 2 geometries (e.g. in `intersection`) and also in addition to that for some temporary point not existing in any of the geometries. In a place where this is done (e.g. in `get_turns`) the information about the original geometries may be unavailable, only arbitrary 3 points. So it is not clear which `Geometry` and how many of them would have to be passed here.

So I propose to remove the parameter at all and have single side strategy only defined by the strategy.

Does it make sense?

include/boost/geometry/algorithms/detail/convex_hull/interface.hpp
```diff
@@ -0,0 +1,394 @@
...
+    static inline OutputIterator apply(Geometry const& geometry,
+                                       OutputIterator& out,
+                                       Strategy const& strategy)
+    {
+        //BOOST_CONCEPT_ASSERT( (geometry::concepts::ConvexHullStrategy<Strategy>) );
+
+        return dispatch::convex_hull_insert<
+                   geometry::point_order<Geometry>::value,
+                   geometry::closure<Geometry>::value
+               >::apply(geometry, out, strategy);
```
awulkiew, line 199: This code takes the new umbrella strategy. There is no backward compatibility.

This is how it is done e.g. for `area` with `services::strategy_converter` "converting" from the old strategy to the new umbrella strategy:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/area.hpp#L240-L252
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/strategies/area/services.hpp#L42-L50
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/strategies/area/cartesian.hpp#L48-L54

Unless this PR breaks the backward compatibility anyway.

include/boost/geometry/algorithms/detail/convex_hull/interface.hpp
```diff
@@ -0,0 +1,394 @@
...
+};
+
+struct convex_hull_insert
+{
+    template <typename Geometry, typename OutputIterator, typename Strategy>
+    static inline OutputIterator apply(Geometry const& geometry,
+                                       OutputIterator& out,
+                                       Strategy const& strategy)
+    {
+        //BOOST_CONCEPT_ASSERT( (geometry::concepts::ConvexHullStrategy<Strategy>) );
```
awulkiew, line 194: So this concept assert was commented out. It's because after changing the old strategy to umbrella strategy you'd have to alter the concept check as well. I'm ok with doing it later.

include/boost/geometry/strategies/convex_hull/cartesian.hpp
vissarion: Yes, I agree, removed.

include/boost/geometry/algorithms/detail/convex_hull/interface.hpp
vissarion: This is a breaking change, the convex hull algorithm use to take an "agnostic" convex_hull strategy "graham_andrew" but now takes a new umbrella strategy to provides a side predicate.

include/boost/geometry/algorithms/detail/convex_hull/interface.hpp
vissarion: OK, I will leave to a future PR.

include/boost/geometry/strategies/convex_hull/geographic.hpp
```diff
@@ -0,0 +1,70 @@
...
+    {}
+
+    static auto side()
+    {
+        return strategy::side::geographic
+            <
+                FormulaPolicy,
+                Spheroid,
+                CalculationType
+            >();
```
awulkiew, line 51: Here spheroid should've been passed (e.g. `base_t::m_spheroid`).

include/boost/geometry/strategies/convex_hull/spherical.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+namespace strategies { namespace convex_hull
+{
+
+template
+<
+    typename RadiusTypeOrSphere = double,
+    typename CalculationType = void
+>
+class spherical : strategies::detail::spherical_base<RadiusTypeOrSphere>
```
awulkiew, line 30: Radius is not needed in this strategy so it could take only `CalculationType`. There is a version `spherical_base` specialized for `void` for that ocasion, i.e.:
```
template <typename CalculationType = void>
class spherical : strategies::detail::spherical_base<void>
```

include/boost/geometry/strategies/geographic/side.hpp
```diff
@@ -120,6 +122,11 @@ class geographic
         typedef typename FormulaPolicy::template inverse
                     <calc_t, false, true, false, false, false> inverse_formula;
 
+        if (geometry::equals(p, p1) || geometry::equals(p, p2))
```
awulkiew, line 125: `equals` needs geographic/spherical point_in_point strategy.

awulkiew: Btw, is it a fix for a bug?

include/boost/geometry/strategies/spherical/side_by_cross_track.hpp
```diff
@@ -71,9 +71,19 @@ public :
         calc_t crs_AD = geometry::formula::spherical_azimuth<calc_t, false>
                              (lon1, lat1, lon, lat).azimuth;
 
+        if (crs_AD == 0)
+        {
+            return 0;
+        }
+
```
awulkiew, line 78: This doesn't look correct. Azimuth 0 means going north which doesn't mean that the point lies at the same great circle as the segment. Or am I missing something?

include/boost/geometry/strategy/cartesian/precise_area.hpp
```diff
@@ -0,0 +1,118 @@
+// Boost.Geometry
+
+// Copyright (c) 2020, Oracle and/or its affiliates.
+
+// Contributed and/or modified by Vissarion Fysikopoulos, on behalf of Oracle
+
+// Licensed under the Boost Software License version 1.0.
+// http://www.boost.org/users/license.html
+
+#ifndef BOOST_GEOMETRY_STRATEGY_CARTESIAN_AREA_ACCURATE_HPP
```
awulkiew, line 10: Macro name does not correspond to file name.

include/boost/geometry/algorithms/detail/convex_hull/graham_andrew.hpp
```diff
@@ -212,31 +211,30 @@ public :
...
         geometry::less<point_type> less;
 
-        detail::get_extremes(geometry, most_left, most_right, less);
+        detail::convex_hull::get_extremes(geometry, most_left, most_right, less);
 
         container_type lower_points, upper_points;
 
-        // TODO: User-defiend CS-specific side strategy
-        typename strategy::side::services::default_strategy<cs_tag>::type side;
+        auto side_strategy = strategy.side();
```
awulkiew, line 218: `auto const`

include/boost/geometry/strategies/convex_hull/cartesian.hpp
```diff
@@ -0,0 +1,49 @@
...
+namespace strategies { namespace convex_hull
+{
+
+template <typename CalculationType = void>
+class cartesian : strategies::detail::cartesian_base
+{
+public:
+    static auto side()
+    {
+        return strategy::side::side_robust<CalculationType>();
```
awulkiew, line 30: So the default side strategy is robust strategy. Is that correct? Typically we're choosing the fastest as the default. Does it mean that the robust side strategy is also faster?

include/boost/geometry/strategy/cartesian/side_non_robust.hpp
```diff
@@ -0,0 +1,76 @@
...
+    left of segment (>0), right of segment (< 0), on segment (0).
+\ingroup strategies
+\tparam CalculationType \tparam_calculation
+\details This predicate determines at which side of a segment a point lies
+*/
+template
+<
+    typename CalculationType = void
+>
+struct side_non_robust
```
awulkiew, line 34: So with this PR we have 3 cartesian side strategies: `side_by_triangle`, `side_non_robust` and `side_robust` with `side_robust` being the default for convex_hull.

And we also have `strategy::area::cartesian` and `strategy::area::precise_cartesian`. With the latter not defined in convex_hull umbrella strategy and probably only needed for testing.

So my questions are:

1. What's the difference between `side_by_triangle` and `side_non_robust`?

2. Why side is called robust and area called precise? Is this distinction actually important or are these names more or less arbitrary? Shouldn't we think about some consistent naming?

3. Are these robust/precise strategies complementary. I.e. is it the case that they should be used together in algorithms requiring both area and side strategies? Or can they be mixed with non-robust/non-accurate strategies?

include/boost/geometry/strategies/convex_hull/geographic.hpp
vissarion: right, done.

include/boost/geometry/strategies/convex_hull/spherical.hpp
vissarion: nice, thanks

include/boost/geometry/strategies/geographic/side.hpp
vissarion: >equals needs geographic/spherical point_in_point strategy.

right

>Btw, is it a fix for a bug?

Indeed, this fixes this bug https://github.com/boostorg/geometry/issues/765.

include/boost/geometry/strategies/spherical/side_by_cross_track.hpp
vissarion: This is a leftover from testing this function. I think the whole file could be removed, since this strategy is not used anywhere and `spherical_side_formula` does the same computation more efficiently.

include/boost/geometry/strategy/cartesian/precise_area.hpp
vissarion: OK this is from PR https://github.com/boostorg/geometry/pull/749 but I am fixing it here.

include/boost/geometry/algorithms/detail/convex_hull/graham_andrew.hpp
vissarion: right

include/boost/geometry/strategies/convex_hull/cartesian.hpp
vissarion: > So the default side strategy is robust strategy. Is that correct? Typically we're choosing the fastest as the default. Does it mean that the robust side strategy is also faster?

Yes, I chose to put robust as default one. It seems that it is also faster, see this discussion https://github.com/boostorg/geometry/issues/699

> EDIT: Is it because robust convex_hull is needed in e.g. in triangulation? If that's the case then convex_hull umbrella strategy could define non-robust strategy and triangulation umbrella strategy could define robust one.

No, this is independent from triangulation.

include/boost/geometry/strategy/cartesian/side_non_robust.hpp
vissarion: > So with this PR we have 3 cartesian side strategies: `side_by_triangle`, `side_non_robust` and `side_robust` with `side_robust` being the default for convex_hull.

> And we also have `strategy::area::cartesian` and `strategy::area::precise_cartesian`. With the latter not defined in convex_hull umbrella strategy and probably only needed for testing.

Actually proposed in this PR https://github.com/boostorg/geometry/pull/749 while the current one build on top of https://github.com/boostorg/geometry/pull/749
 
> So my questions are:
> 
>     1. What's the difference between `side_by_triangle` and `side_non_robust`?

`side_non_robust` is faster but probably less accurate since it does not implement any heuristic it just computes the determinant. 

>     2. Why side is called robust and area called precise? Is this distinction actually important or are these names more or less arbitrary? Shouldn't we think about some consistent naming?

I chose robust keyword for computations of predicates (yes/no result) to declare the guarantee that the result will be correct. I use precise keyword for constructions (numeric result) to declare that the result is "more precise" than the more simple computation. Of course in certain cases you can make it "more precise" (and yet slower) if you use multiprecision. We should definitely use some consistent naming. An alternative to the proposal above could be to use the keyword "robust" to whatever is more involved than straightforward floating points computation. 
 
>     3. Are these robust/precise strategies complementary. I.e. is it the case that they should be used together in algorithms requiring both area and side strategies? Or can they be mixed with non-robust/non-accurate strategies?

They can be mixed. You can compute a convex hull with robust side predicate and then compute the area with the non_robust are strategy. They are independent.

include/boost/geometry/strategies/spherical/side_by_cross_track.hpp
awulkiew: >This is a leftover from testing this function.

Ok, so these lines should be removed, right?

> I think the whole file could be removed, since this strategy is not used anywhere and `spherical_side_formula` does the same computation more efficiently.

Well, I don't know. One is using azimuths and the other 3d vectors and AFAIR the accuracy is different in both cases.

include/boost/geometry/strategy/cartesian/side_non_robust.hpp
awulkiew: Ok, the naming makes sense. I only worry that it may be start to be confusing at some point when there are different strategies called robust, precise or accurate. What if there are two different methods of achieving robustness or precision? Would it make sense or even be possible to use some more explicit names, e.g. indicating how the calculation is performed exactly?

include/boost/geometry/strategies/spherical/side_by_cross_track.hpp
vissarion: OK, so let's keep it. However, since there is an issue which is similar to the geographic side issue I updated  https://github.com/boostorg/geometry/issues/765 and I propose here a similar fix to the one of the geographic side.

include/boost/geometry/strategy/cartesian/side_non_robust.hpp
vissarion: I think for now we can stick to "robust" (for predicates) and "precise" (for constructions). Note that I am also OK with "robust" for all. Then, if at some point we have two different methods of robustness (or "preciseness") for the same problem (which is a bit unlikely at least w.r.t. my current understanding of the BG development) we can think of specialized names.

@barendgehrels what do you think about this PR and the discussed issues on robustness?

include/boost/geometry/strategies/convex_hull/cartesian.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <boost/geometry/strategy/cartesian/side_robust.hpp>
+
+namespace boost { namespace geometry
+{
+
+namespace strategies { namespace convex_hull
+{
+
+template <typename CalculationType = void>
+class cartesian : strategies::detail::cartesian_base
```
awulkiew, line 25: I forgot to mention that the base class should be `public` in order to allow access to its members.

include/boost/geometry/strategies/convex_hull/geographic.hpp
```diff
@@ -0,0 +1,70 @@
...
+namespace strategies { namespace convex_hull
+{
+
+template
+<
+    typename FormulaPolicy = strategy::andoyer,
+    typename Spheroid = srs::spheroid<double>,
+    typename CalculationType = void
+>
+class geographic : strategies::detail::geographic_base<Spheroid>
```
awulkiew, line 31: `public`

include/boost/geometry/strategies/convex_hull/spherical.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+
+namespace boost { namespace geometry
+{
+
+namespace strategies { namespace convex_hull
+{
+
+template <typename CalculationType = void>
+class spherical : strategies::detail::spherical_base<void>
```
awulkiew, line 26: `public`

awulkiew (approved):

include/boost/geometry/strategies/convex_hull/cartesian.hpp
vissarion: It is `public`. Am I missing something or you are looking at a older version?

awulkiew: No, I don't think so. It should either be:
```
class cartesian : public strategies::detail::cartesian_base
```
or
```
struct cartesian : strategies::detail::cartesian_base
```

vissarion: oh, sure, I misunderstood your first comment

vissarion:
@barendgehrels do you agree to merge this to develop to go to 1.75

barendgehrels (approved):
> @barendgehrels do you agree to merge this to develop to go to 1.75

Yes, sorry for the delay. I recently went through this and I'm OK. You can merge it.

PR #761, opened 2020-10-07
Remove ttmath

barendgehrels:
Previous PR removed other numeric types. Here TTMATH is removed in all tests and in headerfiles, where used.
Also ttmath itself is removed (in a separate commit because this is extensions)


vissarion (approved):
Good point, it looks OK to me.

awulkiew (approved):

PR #763, opened 2020-10-09
fixed risk of undefined behavior when parsing a byte_order byte

CG-SD:
tried to parse a spatialite WKB blob with this code but the byte_order byte was replaced by a '69' token :
 parsing this byte is successful but if the value read is not part of the enum, the function returns true but the order parameter is left untouched. 
Calls to that function elsewhere in the code do not initialise the order parameter.
see https://github.com/boostorg/geometry/blob/869e72fc01f2fce5ec3bcdfb91afab65ebd3ec6f/include/boost/geometry/extensions/multi/gis/io/wkb/detail/parser.hpp#L83

another option would be to return false in that case, but it didn't fit my need


mloskot (approved):

awulkiew:
Thanks!

PR #759, opened 2020-10-01
Replace MPL_ASSERT with std static_assert.

awulkiew:
This PR replaces `MPL_ASSERT` and `MPL_ASSERT_MSG` with `BOOST_GEOMETRY_STATIC_ASSERT` or `BOOST_GEOMETRY_STATIC_ASSERT_FALSE` expanded to `static_assert`.

`not_implemented` now takes parameter pack so can take arbitrary number of types.

It also removes `bg::info::XXX` from `not_implemented` because IMO this does not really enhance the output.

Below are examples showing how the compiler output is changed by this PR:

MSVC (`MPL_ASSERT` and `bg::info::XXX`):
```
1>  d:\lib\modular-boost\libs\geometry\include\boost\geometry\algorithms\not_implemented.hpp(70): error C2664: 'int boost::mpl::assertion_failed<false>(boost::mpl::assert<false>::type)': cannot convert argument 1 from 'boost::mpl::failed ************(__cdecl boost::geometry::nyi::not_implemented_error<boost::geometry::info::LINESTRING,boost::geometry::info::MULTI_POLYGON,boost::geometry::info::POLYGON>::THIS_OPERATION_IS_NOT_OR_NOT_YET_IMPLEMENTED::* ***********)(boost::mpl::assert_::types<Term1,Term2,Term3,boost::mpl::na>)' to 'boost::mpl::assert<false>::type'
1>          with
1>          [
1>              Term1=boost::geometry::info::LINESTRING,
1>              Term2=boost::geometry::info::MULTI_POLYGON,
1>              Term3=boost::geometry::info::POLYGON
1>          ]
1>  d:\lib\modular-boost\libs\geometry\include\boost\geometry\algorithms\not_implemented.hpp(77): note: No constructor could take the source type, or constructor overload resolution was ambiguous
1>  d:\lib\modular-boost\libs\geometry\include\boost\geometry\algorithms\not_implemented.hpp(126): note: see reference to class template instantiation 'boost::geometry::nyi::not_implemented_error<boost::geometry::info::LINESTRING,boost::geometry::info::MULTI_POLYGON,boost::geometry::info::POLYGON>' being compiled
1>  d:\lib\modular-boost\libs\geometry\include\boost\geometry\algorithms\union.hpp(55): note: see reference to class template instantiation 'boost::geometry::not_implemented<TagIn1,TagIn2,TagOut>' being compiled
1>          with
1>          [
1>              TagIn1=boost::geometry::linestring_tag,
[...]

PR #766, opened 2020-10-15
Include only necessary headers of Boost.Range.

awulkiew:
Do this in the whole library.

PR https://github.com/boostorg/geometry/pull/759 only did it in several places.


vissarion (approved):
Looks OK to me.

barendgehrels (approved):
Super! Thank you.

PR #767, opened 2020-10-19
Modernize tuple-related code.

awulkiew:
- replace `boost::tuple` with `std::tuple` in some places
- replace `policies::relate::segments_tupled` with `policies::relate::segments_intersection_policy` and instead of `get<I>` to access underlying policies (`policies::relate::segments_intersection_points` and `policies::relate::segments_direction`) use clearly named members `intersection_points` and `direction`. So something like this: `result.template get<1>()` becomes this: `result.direction`, which is more clear.
- replace `boost::tuple` with custom, lightweight tuple in srs projections/transformations which decreases compilation time
- modernize our tuple-related utilities defined in ` include/boost/geometry/util/tuples.hpp`


vissarion (approved):
Looks OK to me, thanks!

include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
```diff
@@ -118,16 +118,16 @@ class linear_intersections
...
                          EqPPStrategy const& strategy)
     {
-        int arrival_a = result.template get<1>().arrival[0];
-        int arrival_b = result.template get<1>().arrival[1];
-        bool same_dirs = result.template get<1>().dir_a == 0
-                      && result.template get<1>().dir_b == 0;
+        int arrival_a = result.direction.arrival[0];
+        int arrival_b = result.direction.arrival[1];
+        bool same_dirs = result.direction.dir_a == 0
+                      && result.direction.dir_b == 0;
```
barendgehrels, line 125: Much more readable, thanks

include/boost/geometry/algorithms/detail/overlay/get_turn_info_helpers.hpp
```diff
@@ -439,22 +429,21 @@ class intersection_info
     typedef typename base::side_calculator_type side_calculator_type;
     typedef typename base::result_type result_type;
     
-    typedef typename boost::tuples::element<0, result_type>::type i_info_type; // intersection_info
-    typedef typename boost::tuples::element<1, result_type>::type d_info_type; // dir_info
+    typedef typename result_type::intersection_points_type i_info_type; // intersection_info
+    typedef typename result_type::direction_type d_info_type;           // dir_info
```
barendgehrels, line 433: The comments can be omitted now, the types are clear enough.

include/boost/geometry/policies/relate/intersection_policy.hpp
```diff
@@ -0,0 +1,215 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+
+// Copyright (c) 2007-2012 Barend Gehrels, Amsterdam, the Netherlands.
```
barendgehrels, line 3: You can make it 2020

include/boost/geometry/policies/relate/intersection_policy.hpp
```diff
@@ -0,0 +1,215 @@
...
+            (
+                pts_policy::error(msg),
+                dir_policy::error(msg)
+            );
+    }
+};
+
+
+// "tupled" to return intersection results together.
+// NOTE: unused, left for backward compatibility
```
barendgehrels, line 136: Is it unused? I think it's still used.
If not used at all, I've no objection to remove it.

include/boost/geometry/srs/projections/spar.hpp
```diff
@@ -41,195 +41,104 @@ namespace spar
...
-    <
-        P0,
-        typename map_params_to_cons
-            <
-                P1, P2, P3, P4, P5, P6, P7, P8, P9, P10, P11, P12, P13, P14,
-                boost::tuples::null_type
-            >::type
-    > type;
+    BOOST_GEOMETRY_STATIC_ASSERT_FALSE(
+        "I out of bounds.",
```
barendgehrels, line 55: Can this be a `std` static assert? But it's TODO anyway so you can leave it for now

barendgehrels (approved):
I glanced through most of it, looks OK to me in general. Thanks.

include/boost/geometry/srs/projections/spar.hpp
awulkiew: It is std `static_assert` which fails when this template is instantiated. This is a utility I added recently which allows to do this. Note that raw `static_assert(false)` would fail always when the class is compiled, even if the template is not instantiated.

The TODO comment is only about making it a metafunction and moving to util/sequence.hpp in the future.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_helpers.hpp
awulkiew: Right.

include/boost/geometry/policies/relate/intersection_policy.hpp
awulkiew: I removed it.

include/boost/geometry/policies/relate/intersection_policy.hpp
awulkiew: Done.

awulkiew:
Thanks for the reviews.

PR #768, opened 2020-10-21
Support umbrella strategies in relops, setops, index, buffer and more.

awulkiew:
Also `is_simple`, `is_valid` and extensions like `dissolve` and `nsphere` relops.

Add umbrella strategies:
 - `relate`: used in setops, relops, `buffer`, *is_xxx*, `dissolve`, etc.
 - `index`: used in `rtree`
 - `io`: used in wkt read/write and extensions/shapefile

Replace `point_box_by_side` agnostic strategy with CS-specific strategies.

Fix passing of strategies (no defaults when user-defined strategy is
passed) in places like buffer and distance.

Alter tests accordingly.


awulkiew:
Right now many algorithms use `relate` strategy. Should I create separate strategy for each algorithm derived from `relate` strategy for naming consistency? If yes, should I do this in this PR which is big already?

include/boost/geometry/algorithms/buffer.hpp
```diff
@@ -242,22 +242,23 @@ inline void buffer(GeometryIn const& geometry_in,
     geometry::envelope(geometry_in, box);
     geometry::buffer(box, box, distance_strategy.max_distance(join_strategy, end_strategy));
 
-    typename strategy::intersection::services::default_strategy
+    typename strategies::relate::services::default_strategy
         <
-            typename cs_tag<GeometryIn>::type
-        >::type intersection_strategy;
+            GeometryIn, GeometryIn
+        >::type strategy;
```
awulkiew, line 248: So buffer uses the default relate umbrella strategy instead of intersection strategy.

Right now buffer strategies are at the same time parameters/options for buffer e.g. holding distance value and CS-specific strategies defining how parts of buffer are calculated. I propose to separate that in the future.

So this would be a separate buffer strategy passed by the user, also defining how specific buffer-related things are calculated (like generation of new points in various coordinate systems). And then I think `buffer()` algorithm should take parameters/options specifying distance/line-end/etc. plus one optional CS-specific umbrella strategy.

include/boost/geometry/strategies/relate/geographic.hpp
```diff
@@ -0,0 +1,344 @@
...
+
+template
+<
+    typename FormulaPolicy = strategy::andoyer,
+    // TODO: Is SeriesOrder argument needed here?
+    std::size_t SeriesOrder = strategy::default_order<FormulaPolicy>::value,
+    typename Spheroid = srs::spheroid<double>,
+    typename CalculationType = void
+>
+class geographic
```
awulkiew, line 44: Here is another question. Do you think that `SeriesOrder` is needed here? The only strategies that requires it are area and segments intersection strategy. The order is used for different series in these two strategies. I'm considering removing this parameter from the umbrella strategies and sticking to the default based on `FormulaPolicy`. What do you think?

barendgehrels (approved):
Sorry the review took so long.

And I can't see that I reviewed every line - but I'm definitely in favour of getting the umbrella strategies!

So of course approved.

vissarion:
Sorry for delays here, I will review in next few days.

include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
```diff
@@ -378,14 +400,15 @@ class multi_point_multi_geometry
...
-
             typedef typename boost::range_value<MultiGeometry>::type single_type;
 
             // The default strategy is enough for Point/Box
             if (! m_intersection_found
-                && ! detail::disjoint::disjoint_point_box(point, box_pair.first, d_pb_strategy_type())
-                && ! dispatch::disjoint<Point, single_type>::apply(point, range::at(m_multi_geometry, box_pair.second), m_strategy))
+                && ! detail::disjoint::disjoint_point_box(point, box_pair.first, m_strategy)
+                && ! dispatch::disjoint
+                        <
```
vissarion, line 409: 4 spaces of indentation here, right?

include/boost/geometry/algorithms/detail/distance/segment_to_box.hpp
```diff
@@ -58,17 +58,20 @@ namespace detail { namespace distance
...
 {
-    typedef typename strategy::disjoint::services::default_strategy
+    // TODO: pass strategy
+    auto const s = strategies::relate::services::strategy_converter
         <
-            Segment, Box
-        >::type strategy_type;
+            // This is the only strategy defined in distance segment/box
+            // strategies that carries the information about the spheroid
+            // so use it for now.
```
vissarion, line 70: can the umbrella strategy carry the spheroid information?

include/boost/geometry/algorithms/detail/equals/implementation.hpp
```diff
@@ -138,32 +137,28 @@ struct area_check
...
-                           strategy.template get_area_strategy<Geometry1>()),
-            geometry::area(geometry2,
-                           strategy.template get_area_strategy<Geometry2>()));
+        return geometry::math::equals(geometry::area(geometry1, strategy),
+                                      geometry::area(geometry2, strategy));
     }
 };
 
 
+/*
```
vissarion, line 146: so this is not needed? do we need it for future use?

include/boost/geometry/algorithms/detail/equals/interface.hpp
```diff
@@ -106,13 +113,37 @@ struct equals
...
+
+template <typename Strategy>
+struct equals<Strategy, false>
+{
+    template <typename Geometry1, typename Geometry2>
+    static inline bool apply(Geometry1 const& geometry1,
+                             Geometry2 const& geometry2,
+                             Strategy const& strategy)
+    {
+        //using strategies::equals::services::strategy_converter;
```
vissarion, line 126: is this going to used in the future? why do you use `relate` instead now?

include/boost/geometry/algorithms/detail/intersection/areal_areal.hpp
```diff
@@ -109,10 +109,10 @@ struct intersection_areal_areal_<TupledOut, tupled_output_tag>
             <
                 areal::index, TupledOut
             >::type areal_out_type;
-        typedef typename geometry::tuples::element
-            <
-                pointlike::index, TupledOut
-            >::type pointlike_out_type;
+        //typedef typename geometry::tuples::element
```
vissarion, line 112: should be removed?

include/boost/geometry/algorithms/detail/is_simple/areal.hpp
```diff
@@ -38,54 +37,34 @@ namespace detail { namespace is_simple
...
-    template <typename InteriorRings>
-    static inline
-    bool are_simple_interior_rings(InteriorRings const& interior_rings)
-    {
-        return
-            detail::check_iterator_range
+    simplicity_failure_policy policy;
+    return ! boost::empty(ring)
+        && ! detail::is_valid::has_duplicates
                 <
```
vissarion, line 46: one more space here ?

include/boost/geometry/algorithms/detail/is_simple/areal.hpp
```diff
@@ -38,54 +37,34 @@ namespace detail { namespace is_simple
...
-            &&
-            are_simple_interior_rings(geometry::interior_rings(polygon));
-    }
-};
+template <typename InteriorRings, typename Strategy>
+inline bool are_simple_interior_rings(InteriorRings const& interior_rings,
+                                      Strategy const& strategy)
+{
+    auto const end = boost::end(interior_rings);
+    return std::find_if(boost::begin(interior_rings), end,
```
vissarion, line 56: indentation looks weird here

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
```diff
@@ -187,11 +187,8 @@ private :
         // So advance to the "non duplicate next"
         // (the check is defensive, to avoid endless loops)
         std::size_t check = 0;
-        while(! detail::disjoint::disjoint_point_point
-                (
-                    current_robust_point, next_robust_point,
-                    disjoint_strategy_type()
-                )
+        while(! detail::disjoint::disjoint_point_point(
```
vissarion, line 190: a space is missing after `while`

vissarion (approved):
Thanks for this PR! It looks OK to me.

include/boost/geometry/algorithms/crosses.hpp
```diff
@@ -65,31 +67,65 @@ struct crosses
...
+template <>
+struct crosses<default_strategy, false>
+{
     template <typename Geometry1, typename Geometry2>
     static inline bool apply(Geometry1 const& geometry1,
                              Geometry2 const& geometry2,
                              default_strategy)
     {
-        typedef typename strategy::relate::services::default_strategy
+        //typedef typename strategies::crosses::services::default_strategy
```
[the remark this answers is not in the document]
awulkiew: Probably not. Initally I though that maybe we'd like to have a separate umbrella strategy for each algorithm. But even if we wanted to implement then we'd probably do this as template aliases which means that the compiler would see all of them as the same umbrella relate strategy. And this means that they'd be interchangeable and relate strategy here could stay.

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
```diff
@@ -921,7 +921,7 @@ template
     typename JoinStrategy,
     typename EndStrategy,
     typename PointStrategy,
-    typename IntersectionStrategy,
+    typename Strategy,
```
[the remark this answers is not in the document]
awulkiew: Because this is the umbrella strategy and the other ones are buffer-only strategies/parameters/selectors. I probably should've called it `Strategies` or `UmbrellaStrategy`.

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
```diff
@@ -962,7 +962,7 @@ inline void buffer_inserter(GeometryInput const& geometry_input, OutputIterator
         >::apply(geometry_input, collection,
             distance_strategy, side_strategy, join_strategy,
             end_strategy, point_strategy,
-            robust_policy, intersection_strategy.get_side_strategy());
+            robust_policy, strategy.side()); // pass strategy?
```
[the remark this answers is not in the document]
awulkiew: `strategy` is the umbrella strategy `strategy.side()` returns side strategy from umbrella strategy (e.g. `side_by_strategy` or `side_by_cross_track`) and `side_strategy` is buffer's parameter/strategy/selector defining how a segment is enlarged by buffer on its side. Again, I should've probably called this variable `strategies`.

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -1087,14 +1029,14 @@ struct buffered_piece_collection
...
+                                                     m_strategy))
             {
                 continue;
             }
 
             int const geometry_code
-                = detail::within::point_in_geometry(point,
-                    original.m_ring, m_point_in_geometry_strategy);
+                = detail::within::point_in_geometry(point, original.m_ring,
+                                                    m_strategy);
```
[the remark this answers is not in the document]
awulkiew: I don't know, I tend to align function arguments with the rest of the arguments if I can.

What exactly do you think should've been used here?
```
int const geometry_code
    = detail::within::point_in_geometry(point, original.m_ring,
        m_strategy);
```
```
int const geometry_code
    = detail::within::point_in_geometry(point,
        original.m_ring, m_strategy);
```
```
int const geometry_code
    = detail::within::point_in_geometry(
        point, original.m_ring, m_strategy);
```
The last one looks ok as well. So I can change it to the last one if you like.

And in this case this is probably also possible:
```
using detail::within::point_in_geometry;
int const geometry_code
    = point_in_geometry(point, original.m_ring, m_strategy);
```
since the same number of lines is used.

Or do you think about something else?

include/boost/geometry/algorithms/detail/disjoint/interface.hpp
```diff
@@ -53,13 +61,37 @@ struct disjoint
...
 
+template <typename Strategy>
+struct disjoint<Strategy, false>
+{
+    template <typename Geometry1, typename Geometry2>
+    static inline bool apply(Geometry1 const& geometry1,
+                             Geometry2 const& geometry2,
+                             Strategy const& strategy)
+    {
+        //using strategies::disjoint::services::strategy_converter;
```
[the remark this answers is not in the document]
awulkiew: Probably not, I'll remove it.

include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
awulkiew: It'd look bad:
```
&& ! dispatch::disjoint
    <
        Point, single_type
    >::apply(point, range::at(m_multi_geometry, box_pair.second), m_strategy))
```

In such cases I put as many tabs/spaces as it is needed to start the arguments list after the name of the function/struct starts.

include/boost/geometry/algorithms/detail/distance/segment_to_box.hpp
awulkiew: It does hold the spheroid, this is why it is constructed from the side strategy here. An alternative would be to construct it from distance strategy but distance strategies were never passed into relops and setops so this backward compatibility mechanism does not support that. I'm now working on umbrella strategies for distance so this code will be replaced anyway.

include/boost/geometry/algorithms/detail/equals/interface.hpp
awulkiew: Probably not. I'll probably leave relate here. I'm using relate everywhere to avoid creating an even bigger PR.

include/boost/geometry/algorithms/detail/equals/implementation.hpp
awulkiew: I don't know it seems not. I probably noticed that this is not used anywhere but was not sure. So in case I'm mistaken I simply comented it out instead of removing it.

include/boost/geometry/algorithms/detail/is_simple/areal.hpp
awulkiew: The rationale is as follows: the number of spaces is the first multiple of 4 starting after the start of the struct's name. AFAIU you're saying that it should start exactly 4 spaces after the start of the struct's name. It requires less number of keys pressed to do it my way (1 or 2 tabs after enter vs tabs followed by spaces or backspaces) and it looks almost the same. But I'm flexible. ;)

include/boost/geometry/algorithms/detail/is_simple/areal.hpp
awulkiew: How about this?
```
    return std::find_if(boost::begin(interior_rings), end,
                        [&](auto const& r)
                        {
                            return ! is_simple_ring(r, strategy);
                        }) == end; // non-simple ring not found
```

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
awulkiew: Indeed, a leftover from the previous version. I'll fix it.

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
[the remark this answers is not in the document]
vissarion: OK, I like `UmbrellaStrategy` naming but it's up to you.

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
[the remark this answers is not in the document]
vissarion: Thanks for the details, I was also aligning function arguments with the rest of the arguments, but I am somehow convinced by https://github.com/boostorg/geometry/wiki/Guidelines-for-Developers to wrap the lines (even this is not explicitly written there) so I am ok with any of your suggestions above.

include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
vissarion: OK

include/boost/geometry/algorithms/detail/equals/implementation.hpp
vissarion: OK

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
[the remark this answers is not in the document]
mloskot: By the way, if it is possible that in future the library will achieve a compromise on using `clang-format` to ensure the consistent code format then, I think, it may be a good idea to pick the ones that `clang-format` can do best.

include/boost/geometry/algorithms/detail/is_simple/areal.hpp
vissarion: I am ok with this rationale

include/boost/geometry/algorithms/detail/is_simple/areal.hpp
vissarion: better, easier to read, thanks

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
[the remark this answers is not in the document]
awulkiew: I looked again and it seems that after putting all of the characters in one line here the line is shorter than some of the other lines in this file. Plus even in the guidelines we define max length as 100 characters. So I think this is the way to go here.

Personally I don't have any preference regarding the formatting besides looking good and allowing to easily catch the sense of an expression. I guess that in various contexts the best formatting would look differently. In any case we should probably aim for the least amount of lines. In case of a function call we should keep the parameters in one block whatever it means. Closer to the max line length this could mean putting all of the parameters at the next line. If there were more space this could mean aligning the parameters to the first one, etc. The number of parameters will also matter because putting each of them at subsequent line will cause the lines bloat. The context will also matter because a function call somewhere in a block should probably be treated differently than in the `if` statement condition. So I'm in general sceptical about automatic formatting and I'd simply leave that to the programmer. But maybe clang-format is intelligent enough? IDK.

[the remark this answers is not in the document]
mloskot: > I don't have any preference regarding the formatting besides looking good

Generally, the only preference I've got is consistency.

> In any case we should probably aim for the least amount of lines.

👍 

> I'm in general sceptical about automatic formatting and I'd simply leave that to the programmer. 

The `clang-format` is not ideal but it get close to almost any formatting style, depending on how sophisticated is team's preferences are. The compelling feature of `clang-format` is that it allows to achieve consistency with almost zero effort and it also helps me to review work by new contributors quicker as I don't have to explain over and over how to format, just point to `.clang-format` file and voila ;) 

Anyway, I meant a brief digression and not to open whole style debate, apologies.

include/boost/geometry/strategies/relate/geographic.hpp
awulkiew: Do you have an opinion regarding it @vissarion @barendgehrels ?

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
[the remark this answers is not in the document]
awulkiew: Ok, I understand, no need to apologise.

include/boost/geometry/algorithms/detail/is_simple/areal.hpp
awulkiew: Done.

include/boost/geometry/algorithms/detail/intersection/areal_areal.hpp
awulkiew: done.

awulkiew:
Thanks for the reviews!

There are some things I'd like to add/modify but I'll do this is the next PR.

PR #769, opened 2020-10-28
Fix/start turns

barendgehrels:
This commit adds start turns (if specified in assign policy, and not for rescaling).
They are necessary in cases where the corresponding arrival turn is missed by accuracy.

Testcases (added) are found by `recursive_polygons_buffer`

Performance: with start turns:
```
$ recursive_polygons_buffer --count 10000 --level 4 --form triangle --validity 0
Test configuration:
  - Using Start Turns
  - Default test type: d
geometries: 300797 errors: 1501 type: d time: 26.026

Test configuration:
  - Default test type: d
geometries: 291375 errors: 2286 type: d time: 24.516
```

Note, with rescaling there are 0 errors, so I'm still concentrating on buffer. Next round I will look in some of the other added cases in more detail, but they're not related to missing turns but to clusters.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
@@ -832,6 +832,65 @@ struct equal : public base_turn_handler
 
 template
 <
+    typename TurnInfo
+>
+struct start : public base_turn_handler
+{
+    template
+    <
+        typename UniqueSubRange1,
+        typename UniqueSubRange2,
+        typename IntersectionInfo,
+        typename DirInfo,
+        typename SideCalculator,
+        typename UmbrellaStrategy
+    >
+    static inline bool apply(UniqueSubRange1 const& /*range_p*/,
+                UniqueSubRange2 const& /*range_q*/,
+                TurnInfo& ti,
+                IntersectionInfo const& info,
+                DirInfo const& dir_info,
+                SideCalculator const& side,
+                UmbrellaStrategy const& )
+    {
+#if defined(BOOST_GEOMETRY_USE_RESCALING)
+        // With rescaling, start turns are not needed.
+        return false;
+#endif
+
+        // Start turns have either how_a = -1, or how_b = -1 (either p leaves or q leaves)
+        BOOST_GEOMETRY_ASSERT(dir_info.how_a != dir_info.how_b);
+        BOOST_GEOMETRY_ASSERT(dir_info.how_a == -1 || dir_info.how_b == -1);
+        BOOST_GEOMETRY_ASSERT(dir_info.how_a == 0 || dir_info.how_b == 0);
+
+        if (dir_info.how_b == -1)
+        {
+            // p --------------->
+            //             |
+            //             | q         q leaves
+            //             v
+            //
+
+            int const side_qj_p1 = side.qj_wrt_p1();
+            ui_else_iu(side_qj_p1 == -1, ti);
+        }
+        else if (dir_info.how_a == -1)
+        {
+            // p leaves
+            int const side_pj_q1 = side.pj_wrt_q1();
+            ui_else_iu(side_pj_q1 == 1, ti);
+        }
+
+        // Copy intersection point
+        assign_point(ti, method_start, info, 0);
+        return true;
+    }
+
+};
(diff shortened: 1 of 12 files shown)


include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -830,6 +830,65 @@ struct equal : public base_turn_handler
...
+
+            int const side_qj_p1 = side.qj_wrt_p1();
+            ui_else_iu(side_qj_p1 == -1, ti);
+        }
+        else if (dir_info.how_a == -1)
+        {
+            // p leaves
+            int const side_pj_q1 = side.pj_wrt_q1();
+            ui_else_iu(side_pj_q1 == 1, ti);
+        }
```
barendgehrels, line 882: This code is relatively simple now, I first had more conditions but they were not relevant in the end.
But I might find cases where some behaviour should be fine-tuned.

test/algorithms/buffer/buffer_multi_polygon.cpp
```diff
@@ -541,7 +605,7 @@ int test_main(int, char* [])
 #endif
 
 #if defined(BOOST_GEOMETRY_TEST_FAILURES)
-    BoostGeometryWriteExpectedFailures(1, 1, 2, 3);
+    BoostGeometryWriteExpectedFailures(1, 14, 2, 11);
```
barendgehrels, line 608: Because it's related to precision, for float these changes were not necessary.
double/long double add more errors, just because I added the failing cases and part of them are not yet solved.

awulkiew (approved):

include/boost/geometry/algorithms/detail/overlay/handle_colocations.hpp
```diff
@@ -507,6 +507,46 @@ inline void discard_interior_exterior_turns(Turns& turns, Clusters& clusters)
...
 }
 
+template<typename Turns, typename Clusters>
+inline void discard_start_turns(Turns& turns, Clusters& clusters)
+{
+    for (auto& nv : clusters)
+    {
+        cluster_info& cinfo = nv.second;
+        auto& indices = cinfo.turn_indices;
+        std::size_t start_count{0};
```
awulkiew, line 517: Out of curiosity, why not `std::size_t start_count = 0;` ?

barendgehrels: Habit, it the office we're preferring the new uniform initialization.

What do you think?

mloskot: I used to be very keen in preferring the uniform initialization heavily and everywhere, until I read the abseil guidelines and tips, in particular https://abseil.io/tips/88

> First, “uniform” is a stretch: there are cases where ambiguity still exists (for the casual reader, not the compiler) in what is being called and how.
> 
> ```
> std::vector<std::string> strings{2}; // A vector of two empty strings.
> std::vector<int> ints{2};            // A vector containing only the integer 2.
> ```

It's a pity the uniform initialization is not corner-case free improvement to C++.

awulkiew: In this particular case I'd use `= 0` (also for numeric classes) because in such case the variable represents `0`.

This is not a case when an object of a class is created by passing some abstract which may describe something like number of elements.

From the same reason I'd probably initialize the vector like the one below. In other words to express that the initializer list on the right represents the vector or that it is the vector (just like with `0` being the number):
```
std::vector<int> ints = {2};
```
This works right? ;)

And then for all other things I don't have an opinion. I'd probably use the old parentheses. But maybe that's just because it's what I'm familiar with. Maybe with the exception of avoiding double parentheses to deal with the most vexing parse.

So I guess I could also express this as a conservative stance where old initialization is used when possible as it was in the past and new brackets only for new features.

barendgehrels: Interesting visions, thanks.
I agree with the statement that it's not that intuitive, `int i{0}` is quite different from all other languages where it is something like `int i=0;`
I'll think about it. For now I will merge this, assuming these were not objections.

PR #772, opened 2020-11-07
Add constexpr and boost::variant workarounds for vs2015.

awulkiew:
If you see our regression tests it looks like the majority of them compiles with VS2015:
https://www.boost.org/development/tests/develop/developer/geometry.html

Therefore I checked how much effort would it take to make all of them to compile. It seems that it was not much. VS2015 doesn't have full `constexpr` support and `boost::variant`'s is implemented using dummy template parameters instead of parameter pack.

For `constexpr` it was sufficient to disable it in some places and add explicitly initialize class member in other.

In order to analyze types of `boost::variant` I decided to not use Boost.MPL as it was before but instead to check raw template parameters, i.e. to check if they are geometries or not. This works for both dummy parameters and types that are not geometries. And while currently non-geometries are not supported in other places of the library we could consider supporting them. An example would be something like:
```
boost::variant<point, polygon, std::string> v = polygon{{{0, 0}, {0, 1}, {1, 1}, {1, 0}, {0, 0}}};
auto a = area(v);
```
Right now this code would not compile. At least I think it would not. But I don't see a semantic reason why it shouldn't. So in the future we could consider expanding the change done in this PR to the whole library.

What do you think about these workarounds?
Does it make sense to support vs2015?

### DIFF
--- a/include/boost/geometry/algorithms/area_result.hpp
+++ b/include/boost/geometry/algorithms/area_result.hpp
@@ -26,6 +26,7 @@
 
 #include <boost/geometry/util/select_most_precise.hpp>
 #include <boost/geometry/util/sequence.hpp>
+#include <boost/geometry/util/type_traits.hpp>
 
 #include <boost/variant/variant_fwd.hpp>
 
@@ -51,17 +52,18 @@ struct area_result
     typedef typename strategy_type::template result_type<Geometry>::type type;
 };
 
+template <typename Geometry, typename Strategy>
+struct area_result<Geometry, Strategy, false>
+{
+    typedef typename Strategy::template result_type<Geometry>::type type;
+};
+
+
 template
 <
     typename Geometry,
-    typename Strategy
+    bool IsGeometry = util::is_geometry<Geometry>::value
 >
-struct area_result<Geometry, Strategy, false>
-    : Strategy::template result_type<Geometry>
-{};
-
-
-template <typename Geometry>
 struct default_area_result
     : area_result
         <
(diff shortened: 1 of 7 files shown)


include/boost/geometry/algorithms/area_result.hpp
```diff
@@ -73,15 +75,37 @@ struct default_area_result
         >
 {};
 
+// Workaround for VS2015
+#if (_MSC_VER < 1910)
```
vissarion, line 79: should it be explicitly `_MSC_VER == 1900` or there is hope that the workaround will also work with older than VS15 versions ?

vissarion (approved):
I do not have a strong opinion against supporting VS2015, so I am ok with this patch.

include/boost/geometry/algorithms/area_result.hpp
awulkiew: No, I don't think that the older VS will compile the library.

This way the checks are symmetric, i.e. `< 1910` vs `>= 1910`.

awulkiew:
@mloskot @barendgehrels Do you have anything against merging it?

barendgehrels (approved):
Looks OK to me

awulkiew:
The #ifdefs were disabling `constexpr` with other compilers incorrectly.
I fixed it here: https://github.com/boostorg/geometry/commit/928603f32ea629f6586f43a2894e6293b908f838

barendgehrels:
Yes, better, thanks.

PR #773, opened 2020-11-12
Fix turns where segments arrive at an intersection point, or leave

barendgehrels:
This happens in rare cases. When they arrive there (or leave), their fraction should be 1 (or 0). The fraction is calculated and is sometimes nearly 1 (or ), but not considered so.
This can be fixed without any threshold because the information on arrival is already there.

It fixes ~10% of the remaining errors in the `recursive_polygons_buffer` test.

Several cases (added last PR) are fixed now. And I added 3 new cases (not yet fixed by this)

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
@@ -179,6 +179,39 @@ struct base_turn_handler
         ti.operations[1].fraction = info.fractions[index].robust_rb;
     }
 
+    template <typename TurnInfo, typename IntersectionInfo, typename DirInfo>
+    static inline void assign_point_and_correct(TurnInfo& ti,
+                method_type method,
+                IntersectionInfo const& info, DirInfo const& dir_info)
+    {
+        ti.method = method;
+
+        // For touch/touch interior always take the intersection point 0 (there is only one).
+        static int const index = 0;
+
+        geometry::convert(info.intersections[index], ti.point);
+
+        for (int i = 0; i < 2; i++)
+        {
+            if (dir_info.arrival[i] == 1)
+            {
+                // The segment arrives at the intersection point, its fraction should be 1
+                // (due to precision it might be nearly so, but not completely, in rare cases)
+                ti.operations[i].fraction = {1, 1};
+            }
+            else if (dir_info.arrival[i] == -1)
+            {
+                // The segment leaves from the intersection point, likewise its fraction should be 0
+                ti.operations[i].fraction = {0, 1};
+            }
+            else
+            {
+                ti.operations[i].fraction = i == 0 ? info.fractions[index].robust_ra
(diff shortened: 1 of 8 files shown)


test/algorithms/set_operations/difference/difference.cpp
```diff
@@ -624,7 +624,7 @@ int test_main(int, char* [])
 #if defined(BOOST_GEOMETRY_TEST_FAILURES)
     // Not yet fully tested for float and long double.
     // The difference algorithm can generate (additional) slivers
-    BoostGeometryWriteExpectedFailures(10, 11, 24, 14);
+    BoostGeometryWriteExpectedFailures(10, 11, 24, 15);
```
barendgehrels, line 627: _(this was already)_

vissarion: This means that there is no extra error here?

vissarion (approved):
It looks OK to me.

test/algorithms/set_operations/difference/difference.cpp
barendgehrels: Nope, these statements are mainly for myself, during transition of removing rescaling. This is one for `long double`. I probably didn't update it correctly last time. It's not updated automatically.
For this PR, I commented the code (updating the fraction) and it was still there - so it's not this PR for sure.

Main purpose is to see if we get there. For set operations, the version without rescaling is on par, or better. For buffer it's still not good enough so I'm concentrating on those cases.

awulkiew (approved):

awulkiew:
@barendgehrels Do you want it to be released with 1.75?

barendgehrels:
> 
> 
> @barendgehrels Do you want it to be released with 1.75?

Hi @awulkiew , no, it's not necessary - only for without rescaling, but that's currently not the default.

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -179,6 +179,39 @@ struct base_turn_handler
...
     }
 
+    template <typename TurnInfo, typename IntersectionInfo, typename DirInfo>
+    static inline void assign_point_and_correct(TurnInfo& ti,
+                method_type method,
+                IntersectionInfo const& info, DirInfo const& dir_info)
+    {
+        ti.method = method;
+
+        // Intersection points are calculated on both segments, either one can be taken.
```
barendgehrels, line 189: This comment was wrong, I'll fix it and merge. Thanks for your reviews

barendgehrels: Done :heavy_check_mark:

awulkiew:
@barendgehrels since you merged it now I'll push it together with other fixes to master for 1.75.

barendgehrels:
> 
> 
> @barendgehrels since you merged it now I'll push it together with other fixes to master for 1.75.

@awulkiew , fine, thank you!

PR #775, opened 2020-11-13
Fix unused parameter warnings

vissarion:


barendgehrels (approved):

include/boost/geometry/strategies/area/cartesian.hpp
```diff
@@ -49,6 +49,7 @@ struct strategy_converter<strategy::area::cartesian<CT> >
 {
     static auto get(strategy::area::cartesian<CT> const& strategy)
     {
+        boost::ignore_unused(strategy);
```
awulkiew, line 52: Here the parameter name could be removed instead, but `ignore_unused` works too.

awulkiew (approved):
Thanks!

include/boost/geometry/algorithms/detail/envelope/segment.hpp
```diff
@@ -60,6 +60,7 @@ struct envelope_segment_call_strategy<Strategy, false>
                              Segment const& segment, Box& mbr,
                              Strategy const& strategy)
     {
+        boost::ignore_unused(segment);
```
awulkiew, line 63: The header defining `ignore_unused` is missing.

PR #778, opened 2020-11-24
Fixes/workarounds for clang 3.5-3.9

awulkiew:
This PR adds fixes/workarounds for two problems that cause compilation failures with clang 3.5-3.9:
1. 3.5 - Class template partial specialization cannot be deduced for `boost::variant` and parameter pack. The solution to that is to use `BOOST_VARIANT_ENUM_PARAMS` in specializations. The list of types can be analyzed with variadic templates afterwards.
2. 3.5-3.9 - The instantiation of `std::is_constructible` for 2d `model::point` and 3 parameters causes the compilation error related to the initialization of 2-element array with too many elements. The solution to that is to enable the constructor only if the number of arguments is lesser or equal to `Dimension`. 

See the [regression matrix](https://www.boost.org/development/tests/develop/developer/geometry.html) for details.

Note that the error 2 was caused partially by this change to `cross_product`: https://github.com/boostorg/geometry/pull/745
in particular by specific `std::is_constructible` checks for 3 parameters. I think they are incorrect themselves and should be changed. There is a PR for that already: https://github.com/boostorg/geometry/pull/747

### DIFF
--- a/include/boost/geometry/algorithms/area_result.hpp
+++ b/include/boost/geometry/algorithms/area_result.hpp
@@ -143,14 +143,14 @@ struct area_result
     : detail::area::area_result<Geometry, Strategy>
 {};
 
-template <typename ...Ts, typename Strategy>
-struct area_result<boost::variant<Ts...>, Strategy>
+template <BOOST_VARIANT_ENUM_PARAMS(typename T), typename Strategy>
+struct area_result<boost::variant<BOOST_VARIANT_ENUM_PARAMS(T)>, Strategy>
     : geometry::area_result
         <
             typename util::select_pack_element
                 <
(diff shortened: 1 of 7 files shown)


vissarion (approved):
Looks OK to me.

mloskot (approved):

barendgehrels:
> 
> 
> Looks OK to me.

Looks OK to me as well - pity that it's necessary, but fine.

awulkiew:
@barendgehrels Thanks! It isn't entirely necessary but it's an easy fix and allows to compile BG with more compilers.
What do you think about this related fix: https://github.com/boostorg/geometry/pull/747 ?

PR #780, opened 2020-12-02
Fix/sort by side point at turn

barendgehrels:
Fixes number of remaining cases (in `recursive_polygons_buffer`) from 1341 to 628 (walking backwards)
and then from 628 to 444 (walking forwards)

It avoids sorting segments around a turn when the segment itself is at the turn (near-zero length)
It adds a new unit test for `copy_segment_points` because I added negative offsets for that.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/copy_segment_point.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/copy_segment_point.hpp
@@ -29,7 +29,6 @@
 #include <boost/geometry/core/static_assert.hpp>
 #include <boost/geometry/core/tags.hpp>
 #include <boost/geometry/geometries/concepts/check.hpp>
-#include <boost/geometry/iterators/ever_circling_iterator.hpp>
 #include <boost/geometry/util/range.hpp>
 #include <boost/geometry/views/closeable_view.hpp>
 #include <boost/geometry/views/reversible_view.hpp>
@@ -48,7 +47,7 @@ template <typename Range, bool Reverse, typename SegmentIdentifier, typename Poi
 struct copy_segment_point_range
 {
     static inline bool apply(Range const& range,
-            SegmentIdentifier const& seg_id, int offset,
+            SegmentIdentifier const& seg_id, signed_size_type offset,
             PointOut& point)
     {
         typedef typename closeable_view
@@ -66,15 +65,14 @@ struct copy_segment_point_range
         cview_type cview(range);
         rview_type view(cview);
 
-        typedef typename boost::range_iterator<rview_type>::type iterator;
-        geometry::ever_circling_iterator<iterator> it(boost::begin(view), boost::end(view),
-                    boost::begin(view) + seg_id.segment_index, true);
+        std::size_t const segment_count = boost::size(view) - 1;
 
-        for (signed_size_type i = 0; i < offset; ++i, ++it)
-        {
-        }
+        while (offset < 0) { offset += segment_count; }
 
-        geometry::convert(*it, point);
+        signed_size_type const target = (seg_id.segment_index + offset) % segment_count;
+
+        geometry::convert(*(boost::begin(view) + target), point);
+
         return true;
     }
 };
@@ -84,7 +82,7 @@ template <typename Polygon, bool Reverse, typename SegmentIdentifier, typename P
 struct copy_segment_point_polygon
 {
     static inline bool apply(Polygon const& polygon,
-            SegmentIdentifier const& seg_id, int offset,
+            SegmentIdentifier const& seg_id, signed_size_type offset,
             PointOut& point)
     {
         // Call ring-version with the right ring
@@ -110,7 +108,7 @@ template <typename Box, bool Reverse, typename SegmentIdentifier, typename Point
 struct copy_segment_point_box
 {
     static inline bool apply(Box const& box,
-            SegmentIdentifier const& seg_id, int offset,
+            SegmentIdentifier const& seg_id, signed_size_type offset,
             PointOut& point)
     {
         signed_size_type index = seg_id.segment_index;
@@ -137,7 +135,7 @@ template
 struct copy_segment_point_multi
 {
     static inline bool apply(MultiGeometry const& multi,
-                             SegmentIdentifier const& seg_id, int offset,
+                             SegmentIdentifier const& seg_id, signed_size_type offset,
                              PointOut& point)
     {
 
@@ -287,7 +285,7 @@ struct copy_segment_point
  */
 template<bool Reverse, typename Geometry, typename SegmentIdentifier, typename PointOut>
 inline bool copy_segment_point(Geometry const& geometry,
-            SegmentIdentifier const& seg_id, int offset,
+            SegmentIdentifier const& seg_id, signed_size_type offset,
             PointOut& point_out)
 {
     concepts::check<Geometry const>();
@@ -316,7 +314,7 @@ template
     typename PointOut
 >
 inline bool copy_segment_point(Geometry1 const& geometry1, Geometry2 const& geometry2,
-            SegmentIdentifier const& seg_id, int offset,
+            SegmentIdentifier const& seg_id, signed_size_type offset,
             PointOut& point_out)
 {
     concepts::check<Geometry1 const>();
--- a/include/boost/geometry/algorithms/detail/overlay/copy_segment_point.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/copy_segment_point.hpp
@@ -397,7 +397,6 @@ inline bool copy_segment_points(Geometry1 const& geometry1, Geometry2 const& geo
 }
 
 
-
 }} // namespace boost::geometry
 
 #endif // BOOST_GEOMETRY_ALGORITHMS_DETAIL_OVERLAY_COPY_SEGMENT_POINT_HPP
--- a/include/boost/geometry/algorithms/detail/overlay/copy_segment_point.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/copy_segment_point.hpp
@@ -42,6 +42,16 @@ namespace boost { namespace geometry
 namespace detail { namespace copy_segments
 {
 
+inline signed_size_type circular_offset(signed_size_type segment_count, signed_size_type index,
+                                        signed_size_type offset)
+{
+    signed_size_type result = (index + offset) % segment_count;
+    if (result < 0)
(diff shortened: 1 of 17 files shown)


include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
```diff
@@ -291,20 +317,34 @@ public :
...
-        add_segment(turn_index, op_index, point1, point_to, op.operation, op.seg_id, is_origin);
+
+        int offset = 0;
+
+        // If the point is in the neighbourhood (the limit itself is not important),
+        // then take a point (or more) further back.
+        // The limit of offset avoids theoretical infinite loops. In practice it currently
+        // walks max 1 point back in all cases.
+        while (approximately_equals(point1, turn.point, 1.0e-6) && offset > -10)
+        {
```
barendgehrels, line 328: This loop is the fix.

test/algorithms/set_operations/difference/difference.cpp
```diff
@@ -624,7 +624,7 @@ int test_main(int, char* [])
 #if defined(BOOST_GEOMETRY_TEST_FAILURES)
     // Not yet fully tested for float and long double.
     // The difference algorithm can generate (additional) slivers
-    BoostGeometryWriteExpectedFailures(10, 11, 24, 15);
+    BoostGeometryWriteExpectedFailures(10, 11, 24, 16);
```
barendgehrels, line 627: I still have to check if this was a regression (and then: why) or something else.

include/boost/geometry/algorithms/detail/overlay/copy_segment_point.hpp
```diff
@@ -66,15 +65,14 @@ struct copy_segment_point_range
...
-        geometry::ever_circling_iterator<iterator> it(boost::begin(view), boost::end(view),
-                    boost::begin(view) + seg_id.segment_index, true);
+        std::size_t const segment_count = boost::size(view) - 1;
 
-        for (signed_size_type i = 0; i < offset; ++i, ++it)
-        {
-        }
+        while (offset < 0) { offset += segment_count; }
+
+        signed_size_type const target = (seg_id.segment_index + offset) % segment_count;
```
awulkiew, line 72: It looks like it should be possible to calculate that without any loops and maybe even with one modulo operation.
Can `-offset` be greater than `segment_count`?

include/boost/geometry/algorithms/detail/overlay/copy_segment_point.hpp
```diff
@@ -66,15 +65,14 @@ struct copy_segment_point_range
...
+
+        geometry::convert(*(boost::begin(view) + target), point);
```
awulkiew, line 74: Or `range::at(view, target)`.

include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
awulkiew: What is this number `1.0e-6`? Will it work for `float` too? Will it work for big point coordinates when the closest next or previous floating point number is greater than `1.0e-6` ? Shouldn't it be scaled by coordinates like epsilon or maybe even based on machine epsilon?

include/boost/geometry/algorithms/detail/overlay/copy_segment_point.hpp
barendgehrels: Yes, modulo is probably possible. Offset is usually 1 (I've never encountered higher values)

include/boost/geometry/algorithms/detail/overlay/copy_segment_point.hpp
barendgehrels: Good idea, will change this

include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
barendgehrels: I wrote that in the comments. Machine epsilon won't work (I tried), it's too small and will still result in non-distinguishable sides (those are only indirectly based on these coordinates) - using them for side calculations afterwards.

Your concern about big coordinates might be valid, I have to check that.

awulkiew: I mean **based on** machine epsilon. So taking both of my remarks into account e.g.:
```
n * epsilon * max_coordinate(turn.point)
```
where `n` is your constant for tweaking. As an alternative `approximately_equals` could check relatively to coordinates because AFAIU right now `1.0e-6` is absolute value. So it could take epsilon-based-error, e.g. `n * epsilon` and then internally use this value together with each one coordinate to check for approximate equality. So use separate error-thresholds for each coordinate, basically similar to `equals()` but with epsilon multiplied by some factor.

Side note: there are only 3 or 4 values of machine epsilon used in Boost.Geometry for a given machine. One for each floating point number or `0` for integers and user-defined types. So the above would not work well for user-defined floating point types unless `numeric_limits` are specialized for such type. And this is not always possible.

barendgehrels: Thanks @awulkiew !
I get the point and will apply it as you suggest.

barendgehrels: I'm afraid this doesn't work for the range of values I'm using.

For `double`: 1.0e9 * epsilon is 2.22045e-07, this is around the value I need. But it doesn't translate to other values.
For `float`: 1.0e9 * epsilon is 119.209 and of course we can't work with that limit...
For `long double`: 1.0e9 * epsilon is 1.0842e-10, this works still fine.

I can use it if I use the double epsilon for float, which is kind of awkward.

My implementation (but I won't push it):
```
    // Returns true if two points are approximately equal, tuned by a giga-epsilon constant
    // (if constant is 1.0, for type double, the boundary is about 1.0e-6)
    template <typename Point1, typename Point2, typename T>
    static inline bool approximately_equals(Point1 const& a, Point2 const& b, T giga_epsilon)
    {
        // Including distance would introduce cyclic dependencies.
        using coor_t = typename select_coordinate_type<Point1, Point2>::type;
        constexpr coor_t epsilon = std::numeric_limits<coor_t>::epsilon();

        coor_t const& a0 = geometry::get<0>(a);
        coor_t const& b0 = geometry::get<0>(b);
        coor_t const& a1 = geometry::get<1>(a);
        coor_t const& b1 = geometry::get<1>(b);
        coor_t const one = 1.0;
        auto const c = math::detail::greatest(a0, b0, a1, b1, one);
        auto const limit = 1.0e9 * giga_epsilon * epsilon * c;

        return std::abs(a0 - b0) <= limit && std::abs(a1 - b1) <= limit;
    }

```

And I call it with 
`while (approximately_equals(point1, turn.point, 1.0) && offset > -10)`

test/algorithms/set_operations/difference/difference.cpp
barendgehrels: It fails in another (better) way and therefore the #expectation differs (one test can have multiple failures, for area, number of output polygons, etc)

include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
barendgehrels: OK, using a `min` I can fix it. I will submit this.

barendgehrels: :heavy_check_mark:

include/boost/geometry/algorithms/detail/overlay/copy_segment_point.hpp
barendgehrels: Fixed :heavy_check_mark:

include/boost/geometry/algorithms/detail/overlay/copy_segment_point.hpp
barendgehrels: :heavy_check_mark:

barendgehrels:
I still had to use `(std::min)`.
But now it's ready, and review comments are processed. Are you OK with this PR?

include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
```diff
@@ -331,25 +331,32 @@ public :
...
+            point_from = walk_over_ring(op, --offset, geometry1, geometry2);
+        }
+
+        // Similarly for the point to, walk forward
+        offset = 0;
+        while (approximately_equals(point_to, turn.point, 1.0) && offset < 10)
         {
-            point1 = walk_back(op, --offset, geometry1, geometry2);
+            point_to = walk_over_ring(op, ++offset, geometry1, geometry2);
```
barendgehrels, line 354: Like for `from`, we should also walk over the ring for `to`
(this is testcase `nores_wt_2`)

barendgehrels:
I want to continue tomorrow with the follow up.
@awulkiew @vissarion are you OK with this PR?

include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
vissarion: OK, so now `limit` is in the order of `1.0e-3`, `1.0e-7`, `1.0e-10`, for `float`, `double`, `long double` resp. 

Does the current solution work well (or better than the original with fixed `1.0e-6`) for large coordinates? What about user defined floating point numbers? 

A similar question, is `limit_giga_epsilon` ever needed to be different than `1.0`?

vissarion (approved):
In general I am ok with merging.

include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
barendgehrels: No, the solution doesn't work better, it works about the same (there's one case different for long double which I still have to investigate but I think it's not really related).

However, I didn't test with huge coordinates (maybe I should) or very very small coordinates. These testcases have values in the range of -100 .. 100. And Adam pointed out that it might turn out differently for other ranges, and therefore this solution is probably better.

About the limit: `1.0` works fine. But the limit is not crucial here: 2.0 will also work fine. Because it walks forwards or backwards to the next coordinate, and that will be fine unless there is a turn within the range of that limit...

barendgehrels:
I will merge this, next PR will be built on top of this. If you've still concerns let me know and we can fix that afterwards.

Issue #664, opened 2020-02-05
Writing 3D WKT: Z in prefix?


sgiraudot:
When writing 3D objects in WKT, a "Z" is supposed to be added to the prefix, such that a 2D point and a 3D point are differentiated:
```
POINT (1,2)
POINT Z (1,2,3)
```
But as far as I tested, when writing 3D objects using the WKT writer of Boost Geometry, the "Z" prefix is never written, leading to this kind of output:

```
POINT (1,2)
POINT (1,2,3)
```
It seems to be similar for other types of primitives (`LINESTRING`, etc.).

Reading 3D objects with Boost both works with and without "Z" in the input.

Wouldn't it make more sense to add "Z" to the prefix whenever a 3D object is written?

mloskot:
> When writing 3D objects in WKT, a "Z" is supposed to be added to the prefix,

Yes, only when writing WKT as defined the ISO SQL/MM specification.
Boost.Geometry is not compliant with the SQL/MM. The WKT I/O is based on the SFSQL 1.0 specification (still, I'm not betting my head it's fully compliant with it either).

I'd suggest to consider the WKT I/O is a utility for user's convenience and for convenience of developers testing the library features itself. It is not really a fully-featured up-to-date production-ready and SQL/MM WKT implementation.

awulkiew:
So we'd need a separate function for that.

We should also probably do it for `M`. The problem is currently we are unable between distinguishing between coordinates, measures, time, etc. It's all coordinates for us.

mloskot:
I thought perhaps `detail::wkt::prefix_point` could be a template taking `Point` then it could dispatch based on dimension, `POINT` for 2 or `POINT Z` for 3.

I think `Z` is more important than other measures. Another issue is that if BG advertises support of other measures, then users will likely expect e.g. why `M` is not copied/preserved/synthesised etc. by operation X :-)

sudo-panda:
I would like to work on this. Is what @mloskot suggested final or does this need more inputs?

mloskot:
@sudo-panda If my comment does not receive any responses and if my suggestion _seems sensible to you_, and you are willing to give it a go, then I'd suggest you to submit PR.

Once there is a (working) proof of concept ready to review, it will be much easier for more people to offer specific feedback. It may turn out the idea is a good one and your PR will be accepted or we will find problems and your PR will need rework. You need to decide if you want to take the risk :)

sudo-panda:
[...]

PR #782, opened 2020-12-09
Fixed Writing 3D WKT:Z prefix

Siddharth-coder13:
Changed the `wkt_point` struct to take 3 parameters rather than 2, for one for 2d and one for 3d, added the `prefix_z` function in `prefix.hpp,` and also added a line of code to print 3d points in `01_point_example.cpp` .

### DIFF
--- a/example/01_point_example.cpp
+++ b/example/01_point_example.cpp
@@ -129,7 +129,7 @@ int main()
     assign_values(d3a, 1, 2, 3);
     assign_values(d3b, 4, 5, 6);
     d3 = distance(d3a, d3b);
-
+    std::cout << wkt(d3a) << std::endl;
 
 
     // Other examples show other types of points, geometries and more algorithms
--- a/include/boost/geometry/io/wkt/detail/prefix.hpp
+++ b/include/boost/geometry/io/wkt/detail/prefix.hpp
@@ -32,6 +32,11 @@ struct prefix_point
     static inline const char* apply() { return "POINT"; }
 };
 
+struct prefix_point_z
+{
+    static inline const char* apply() { return "POINT Z"; }
+};
+
 struct prefix_polygon
 {
     static inline const char* apply() { return "POLYGON"; }
--- a/include/boost/geometry/io/wkt/write.hpp
+++ b/include/boost/geometry/io/wkt/write.hpp
@@ -113,13 +113,14 @@ struct double_closing_parenthesis
 /*!
 \brief Stream points as \ref WKT
 */
-template <typename Point, typename Policy>
+template <typename Point, typename Policy1, typename Policy2>
 struct wkt_point
 {
     template <typename Char, typename Traits>
     static inline void apply(std::basic_ostream<Char, Traits>& os, Point const& p, bool)
     {
-        os << Policy::apply() << "(";
+        if(dimension<Point>::type::value == 3) os << Policy2::apply() << "(";
+        else os << Policy1::apply() << "(";
         stream_coordinate<Point, 0, dimension<Point>::type::value>::apply(os, p);
         os << ")";
     }
@@ -350,7 +351,8 @@ struct wkt<Point, point_tag>
     : detail::wkt::wkt_point
         <
             Point,
-            detail::wkt::prefix_point
+            detail::wkt::prefix_point,
+            detail::wkt::prefix_point_z
         >
 {};
 
@@ -417,6 +419,7 @@ struct wkt<Multi, multi_point_tag>
             detail::wkt::wkt_point
                 <
                     typename boost::range_value<Multi>::type,
+                    detail::wkt::prefix_null,
                     detail::wkt::prefix_null
                 >,
             detail::wkt::prefix_multipoint


mloskot (changes requested):
Thanks for the contribution.

Since every new improvement or feature needs to be covered with test case(s), you may consider adding some to https://github.com/boostorg/geometry/tree/develop/test/io/wkt

mloskot:
@awulkiew In https://github.com/boostorg/geometry/issues/664#issuecomment-583736980 you suggested
> we'd need a separate function for that.

What did you have in mind?

This PR seems to follow my suggestion I made in https://github.com/boostorg/geometry/issues/664#issuecomment-583773745

@Siddharth-coder13 The commit and description of PR that addresses an existing issue should contain reference to the issue (i.e. `#664`)

Siddharth-coder13:
Yeah, sure from next time I will make sure this.

Siddharth-coder13:
Do this PR needs more changes? @mloskot

mloskot:
@Siddharth-coder13 No, for now it does not, but we need to wait and clarify my question https://github.com/boostorg/geometry/pull/782#issuecomment-743275433 in order to confirm this is actually the kind of change we need and want.

Siddharth-coder13:
@awulkiew,  your reviews are needed on this PR.

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -113,13 +113,14 @@ struct double_closing_parenthesis
...
-template <typename Point, typename Policy>
+template <typename Point, typename Policy1, typename Policy2>
 struct wkt_point
 {
     template <typename Char, typename Traits>
     static inline void apply(std::basic_ostream<Char, Traits>& os, Point const& p, bool)
     {
-        os << Policy::apply() << "(";
+        if(dimension<Point>::type::value == 3) os << Policy2::apply() << "(";
+        else os << Policy1::apply() << "(";
```
barendgehrels, line 123: Question: do we always want a `Z` in our 3D point? Until now we just did without and it works without.
If we suddenly add a `Z` here it might be incompatible for users (if any) who use this functionality. So it's a breaking change.

Apart from this it should be a compile time condition i/o a runtime condition. So we should dispatch based on dimension. But I'm not sure about that either, because if it is optional (as it probably should be), that might be inconvenient.

barendgehrels (changes requested):
Thanks for the PR but it's not yet clear enough if we want to have a `Z` by default.

include/boost/geometry/io/wkt/write.hpp
barendgehrels: As far as I can see the specification indeed mentions a `Z` for three dimensional points. It's unclear if a three dimensional point without a `Z` is acceptable or not.
But what do we do with four dimensional points? Or even more? We've not much support for them of course, but we can calculate distances for them...

In my personal opinion, for a three dimensional point, a list with three coordinates is already enough. In reading we can read or not read the `Z` (that is already implemented).

vissarion: I would agree with @barendgehrels 

Another issue here is that after this change the following code 
```
boost::geometry::read_wkt("POINT M(1 2 3)", p);
std::cout << boost::geometry::wkt(p) << std::endl;
```
will output
    
```POINT Z(1 2 3)```

I think we should decide how we want to support more than 2 dimensions and probably issue https://github.com/boostorg/geometry/issues/664 is the right place for that.

**Review by vissarion — CHANGES_REQUESTED** [review]

File: include/boost/geometry/io/wkt/write.hpp (hunk 1 above)
Comment (mloskot) [reply include/boost/geometry/io/wkt/write.hpp:123]: @barendgehrels points out a valid issue.

> But what do we do with four dimensional points?

From the OGC SFS point, we have:
- OGC SFS 1.1.0 restricted to 2D geometries 
- The extended WKT/ or 2.5D WKT implemented by OGR, PostGIS, etc. allow 2, 3, 4 dimension coordinates and do not use `Z` or `M` tags, see [Adam Gawne-Cain, 1999, Simple Features Revision Proposal](https://gis.stackexchange.com/a/79189/939)
- OGC SFS 1.2.x allows 2, 3 and 4 dimensions using the `Z` or `M` tags:  `POINT Z(x, y, z)`,   `POINT M(x, y, m)`,   `POINT ZM(x, y, z, m)`. I think many may consider `M` as pseudo-dimension or not a dimension but a measure value.

>  a list with three coordinates is already enough

It seems desirable indeed to keep the current WKT I/O that does not read/output the `Z` and `M` tags.

Perhaps, an acceptable solution would be an optional `flavor` parameter, or similar, with current

- `boost::geometry::read_wkt(wkt, geometry, flavor`
- `boost::geometry::wkt(geometry, flavor)`

where `flavor` is enum, something like:

```cpp
enum /*class*/ wkt_flavour
{
  ogc2d,        // OGC SFS 1.1.0
  ogc3d,        // OGC SFS 1.2.x
  ogc25d,       // Adam Gawne-Caine, 1999, Simple Features Revision Proposal
  ewkt = ogc25d // PostGIS terminology
};
```

or as an entirely separate OGC-compliant `boost/geometry/gis/io/wkt/` extension.

example/01_point_example.cpp
```diff
@@ -107,6 +107,29 @@ int main()
 
 
     // Some ways of getting point values
+    typedef model::point<double, 3, cs::cartesian> point_t;
+    typedef model::ring<point_t> ring_t;
+    typedef model::PolyhedralSurface<ring_t> poly_t;
```
mloskot: Since the library is now C++14, the aliases could be declared with the `using` declarations, which is a bit neater, IMHO, as it puts the most important part, the name, to the left.

example/01_point_example.cpp
```diff
@@ -107,6 +107,29 @@ int main()
...
+
+    poly_t polyhedron1;
+    poly_t polyhedron2 = {{{0,0,0}, {0, 1, 0}, {1, 1, 0}, {1, 0, 0}, {0, 0, 0}}, {{0, 0, 0}, {0, 1, 0}, {0, 1, 1}, {0, 0, 1}, {0, 0, 0}},
+    {{0, 0, 0}, {1, 0, 0}, {1, 0, 1}, {0, 0, 1}, {0, 0, 0}}, {{1, 1, 1}, {1, 0, 1}, {0, 0, 1}, {0, 1, 1}, {1, 1, 1}}, {{1, 1, 1}, {1, 0, 1}, {1, 0, 0}, {1, 1, 0}, {1, 1, 1}}, 
+    {{1, 1, 1}, {1, 1, 0}, {0, 1, 0}, {0, 1, 1}, {1, 1, 1}} };
```
mloskot: Could you reformat it to something easier to distinguish rows and columns? e.g.

```
{
  {{...}},
  {{...}},
  {{...}}
};
```

include/boost/geometry/core/point_type.hpp
```diff
@@ -92,6 +92,16 @@ struct point_type<ring_tag, Ring>
...
 
+// Specialization for PolyhedralSurface: the point-type is the point-type of its rings
+template <typename PolyhedralSurface>
+struct point_type<polyhedral_surface_tag, PolyhedralSurface>
+{
+    typedef typename point_type
+        <
+            ring_tag,
+            typename ring_type<polyhedral_surface_tag, PolyhedralSurface>::type
+        >::type type;
```
mloskot: The `using` too?

include/boost/geometry/geometries/PolyhedralSurface.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+#include <boost/geometry/geometries/concepts/point_concept.hpp>
+#include <boost/geometry/geometries/ring.hpp>
+#include <boost/geometry/core/tag.hpp>
+#include <boost/geometry/core/tags.hpp>
+
+#include <boost/config.hpp>
+#ifndef BOOST_NO_CXX11_HDR_INITIALIZER_LIST
+#include <initializer_list>
+#endif
```
mloskot: I'd move this in  these between

```
#include <vector>
<---
#include <boost/concept/assert.hpp>
```

include/boost/geometry/geometries/PolyhedralSurface.hpp
```diff
@@ -0,0 +1,113 @@
...
+>
+class PolyhedralSurface : public Container<Ring, Allocator<Ring> >
+{
+	BOOST_CONCEPT_ASSERT( (concepts::Ring<Ring>) );
+
+public :
+
+    using point_type = model::point<double, 3, boost::geometry::cs::cartesian>;
+    using ring_type = ring<point_type, true, true>;
+    using ph = Container<ring_type, Allocator<ring_type> >;
```
mloskot: What does `ph` stand for?
I'd suggest to use full, descriptive names, avoid abbreviations, unless others agree those are fine.

include/boost/geometry/geometries/concepts/PolyhedralSurface_concept.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+template <typename Geometry>
+class PolyhedralSurface
+{
+
+#ifndef DOXYGEN_NO_CONCEPT_MEMBERS
+
+	typedef typename ring_type<Geometry>::type ring_type;
+	
+
```
mloskot: Too many blank lines

include/boost/geometry/geometries/concepts/PolyhedralSurface_concept.hpp
```diff
@@ -0,0 +1,40 @@
...
+	typedef typename ring_type<Geometry>::type ring_type;
+	
+	
+	BOOST_CONCEPT_ASSERT( (concepts::Ring<ring_type>) );
+
+public:
+	
+	BOOST_CONCEPT_USAGE(PolyhedralSurface)
+	{
+
```
mloskot: Superfluous blank

include/boost/geometry/geometries/concepts/PolyhedralSurface_concept.hpp
```diff
@@ -0,0 +1,40 @@
...
+	}
+#endif
+	
+};
+
+} // namepspace concepts
+
+} // namespace geometry
+
+} // namespace boost
```
mloskot:
```
}}} // boost::geometry::concepts
```

include/boost/geometry/geometries/concepts/check.hpp
```diff
@@ -39,7 +39,7 @@
 #include <boost/geometry/geometries/concepts/polygon_concept.hpp>
 #include <boost/geometry/geometries/concepts/ring_concept.hpp>
 #include <boost/geometry/geometries/concepts/segment_concept.hpp>
-
+#include <boost/geometry/geometries/concepts/PolyhedralSurface_concept.hpp>
```
mloskot: Please, do not use CamelCase in file names (i.e. `PolyhedralSurface_concept.hpp` should read `polyhedralsurface_concept.hpp`

include/boost/geometry/geometries/geometries.hpp
```diff
@@ -29,5 +29,6 @@
 #include <boost/geometry/geometries/polygon.hpp>
 #include <boost/geometry/geometries/ring.hpp>
 #include <boost/geometry/geometries/segment.hpp>
+#include <boost/geometry/geometries/PolyhedralSurface.hpp>
```
mloskot: CamelCase

include/boost/geometry/geometries/register/PolyhedralSurface.hpp
```diff
@@ -0,0 +1,17 @@
...
+
+#define BOOST_GEOMETRY_REGISTER_POLYHEDRALSURFACE(PolyhedralSurface) \
+namespace boost { namespace geometry { namespace traits {  \
+    template<> struct tag<PolyhedralSurface> { typedef PolyhedralSurface_tag type; }; \
+}}}
+
+
+#define BOOST_GEOMETRY_REGISTER_POLYHEDRALSURFACE_TEMPLATED(PolyhedralSurface) \
+namespace boost { namespace geometry { namespace traits {  \
+    template<typename P> struct tag< PolyhedralSurface<P> > { typedef PolyhedralSurface_tag type; }; \
```
mloskot:
```
< PolyhedralSurface<P> >
```
to
```
<PolyhedralSurface<P>>
```

include/boost/geometry/io/wkt/read.hpp
```diff
@@ -453,6 +453,32 @@ struct polygon_parser
...
+    typedef container_appender<ring_return_type> appender;
+
+    static inline void apply(tokenizer::iterator& it,
+                             tokenizer::iterator const& end,
+                             std::string const& wkt,
+                             PolyhedralSurface& Poly)
+    {
+        handle_open_parenthesis(it, end, wkt);
+        typename ring_type<PolyhedralSurface>::type ring;
+        while(it != end && *it != ")"){
```
mloskot: `{` in separate line

mloskot (approved):

include/boost/geometry/io/wkt/detail/prefix.hpp
```diff
@@ -42,6 +47,11 @@ struct prefix_linestring
     static inline const char* apply() { return "LINESTRING"; }
 };
 
+struct prefix_polyhedral_surface
+{
+	static inline const char* apply() { return "POLYHEDRALSURFACE Z"; }
```
mloskot: Not sure about the `Z` as per @barendgehrels  https://github.com/boostorg/geometry/pull/782#discussion_r550097589

mloskot (changes requested):

mloskot:
@Siddharth-coder13 Please, see https://github.com/boostorg/geometry/pull/789#issuecomment-777841388

vissarion:
@Siddharth-coder13 are you planing to work on this? Otherwise we should close it, it is inactive for 4 years.

PR #786, opened 2020-12-22
Catch up with dms_parser change from function call operator to apply

mloskot:
The change to `dms_parser` was applied in PR #394

------

This PR is required to make all examples compile, see #787


barendgehrels (approved):
Thanks - a bit weird that this was apparently untested.
But it's an extension.

mloskot:
It surprised me too.

I've taken a note to add one in https://github.com/boostorg/geometry/issues/788

PR #790, opened 2020-12-30
Fix/get turn info collinear

barendgehrels:
This PR handles, in rare cases, collinear as if it was equal, because the arriving segment arrives at the intersection point in the same way as it is in equal (but at an undetectable distance).

In terms of statistics this reduces the error count from 444 to 241 (in recursive_polygons_buffer)

There are several tests added, some of them are fixed by this test, some remain to be fixed but it's more convenient to have them in this PR already.

There is also a bit cleanup in get_turn_info in the same way I added the new code.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
@@ -186,7 +186,8 @@ struct base_turn_handler
     {
         ti.method = method;
 
-        // For touch/touch interior always take the intersection point 0 (there is only one).
+        // For touch/touch interior always take the intersection point 0
+        // (usually there is only one - but if collinear is handled as touch, both could be taken).
         static int const index = 0;
 
         geometry::convert(info.intersections[index], ti.point);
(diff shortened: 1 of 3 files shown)


include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -965,8 +922,57 @@ template
...
+#endif
+        int const arrival_p = dir_info.arrival[0];
+        int const arrival_q = dir_info.arrival[1];
+        if (arrival_p * arrival_q != -1 || info.count != 2)
+        {
+            // Code below assumes that either p or q arrives in the other segment
+            return false;
+        }
+
+        auto const location = distance_measure(info.intersections[1],
```
vissarion, line 948: regarding the name `location` maybe `distance_from_location` or something is more relevant since this is not a location but some kind of distance.

vissarion (approved):
Looks ok to me, thanks!

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
barendgehrels: Agreed, I don't know why I took `location`. I will change it, and merge. Thanks for your review.

barendgehrels: In other places I used `dm` for this purpose so I use that now here too.

Issue #771, opened 2020-11-03
Azimuth algorithm is undocumented and it's API is not exposed at the same level as distance


kenba:
The OP of this stack overflow question: [boost geometry angle between two points](https://stackoverflow.com/questions/64639137/boost-geometry-angle-between-two-points) proposed using the `azimuth` algorithm to calculate the angle (bearing) between two points. 

The `azimuth` algorithm is only defined in the file <boost/geometry/algorithms/detail/azimuth.hpp>, it is not exposed on the API at the same level as `distance`: <boost/geometry/algorithms/distance.hpp> nor is it listed in the documentation: [algorithms](https://www.boost.org/doc/libs/1_74_0/libs/geometry/doc/html/geometry/reference/algorithms.html). A commenter pointed out that it is therefore an "undocumented" algorithm.

For the purposes of navigation, the `azimuth` algorithm is as important, if not more so, than the `distance` algorithm. Therefore it's API and documentation should at least match that of the `distance` algorithm. Note, also for the purposes of navigation it should include geographic azimuth algorithms, e.g.: Andoyer, Vincenty and especially, Karney.

BTW the `course` algorithm in <boost/geometry/algorithms/detail/course.hpp> simply refers to the `azimuth` algorithm. Should this also be exposed properly on the API and documented or should it be deleted?

vissarion:
Indeed azimuth is computed internally in many places of the library but there is no (proper) algorithm exposed on the API. Such an algorithm would look similar to distance and use strategies to choose between andoyer, thomas, vincenty or karney "formulas" in geographic systems.

 >BTW the course algorithm in <boost/geometry/algorithms/detail/course.hpp> simply refers to the azimuth algorithm. Should this also be exposed properly on the API and documented or should it be deleted?

It is left there for backward compatibility but it seems that it is not used at all.

PR #797, opened 2021-01-20
Add new algorithm azimuth.

awulkiew:
This PR addresses the issue: https://github.com/boostorg/geometry/issues/771

Usage:
```
auto azimuth = boost::geometry::azimuth(p1, p2);
```
or
```
bg::srs::spheroid<double> spheroid(6378137.0, 6356752.3142451793);
bg::strategies::azimuth::geographic<> strategy(spheroid);
auto azimuth = boost::geometry::azimuth(p1, p2, strategy);
```

After:
https://postgis.net/docs/ST_Azimuth.html

Note:
It also adds implementation of cartesian azimuth strategy and azimuth umbrella strategies.


doc/src/examples/algorithms/azimuth.cpp
```diff
@@ -0,0 +1,42 @@
...
+//` Shows how to calculate azimuth
+
+#include <iostream>
+
+#include <boost/geometry.hpp>
+#include <boost/geometry/geometries/point_xy.hpp>
+
+int main()
+{
+    typedef boost::geometry::model::d2::point_xy<double> point_type;
```
barendgehrels, line 21: We could use `using` - but it's fine anyway. Minor.

include/boost/geometry/algorithms/azimuth.hpp
```diff
@@ -0,0 +1,226 @@
...
+            <
+                typename coordinate_type<Point1>::type,
+                typename coordinate_type<Point2>::type
+            >::type calc_t;
+
+        calc_t result = 0;
+        calc_t x1 = geometry::get_as_radian<0>(p1);
+        calc_t y1 = geometry::get_as_radian<1>(p1);
+        calc_t x2 = geometry::get_as_radian<0>(p2);
+        calc_t y2 = geometry::get_as_radian<1>(p2);
```
barendgehrels, line 76: Can you make them `const` ?

include/boost/geometry/algorithms/azimuth.hpp
```diff
@@ -0,0 +1,226 @@
...
+\qbk{[include reference/algorithms/azimuth.qbk]}
+
+\qbk{
+[heading Example]
+[azimuth]
+[azimuth_output]
+}
+*/
+template <typename Point1, typename Point2>
+inline auto azimuth(Point1 const& point1, Point2 const& point2)
```
barendgehrels, line 181: Cool, this `auto` here!
:smiley:

include/boost/geometry/strategies/spherical.hpp
```diff
@@ -55,6 +56,13 @@ class spherical : strategies::detail::spherical_base<RadiusTypeOrSphere>
             >(base_t::m_radius);
     }
 
+    // azimuth
```
barendgehrels, line 59: This comment is probably redundant

test/algorithms/azimuth.cpp
```diff
@@ -0,0 +1,62 @@
...
+    test_one(P(0, 0), P(100, 1), 88.986933066023497);
+    test_one(P(0, 0), P(-1, 1), -45.187718848049521);
+    test_one(P(0, 0), P(-100, 1), -88.986933066023497);
+}
+
+int test_main(int, char* [])
+{
+    test_car< bg::model::point<double, 2, bg::cs::cartesian> >();
+    test_sph< bg::model::point<double, 2, bg::cs::spherical_equatorial<bg::degree> > >();
+    test_geo< bg::model::point<double, 2, bg::cs::geographic<bg::degree> > >();
```
barendgehrels, line 59: I like it.

barendgehrels (approved):
Thanks a lot @awulkiew , looks really good.
I had only a few really minor remarks.

include/boost/geometry/algorithms/azimuth.hpp
```diff
@@ -0,0 +1,226 @@
...
+struct azimuth : not_implemented<Tag1, Tag2>
+{};
+
+template <typename Point1, typename Point2>
+struct azimuth<Point1, Point2, point_tag, point_tag>
+{
+    template <typename Strategy>
+    static auto apply(Point1 const& p1, Point2 const& p2, Strategy const& strategy)
+    {
+        typedef typename decltype(strategy.azimuth())::template result_type
```
vissarion, line 66: you can avoid it by using `auto` below, right?

test/algorithms/azimuth.cpp
vissarion: nice, it could contain a call with the non default formula for azimuth (i.e. andoyer).

vissarion (approved):
Thanks! OK with merging.

include/boost/geometry/algorithms/azimuth.hpp
awulkiew: Not without changes in azimuth strategies. Azimuth strategy `CalculationType`, Point1's coordinate type and Point2's coordinate type can be different types and right now this strategy `apply()` requires all arguments of the same type.

It would be doable if azimuth strategy `apply()` allowed to pass arguments of arbitrary type, was doing the conversion to the calculation type inside and returning the result from the function instead of using output argument for that. Then there would be a problem with the overloads of `apply()` allowing to calculate both forward and reverse azimuths at the same time. These functions would have to have different name than `apply()` which would be ok I guess since there already is `apply_reverse()`.

So yes, it could be avoided but would require more changes. I think this is a good idea in general to remove all of these utilities like `result_type`, `state_type`, etc. and replace them with a mechanism using `auto`. But this would mean changes in the interfaces of strategies. If we agreed to do this I'd rather do this in a different PR.

vissarion: sure, better doing this in separate PR

include/boost/geometry/algorithms/azimuth.hpp
awulkiew: Done

doc/src/examples/algorithms/azimuth.cpp
awulkiew: If `using` is prefered we should probably do this consistently in all examples at once.

include/boost/geometry/algorithms/azimuth.hpp
awulkiew: Yes, the same could be done in `distance()` and `area()`!

test/algorithms/azimuth.cpp
awulkiew: Done, I added cases with vincenty.

include/boost/geometry/strategies/spherical.hpp
awulkiew: Done.

awulkiew:
Thanks for the reviews!
I addressed all of the raised issues and am going to merge now.

PR #798, opened 2021-01-20
Fix MSVC warning C4267: '=': conversion from 'size_t' to 'int'

mloskot:
Reported via ML https://lists.boost.org/geometry/2021/01/3829.php


include/boost/geometry/extensions/algorithms/detail/overlay/dissolver.hpp
```diff
@@ -433,7 +433,7 @@ struct dissolver_generic
 
         // There are less then 16 elements, handle them quadraticly
 
-        int n = boost::size(output_collection);
+        std::size_t n = boost::size(output_collection);
```
barendgehrels, line 436: Thanks

barendgehrels (approved):

PR #800, opened 2021-01-27
[overlay] move segment_count_on_ring and segment_distance to get_ring, and add unit tests

barendgehrels:
This is a preparation step for next PR, which will be larger.
Because I need `segment_distance` for an update in turn clusters too.

This smaller PR is just a move, just refactoring.


test/count_set.hpp
```diff
@@ -54,7 +52,7 @@ struct count_set
     friend std::ostream &operator<<(std::ostream &os, const count_set& s)
     {
        os << "{";
-       BOOST_FOREACH(std::size_t const& value, s.m_values)
+       for(std::size_t const& value : s.m_values)
```
barendgehrels, line 55: This is non related but I just got it in too

barendgehrels:
Are you OK with this PR?

test/count_set.hpp
vissarion: great! But there is a space missing after `for`

test/count_set.hpp
```diff
@@ -84,7 +82,7 @@ private :
         else if (a.size() > 1 && b.size() == 1)
         {
             // One of them is optional, add the second
-            BOOST_FOREACH(std::size_t const& value, a)
+            for(std::size_t const& value : a)
```
vissarion, line 85: here too

vissarion (approved):
thanks! OK for merging!

awulkiew (approved):

test/count_set.hpp
barendgehrels: :heavy_check_mark: thanks, fixed

test/count_set.hpp
barendgehrels: :heavy_check_mark:

Issue #799, opened 2021-01-21
bg::area underflow with geographic coordinates


CJxD:
`bg::area()` with a geographic polygon yields unexpected results when the area of the polygon is less than ~20km² when using `double` for the `point` precision.

In this example, a 3m radius octagon is created with `double` precision points. The expected area is about 26m², but instead we get -12,375m².

This goes against expectations because `double` is able to define GPS coordinates to around 1mm of precision.

When `precision_t` is set to `long double`, this example works. However, `long double` is not available on some architectures, and I discovered this bug after strange behaviour when running the code on both arm32 and arm64 smartphones.

My best guess is that there is an avoidable numerical underflow somewhere in `boost/geometry/formulas/area_formulas.hpp:418-547`.

```
#include <cmath>
#include <iostream>

#include <boost/geometry/geometry.hpp>

#define EARTH_CIRCUMFERENCE 40075017
#define DEG_LAT_METRES 111320

namespace bg = boost::geometry;

using precision_t = double; // Works with `long double`, which not available on some architectures (notably AArch64)
using point_t = bg::model::point<precision_t, 2, bg::cs::geographic<bg::degree>>;
using polygon_t = bg::model::polygon<point_t, false, true>;

template <typename T, std::enable_if_t<std::is_floating_point<T>::value, void*> = nullptr>
inline T deg2rad(T degrees) {
    return degrees * static_cast<T>(M_PI) / static_cast<T>(180);
}

template <typename T, std::enable_if_t<std::is_floating_point<T>::value, void*> = nullptr>
inline T metresToDegreesLatitude(T metres) {
    return metres / static_cast<T>(DEG_LAT_METRES);
}

template <typename T, std::enable_if_t<std::is_floating_point<T>::value, void*> = nullptr>
inline T metresToDegreesLongitude(T metres, T latitude) {
    return metres * static_cast<T>(360) /
        (static_cast<T>(EARTH_CIRCUMFERENCE) * std::cos(deg2rad(latitude)));
}

int main() {
    precision_t la = 51.47;
    precision_t lo = -0.02;
    precision_t radius = 3.0;

    precision_t dLat = metresToDegreesLatitude(radius);
    precision_t dLon = metresToDegreesLongitude(radius, la);
    constexpr precision_t invrt2 = M_SQRT1_2;

    polygon_t poly {
            {
                { lo, la + dLat },
                { lo - dLon * invrt2, la + dLat * invrt2 },
                { lo - dLon, la },
[...]

PR #801, opened 2021-02-01
Improve accuracy in geographic area computations

vissarion:
This PR proposes a fix for https://github.com/boostorg/geometry/issues/799 

In short this PR

- enables the series expansions formulas and make them available for area computations
- changes the spherical part of the geogrpahic area computation
- adds an error threshold in the geographic area computation and when the ellipsoidal part is far from the spherical part (indicates a possible inaccuracy in azimuth computation) discards the ellipsoidal part. This is a heuristic that improves the accuracy in many cases.
- adds tests


include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -442,58 +447,52 @@ class area_formulas
...
+        CT const cos_omg12 =  cos_omg1 * cos_omg2 + sin_omg1 * sin_omg2;
         CT excess;
 
         bool meridian = get<0>(p2) - get<0>(p1) == CT(0)
               || get<1>(p1) == CT(90) || get<1>(p1) == -CT(90)
               || get<1>(p2) == CT(90) || get<1>(p2) == -CT(90);
 
-        if (!meridian && cos_omg12 > -CT(0.7)
-                      && sin_bet2 - sin_bet1 < CT(1.75)) // short segment
+        if (!meridian && cos_omg12 > -CT(0.7071)
```
CJxD, line 474: Is this supposed to be 1/sqrt(2)? Perhaps use a constant for this?

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -503,16 +502,17 @@ class area_formulas
...
         CT coeffs[SeriesOrder + 1];
-        const std::size_t coeffs_var_size = (series_order_plus_two
-                                            * series_order_plus_one) / 2;
+        std::size_t const coeffs_var_size = (series_order_plus_two
+            * series_order_plus_one) / 2;
+
         CT coeffs_var[coeffs_var_size];
 
-        if(ExpandEpsN){ // expand by eps and n
+        if (ExpandEpsN){ // expand by eps and n
```
barendgehrels, line 510: Minor - since you corrected the format, the `{` should be changed too

include/boost/geometry/strategy/geographic/area.hpp
```diff
@@ -127,8 +127,14 @@ protected :
         {
             return_type result;
 
-            return_type sum = spheroid_const.m_c2 * m_excess_sum
-                   + spheroid_const.m_e2 * spheroid_const.m_a2 * m_correction_sum;
+            return_type spherical_term = spheroid_const.m_c2 * m_excess_sum;
+            return_type ellipsoidal_term = spheroid_const.m_e2
+                * spheroid_const.m_a2 * m_correction_sum;
```
barendgehrels, line 132: Maybe you can make them `const`

test/algorithms/area/area_geo.cpp
```diff
@@ -81,20 +81,20 @@ void test_geo_strategies()
...
+    BOOST_CHECK_CLOSE(area, 63316423410597.016, err);
     area = bg::area(geometry_geo, geographic_andoyer3);
-    BOOST_CHECK_CLOSE(area, 63316309224411.195, err);
+    BOOST_CHECK_CLOSE(area, 63316423410701.703, err);
     area = bg::area(geometry_geo, geographic_andoyer4);
-    BOOST_CHECK_CLOSE(area, 63316309224411.094, err);
+    BOOST_CHECK_CLOSE(area, 63316423410701.602, err);
     area = bg::area(geometry_geo, geographic_andoyer5);
-    BOOST_CHECK_CLOSE(area, 63316309224411.094, err);
+    BOOST_CHECK_CLOSE(area, 63316423410701.602, err);
```
barendgehrels, line 97: Looks cool.
I'm just wondering, how do we know it's better?

test/algorithms/distance/distance_geo_linear_box.cpp
```diff
@@ -564,4 +564,5 @@ BOOST_AUTO_TEST_CASE( test_all_linear_areal )
     test_all_l_ar<geo_point>(vincenty_pp(), vincenty_ps(), vincenty_sb());
     test_all_l_ar<geo_point>(thomas_pp(), thomas_ps(), thomas_sb());
     test_all_l_ar<geo_point>(andoyer_pp(), andoyer_ps(), andoyer_sb());
+    //test_all_l_ar<geo_point>(karney_pp(), karney_ps(), karney_sb());
```
barendgehrels, line 567: Why are these added and commented? Still to do?

barendgehrels (approved):
Thanks, in general I'm OK, few questions

include/boost/geometry/formulas/area_formulas.hpp
vissarion: You are right, this is to select between trapezoid formula and azimuth subtraction (i.e. alp2 - alp1) for spherical excess computation. 

With trapezoid formula we have the following relative errors (using the dataset from https://zenodo.org/record/32156#.WOyaMrUmv7C

maximum legnth | 2 | 10 | 100 | 1000 | 2000 | 1000000 | 2000000 | 5000000 | 7500000 | 10000000 | 12500000 | 15000000 | 17500000 | 19500000 | 20000000
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
andoyer | 0.002051124711 | 5.00E-06 | 1.80E-07 | 6.16E-06 | 1.30E-09 | 1.37E-05 | 5.49E-05 | 0.0003688702402 | 0.0009236612365 | 0.001903669355 | 0.003755516045 | 0.007758503031 | 0.01998210139 | 0.1029663252 | 0.4328622397
thomas | 0.004475880816 | 2.88E-10 | 3.86E-09 | 3.85E-10 | 1.61E-10 | 1.37E-05 | 5.49E-05 | 0.0003688700554 | 0.0009236610226 | 0.00190366909 | 0.003755515698 | 0.007758502504 | 0.0199821028 | 0.102966375 | 0.4329122004
vincenty | 0.004729452869 | 4.34E-06 | 1.63E-07 | 1.53E-07 | 1.52E-07 | 1.36E-05 | 5.49E-05 | 0.0003689026972 | 0.0009236909016 | 0.001903649421 | 0.003755454646 | 0.007757974129 | 0.0199798291 | 0.1029193871 | 0.4322189917

While with azimuth subtraction (i.e. alp2 - alp1)

maximum legnth | 2 | 10 | 100 | 1000 | 2000 | 1000000 | 2000000 | 5000000 | 7500000 | 10000000 | 12500000 | 15000000 | 17500000 | 19500000 | 20000000
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
andoyer | 0.9997504932 | 0.001255017513 | 7.88E-05 | 0.00214284374 | 9.84E-07 | 2.65E-10 | 5.72E-09 | 1.49E-09 | thomas | 1.09E-08 | 5.73E-09 | 4.34E-11 | 1.50E-10 | 3.42E-11 | 1.60E-10 | 0.06630115469
vincenty| 1 | 1.69E-07 | 6.84E-06 | 8.66E-06 | 5.48E-08 | 3.78E-08 | 3.83E-08 | 4.13E-08 | 4.77E-08 | 5.89E-08 | 9.15E-08 | 3.70E-07 | 3.22E-06 | 0.0002948330798 | 0.214268049
1.689545791 | 0.001697061092 | 9.38E-05 | 8.60E-06 | 1.59E-07 | 1.67E-07 | 3.23E-07 | 2.04E-06 | 5.61E-06 | 1.37E-05 | 3.45E-05 | 0.0001037443284 | 0.0005077352782 | 0.01102688439 | 0.1776476855

So we want to keep the best of both worlds which is not very simple by using the trigonometric thresholds and since we have already computed the azimuths in a previous step of the computation, we can compute the distance too and use it as a threshold. Thus, we have to select the former formula for segments with length < 2000 and the latter for the rest. In lack of data between 2000 and 1000000 segment lengths I am choosing 10000 as a threshold.

include/boost/geometry/strategy/geographic/area.hpp
vissarion: sure

test/algorithms/area/area_geo.cpp
vissarion: it is with respect to geographiclib that gives `63316536351834.289`

I added also comparisons with geographiclib in `area_sph_geo.cpp` tests showing that BG after this PR is more accurate than before.

test/algorithms/distance/distance_geo_linear_box.cpp
vissarion: Yes, actually only `karney_pp()` strategy is working after activating the series formulas. This should be fixed in another PR.

vissarion:
@barendgehrels @CJxD thanks for your comments.

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -407,93 +409,107 @@ class area_formulas
...
-        typedef Inverse<CT, false, true, true, false, false> inverse_type;
-        typedef typename inverse_type::result_type inverse_result;
+        // Azimuth Approximation
 
-        inverse_result i_res = inverse_type::apply(get_as_radian<0>(p1),
-                                                   get_as_radian<1>(p1),
-                                                   get_as_radian<0>(p2),
-                                                   get_as_radian<1>(p2),
-                                                   spheroid_const.m_spheroid);
+        auto const i_res = FormulaPolicy::template inverse
```
awulkiew, line 431: FormulaPolicy is strategies-related. It was implemented in order to allow t pick formula in strategies. They are implemented in strategies directory. I think bare formula should be used here.

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -407,93 +409,107 @@ class area_formulas
...
-              || get<1>(p1) == CT(90) || get<1>(p1) == -CT(90)
-              || get<1>(p2) == CT(90) || get<1>(p2) == -CT(90);
+        auto const half_pi = math::pi<CT>() / 2;
+        bool meridian = lon2r - lon1r == CT(0)
+            || lat1r == half_pi || lat1r == -half_pi
+            || lat2r == half_pi || lat2r == -half_pi;
 
-        if (!meridian && cos_omg12 > -CT(0.7)
-                      && sin_bet2 - sin_bet1 < CT(1.75)) // short segment
+        if (!meridian && (i_res.distance) < 10000)  // short segment
```
awulkiew, line 481: Two problems with it:
- it's looks like an arbitrary number based on an (informed) guess but instead should probably be empirically found by experimenting
- it should at least depend on the size of the globe, it is possible to create a spheroid using km instead of meters for axes, in such case this number should probably be `10`, not to mention that the flattening is probably also a factor

include/boost/geometry/formulas/karney_inverse.hpp
```diff
@@ -465,12 +463,12 @@ class karney_inverse
         {
             if (BOOST_GEOMETRY_CONDITION(CalcFwdAzimuth))
             {
-                result.azimuth = atan2(sin_alpha1, cos_alpha1) * r2d;
+                result.azimuth = atan2(sin_alpha1, cos_alpha1);
```
awulkiew, line 466: Was this a bug? A leftover from the previous version returning degrees?

include/boost/geometry/strategies/geographic/parameters.hpp
```diff
@@ -178,7 +178,7 @@ struct karney
             >
     {};
 };
-*/
+
```
awulkiew, line 181: Ok, so this also enables karney formula policy.

include/boost/geometry/strategy/geographic/area.hpp
```diff
@@ -127,8 +144,14 @@ protected :
...
 
-            return_type sum = spheroid_const.m_c2 * m_excess_sum
-                   + spheroid_const.m_e2 * spheroid_const.m_a2 * m_correction_sum;
+            return_type const spherical_term = spheroid_const.m_c2 * m_excess_sum;
+            return_type const ellipsoidal_term = spheroid_const.m_e2
+                * spheroid_const.m_a2 * m_correction_sum;
+
+            // ignore ellipsoidal term if is large (probably from an azimuth
+            // inaccuracy)
+            return_type sum = math::abs(ellipsoidal_term/spherical_term) > 0.01
```
awulkiew, line 153: Here is another magic number. If it's from azimuth inaccuracy shouldn't it be based on formula?

include/boost/geometry/strategy/geographic/area.hpp
```diff
@@ -80,6 +80,9 @@ class geographic
 protected :
     struct spheroid_constants
     {
+        // Area formula is implemented for a maximum series order 5
+        static constexpr auto SeriesOrderNorm = SeriesOrder > 5 ? 5 : SeriesOrder;
```
awulkiew, line 84: Why not define it in the class scope together with other static const variables?

include/boost/geometry/strategy/geographic/area.hpp
```diff
@@ -173,26 +196,35 @@ public :
                       PointOfSegment const& p2,
                       state<Geometry>& st) const
     {
+        using CT = typename result_type<Geometry>::type;
+
+        // if the segment in not on a meridian
         if (! geometry::math::equals(get<0>(p1), get<0>(p2)))
         {
+            // Area formula is implemented for a maximum series order 5
+            constexpr auto SeriesOrderNorm = SeriesOrder > 5 ? 5 : SeriesOrder;
```
awulkiew, line 205: This would not be needed if it was defined as static const class member.

awulkiew (approved):

include/boost/geometry/formulas/area_formulas.hpp
vissarion: > it's looks like an arbitrary number based on an (informed) guess but instead should probably be empirically found by experimenting

see https://github.com/boostorg/geometry/pull/801#discussion_r570871224 above, the experiments are with WGS84, of course one could experiment more, but for the purpose of this PR I think is enough

> it should at least depend on the size of the globe, it is possible to create a spheroid using km instead of meters for axes, in such case this number should probably be 10, not to mention that the flattening is probably also a factor

right, I replaced `10000` with `mean_radius<CT>(spheroid_const.m_spheroid) / CT(638)`

include/boost/geometry/formulas/area_formulas.hpp
vissarion: I am aware that `FormulaPolicy` was implemented to pick formulas in strategies and we are in formulas here, but I found it handy since base formula cannot be passed here since karney inverse formula has a different templated type i.e. 
`inverse<typename CT, bool, bool, bool, bool, bool, size_t >` while the other inverse formulas have `inverse<typename CT, bool, bool, bool, bool, bool>`

include/boost/geometry/formulas/karney_inverse.hpp
vissarion: Probably the second.

include/boost/geometry/strategy/geographic/area.hpp
vissarion: It is independent from the choice of formula, i.e. for an accurate formula this condition should always be false. It describes  how  tolerate we are with errors in inaccurate formulas.

include/boost/geometry/strategy/geographic/area.hpp
vissarion: thanks I overlooked it!

include/boost/geometry/formulas/area_formulas.hpp
awulkiew: If you instantiate it here using FormulaPolicy the default `SeriesOrder = 8` of karney's formula is used anyway. The interface of formulas and policies are the same, you still have to pass the same parameters because other formulas does no take `SeriesOrder`.

So this is not about passing `SeriesOrder` but only about the template parameter.

So why not simply have a `detail` version allowing to pass `SeriesOrder` for testing purposes and then the "official" version implementing the same interface as all other formulas?

Btw, there is also the `SeriesOrder` passed into the strategy by the user. Are they corresponding? If not shouldn't we simply remove them and always use our defaults?

vissarion: > So why not simply have a detail version allowing to pass SeriesOrder for testing purposes and then the "official" version implementing the same interface as all other formulas?

This is a good idea. I implemented a `detail` version of `karney_inverse`. 

I kept the interface of area strategy the same. `SeriesOrder` in area strategy are unrelated to series order in inverse formulas. I added some examples in area tests to show this.

Issue #788, opened 2020-12-23
Add test for extensions/test/algorithms/parse.hpp


mloskot:
It would be helpful to have a test covering features in the [extensions/test/algorithms/parse.hpp](https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/extensions/algorithms/parse.hpp) extension, see https://github.com/boostorg/geometry/pull/786

GauravSarkar:
@mloskot  if you can give a little bit more details on this issue, I was willing to take up the issue.

GauravSarkar:
As far as I have seen there is no parse.hpp under  ```extensions/test/algorithms```. So are you asking to create it and add test results of features as per #786 ?

mloskot:
The file is there, see the link in the issue description that I've just updated.

I don't really have anything to add.
It's a typical procedure: read the code, learn what's the public interface, add test cases and i exercise functions of the public interface.

GauravSarkar:
Thanks @mloskot  I got it. I  will go through the code

ayushgupta138:
@mloskot  could you please provide some more details regarding the issue and about the test cases that needs to added. I am willing to work on this issue, since there hasn't been a pull request lately.

mloskot:
@ayushgupta138 See my earlier comment.

RajWorking:
You want us to create X?
include/boost/geometry/extensions/algorithms/distance_info.hpp : extensions/test/algorithms/distance_info.cpp
include/boost/geometry/extensions/algorithms/parse.hpp : X
X is extensions/test/algorithms/parse.cpp, am I correct?

mloskot:
I think, you've decoded the file structure quite right.

ayushgupta138:
(Assuming parse.hpp is included)
`typedef bg::model::point<double, 2, bg::cs::spherical_equatorial<radian>> earth;`
`earth ear;`
`parse<earth>(ear, "17.65R N", "23.2R W");`
`cout << get<0>(ear) << " " << get<1>(ear) << "\n";`
The above code snippet outputs `-23.2 17.65`
but if space is added between the value and 'R' 
`parse<earth>(ear, "17.65 R N", "23.2 R W");`
the values 17.65 and 23.2 are assumed as degree values and converted to radian outputting  
` -0.404916  0.308051`
Since N/E/W/S are working properly even if spaces are added,
@mloskot  is this a bug or is it functioning the way it should?

ayushgupta138:
[...]

PR #802, opened 2021-02-05
[test] Add tests for /extensions/algorithms/parse.hpp

jkerkela:
For https://github.com/boostorg/geometry/issues/788
@mloskot

PR #806, opened 2021-02-12
Add unit test for parse.hpp

ayushgupta138:
Unit test for extensions/test/algorithms/parse.hpp for the issue #788

### DIFF
--- a/extensions/test/algorithms/Jamfile
+++ b/extensions/test/algorithms/Jamfile
@@ -12,7 +12,6 @@ test-suite boost-geometry-extensions-algorithms
     :
 #    [ run dissolve.cpp ]
     [ run distance_info.cpp ]
-    [ run parse.cpp]
     [ run connect.cpp ]
 #    [ run offset.cpp ]
     [ run parse.cpp ]
--- a/extensions/test/algorithms/Jamfile
+++ b/extensions/test/algorithms/Jamfile
@@ -15,6 +15,7 @@ test-suite boost-geometry-extensions-algorithms
     [ run parse.cpp]
     [ run connect.cpp ]
 #    [ run offset.cpp ]
+    [ run parse.cpp ]
     [ run midpoints.cpp ]
 #    [ run selected.cpp ]
     ;
--- a/extensions/test/algorithms/Jamfile
+++ b/extensions/test/algorithms/Jamfile
@@ -12,6 +12,7 @@ test-suite boost-geometry-extensions-algorithms
     :
 #    [ run dissolve.cpp ]
     [ run distance_info.cpp ]
+    [ run parse.cpp]
     [ run connect.cpp ]
 #    [ run offset.cpp ]
     [ run midpoints.cpp ]
--- a/extensions/test/algorithms/parse.cpp
+++ b/extensions/test/algorithms/parse.cpp
@@ -1,137 +1,219 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 // Unit Test
 
-// Copyright (c) 2007-2013 Barend Gehrels, Amsterdam, the Netherlands.
-// Copyright (c) 2008-2013 Bruno Lalande, Paris, France.
-// Copyright (c) 2009-2013 Mateusz Loskot, London, UK.
-// Copyright (c) 2013 Adam Wulkiewicz, Lodz, Poland.
+// Copyright (c) 2021 Joni Kerkelä, Oulu, Finland.
+// Copyright (c) 2021 Ayush Gupta, Gujarat, India
 
 // Use, modification and distribution is subject to the Boost Software License,
 // Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
 // http://www.boost.org/LICENSE_1_0.txt)
 
-#include<string>
+#include <string>
+#include <geometry_test_common.hpp>
 
-#include<geometry_test_common.hpp>
+#include <boost/geometry/core/coordinate_type.hpp>
+#include <boost/geometry/extensions/algorithms/parse.hpp>
+#include <boost/geometry/geometries/point.hpp>
+#include <boost/geometry/util/math.hpp>
 
-#include<boost/geometry/geometries/geometries.hpp>
-#include<boost/geometry/extensions/algorithms/parse.hpp>
+#include <test_common/test_point.hpp>
 
+using namespace boost::geometry;
 
-template<typename T, typename Geometry>
-void check_parse(Geometry point, T expected_latitude, T expected_longitude)
-{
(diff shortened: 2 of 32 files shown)


ayushgupta138:
@mloskot  @vissarion  @awulkiew  could you please review the pr and merge the changes.

mloskot:
@ayushgupta138 I have just merged PR #802 with initial (first) submission of the tests by @jkerkela 

Next, I'd like to combine your tests, which also cover the parser from different angle, into those tests.
Would you be interested to do it?

ayushgupta138:
Yes , I am interested in it. Let me know what you need from my side.

mloskot:
1. Update your `unit_test_parse` branch with the latest changes from `develop`.
2. Resolve any conflicts - you will get some, obviously, as `extensions/test/algorithms/parse.cpp` is a new file added by both, #802 and #806
3. Copy/Add your test cases to the `extensions/test/algorithms/parse.cpp`
4. `git push --force` to update this PR

ayushgupta138:
On it.

ayushgupta138:
As mentioned by you, I made commits on this pr.

extensions/test/algorithms/Jamfile
```diff
@@ -12,6 +12,7 @@ test-suite boost-geometry-extensions-algorithms
     :
 #    [ run dissolve.cpp ]
     [ run distance_info.cpp ]
+    [ run parse.cpp]
```
mloskot, line 15: This is no longer necessary since parse.cpp has already been added, in #802

extensions/test/algorithms/parse.cpp
```diff
@@ -2,36 +2,38 @@
...
-	BOOST_CHECK_CLOSE(((double)bg::get<1>(geo_point_result)), expected_result_coord_2, 1.0e-6);
-
-	P p;
-	parse(p, coordinate_1, coordinate_2);
-	BOOST_CHECK_CLOSE(((double)bg::get<0>(p)), expected_result_coord_1, 1.0e-6);
-	BOOST_CHECK_CLOSE(((double)bg::get<1>(p)), expected_result_coord_2, 1.0e-6);
+	Geometry geometry;
+	geometry = parse<Geometry>(coordinate_1, coordinate_2);
+	BOOST_CHECK_CLOSE(((double)bg::get<0>(geometry)), expected_result_coord_1, 1.0e-4);
+	BOOST_CHECK_CLOSE(((double)bg::get<1>(geometry)), expected_result_coord_2, 1.0e-4);
```
mloskot, line 32: Such change from `1.0e-6` to `1.0e-4`  needs a word of comment, ideally, in commit message

extensions/test/algorithms/parse.cpp
```diff
@@ -104,6 +106,26 @@ void test_all_complex_dms_strategy()
 
 	test_parse_with_point_and_parse_strategy<P, parser_t>(std::string("5.15R E"), std::string("5.15R N"),
 		expected_dms_result9, expected_dms_result10, parser_t());
+
+	// test with random characters in strategy(for generality check)
+	typedef bg::projections::detail::dms_parser<coord_t, true, 'A', 'B', 'C', 'D', 'm', 's', 'o', 'p'> parser_t1;
```
mloskot, line 111: Since we moved to C++14, we should prefer
```
using parser_t1 = bg::projections::detail::dms_parser<coord_t, true, 'A', 'B', 'C', 'D', 'm', 's', 'o', 'p'>;
```

which places the alias name to the left, it's easier to read.

extensions/test/algorithms/parse.cpp
```diff
@@ -140,32 +162,49 @@ void test_all_with_simple_dms_strategy()
...
-		std::string("2S"), -1 * bg::math::d2r<double>(), -2 * bg::math::d2r<double>());
-	test_parse<bg::model::point<double, 2, bg::cs::spherical<bg::radian> > >(std::string("1W"),
-		std::string("2S"), -1 * bg::math::d2r<double>(), -2 * bg::math::d2r<double>());
-	test_parse<bg::model::point<double, 2, bg::cs::spherical_equatorial<bg::radian> > >(std::string("1W"),
-		std::string("2S"), -1 * bg::math::d2r<double>(), -2 * bg::math::d2r<double>());
-	test_parse<bg::model::point<double, 2, bg::cs::undefined> >(std::string("1W"),
-		std::string("2S"), -1 * bg::math::d2r<double>(), -2 * bg::math::d2r<double>());
-	test_parse<bg::model::point<double, 2, bg::cs::cartesian> >(std::string("1W"),
-		std::string("2S"), -1 * bg::math::d2r<double>(), -2 * bg::math::d2r<double>());
+	typedef typename bg::coordinate_type<Geometry>::type T;
```
mloskot, line 171: `using`

mloskot (changes requested):

mloskot:
@ayushgupta138 Please hit the "Resolve conversation" button for the comment threads that you have addressed.

extensions/test/algorithms/Jamfile
ayushgupta138: removed in new commit

extensions/test/algorithms/parse.cpp
ayushgupta138: mentioned now

ayushgupta138:
Actually I am new to contributing in orgs based settings. It will take some time to get used to all the protocols involved.

extensions/test/algorithms/parse.cpp
mloskot: Yes, the commit https://github.com/boostorg/geometry/pull/806/commits/138d3d53e66a65dfd12be0af7815bc9493ed49fc  does mention it:

```
and reduce the error range from 1e-6 to 1e-4
```

but this is just *repeating **what** the commit does change* (i.e. it is equivalent to what everyone can see in the diff :-)), but it does not explain **why** the range changed. The **why** is what the commit messages are for. The **what** everyone can see by looking at the diff of a commit.

So, I'm hitting "Unresolve conversation" as this still needs clarification.

ayushgupta138: The range changed because I have also included test cases where the final value is very close to zero, so even though the difference between the actual value and expected value is of order 1e-9 , the percentage error exceed the previous error range of 1e-6 and had to be reduced to 1e-4. This was the case when 10 precision points were specified in the expected value. Is this explanation satisfying enough? If so I will add the explanation in the commit message after your confirmation.

mloskot: Yes, it is indeed.

mloskot (approved):

mloskot:
@ayushgupta138 Thank you

Issue #803, opened 2021-02-05
bg::buffer() returns wrong result


Mitsuhiko-Matsukawa:
In this code, bg::buffer() returns empty result.
However if I change 6.6613381477509392e-16 to 0.0, it returns correct result.

boost1.75.0, Visual Studio 2019

```cpp
#include <boost/geometry.hpp>
#include <boost/geometry/geometries/point_xy.hpp>
#include <boost/geometry/geometries/geometries.hpp>

int main()
{
    typedef double coordinate_type;
    typedef boost::geometry::model::d2::point_xy<coordinate_type> point;
    typedef boost::geometry::model::polygon<point> polygon;

    // Declare strategies
    const double buffer_distance = 10.0;
    const int points_per_circle = 36;
    boost::geometry::strategy::buffer::distance_symmetric<coordinate_type> distance_strategy(buffer_distance);
    boost::geometry::strategy::buffer::join_miter join_strategy(points_per_circle);
    boost::geometry::strategy::buffer::end_round end_strategy(points_per_circle);
    boost::geometry::strategy::buffer::point_circle circle_strategy(points_per_circle);
    boost::geometry::strategy::buffer::side_straight side_strategy;

    // Declare output
    boost::geometry::model::multi_polygon<polygon> result;

    // Declare/fill a linestring
    boost::geometry::model::linestring<point> ls = {
        { 2773.6899360413681, -17.493356405074767 },
        { 2767.9084041267429, -9.0288291122330797 },
        { 2765.7403296587586, -1.6929054585683376 },
#if 1   // assertion failed (empty results)
        { 2765.3273630934282, 6.6613381477509392e-16 },
#else   // assertion successful
        { 2765.3273630934282, 0.0 },
#endif
        { 2762.8495637014462, 10.157432751410026 },
        { 2759.9587977441333, 24.829280059133669 },
        { 2756.3453402974928, 38.372523727680374 },
        { 2752.0091913615238, 45.708447381345117 }
    };

    // Create the buffer of a linestring
    boost::geometry::buffer(ls, result,
        distance_strategy, side_strategy,
        join_strategy, end_strategy, circle_strategy);

    assert(result.size() == 1);

    return 0;
}
```

barendgehrels:
Hi @Mitsuhiko-Matsukawa , thanks for the report. I can reproduce it and will at it to our testsuite.

[...]

PR #815, opened 2021-02-19
[buffer] fix case for issue 803 

barendgehrels:
Issue #803 had an artifact, generated on nearly collinear lines in the buffer

The intersection point (calculated from the green and the blue line) was quite off (magenta) as you can see below.
The artifact itself is also visible below.

Now I added an option to calculate the intersection point from either those lines, or from the angle bisector between the two perpendicular lines, so from the green line and the center between the two orange lines (here they're nearly collinear)

ip_wrong

buffer

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
@@ -97,6 +97,7 @@ struct buffer_range
         typename Collection,
         typename Point,
         typename DistanceStrategy,
+        typename SegmentStrategy,
         typename JoinStrategy,
         typename EndStrategy,
         typename RobustPolicy,
@@ -113,6 +114,7 @@ struct buffer_range
             output_point_type const& perp2,
             geometry::strategy::buffer::buffer_side_selector side,
             DistanceStrategy const& distance,
+            SegmentStrategy const& segment_strategy,
             JoinStrategy const& join_strategy,
             EndStrategy const& end_strategy,
             RobustPolicy const& ,
@@ -155,8 +157,10 @@ struct buffer_range
                     // The corner is convex, we create a join
                     // TODO (future) - avoid a separate vector, add the piece directly
                     output_point_type intersection_point;
-                    if (line_line_intersection::apply(perp1, perp2,
-                                    prev_perp1, prev_perp2, intersection_point))
+                    if (line_line_intersection::apply(prev_perp1, prev_perp2,
+                                                      perp1, perp2, previous_input,
+                                                      segment_strategy.equidistant(),
+                                                      intersection_point))
                     {
                         std::vector<output_point_type> range_out;
                         if (join_strategy.apply(intersection_point,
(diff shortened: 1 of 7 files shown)


include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
```diff
@@ -202,28 +206,28 @@ struct buffer_range
         typename Collection,
         typename Iterator,
         typename DistanceStrategy,
-        typename SideStrategy,
+        typename SegmentStrategy,
```
barendgehrels, line 209: This is slightly related (because I had to pass it too), and fix the issue that we had two side-strategies (one real, and one to generate the sides along buffer segments). The last one is now renamed to `SegmentStrategy` in this code.

include/boost/geometry/algorithms/detail/buffer/line_line_intersection.hpp
```diff
@@ -24,23 +24,97 @@ namespace boost { namespace geometry
...
+            auto const denominator_pr = p.b * r.a - p.a * r.b;
+
+            if (math::equals(denominator_pq, zero)
+                && math::equals(denominator_pr, zero))
+            {
+                // Degenerate case (for example when length results in <inf>)
+                return false;
+            }
+
+            ip = std::abs(denominator_pq) > std::abs(denominator_pr)
```
barendgehrels, line 104: This is the fix

include/boost/geometry/algorithms/detail/buffer/line_line_intersection.hpp
```diff
@@ -24,23 +24,97 @@ namespace boost { namespace geometry
...
+        auto const denominator_pq = p.b * q.a - p.a * q.b;
+        constexpr decltype(denominator_pq) const zero = 0;
+
+        if (equidistance)
+        {
+            auto const between = between_point(pj, qi);
+
+            auto const r = detail::make::make_infinite_line<ct>(vertex, between);
+
+            auto const denominator_pr = p.b * r.a - p.a * r.b;
```
vissarion, line 95: why not using the determinant function?

include/boost/geometry/algorithms/detail/buffer/line_line_intersection.hpp
```diff
@@ -24,23 +24,97 @@ namespace boost { namespace geometry
...
+        //
+        // If generated sides along the segments can have an adapted distance,
+        // in a custom strategy, then the calculation of the point in between
+        // might be incorrect and whe optimization is not used.
+
+        using ct = typename coordinate_type<Point>::type;
+
+        auto const p = detail::make::make_infinite_line<ct>(pi, pj);
+        auto const q = detail::make::make_infinite_line<ct>(qi, qj);
+        auto const denominator_pq = p.b * q.a - p.a * q.b;
```
vissarion, line 86: why not using the determinant function?

vissarion (approved):
I am ok with merging.

include/boost/geometry/algorithms/detail/buffer/line_line_intersection.hpp
barendgehrels: > why not using the determinant function?

:+1: I'll change it.
It was not in the code where I adapted it from - but I can change it there too

include/boost/geometry/algorithms/detail/buffer/line_line_intersection.hpp
awulkiew: So are we using `std::abs` now instead of `math::abs`?

awulkiew (approved):

include/boost/geometry/algorithms/detail/buffer/line_line_intersection.hpp
barendgehrels: Thanks, that was an error. Fixed. :heavy_check_mark:

include/boost/geometry/algorithms/detail/buffer/line_line_intersection.hpp
barendgehrels: :heavy_check_mark: done
Calling it directly looked a bit messy, I now added a determinant variant for the infinite line and that looks good IMO.
Do you agree?

include/boost/geometry/arithmetic/infinite_line_functions.hpp
```diff
@@ -21,25 +23,43 @@ namespace boost { namespace geometry
...
+inline Point assign_intersection_point(Line const& p, Line const& q, Type const& denominator)
+{
+    BOOST_ASSERT(denominator != Type(0));
+
+    // x = | pb pc | / d  and y = | pc pa | / d
+    //     | qb qc |              | qc qa |
+
+    Point result;
+    geometry::set<0>(result, determinant<Line, &Line::b, &Line::c>(p, q) / denominator);
+    geometry::set<1>(result, determinant<Line, &Line::c, &Line::a>(p, q) / denominator);
```
barendgehrels, line 43: This shows how determinant (of infinite lines) is called now

PR #820, opened 2021-02-24
[svg] changes newly added scale function in a constructor parameter

barendgehrels:
This addresses the comments of @awulkiew 

Note that it is not 100% backward compatible: all generated SVG's will now have a default margin of 5%. I'm not 100% sure about that. Maybe we should revert that (having a 100% scale initially). There might be customers having these as test results.


include/boost/geometry/io/svg/svg_mapper.hpp
```diff
@@ -292,37 +292,85 @@ class svg_mapper : boost::noncopyable
...
+        , SvgCoordinateType height
+        , calculation_type scale
+        , std::string const& width_height = "width=\"100%\" height=\"100%\"")
+        : m_stream(stream)
+        , m_width(width)
+        , m_height(height)
+        , m_scale(scale)
+    {
+        assign_inverse(m_bounding_box);
+        write_header(width_height);
```
barendgehrels, line 369: The header is now written immediately. Therefore uses can stream, for example arrow-definitions, right after the constructor (in previous version it needed to be done after `scale`).

Also the scale is here now and therefore I had to add a second version of the constructor (because having it after the width_height is inconvenient)

awulkiew (approved):

include/boost/geometry/io/svg/svg_mapper.hpp
awulkiew: Yes, the default `scale = 1` seems to be more intuitive.

barendgehrels: I'll change it

vissarion (approved):
Thanks, I am OK.

include/boost/geometry/io/svg/svg_mapper.hpp
barendgehrels: :heavy_check_mark: changed to `1.0`

PR #824, opened 2021-03-03
[test][area] Tweak error thresholds for small areas.

awulkiew:
After merging https://github.com/boostorg/geometry/pull/801 some of the tests started to fail.
This PR changes error margins for these cases.

I don't know if this is a temporary or permanent solution. @vissarion do you think this is worth investigating more?


vissarion (approved):
It seems minor and the errors acceptable. We could leave it for now. Thanks for the PR!

Issue #825, opened 2021-03-05
Inconsistency in #define directives in geometry/include/boost/geometry/algorithms/area_result.hpp


ayushgupta138:
As per the developer guidelines https://github.com/boostorg/geometry/wiki/Guidelines-for-Developers , 
 Every header shall have `#include` guard based on header path and file name:
`#ifndef BOOST_GEOMETRY_<DIR1>_<DIR2>_<FILE>_HPP`
`#define BOOST_GEOMETRY_<DIR1>_<DIR2>_<FILE>_HPP`
area_result.hpp begins with
https://github.com/boostorg/geometry/blob/6918276567a16d1699f5e27f05d9bcd7c6512a43/include/boost/geometry/algorithms/area_result.hpp#L13
https://github.com/boostorg/geometry/blob/6918276567a16d1699f5e27f05d9bcd7c6512a43/include/boost/geometry/algorithms/area_result.hpp#L14
instead of 
`#ifndef BOOST_GEOMETRY_ALGORITHMS_AREA_RESULT_HPP`
`#define BOOST_GEOMETRY_ALGORITHMS_AREA_RESULT_HPP`

awulkiew:
Good catch!

PR #826, opened 2021-03-05
Fix #define directive convention in area_result.hpp and default_area_result.hpp

ayushgupta138:
For #825


vissarion (approved):
Thanks. Could you please also update the comment of `#endif` at the end of the file ?

awulkiew:
Thanks!

PR #828, opened 2021-03-17
[clusters] detection is refactored to a simpler sweeping algorithm

barendgehrels:
This fixes a part (around 30%) of the remaining errors in the recursive buffer check

There are some (small) changes in expectations, because the previous clustering algorithm did not generate a cluster for some cases (for example in the distance_zero case) while it should have (they are at exactly the same locations).

Apart from fixing several buffer cases, the new algorithm is also much simpler and more maintainable.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/approximately_equals.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/approximately_equals.hpp
@@ -1,0 +1,49 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+
+// Copyright (c) 2021 Barend Gehrels, Amsterdam, the Netherlands.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_ALGORITHMS_DETAIL_OVERLAY_APPROXIMATELY_EQUALS_HPP
+#define BOOST_GEOMETRY_ALGORITHMS_DETAIL_OVERLAY_APPROXIMATELY_EQUALS_HPP
+
+#include <boost/geometry/core/access.hpp>
+#include <boost/geometry/util/math.hpp>
+#include <boost/geometry/util/select_coordinate_type.hpp>
+#include <boost/geometry/util/select_most_precise.hpp>
+
+namespace boost { namespace geometry
+{
+
+#ifndef DOXYGEN_NO_DETAIL
+namespace detail { namespace overlay
+{
+
+template <typename Point1, typename Point2, typename E>
+inline bool approximately_equals(Point1 const& a, Point2 const& b,
+                                 E const& multiplier)
+{
+    using coor_t = typename select_coordinate_type<Point1, Point2>::type;
+    using calc_t = typename geometry::select_most_precise<coor_t, E>::type;
+
+    calc_t const& a0 = geometry::get<0>(a);
+    calc_t const& b0 = geometry::get<0>(b);
+    calc_t const& a1 = geometry::get<1>(a);
+    calc_t const& b1 = geometry::get<1>(b);
+
+    math::detail::equals_factor_policy<calc_t> policy(a0, b0, a1, b1);
+    policy.factor *= multiplier;
+
+    return math::detail::equals_by_policy(a0, b0, policy)
+        && math::detail::equals_by_policy(a1, b1, policy);
+}
+
+}} // namespace detail::overlay
+#endif //DOXYGEN_NO_DETAIL
+
+
+}} // namespace boost::geometry
+
+#endif // BOOST_GEOMETRY_ALGORITHMS_DETAIL_OVERLAY_APPROXIMATELY_EQUALS_HPP
--- a/include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
@@ -1,0 +1,218 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+
+// Copyright (c) 2021 Barend Gehrels, Amsterdam, the Netherlands.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_ALGORITHMS_DETAIL_OVERLAY_GET_CLUSTERS_HPP
+#define BOOST_GEOMETRY_ALGORITHMS_DETAIL_OVERLAY_GET_CLUSTERS_HPP
+
+#include <algorithm>
+#include <map>
+#include <vector>
+
+#include <boost/geometry/core/access.hpp>
+#include <boost/geometry/algorithms/detail/overlay/approximately_equals.hpp>
+#include <boost/geometry/algorithms/detail/overlay/cluster_info.hpp>
+#include <boost/geometry/algorithms/detail/overlay/get_ring.hpp>
+#include <boost/geometry/algorithms/detail/recalculate.hpp>
+#include <boost/geometry/policies/robustness/rescale_policy_tags.hpp>
+#include <boost/range/value_type.hpp>
+#include <boost/geometry/util/math.hpp>
+
+#define BOOST_GEOMETRY_USE_RESCALING_IN_GET_CLUSTERS
+
+
+namespace boost { namespace geometry
+{
+
+#ifndef DOXYGEN_NO_DETAIL
+namespace detail { namespace overlay
+{
+
+template <typename Tag = no_rescale_policy_tag, bool Integral = false>
+struct sweep_equal_policy
+{
+    template <typename P>
+    static inline bool equals(P const& p1, P const& p2)
+    {
+        // Points within a kilo epsilon are considered as equal
+        return approximately_equals(p1, p2, 1000.0);
+    }
+
+    template <typename T>
+    static inline bool exceeds(T value)
+    {
+        // This threshold is an arbitrary value
+        // as long as it is than the used kilo-epsilon
+        return value > 1.0e-3;
+    }
+};
+
+template <>
+struct sweep_equal_policy<no_rescale_policy_tag, true>
+{
+    template <typename P>
+    static inline bool equals(P const& p1, P const& p2)
+    {
+        return geometry::get<0>(p1) == geometry::get<0>(p2)
+            && geometry::get<1>(p1) == geometry::get<1>(p2);
+    }
+
+    template <typename T>
+    static inline bool exceeds(T value)
+    {
+        return value > 0;
+    }
+};
+
+#ifdef BOOST_GEOMETRY_USE_RESCALING_IN_GET_CLUSTERS
+template <>
+struct sweep_equal_policy<rescale_policy_tag, true>
+{
+    template <typename P>
+    static inline bool equals(P const& p1, P const& p2)
+    {
+        // Neighbouring cells in the "integer grid" are considered as equal
+        return math::abs(geometry::get<0>(p1) - geometry::get<0>(p2)) <= 1
+            && math::abs(geometry::get<1>(p1) - geometry::get<1>(p2)) <= 1;
+    }
+
+    template <typename T>
+    static inline bool exceeds(T value)
+    {
+        return value > 1;
+    }
+};
+#endif
+
+template <typename Point>
+struct turn_with_point
+{
+    std::size_t turn_index;
+    Point pnt;
+};
+
+template <typename Cluster, typename Point>
+struct cluster_with_point
+{
+    Cluster cluster;
+    Point pnt;
+};
+
+// Use a sweep algorithm to detect clusters
+template
+<
+    typename Turns,
+    typename Clusters,
+    typename RobustPolicy
+>
+inline void get_clusters(Turns& turns, Clusters& clusters,
+                         RobustPolicy const& robust_policy)
+{
+    using turn_type = typename boost::range_value<Turns>::type;
+    using cluster_type = typename Clusters::mapped_type;
+
+#ifdef BOOST_GEOMETRY_USE_RESCALING_IN_GET_CLUSTERS
+    // For now still use robust points for rescaled, otherwise points do not match
+    using point_type = typename geometry::robust_point_type
+    <
+        typename turn_type::point_type,
+        RobustPolicy
+    >::type;
+#else
+    using point_type = typename turn_type::point_type;
+#endif
+
+    sweep_equal_policy
+        <
+            typename rescale_policy_type<RobustPolicy>::type,
+            std::is_integral<typename coordinate_type<point_type>::type>::value
+        > equal_policy;
+
+    std::vector<turn_with_point<point_type>> points;
+    std::size_t turn_index = 0;
+    for (auto const& turn : turns)
+    {
+        if (! turn.discarded)
+        {
+#ifdef BOOST_GEOMETRY_USE_RESCALING_IN_GET_CLUSTERS
+            point_type pnt;
+            geometry::recalculate(pnt, turn.point, robust_policy);
+            points.push_back({turn_index, pnt});
+#else
+            points.push_back({turn_index, turn.point});
+#endif
+        }
+        turn_index++;
+    }
+
+    // Sort the points from top to bottom
+    std::sort(points.begin(), points.end(), [](auto const& e1, auto const& e2)
+    {
+       return geometry::get<1>(e1.pnt) > geometry::get<1>(e2.pnt);
+    });
+
+    // The output vector will be sorted from bottom too
+    std::vector<cluster_with_point<cluster_type, point_type>> clustered_points;
+
+    // Compare points with each other. Performance is O(n log(n)) because of the sorting.
+    for (auto it1 = points.begin(); it1 != points.end(); ++it1)
+    {
+        // Inner loop, iterates until it exceeds coordinates in y-direction
+        for (auto it2 = it1 + 1; it2 != points.end(); ++it2)
+        {
+            auto const d = geometry::get<1>(it1->pnt) - geometry::get<1>(it2->pnt);
+            if (equal_policy.exceeds(d))
+            {
+                // Points at this y-coordinate or below cannot be equal
+                break;
+            }
+            if (equal_policy.equals(it1->pnt, it2->pnt))
+            {
+                std::size_t cindex = 0;
+
+                // Most recent clusters (with this y-value) are at the bottom
+                // therefore we can stop as soon as the y-value is out of reach (TODO)
+                bool found = false;
+                for (auto cit = clustered_points.begin();
+                     cit != clustered_points.end(); ++cit)
+                {
+                    found = equal_policy.equals(cit->pnt, it1->pnt);
+                    if (found)
+                    {
+                        break;
+                    }
+                    cindex++;
+                }
+
+                // Add new cluster
+                if (! found)
+                {
+                   cindex = clustered_points.size();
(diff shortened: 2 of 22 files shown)


include/boost/geometry/algorithms/detail/overlay/handle_colocations.hpp
```diff
@@ -51,289 +52,6 @@ namespace boost { namespace geometry
 namespace detail { namespace overlay
 {
```
barendgehrels, line 54: Note that complex code below is all gone

include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
```diff
@@ -290,31 +291,6 @@ public :
         add_segment_to(turn_index, op_index, point_to, op);
     }
```
barendgehrels, line 293: moved to separate headerfile (now kilo i/o giga)

include/boost/geometry/algorithms/detail/overlay/approximately_equals.hpp
```diff
@@ -0,0 +1,53 @@
...
+    calc_t const& b1 = geometry::get<1>(b);
+    calc_t const one = 1.0;
+    calc_t const c = math::detail::greatest(std::abs(a0), std::abs(b0),
+                                            std::abs(a1), std::abs(b1), one);
+
+    // The maximum limit is avoid, for floating point, large limits like 400
+    // (which are be calculated using eps)
+    constexpr calc_t maxlimit = 1.0e-3;
+    auto const limit = (std::min)(maxlimit, limit_kilo_epsilon * machine_kilo_epsilon * c);
+    return std::abs(a0 - b0) <= limit && std::abs(a1 - b1) <= limit;
```
awulkiew, line 44: The comment suggests that an alternative for this algorithm is the one using `distance`. `distance` would take all of the dimensions into account. This algorithm takes only 2 dimensions into account. Which one is correct? In general case should this be replaced with a N-dimensional version?

----

In `namespace math` there are:
- `equals_by_policy` - https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/util/math.hpp#L196
- `equals_with_epsilon` - https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/util/math.hpp#L575

Maybe they could be used here instead of calculating all of this manually?
If needed the policy in `equals_by_policy` could be replaced with lambda expression for convenience.

awulkiew: Furthermore, it would be possible to use distance here. You could simply pass Pt/Pt distane strategy or newly proposed umbrella strategy (https://github.com/boostorg/geometry/pull/812) and then use it directly here without including the algorithm.

But have in mind that this computionally simple code would become a lot more complex if distance was calculated. There would be additional `sqrt` call for cartesian and `sin/cos` calls for other coordinate systems.

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
```diff
@@ -0,0 +1,302 @@
...
+{
+
+template <typename T>
+struct sweep_equal_policy
+{
+    template <typename P>
+    static inline bool equals(P const& p1, P const& p2)
+    {
+        // Using machine epsilon results in too much points considered as inequal
+        return approximately_equals(p1, p2, 1.0); // 1000
```
awulkiew, line 38: I think it would be more intuitive if here a multiplier for machine epsilon was passed and there was no requirement on the caller's part to know that internally this number is multiplied by 1000 again before multiplying it by machine epsilon.

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
```diff
@@ -0,0 +1,302 @@
...
+    {
+        // Its value is should be LARGER than the used kilo-epsilon
+        return value > 1.0e-3;
+    }
+
+};
+
+#ifdef BOOST_GEOMETRY_USE_RESCALING_IN_GET_CLUSTERS
+template <>
+struct sweep_equal_policy<boost::long_long_type>
```
awulkiew, line 51: So this is a specialization for the type used for rescaling. What if a user passed a geometry using integral coordinates with rescaling disabled? There are bigger integral types than long long including e.g. MultiPrecision. My point is, shouldn't this be done for all integral coordinates?

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
```diff
@@ -0,0 +1,302 @@
...
+    for (auto it1 = points.begin(); it1 != points.end(); ++it1)
+    {
+        // Inner loop, walks until too far away in y-direction
+        for (auto it2 = it1 + 1; it2 != points.end(); ++it2)
+        {
+            auto const d = geometry::get<1>(it1->pnt) - geometry::get<1>(it2->pnt);
+            if (equal_policy.exceeds(d))
+            {
+                // Too far away in y-direction to be considered as equal
+                break;
```
awulkiew, line 145: Why is y direction special?

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
```diff
@@ -0,0 +1,302 @@
...
+    {
+        if (turn.method != method_start)
+        {
+            for (const auto& op : turn.operations)
+            {
+                auto it = start_turns_per_segment.find(op.seg_id);
+                if (it != start_turns_per_segment.end())
+                {
+                    auto v = it->second;
+                    for (auto i : v)
```
awulkiew, line 281: I understand using `auto` for turns, iterators and operations but here, is it really needed? Furthermore at line 280 a vector is copied. How about this?
```
for (std::size_t i : it->second)
```

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
```diff
@@ -0,0 +1,302 @@
...
+{
+    if (first.source_index == second.source_index
+        && first.ring_index == second.ring_index
+        && first.piece_index == second.piece_index
+        && first.multi_index == second.multi_index)
+    {
+        auto const sd = first.source_index == 0
+                        ? segment_distance(geometry0, first, second)
+                        : segment_distance(geometry1, first, second);
+        return sd == 1;
```
awulkiew, line 217: This code looks confusing to me. The function naming suggests that some kind distance is calculated. Since the geometry is passed and `segment` is in the name I jump to the conclusion that this is some kind of geometric distance. We can't deduce what's happening from the return type neither because `auto` is used. On the other hand the distance is compared with `1` which makes little sense.

I think that explicitly stating the type `signed_size_type` would be a good start. I guess this name refers to the `std::distance()` but if we could change the name to something more descriptive I think this would be beneficial.

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
```diff
@@ -0,0 +1,302 @@
...
+    }
+    return false;
+}
+
+
+template <typename Turn, typename Geometry0, typename Geometry1>
+inline bool corresponding_turn(Turn const& turn, Turn const& start_turn,
```
awulkiew, line 224: Could this function be renamed? It's not clear in what way the turns correspond to each other.

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
```diff
@@ -0,0 +1,302 @@
...
+    // Start turns are now generated, in case the previous turn is missed
+    // But often it is not missed, and then it should still be deleted
+    // This is how it can be (in float, collinear, points far apart due to floating point precision)
+    //   [m, i s:0, v:6 1/1 (1) // u s:1, v:5  pnt (2.54044, 3.12623)]
+    //   [s, i s:0, v:7 0/1 (0) // u s:1, v:5  pnt (2.70711, 3.29289)]
+
+    // 1 Build map (segment->turnids) of start turns
+    std::map<segment_identifier, std::vector<std::size_t>> start_turns_per_segment;
+    std::size_t index = 0;
+    for (auto& turn : turns)
```
awulkiew, line 257: `auto const&` ?

awulkiew:
Thanks! I labeled it as a bugfix but I don't know if that's correct.

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
barendgehrels: Sure, much better. Thanks for your review! I'll address your comments tomorrow.

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
barendgehrels: Yes, it's an arbitrary choice.

Clusters are locations with multiple turns. They are crucial to the traversal algorithm. It should work for either cartesian or non cartesian coordinates.

I don't understand that x-coordinate would be better, in my opinion y-direction is better just because of those antimeridian points.
It's all turn-based, so it's point-based. There are no segments or polygons involved (in the clusters). Of course turns are calculated from lines or polygons intersections, and they may span anything.

Polygons surrounding poles or crossing the antimeridian are not problematic, unless there would be a cluster at (90) or (-90) (one point). I think that's a theoretical problem, it won't happen in practice (unless you make a testcase for it). The chance for a cluster on the -180 meridian is low too, but higher because it's a globe crossing line.

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
barendgehrels: :heavy_check_mark:

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
barendgehrels: I think the name is OK but I understand your question.
Note that the turns with `method_start` are additional. They were not necessary in rescaling, but they are added because sometimes (rare cases) the other turn (`method_touch` or `method_touch_interior`) is missing. Therefore I call it the corresponding turn.

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
barendgehrels: It's the distance in terms of segments (including circular behavior). So two segments are consecutive if that distance is 1.
I actually don't see much confusion here. Of course I can replace auto with `signed_size_type` indeed.

barendgehrels: :heavy_check_mark:

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
barendgehrels: :heavy_check_mark: added description to the function

include/boost/geometry/algorithms/detail/overlay/approximately_equals.hpp
barendgehrels: I moved this function. It's not new in this PR.
But yes, I recently created it and it could be enhanced or replaced.
The problem with the code from math is that it gives too high values for floating point (as indicated in the comment).
I tried to replace it with comparable distance, and that works for cartesian but (again) requires an epsilon defined by the coordinate system, which will result in high values for float and therefore in similar code as is written here.
The 3d variant we could make with your new `for_each_coordinate` functionality.

For now I keep it here but we can discuss further about it.

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
barendgehrels: The comment was left there by mistake

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
barendgehrels: :heavy_check_mark:

barendgehrels:
> Thanks! I labeled it as a bugfix but I don't know if that's correct.

It's fine.

include/boost/geometry/algorithms/detail/overlay/approximately_equals.hpp
barendgehrels: I did a test using `distance`. W.r.t. performance in Cartesian there is no measurable difference in this use case. Probably because it doesn't happen so often.
But using geographical distance will cause problems:
```
false 1.6948370905798346708337476229644
false 0.84474950387190805844994656581548
false 0.41427736018131078310489101568237
false 0.19008348922390885804567517425312
false 0
false 0
```
From the small test, the distance already returns zero if they are still far away (in terms of epsilon). Apart from that it's probably an overkill to use geographic distance there.

So I propose to still keep it as it is here (and using `for_each` later)

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
barendgehrels: No, it's made for the rescaled integral grid.
But indeed, other integral coordinates should return "exactly equal".

barendgehrels:
@awulkiew : will commands marked as "fixup:" be squashed automatically on merge?

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
awulkiew: > Clusters are locations with multiple turns.

So these are points that may be close to each other. These points are either endpoints or intersection points. And the purpose of this scanning is to find them and group close ones into clusters. Correct?

If that's the case then I think indeed this particular algorithm should work in any CS. I guess the coordinates might be normalized but technically valid range is in (-180 -90, 180 90) so ok.

----

I'll explain what I was thinking about regarding vertices, segments, etc. Consider this:

path10

In spherical/geographic CS, points of a segment can lie "above" its endpoints and any detected intersection point on this segment unless the intersection point is at the vertex or vertex of a geodesic is not between the endpoints. So in a hypothetical scenario of looking for segments/rings/etc. by scanning for endpoints or intersection points it is not possible to detect the segment correctly when the algorithm is scanning the latitudes.

But since your goal (AFAIU) is only to group close points together, this is not a problem I think.

awulkiew:
@barendgehrels I don't know. If you want to squash the commits maybe simply use "Squash and merge" instead of "Create a merge commit"?

Alternatively you could manually amend and force push:
```
git reset HEAD^
git add .
git commit --amend
git push -f
```

barendgehrels:
> 
> 
> @barendgehrels I don't know. If you want to squash the commits maybe simply use "Squash and merge" instead of "Create a merge commit"?
> 
> Alternatively you could manually amend and force push:
> 
> ```
> git reset HEAD^
> git add .
> git commit --amend
> git push -f
> ```

OK, sure, I'll squash it before merging. I made a separate commit to show you what comments I processed.

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
barendgehrels: > And the purpose of this scanning is to find them and group close ones into clusters. Correct?

Correct

> I'll explain what I was thinking about regarding vertices, segments, etc.

Thanks! It's good to keep in mind if this algorithm (the sweeping part) is used more often. And yes, I then see what you mean and going in x-direction will be more convenient.

awulkiew: Ok, so the only problem is the normalization/winding because any point with latitude close to -90 or 90 should be considered equal to any other regardless of the longitude. And points with londitude -180 should be considered equal to points with longitude 180. The normalization functions we have would AFAIR set the longitudes to 0 for points at poles and longitude to 180 if points are -180. All of this is done WRT machine epsilon, this is done for points close to poles or antimeridian. So if your epsilon is greater than the machine epsilon then you'd have to normalize coordinates differently or take the CS into consideration by manually checking the closeness in corresponding way.

awulkiew:
> I made a separate commit to show you what comments I processed.

Ok, right.

It seems my fix for compilation error in expectation_limits.hpp is conflicting with your fix. Sorry about that. I'm working on the release.

include/boost/geometry/algorithms/detail/overlay/approximately_equals.hpp
awulkiew: `equals_with_epsilon` and `equals_with_policy` allows you to pass a value which is multiplied by machine epsilon before performing the test. Just like you did.

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
awulkiew: My point was that I propose to pass the value of epsilon multiplier here, so (1000), instead of a multiplier for another multiplier which is unknown unless you actually see the `approximately_equals` function definition.

barendgehrels:
> It seems my fix for compilation error in expectation_limits.hpp is conflicting with your fix. Sorry about that. I'm working on the release.

No problem. Thanks for your work on the release.

include/boost/geometry/algorithms/detail/overlay/approximately_equals.hpp
barendgehrels: I'll have a look again.

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
awulkiew: I started wondering about one more thing. Since this is only for the checks if points are close to each other. Have you considered to do what we usually do when checking points of equality (which is actually closeness WRT machine epsilon), i.e. simply using `std::set` with a custom `less` making sure that close points are considered equal?

barendgehrels: It's a good idea. I'll check it on Wednesday together with possible epsilon changes. Even with the sorting (which indeed should stay), I maybe can still use the set for the clusters.

include/boost/geometry/algorithms/detail/overlay/approximately_equals.hpp
barendgehrels: :heavy_check_mark: thanks for the comments, done in new commit. I added a unit test too.
_(I will squash commits before merging)_

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
barendgehrels: :heavy_check_mark: done

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
barendgehrels: I have had a look. I'm not finished yet but I doubt if it is more convenient. 

Can you point me to the code where we actually really do this? Because with a custom less operator in two dimensions, points approximately considered equal, it's not that trivial. I've some testcode which (still) results in slightly different results and is more complex than what I had written originally.

However, a unit test is added already and that is useful anyway. Plus code is split better. (see last commit)

awulkiew: Ok, not `std::set` but `std::binary_search` but it's the same story: https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp#L80-L83
And I agree that this is technically incorrect because for 3 points: with coord1 = 0, coord2 = eps and coord3 = 2eps. The first one is equal to the second, the second to the third but the third is not equal to the first one. Which is inconsistent and I guess might cause assertion failure in some STL implementations.

awulkiew (approved):

include/boost/geometry/algorithms/detail/overlay/get_clusters.hpp
barendgehrels: OK, thanks. Then I keep it as it was.

Thanks for the review comments and the approval.

barendgehrels:
@vissarion I will squash, clean up a bit (there are some debug comments in the new unit test) and merge next Wednesday, probably, if you are OK with this PR.

PR #840, opened 2021-04-01
Umbrella strategies for various algorithms.

awulkiew:
Support is added in the following algorithms:
- buffer (now it has custom strategy because simplify is used internally so it cannot use relate strategy)
- centroid
- densify
- discrete_frechet_distance
- discrete_hausdorff_distance
- length
- perimeter
- simplify

Centroid legacy strategies now does not require passing point types as template parameters. These parameters are now ignored.

Simplify agnostic strategy is now algorithm's detail.

Added dummy geometries which allow to create strategies in cases when geometry variables are not accesible or the access is not straightforward (e.g. would require to call operator/function or rearrange the algorithm).

Strategies are passed correctly in some places where default one was used improperly.


barendgehrels:
This PR is very welcome. I will review it soon.

You asked me what compiler I'm using: `clang version 5.0.2 (tags/RELEASE_502/final)`

include/boost/geometry/algorithms/detail/distance/segment_to_box.hpp
```diff
@@ -114,7 +115,7 @@ class segment_to_box_2D_generic
         std::vector<box_point> box_points(4);
         detail::assign_box_corners_oriented<true>(box, box_points);
  
-        ps_strategy_type const strategy = strategies.distance(box_points.front(), segment);
+        ps_strategy_type const strategy = strategies.distance(dummy_point(), dummy_segment());
```
barendgehrels, line 118: So we need an instance of a specific type of strategy and therefore we pass the dummies.
Can't we say, for example: `stragies.template distance<point_tag, segment_tag>()` instead? Or are there disadvantages?
What if the returned strategy depends on coordinate system?

include/boost/geometry/strategies/buffer/services.hpp
```diff
@@ -0,0 +1,46 @@
+// Boost.Geometry
+
+// Copyright (c) 2021, Oracle and/or its affiliates.
```
barendgehrels, line 3: Apparently the copyright message (with my name) is gone, can you keep the copyright of the moved messages?
I would like to keep seeing my name in the files I wrote :-)

Or actually, these files are probably brand new files, new syntactic sugar. Then it's fine, you can ignore this comment.

barendgehrels (approved):
I'm OK, I didn't review all lines, but the move to umbrella strategies is welcome.

include/boost/geometry/strategies/buffer/services.hpp
awulkiew: Yes, they are new. You can always add something in this file together with copyright message. :)

include/boost/geometry/algorithms/detail/distance/segment_to_box.hpp
awulkiew: > Can't we say, for example: `stragies.template distance<point_tag, segment_tag>()` instead?

Yes, we could. At least I think it should be possible to implement this directly inside the umbrella strategies like the current version is. It'd require to modify all of the umbrella strategies.

> Or are there disadvantages?

It's longer and requires the additional `template` keyword. In most of the cases it's enough to simply pass geometry that you already have. So:
```
strategies.distance(point, segment);
strategies.distance(dummy_point(), dummy_segment());
```
vs
```
strategies.template distance<typename tag<point>::type, typename tag<segment>::type>();
strategies.template distance<point_tag, segment_tag>();
```

> What if the returned strategy depends on coordinate system?

The umbrella strategy defines the coordinate system. It's created on the very top of the algorithm. By default it's created based on the coordinate type of the geometries (like it was with legacy strategies). All of the legacy strategies returned by it should use the same coordinate system in order to be consistent with each other. Umbrella strategy is passed internally into all algorithm that may be called by upper-level algorithms. The coordinate systems of geometries are irrelevant. It's because otherwise one could pass upper-level strategy of one CS into an upper-level algorithm which then would call lowel-level algorithm where inconsistent CS would be taken from geometry.

barendgehrels: Right, thanks for your answers! I'm OK with passing the dummies to avoid longer usage. Usually the geometry will be available, and otherwise, the dummy-definition is quite concise, I noticed. So it's fine from my perspective.

vissarion (approved):

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -449,83 +676,127 @@ struct simplify<MultiPolygon, multi_polygon_tag>
...
-            point_tag, segment_tag, point_type
-        >::type ds_strategy_type;
-
-        typedef strategy::simplify::douglas_peucker
-        <
-            point_type, ds_strategy_type
-        > strategy_type;
+        using strategies::simplify::services::strategy_converter;
 
-        BOOST_CONCEPT_ASSERT(
```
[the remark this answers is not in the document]
awulkiew: I guess it'd still make sense to check the legacy strategy after it's retrieved from umbrella strategy. I'll look into it.

include/boost/geometry/algorithms/detail/dummy_geometries.hpp
```diff
@@ -0,0 +1,56 @@
...
+} // namespace detail
+#endif // DOXYGEN_NO_DETAIL
+
+namespace traits
+{
+
+template <> struct tag<geometry::detail::dummy_point> { typedef point_tag type; };
+template <> struct tag<geometry::detail::dummy_segment> { typedef segment_tag type; };
+template <> struct tag<geometry::detail::dummy_box> { typedef box_tag type; };
+template <> struct tag<geometry::detail::dummy_linestring> { typedef linestring_tag type; };
```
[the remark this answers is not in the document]
awulkiew: Yes, right now they are not used but technically they could. I defined them for the sake of completeness.

include/boost/geometry/algorithms/simplify.hpp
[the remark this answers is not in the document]
awulkiew: What I wrote before is wrong. There is no simplify strategy anymore. It was an agnostic strategy moved to algorithms.

include/boost/geometry/algorithms/detail/dummy_geometries.hpp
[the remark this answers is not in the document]
vissarion: Yes, I was thinking whether it was clearer to define them on demand but I am OK with leaving it as is.

Issue #838, opened 2021-03-29
union_ produces wrong result


amagdum:
Result of simple union of two polygon is failing when BOOST_GEOMETRY_NO_ROBUSTNESS is enabled. I have been used no robustness as using robustness causes other issues with buffer operation. This is tested with latest boost 1.75.

std::string polygon1_wkt = "POLYGON((27162.5232832765 -40045.435290614478,27162.764636849268 -40045.321663646188,27164.74034168441 -40044.391828870466,27164.854765308635 -40044.634874609801,27164.613411735867 -40044.748501578091,27162.637706900725 -40045.678336353812,27162.5232832765 -40045.435290614478))";

using Point2d =
    boost::geometry::model::d2::point_xy<double>
boost::geometry::model::polygon<Point2d> polygon1;
  boost::geometry::read_wkt(polygon1_wkt, polygon1); 

  std::string polygon2_wkt = "POLYGON((27164.783474018484 -40044.371515841653,27165.010815897258 -40044.854409784071,27162.750625155277 -40045.918184556889,27162.523283276503 -40045.43529061447,27164.783474018484 -40044.371515841653))";
boost::geometry::model::polygon<Point2d> polygon2;
  boost::geometry::read_wkt(polygon2_wkt, polygon2); 

boost::geometry::model::multi_polygon<Point2d> result;
  boost::geometry::union_(polygon1, polygon2, result);
  std::string message;
  std::cout<< "is result valid = " << boost::geometry::is_valid(result, message) << message << std::endl;

barendgehrels:
Thanks for the issue I will check it next Wednesday

barendgehrels:
I can reproduce it. The union is successful (i.e. give the right area) but is indeed invalid. I don't know yet why it is invalid.
For the moment I don't know a workaround, apart from not checking validity.
I added your testcase to our suite.

amagdum:
Thanks. BTW, if you plot resulting polygon, you can see a spike. 

![image](https://user-images.githubusercontent.com/51491348/113183131-8abd7d80-9208-11eb-953b-a5dd83826b98.png)

barendgehrels:
> Thanks. BTW, if you plot resulting polygon, you can see a spike.

Thank you! That might give some information. I did plot the polygon (I always do when creating a testcase) but I didn't see it. I will check it again next week.

barendgehrels:
>  I will check it again next week.
[...]

PR #842, opened 2021-04-14
[intersection] fix intersection points outside of the segments

barendgehrels:
This fixes bug #838 (the visible artifact was an intersection point far away from both segments)

It also fixes some other errors (probably artifacts)

### DIFF
--- a/include/boost/geometry/policies/robustness/segment_ratio.hpp
+++ b/include/boost/geometry/policies/robustness/segment_ratio.hpp
@@ -94,6 +94,35 @@ struct equal<Type, false>
     }
 };
 
+template
+<
+    typename Type,
+    bool IsFloatingPoint = std::is_floating_point<Type>::type::value
+>
+struct possibly_collinear {};
+
+template <typename Type>
+struct possibly_collinear<Type, true>
+{
+    template <typename Ratio, typename Threshold>
+    static inline bool apply(Ratio const& ratio, Threshold threshold)
+    {
+        return std::abs(ratio.denominator()) < threshold;
+    }
+};
+
+// Any ratio based on non-floating point (or user defined floating point)
+// is collinear if the denominator is exactly zero
+template <typename Type>
+struct possibly_collinear<Type, false>
+{
+    template <typename Ratio, typename Threshold>
+    static inline bool apply(Ratio const& ratio, Threshold)
+    {
+        return ratio.denominator() == 0;
+    }
+};
+
 }}
 
 //! Small class to keep a ratio (e.g. 1/4)
@@ -240,6 +269,12 @@ public :
     inline bool close_to(thistype const& other) const
     {
         return geometry::math::abs(m_approximation - other.m_approximation) < 50;
+    }
+
+    template <typename Threshold>
+    inline bool possibly_collinear(Threshold threshold) const
+    {
+        return detail::segment_ratio::possibly_collinear<Type>::apply(*this, threshold);
     }
 
     inline bool operator< (thistype const& other) const
--- a/include/boost/geometry/strategies/cartesian/intersection.hpp
+++ b/include/boost/geometry/strategies/cartesian/intersection.hpp
@@ -220,7 +220,9 @@ struct cartesian_segments
         }
 
         template <typename Point, typename Segment>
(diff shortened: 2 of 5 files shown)


include/boost/geometry/strategies/cartesian/intersection.hpp
```diff
@@ -282,6 +298,49 @@ struct cartesian_segments
...
+
+            // The segments are nearly collinear and because of the calculation
+            // method with very small denominator, the IP appears outside the
+            // segment(s). Correct it to the end point.
+            // Because they are nearly collinear, it doesn't really matter to
+            // to which endpoint (or it is corrected twice).
+            if (exceeds<0>(point, a))
+            {
+                geometry::set<0>(point, geometry::get<0, 0>(a));
+                geometry::set<1>(point, geometry::get<0, 1>(a));
```
barendgehrels, line 326: I'll probably change this to `assign` or something we have for this

include/boost/geometry/strategies/cartesian/intersection.hpp
```diff
@@ -282,6 +298,49 @@ struct cartesian_segments
             {
                 assign_b(point, a, b);
             }
+
+            // Verify
```
barendgehrels, line 302: Will remove

include/boost/geometry/strategies/cartesian/intersection.hpp
```diff
@@ -245,6 +247,20 @@ struct cartesian_segments
...
+        static bool exceeds(Point& point, Segment const& s)
+        {
+            return (geometry::get<Index, 0>(s) < geometry::get<1 - Index, 0>(s)
+                    && geometry::get<0>(point) < geometry::get<Index, 0>(s))
+                || (geometry::get<Index, 0>(s) > geometry::get<1 - Index, 0>(s)
+                    && geometry::get<0>(point) > geometry::get<Index, 0>(s))
+                || (geometry::get<Index, 1>(s) < geometry::get<1 - Index, 1>(s)
+                    && geometry::get<1>(point) < geometry::get<Index, 1>(s))
+                || (geometry::get<Index, 1>(s) > geometry::get<1 - Index, 1>(s)
+                    && geometry::get<1>(point) > geometry::get<Index, 1>(s));
```
barendgehrels, line 260: I can make an extra function per dimension such that will look simpler, and add some ascii art

include/boost/geometry/strategies/cartesian/intersection.hpp
barendgehrels: :heavy_check_mark: done

include/boost/geometry/strategies/cartesian/intersection.hpp
barendgehrels: :heavy_check_mark:

include/boost/geometry/strategies/cartesian/intersection.hpp
barendgehrels: :heavy_check_mark: done

include/boost/geometry/policies/robustness/segment_ratio.hpp
```diff
@@ -94,6 +94,35 @@ struct equal<Type, false>
...
+{
+    template <typename Ratio, typename Threshold>
+    static inline bool apply(Ratio const& ratio, Threshold threshold)
+    {
+        return std::abs(ratio.denominator()) < threshold;
+    }
+};
+
+// Any ratio based on non-floating point (integer or arithmetic)
+// will not be collinear.
```
awulkiew, line 115: What do you mean by arithmetic? Do you mean user-defined types? They could represent FP number right?

Also why integers won't be possibly collinear? I mean for them possibly collinear and strictly collinear would be the same, right? My point is I don't understand why does this specialization return always `false`.

awulkiew (approved):

include/boost/geometry/policies/robustness/segment_ratio.hpp
barendgehrels: Good point I will have a look

barendgehrels: :heavy_check_mark: I changed it. Thanks. I will merge it because next PR depends on it.

Issue #843, opened 2021-04-21
Proposal: GeometryCollection, DynamicGeometry


awulkiew:
I'd like to propose addition of GeometryCollection, DynamicGeometry and making the support for TupledGeometry public. I made a prototype for these concepts here: https://github.com/awulkiew/test-geometry-collection

**DynamicGeometry** is any type using type-erasure idiom allowing to store various StaticGeometries transparently. Examples:
- `boost::variant`
- `boost::variant2::variant2`
- `std::variant`
- `boost::any`
- `std::any`
- User-defined class internally using abstract class and hierarchy of geometries

The concept:
- `traits::tag<G>::type` is `dynamic_geomery_tag`
- `traits::geometry_types<G>::type` is `util::type_sequence` storing all of the StaticGeometries G can hold
- (mutable) G's assignment operator taking any of the types from `traits::geometry_types<G>::type` is defined
- `traits::visit<G>::apply` visits geometry
- (optional) `traits::visit<G1, G2>::apply` visits two geometries, by default 1-argument `visit` is applied 2 times

Alternatives:
- replace `traits::geometry_types<G>::type` defining `util::type_sequence` with traits returning number of types and allowing to get a type defined by some Index, mimicking `std::variant_size` and `std::variant_alternative`.
- replace `traits::visit` with `get<>` (then it wouldn't be possible to rely on multiple variants visitation)

**GeometryCollection** is a ForwardRange storing any type that allows the use of DynamicGeometry, even by proxy, so not necessarily storing DynamicGeometry itself. Examples:
- `std::vector<boost::any>`
- `bg::model::dynamic_geometry<boost::variant<...>>`
- UserDefined class holding e.g. a container of DynamicGeometries, a container of pointers to DynamicGeometries, etc.

The concept:
- `traits::tag<GC>::type` is `geometry_collection_tag`
- adapted to Boost.Range ForwardRange concept.
- (mutable) `traits::clear`
- (mutable) `traits::emplace_back` adding StaticGeometry to the collection (probably could be called `emplace` for even more genericity, e.g. to work well with sets/maps)
- (optional) `traits::visit_iterator<GC>::apply` visits geometry pointed by the iterator of GC, by default this trait calls `traits::visit<boost::range_value<GC>::type>::apply`, so ranges of DynamicGeometries
[...]

PR #850, opened 2021-05-17
Add DynamicGeometry and GeometryCollection.

awulkiew:
After: https://github.com/boostorg/geometry/issues/843

This PR adds DynamicGeometry and GeometryCollection and usage in several algorithms.

List of changes:
- Added DynamicGeometry concept
  - Required specialization of `traits::tag<>` defining `type` as `dynamic_geometry_tag`
  - Required specialization of 1-parameter `traits::visit<>`
  - Required specialization of `traits::geometry_types<>`
  - Optional specialization of 2-parameter `traits::visit<>`
  - [mutable] Must be assignable from types specified in `traits::geometry_types<>`
- Added GeometryCollection concept
  - Required specialization of `traits::tag<>` defining `type` as `geometry_collection_tag`
  - Required adaptation to ForwardRange concept
  - Optional specialization of `traits::geometry_types<>` (by default using `traits::geometry_types<>` for `range_value<>::type`)
  - Optional specialization of `traits::iter_visit<>` (by default calling `traits::visit<>` for `range_value<>::type`)
  - [mutable] Required specialization of `traits::emplace_back<>` allowing to add objects of types specified in `traits::geometry_types<>`
  - [mutable] Optional specialization of `traits::clear<>` (by default calling `collection.clear()`)
  - May be recursive
- Added adaptations of DynamicGeometry
  - `boost::variant<>`
  - `boost::variant2::variant<>`
  - `boost::any<>` (requires additional user specialization of `traits::geometry_types<>`)
  - `std::variant<>`
  - `std::any<>` (requires additional user specialization of `traits::geometry_types<>`)
- Added GeometryCollection models
  - `model::geometry_collection<>`
- Added concept checks for DynamicGeometry and GeometryCollection
  - added new util `concepts::concept_type<>` defining concept of Geometry in a generic way
  - refactored `concepts::check()`, dispatches replaced with `concepts::concept_type<>`
- Added algorithms:
  - `detail::visit()`
  - `detail::visit_breadth_first()`
- Added support for DynamicGeometry and GeometryCollection in:
  - `append()` (only DynamicGeometry)
  - `area()`
  - `clear()`
  - `disjoint()` (the algorithm for 2 GeometryCollections has quadratic complexity so should probably be refactored)
  - `length()`
  - `read_wkt()`
  - `wkt()`
  - `comparable_distance_result<>`
  - `distance_result<>`
- Other
  - Added support for rvalue references to range `value_type` in `traits::push_back` and `range::push_back` to support move semantics (the support for rvalue reference for the Range itself should still be added in a different PR)
  - `boost::variant` workarounds previously implemented for VS2015 now works with all compilers `#ifdef BOOST_VARIANT_DO_NOT_USE_VARIADIC_TEMPLATES` (in various algorithms)

I'm open to suggestions.


include/boost/geometry/algorithms/detail/disjoint/interface.hpp
```diff
@@ -104,13 +102,21 @@ struct disjoint<default_strategy, false>
...
-namespace resolve_variant {
+namespace resolve_dynamic {
 
-template <typename Geometry1, typename Geometry2>
+template
+<
+    typename Geometry1, typename Geometry2,
+    bool IsDynamic = util::is_dynamic_geometry<Geometry1>::value
+                  || util::is_dynamic_geometry<Geometry2>::value,
+    bool IsGC = util::is_geometry_collection<Geometry1>::value
```
mloskot, line 112: Nitpick: since there is non-abbreviated `IsDynamic`, then `IsGC` could be `IsCollection` ;)

include/boost/geometry/algorithms/visit.hpp
```diff
@@ -0,0 +1,307 @@
...
+    BOOST_GEOMETRY_STATIC_ASSERT_FALSE(
+        "Not implemented for these types.",
+        Geometry1, Geometry2);
+};
+
+template <typename Geometry1, typename Geometry2, typename Tag2>
+struct visit_two<Geometry1, Geometry2, dynamic_geometry_tag, Tag2>
+{
+    template <typename F, typename G1, typename G2>
+    static void apply(F && f, G1 & geom1, G2 & geom2)
```
mloskot, line 102:
```cpp
static void apply(F && f, G1 & geom1, G2 & geom2) // visit left-most geometry
```

include/boost/geometry/algorithms/visit.hpp
```diff
@@ -0,0 +1,307 @@
...
+    {
+        return f(g);
+    }
+
+    template
+    <
+        typename Geom, typename Iterator,
+        std::enable_if_t<util::is_geometry_collection<Geom>::value, int> = 0
+    >
+    static void set_iterators(Geom & g, Iterator & first, Iterator & last)
```
mloskot, line 241: I guess it could not be `Geom const& g`

mloskot (approved):
I've skimmed the code trying to understand the core parts and nothing looks suspicious.
Nice feature!

include/boost/geometry/algorithms/visit.hpp
awulkiew: The point is to support both const and non-const references here. This is also why a function template argument is used instead of struct template argument `Geometry`.

awulkiew:
@mloskot Thanks for the review!

The thing that I don't like about it is the fact that `traits::visit` has to be specialized for a specific number of geometries, one or two. Do you have a different idea how this trait could be implemented to allow specialization for e.g. any number of arbitrary variants (potentially holding different geometries)?

Do you think that rvalue references should be handled in visit too?

Do you think visit should be able to return a value?

mloskot:
> Do you have a different idea how this trait could be implemented to allow specialization for e.g. any number of arbitrary variants

My C++ is getting rusty, but it generally should be [feasible](https://stackoverflow.com/q/24687026/151641), but I'm sure you've looked at the variadics.

> Do you think that rvalue references should be handled in visit too?
> Do you think visit should be able to return a value?

I think it is generally a good idea to keep interfaces aligned with the standard stuff, i.e. similar to `std::visit`

I wish I could be more helpful, but can't afford any experiments on my own right now.

include/boost/geometry/algorithms/append.hpp
```diff
@@ -148,206 +156,103 @@ struct range_to_polygon
...
-    : detail::append::append_no_action<Geometry, Point>
+template
+<
+    typename Geometry, typename RangeOrPoint,
+    typename Tag = typename geometry::tag<Geometry>::type,
+    typename OtherTag = typename geometry::tag<RangeOrPoint>::type
+>
+struct append
+    : detail::append::append_no_action
 {};
```
barendgehrels, line 167: Looks fine, thanks for the clean up

include/boost/geometry/algorithms/append.hpp
```diff
@@ -148,206 +156,103 @@ struct range_to_polygon
...
-    {
-        boost::apply_visitor(
-            visitor<RangeOrPoint>(
-                range_or_point,
-                ring_index,
-                multi_index
-            ),
-            variant_geometry
-        );
+            // TODO: move this to bg::append()?
```
barendgehrels, line 240: Probably, yes, we usually don't do this in the dispatch

include/boost/geometry/algorithms/area.hpp
```diff
@@ -287,37 +286,41 @@ struct area
...
+        apply(Geometry const& geometry, Strategy const& strategy)
     {
-        Strategy const& m_strategy;
-
-        visitor(Strategy const& strategy): m_strategy(strategy) {}
-
-        template <typename Geometry>
-        typename area_result<variant_type, Strategy>::type
-        operator()(Geometry const& geometry) const
+        // TODO: Technically value could be returned directly from visit
```
barendgehrels, line 296: Current version looks fine to me and consistent with the version below

include/boost/geometry/algorithms/detail/disjoint/interface.hpp
```diff
@@ -125,104 +131,54 @@ struct disjoint
...
+        geometry::visit_breadth_first([&](auto const& g1)
+        {
+            geometry::visit_breadth_first([&](auto const& g2)
+            {
+                result = disjoint
+                    <
+                        util::remove_cref_t<decltype(g1)>, util::remove_cref_t<decltype(g2)>
+                    >::apply(g1, g2, strategy);
+                // If any of the combination intersects then the final result is not disjoint
+                //   so break testing.
```
barendgehrels, line 172: I think the last part `so break testing` can be omitted

include/boost/geometry/algorithms/detail/disjoint/interface.hpp
```diff
@@ -125,104 +131,54 @@ struct disjoint
...
+                return result;
+            }, geometry2);
+            return result;
+        }, geometry1);
+        return result;
```
barendgehrels, line 177: I don't get the logic yet but I didn't dive into `visit_breadth_first` yet

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -294,7 +294,7 @@ struct simplify_polygon
             simplify_ring::apply(*it, out, max_distance, strategy);
             if (! geometry::is_empty(out))
             {
-                range::push_back(interior_rings_out, out);
+                range::push_back(interior_rings_out, std::move(out));
```
barendgehrels, line 297: `move` is better, thanks

include/boost/geometry/algorithms/visit.hpp
```diff
@@ -0,0 +1,307 @@
...
+        {
+            for (; it != end; ++it)
+            {
+                bool result = true;
+                traits::visit_iterator<util::remove_cref_t<Geom>>::apply([&](auto & g)
+                {
+                    result = visit_or_enqueue(function, g, queue, it);
+                }, it);
+                if (! result)
+                    return false;
```
barendgehrels, line 195: Curly braces are missing

include/boost/geometry/algorithms/visit.hpp
```diff
@@ -0,0 +1,307 @@
...
+    >
+    static void set_iterators(Geom & g, Iterator & first, Iterator & last)
+    {
+        first = boost::begin(g);
+        last = boost::end(g);
+    }
+    template
+    <
+        typename Geom, typename Iterator,
+        std::enable_if_t<!util::is_geometry_collection<Geom>::value, int> = 0
```
barendgehrels, line 249: Minor, space after `!` is missing

include/boost/geometry/algorithms/visit.hpp
```diff
@@ -0,0 +1,307 @@
...
+        iter_t it = boost::begin(geom);
+        iter_t end = boost::end(geom);
+        for(;;)
+        {
+            for (; it != end; ++it)
+            {
+                bool result = true;
+                traits::visit_iterator<util::remove_cref_t<Geom>>::apply([&](auto & g)
+                {
+                    result = visit_or_enqueue(function, g, queue, it);
```
barendgehrels, line 192: OK, I see, it visits all elements in the collection without recursion.
I'm not a fan of storing iterators but I think it's OK here.

include/boost/geometry/algorithms/visit.hpp
```diff
@@ -0,0 +1,307 @@
...
+//   StaticGeometries.
+// The number of elements visited may be different than the size of the top-most GeometryCollection
+//   if it's recursive.
+// TODO: This algorithm is similar to std::for_each so it could be called e.g. std::visit_each or
+//   std::for_each_visit but it doesn't work the same way, i.e. it traverses recursive
+//   GeometryCollections.
+// TODO: Should this algorithm work only for GeometryCollection?
+// TODO: We could change the order of the arguments but I'm not sure if that's a good idea.
+template <typename UnaryFunction, typename Geometry>
+inline void visit_breadth_first(UnaryFunction function, Geometry & geometry)
```
barendgehrels, line 299: Is this a function for the end user?
If so we should resolve all todo's and make a decision about the order.
Or should we move it to `namespace detail` first?

barendgehrels: Same for the `visit` functions above

include/boost/geometry/core/mutable_range.hpp
```diff
@@ -85,8 +111,13 @@ struct push_back
 template <typename Range>
 struct resize
 {
+    typedef typename boost::range_size
+        <
+            typename std::remove_reference<Range>::type
+        >::type size_type;
```
barendgehrels, line 117: New types should actually be defined by `using`

include/boost/geometry/core/mutable_range.hpp
```diff
@@ -75,6 +76,31 @@ struct push_back
...
+{
+    // When specializing it'd be enough to only implement it for one argument
+    // because we'll use it to pass one object of type potentially different
+    // than range_value but which range_value can be constructed from.
+    template <typename ...Args>
+    static inline void apply(typename rvalue_type<Range>::type range,
+                             Args&&... args)
+    {
+        range.emplace_back(std::forward<Args>(args)...);
+    }
```
barendgehrels, line 103: This will break older code? (with own types)? Or is it optional?

include/boost/geometry/extensions/gis/io/shapefile/read.hpp
```diff
@@ -614,27 +614,23 @@ struct read_polygon_policy
...
             if (num_parts == 1)
-                range::push_back(outer_rings, ring); // order could be checked here too
+                range::push_back(outer_rings, std::move(ring)); // order could be checked here too
             // check order
             else if ( is_outer_ring(ring, order_strategy) )
-                range::push_back(outer_rings, ring);
+                range::push_back(outer_rings, std::move(ring));
             else
-                range::push_back(inner_rings, ring);
+                range::push_back(inner_rings, std::move(ring));
```
barendgehrels, line 624: curly braces are missing (was already)

include/boost/geometry/geometries/adapted/boost_any.hpp
```diff
@@ -0,0 +1,104 @@
...
+// Licensed under the Boost Software License version 1.0.
+// http://www.boost.org/users/license.html
+
+#ifndef BOOST_GEOMETRY_GEOMETRIES_ADAPTED_BOOST_ANY_HPP
+#define BOOST_GEOMETRY_GEOMETRIES_ADAPTED_BOOST_ANY_HPP
+
+
+#include <utility>
+
+#include <boost/any.hpp>
```
barendgehrels, line 16: Do we still need it? `std::any` will be there in C++17
But I'm fine with this one too

include/boost/geometry/geometries/adapted/boost_variant.hpp
```diff
@@ -0,0 +1,157 @@
...
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_GEOMETRIES_ADAPTED_BOOST_VARIANT_HPP
+#define BOOST_GEOMETRY_GEOMETRIES_ADAPTED_BOOST_VARIANT_HPP
+
+
+#include <boost/variant/apply_visitor.hpp>
+#include <boost/variant/static_visitor.hpp>
+#include <boost/variant/variant.hpp>
```
barendgehrels, line 24: Same here, `std::variant` in C++17

include/boost/geometry/geometries/concepts/check.hpp
```diff
@@ -181,6 +181,16 @@ struct check<Geometry, multi_polygon_tag, false>
     : detail::concept_check::check<concepts::MultiPolygon<Geometry> >
 {};
 
+template <typename Geometry, bool IsConst>
+struct check<Geometry, dynamic_geometry_tag, IsConst>
+    // TODO traits::visit
```
barendgehrels, line 186: There are still many `TODO` 's in this PR.
I don't know how important they are, and when this should be done.

include/boost/geometry/geometries/geometry_collection.hpp
```diff
@@ -0,0 +1,79 @@
...
+{
+    typedef Container<DynamicGeometry, Allocator<DynamicGeometry>> base_type;
+
+public:
+    geometry_collection() = default;
+
+//    template <typename Iterator>
+//    geometry_collection(Iterator begin, Iterator end)
+//        : base_type(begin, end)
+//    {}
```
barendgehrels, line 48: Delete this?

barendgehrels (changes requested):
Thanks a lot!

I think at least some of the `TODO`'s should be done, or rephrased, and some commented code can go or be uncommented, so I set this to `Request changes` for now, but apart from that, I'm fine.

include/boost/geometry/algorithms/area.hpp
awulkiew: Plus there is a problem with the old `variant` which forces us to define the return type as the template parameter of `boost::static_visitor<>`. I tried to define it automatically based on the return type of the functor for the first type stored in a variant and it would be doable but the adaptation of `boost::variant` would be a lot more complex. We would be able to write:
```
    return visit<...>(...);
```
but I think in general `area_result<>` would still be needed so I doubt it'd be worth it to make visit more complex only to be able to ommit two lines in some places of the code. I only worry about cutting of the possibility because it's possible that in the (most likely distant) future we could need this for something. But that's only that.

include/boost/geometry/algorithms/detail/disjoint/interface.hpp
awulkiew: It's based on the assumption that geometry collections can be recursive. So a GC can store GCs. AFAIU this can be the case. So this function traverses the recursive GC in a breadth-first fashion (as opposed to depth-first). And as long as the function/lambda returns `true` the traversal is continued. Maybe I should've called it differently. E.g. `visit_breadth_first_if`?

So here as long as pairs of geometries are keep being disjoint then the traversal should continue. Hence the `result` is returned because it's `true` if pairs of geometries are being disjoint.

Or is it not clear from a different reason?

include/boost/geometry/algorithms/visit.hpp
awulkiew: What do you think about relaxing this guideline? I.e. to allow single lines under if-else? From my point of view adding curly braces around this one line makes the whole algorithm less readable because there are 2x as much lines of code that can be seen by the eye, the information is the same and this space could be used for other things.

It'd be a different thing if this was very long line which potentially could be wrapped into yet another line. In this case I'd argue that curly braces are indeed desireable. But in case of a short call or return statement. I don't think that it is.

Do you strongly feel that these braces are necessary here?

include/boost/geometry/algorithms/visit.hpp
awulkiew: Do you mean you don't like dealing with recursion this way (i.e. using manual stack or queue) and prefer recursion using function stack?
Or would you prefer storing e.g. pointers or something else?

include/boost/geometry/algorithms/visit.hpp
awulkiew: Yes, moving it to `detail` is a good idea regardless of the order of arguments.

Regarding the order, it depends if we think it's more like `std::visit` or `std::for_each`. I'm leaning more towards having the same interface for the whole family of `visit` functions also consistent with `std::visit`.

include/boost/geometry/core/mutable_range.hpp
awulkiew: Not by itself. It could break the code if we replaced `push_back` with it and a user-defined geometry wouldn't also define `emplace_back`.

This is why `emplace_back` should be used only with GCs. Why not `push_back`? Because a GC can store a different type than the one that is passed. E.g. variant is stored but we're emplacing back a point, or GC stores pointers to geometries but we pass a linestring.

For all other geometries we know the type that is stored and we should keep using `push_back`. However now we can also move which will perform copy if the user-defined geometry doesn't define move ctor.

include/boost/geometry/geometries/adapted/boost_any.hpp
awulkiew: I don't know. BG is a C++14 library, so I guess? I implemented it more like a proof of concept because I doubt that anyone will use `boost::any` with BG since there is `variant2` but who knows. In general no adaptations are needed becasue a user could write them.

include/boost/geometry/geometries/adapted/boost_variant.hpp
awulkiew: The adaptation for `boost::variant` is for backward compatibility.
For `boost::variant2::variant` is there because BG is C++14 library.

And yes, the adaptation for `std::any` and `std::variant` should be added as well. But then this PR would be even bigger.

awulkiew: The adaptation for `boost::variant` is also included by default by the algorithms for backward compatibility. We could change that forcing the users to include this optional file but technically this would be a breaking change.

include/boost/geometry/geometries/geometry_collection.hpp
awulkiew: Or add ctors in all models (linestring, ring, etc.) taking a pair of iterators?

include/boost/geometry/geometries/concepts/check.hpp
awulkiew: There are many TODOs because there are many ways how to proceed. We also didn't talk about it on the Issues page. So I'm treating this PR as a ground for brainstorming.

The question here is: should we check all of the geometries stored in a dynamic geometry here or rather do this in a `variant_dispatch` (see above) and here ignore the check?

Btw, the `traits::visit` is wrong. It'd have to be a`concept_check` for `concepts::DynamicGeometry` checking all of the types defined in `traits::geometry_types<>::type`.

include/boost/geometry/algorithms/visit.hpp
awulkiew: I moved them to `detail`.

include/boost/geometry/geometries/adapted/boost_variant.hpp
awulkiew: I added adaptations for `std::any` and `std::variant`.

awulkiew:
@mloskot I added support for rvalue references because it is possible we'll want to move something inside visit. I didn't implement returning values though because:
- in some cases the implementation would be complicated and still not 100% bulletproof, e.g. for `boost::variant` forcing to define return value as template argument of `boost::static_visitor<>`,
- this would be a requirement for the users adapting their classes and this requirement would make this more complicated as well.

So I don't think it's worth it only to be able to save 2 lines of code now and then.

awulkiew:
@barendgehrels I made the PR more complete. I added the list of changes in the description.

include/boost/geometry/geometries/concepts/check.hpp
awulkiew: There are the concepts for DG and GC now, though only the first geometry from the `geometry_types<>` sequence is checked there because we also need it to check whether or not it can be assigned to DG or emplaced_back to GC.

So I see two possibilities, as I mentioned:
- check the first type in the concept check for DG and GC and the rest when they are actually used, i.e. in the `variant_dispatch` portion of an algorithm
- check all of the types in the concept check

Do you have any preference?

include/boost/geometry/geometries/adapted/boost_any.hpp
awulkiew: With that said I think it'd better that we do this both for user convenience and to make sure the the adaptations are correct.

barendgehrels: Thank you, it's fine to keep them and be sure about it. It was just a consideration.

include/boost/geometry/geometries/concepts/check.hpp
barendgehrels: As mentioned above, I'm fine either way. If I would have a preference, I would select the option which gives the most clear error message if users pass a wrong (not fulfilling) geometry. Unless that requires a lot of extra code.

include/boost/geometry/geometries/geometry_collection.hpp
barendgehrels: We might do that, but as far as I know, no user asked for this.

include/boost/geometry/geometries/adapted/std_any.hpp
```diff
@@ -0,0 +1,82 @@
...
+// Licensed under the Boost Software License version 1.0.
+// http://www.boost.org/users/license.html
+
+#ifndef BOOST_GEOMETRY_GEOMETRIES_ADAPTED_STD_ANY_HPP
+#define BOOST_GEOMETRY_GEOMETRIES_ADAPTED_STD_ANY_HPP
+
+
+#include <boost/config.hpp>
+
+#ifndef BOOST_NO_CXX17_HDR_ANY
```
barendgehrels, line 16: This looks good, thanks.

include/boost/geometry/geometries/adapted/boost_variant2.hpp
```diff
@@ -0,0 +1,73 @@
...
+namespace boost { namespace geometry
+{
+
+namespace traits
+{
+
+template <typename ...Ts>
+struct tag<boost::variant2::variant<Ts...>>
+{
+    typedef dynamic_geometry_tag type;
```
barendgehrels, line 34: Suggestion: `using type = ...`

include/boost/geometry/geometries/geometry_collection.hpp
awulkiew: No user asked for the `initializer_list` either. This is just something I found useful at some point during testing. But I'm ok with leaving it as it is.

barendgehrels (approved):
Thanks!

include/boost/geometry/geometries/concepts/check.hpp
awulkiew: It wouldn't require a lot of extra code. I'm not sure about the error message, It'd probably be the same but I think I'll do this in concepts to keep all of the checks in one place.

barendgehrels: OK

include/boost/geometry/geometries/geometry_collection.hpp
barendgehrels: OK (we can always add it later if necessary)

include/boost/geometry/algorithms/detail/disjoint/interface.hpp
barendgehrels: It's clear now. Thank you.

include/boost/geometry/algorithms/visit.hpp
awulkiew: I added them but I'll probably create an issue to discuss this later.

include/boost/geometry/geometries/concepts/check.hpp
awulkiew: Done.

test/algorithms/clear_multi.cpp
```diff
@@ -42,12 +56,12 @@ void test_all()
     typedef bg::model::multi_linestring<ls> mls;
     typedef bg::model::multi_polygon<poly> mpoly;
 
-    test_geometry<Point>("POINT(0 0)", 1);
-    test_geometry<ls>("LINESTRING(0 0,0 1)", 0);
-    test_geometry<poly>("POLYGON((0 0,0 1,1 0,0 0))", 0);
-    test_geometry<mpoint>("MULTIPOINT((0 0),(0 1),(1 0),(0 0))", 0);
-    test_geometry<mls>("MULTILINESTRING((0 0,0 1),(1 0,0 0))", 0);
-    test_geometry<mpoly>("MULTIPOLYGON(((0 0,0 1,1 0,0 0)),((10 0,10 1,11 0,10 0)))", 0);
+    test_geometry<Point>("POINT(0 0)", 1, 1);
```
vissarion, line 59: why not `0` after clearing?

test/algorithms/disjoint/test_disjoint.hpp
```diff
@@ -88,17 +89,27 @@ void test_disjoint(std::string const& id,
...
     check_disjoint(id, wkt1, wkt2, v1, g2, expected, no_strategy());
     check_disjoint(id, wkt1, wkt2, g1, v2, expected, no_strategy());
     check_disjoint(id, wkt1, wkt2, v1, v2, expected, no_strategy());
+
+    check_disjoint(id, wkt1, wkt2, gc1, g2, expected, no_strategy());
+    check_disjoint(id, wkt1, wkt2, g1, gc2, expected, no_strategy());
+    check_disjoint(id, wkt1, wkt2, gc1, gc2, expected, no_strategy());
+
+    check_disjoint(id, wkt1, wkt2, v1, gc2, expected, no_strategy());
+    check_disjoint(id, wkt1, wkt2, gc1, v2, expected, no_strategy());
```
vissarion, line 112: maybe add a test with gc and strategy?

vissarion (approved):
Thanks and sorry for late reviewing, I am OK for merging.

test/algorithms/clear_multi.cpp
awulkiew: Because Boost.Geometry Point can't be empty.

include/boost/geometry/algorithms/area_result.hpp
```diff
@@ -74,37 +72,16 @@ struct default_area_result
         >
 {};
 
-// Workaround for VS2015
```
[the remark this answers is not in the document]
awulkiew: No, the workaround was replaced by proper handling of `#ifdef BOOST_VARIANT_DO_NOT_USE_VARIADIC_TEMPLATES` in `geometries/adapted/boost_variant.hpp`

include/boost/geometry/algorithms/append.hpp
```diff
@@ -148,206 +157,100 @@ struct range_to_polygon
...
 
-namespace splitted_dispatch
-{
-
-template <typename Tag, typename Geometry, typename Point>
-struct append_point
-    : detail::append::append_no_action<Geometry, Point>
+template
+<
+    typename Geometry, typename RangeOrPoint,
```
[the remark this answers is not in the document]
awulkiew: This was done on purpose, there is enough free space so I kept the mandatory arguments in one line. Do you think this is wrong and I should put each one of them in separate lines?

include/boost/geometry/algorithms/detail/buffer/buffered_ring.hpp
```diff
@@ -83,11 +83,14 @@ struct buffered_ring_collection : public std::vector<Ring>
...
-namespace dispatch
+namespace concepts
 {
-template <typename Geometry, bool IsConst>
-struct check<Geometry, detail::buffer::buffered_ring_collection_tag, IsConst>
+
+template <typename Geometry>
+struct concept_type<Geometry, geometry::detail::buffer::buffered_ring_collection_tag>
 {
+    struct dummy {};
```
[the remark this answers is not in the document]
awulkiew: Yes this is a dummy, empty concept check.

include/boost/geometry/strategies/default_length_result.hpp
```diff
@@ -20,8 +20,7 @@
 #define BOOST_GEOMETRY_STRATEGIES_DEFAULT_LENGTH_RESULT_HPP
 
 
-#include <boost/variant/variant_fwd.hpp>
-
+#include <boost/geometry/algorithms/detail/select_geometry_type.hpp>
```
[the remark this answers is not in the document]
awulkiew: Yes, it defines `detail::geometry_types<>` used at line 63.

include/boost/geometry/algorithms/detail/select_geometry_type.hpp
```diff
@@ -0,0 +1,106 @@
...
+{
+
+template
+<
+    typename Geometry,
+    template <typename, typename> class LessPred,
+    bool IsDynamicOrCollection = util::is_dynamic_geometry<Geometry>::value
+                              || util::is_geometry_collection<Geometry>::value
+>
+struct select_geometry_type
```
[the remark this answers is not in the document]
awulkiew: Technically they are used in files that are defined in `strategies/` directory, yes. These are `default_length_result`, `distance_result` and `comparable_distance_result`. They define the result of the algorithm defined by both geometries and strategy based on tag dispatching. This is why these files belong in the `algorithms/` directory, just like `area_result` which was already moved. I plan to move them there too as well, but not in this PR.

`util::is_xxx` are multi-purpose traits that can be used anywhere.

`detail::select_geometry_type` is a utility used only in algorithms and it allows to pick a type according to a predicate. And this type can either be the geometry itself or a type of a sequence defined by `traits::geometry_type<>` for GC or DG. So this is a single utility allowing us e.g. to choose a Geometry which has the most precise `coordinate_type` (used in `default_length_result`) or the most precise result of a strategy for a pair of geometries (`select_geometry_types` used in `distance_result`) for any Geometry the library supports.

include/boost/geometry/strategies/distance_result.hpp
```diff
@@ -78,12 +78,7 @@ struct distance_result_strategy_type<Geometry1, Geometry2, Strategy, true>
 {};
 
 
-template
```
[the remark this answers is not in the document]
awulkiew: It's because this was a part of the workaround for VS2015. With this compiler Boost.Variant disables variadic templates and emulates them instead. This means that at the end of the list of arguments there are types `boost::detail::variant::null` or something like that which are not geometries. `AreGeometries` was used to ignore them with VS2015 by setting the distance result to `int` for them. Now the adaptation for Boost.Variant handles it so this workaround including `AreGeometries` is not needed anymore.

include/boost/geometry/algorithms/detail/select_geometry_type.hpp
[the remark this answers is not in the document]
vissarion: OK, thanks for the explanations.

include/boost/geometry/strategies/default_length_result.hpp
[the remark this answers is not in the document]
vissarion: right, select_geometry_type.hpp also includes the geometry_types ! thanks

include/boost/geometry/algorithms/append.hpp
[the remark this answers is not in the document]
vissarion: It is different from the guidelines but it makes sense so I am OK.

awulkiew:
Thanks for the reviews!

Issue #851, opened 2021-05-21
Geographic Intersection of LineString with Box yields wrong result for boost version > 1.67.0


mapman63:
The following code

``` c++
#include <iostream>
#include <boost/geometry.hpp>
#include <boost/geometry/core/coordinate_system.hpp>
#include <boost/geometry/geometries/point_xy.hpp>

namespace bg = boost::geometry;
namespace bm = bg::model::d2;

using Pt = bm::point_xy<double, bg::cs::geographic<bg::degree> >;
using Box = bg::model::box<Pt>;
using LineString = bg::model::linestring<Pt>;

 Pt Stuttgart() { return { 9.18, 48.78}; }
 Pt Berlin() { return { 13.4, 52.52 }; }
 Box boxInCzechRepublic() {
 return { {13.381347, 50.625}
        , {13.40332, 50.646972}};
}

LineString StuttgartBerlin() {
    return {Stuttgart(), Berlin()};
}


int main() {
  bool const result = bg::intersects (StuttgartBerlin(), boxInCzechRepublic());
  if (result) {
    std::cout << "The line from Stuttgart to Berlin intersects the Czech Republic" << std::endl;
  }
  else {
    std::cout << "The line from Stuttgart to Berlin does not intersect the Czech Republic" << std::endl;
  }
  return 0;
}
```

yields the correct result for boost version <= 1.67 and the wrong result for boost version >= 1.68:

- version 1.67.0: https://godbolt.org/z/fMxEjo86P
- version 1.68.0: https://godbolt.org/z/nhbqE9sz8
- (...) 
- version 1.76.0: https://godbolt.org/z/r1r8r6YhY

I also observed this behavior for other compilers.

awulkiew:
Thanks for the report!

I can reproduce it. The algorithm returns `true` for the case:

![obraz](https://user-images.githubusercontent.com/1226951/119262664-4a60f700-bbdc-11eb-8034-a342cbe36512.png)

The issue seems to be located in `disjoint_segment_box_sphere_or_spheroid`, somewhere around these lines:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp#L206-L213

`vertex_lat` is the same as the latitude of `box_seg`'s min/max only if min/max is lower/higher than both of the endpoints.

I also don't know why the function doesn't return when all of the azimuth sides are the same here:
[...]

PR #853, opened 2021-05-24
Add missing case for geo segment-box disjoint

vissarion:
This PR proposes a fix for https://github.com/boostorg/geometry/issues/851

### DIFF
--- a/include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
+++ b/include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
@@ -181,23 +181,23 @@ struct disjoint_segment_box_sphere_or_spheroid
         azimuth_strategy.apply(lon1, lat1, b_lon_min, b_lat_max, a_b2);
         azimuth_strategy.apply(lon1, lat1, b_lon_max, b_lat_max, a_b3);
 
-        auto b0 = formula::azimuth_side_value(alp1, a_b0);
-        auto b1 = formula::azimuth_side_value(alp1, a_b1);
-        auto b2 = formula::azimuth_side_value(alp1, a_b2);
-        auto b3 = formula::azimuth_side_value(alp1, a_b3);
+        int s0 = formula::azimuth_side_value(alp1, a_b0);
+        int s1 = formula::azimuth_side_value(alp1, a_b1);
+        int s2 = formula::azimuth_side_value(alp1, a_b2);
+        int s3 = formula::azimuth_side_value(alp1, a_b3);
 
-        if (b0 == 0 || b1 == 0 || b2 == 0 || b3 == 0)
+        if (s0 == 0 || s1 == 0 || s2 == 0 || s3 == 0)
         {
             return disjoint_info::intersect;
         }
 
-        bool b0_positive = b0 > 0;
-        bool b1_positive = b1 > 0;
-        bool b2_positive = b2 > 0;
-        bool b3_positive = b3 > 0;
+        bool s0_positive = s0 > 0;
+        bool s1_positive = s1 > 0;
+        bool s2_positive = s2 > 0;
+        bool s3_positive = s3 > 0;
 
-        bool all_positive = b0_positive && b1_positive && b2_positive && b3_positive;
-        bool all_non_positive = !(b0_positive || b1_positive || b2_positive || b3_positive);
+        bool all_positive = s0_positive && s1_positive && s2_positive && s3_positive;
+        bool all_non_positive = !(s0_positive || s1_positive || s2_positive || s3_positive);
         bool vertex_north = lat1 + lat2 > 0;
 
         if ((all_positive && vertex_north) || (all_non_positive && !vertex_north))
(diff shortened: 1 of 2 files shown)


include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
```diff
@@ -186,7 +186,17 @@ struct disjoint_segment_box_sphere_or_spheroid
         bool b2 = formula::azimuth_side_value(alp1, a_b2) > 0;
         bool b3 = formula::azimuth_side_value(alp1, a_b3) > 0;
 
-        if (!(b0 && b1 && b2 && b3) && (b0 || b1 || b2 || b3))
+        bool all_ones = b0 && b1 && b2 && b3;
+        bool all_zeros = !(b0 || b1 || b2 || b3);
```
awulkiew, line 190: `azimuth_side_value` can return values in {-1, 0, 1} meaning {right, on_segment, left} respectively so this condition doesn't mean `all_zeros` It means `not_all_ones` or `either_minus_ones_or_zeros`.

include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
```diff
@@ -186,7 +186,17 @@ struct disjoint_segment_box_sphere_or_spheroid
...
+        bool vertex_north = lat1 + lat2 > 0;
+
+        if ((all_ones && vertex_north) || (all_zeros && !vertex_north))
+        {
+            return disjoint_info::disjoint_no_vertex;
+        }
+
+        if ((!all_ones && (b0 || b1 || b2 || b3) && vertex_north) ||
+            (!all_zeros && !(b0 && b1 && b2 && b3) && !vertex_north))
```
awulkiew, line 199: For now leaving the fact that `all_zeros` can be `true` if side values are `-1` this expression could be simplified to:
```
(!all_ones && !all_zeros && vertex_north) || (!all_zeros && all_ones && !vertex_north)
```

include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
vissarion: By "zeros" and "ones" I had in mind the values of b1, ..., b3. I see how this is misleading. I rephrased it.

include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
vissarion: nice catch! actually 
```
(!all_ones && !all_zeros && vertex_north) || (!all_zeros && !all_ones && !vertex_north)
```
(there was a typo above) 
Which can be further simplified to `(!all_ones && !all_zeros)`

I also added the special case of some side value is equal to `0`.

include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
```diff
@@ -181,12 +181,31 @@ struct disjoint_segment_box_sphere_or_spheroid
...
         azimuth_strategy.apply(lon1, lat1, b_lon_max, b_lat_max, a_b3);
 
-        bool b0 = formula::azimuth_side_value(alp1, a_b0) > 0;
-        bool b1 = formula::azimuth_side_value(alp1, a_b1) > 0;
-        bool b2 = formula::azimuth_side_value(alp1, a_b2) > 0;
-        bool b3 = formula::azimuth_side_value(alp1, a_b3) > 0;
+        auto b0 = formula::azimuth_side_value(alp1, a_b0);
+        auto b1 = formula::azimuth_side_value(alp1, a_b1);
+        auto b2 = formula::azimuth_side_value(alp1, a_b2);
+        auto b3 = formula::azimuth_side_value(alp1, a_b3);
```
awulkiew, line 187: The result type is known native type - `int`. It doesn't required to instantiate any template. It's even shorter than `auto`. I think that in such cases `auto` shouldn't be used.

awulkiew: Btw, since you moved from `bool` to `int` these variables are now side value so maybe it'd be good to rename them to `s0`, `s1`, etc.?

include/boost/geometry/algorithms/detail/disjoint/segment_box.hpp
```diff
@@ -181,12 +181,31 @@ struct disjoint_segment_box_sphere_or_spheroid
...
+        }
+
+        bool b0_positive = b0 > 0;
+        bool b1_positive = b1 > 0;
+        bool b2_positive = b2 > 0;
+        bool b3_positive = b3 > 0;
+
+        bool all_positive = b0_positive && b1_positive && b2_positive && b3_positive;
+        bool all_non_positive = !(b0_positive || b1_positive || b2_positive || b3_positive);
+        bool vertex_north = lat1 + lat2 > 0;
```
awulkiew, line 201: This will return `false` strictly (not using epsilon) for a segment at equator right? Can this be a problem?

awulkiew (approved):

PR #854, opened 2021-05-24
Update line_interpolate tests

vissarion:


awulkiew:
Thanks! Do you know why have the result changed?

awulkiew (approved):

vissarion:
> Thanks! Do you know why have the result changed?

Maybe the changes in line_interpolate umbrella strategies. Line interpolate tests that contain in the output the last point of the input linestring are not robust since getting that point in the computation depends on computations of distances and summation of distances. I made the tests more robust by removing or changing all those case.

awulkiew:
Ok, thanks!

PR #855, opened 2021-05-24
Umbrella strategies improvements

awulkiew:
This PR adds the following changes:
- SeriesOrder template parameter is removed from geographic umbrella strategies
- Umbrella strategy is propagated in `is_simple` and `is_valid` 
- `is_convex` umbrella strategy is added and supported in the algorithm (using the same side strategy as `convex_hull`)
- `is_convex` for Polygon is implemented
- Strategy getters (e.g. `get_point_in_geometry_strategy()`) are removed from legacy strategies
- Default umbrella strategies are included with algorithms so the user no longer needs to include strategies explicitly if only the default one is needed.

Todo (probably in a different PR or in this one if you insist):
- Removal of strategy includes in tests when they are not needed (since the defaults are included with algorithms)
- There is still one umbrella strategy missing for `remove_spikes` (`convex_hull` strategy could be reused for that, e.g. as an alias)
- Aliases of umbrella strategies for all algorithms so they are not confusing and the user knows that e.g. `within` algorithm expects `within` strategy (right now `relate` strategy is used for all relops and setops).
- Relocation of legacy strategies into `strategy/` directory

Question:
- should we use `side_robust` strategy in all algorithms by default, e.g. also in relops and setops?


include/boost/geometry/algorithms/is_convex.hpp
```diff
@@ -142,38 +161,93 @@ struct is_convex<Box, box_tag>
...
+        //   - line segments are defined by geodesics and box edges by parallels and meridians
+        //   - we use this definition: A convex polygon is a simple polygon (not self-intersecting)
+        //     in which no line segment between two points on the boundary ever goes outside the
+        //     polygon.
+        //   Then a geodesic segment would go into the exterior of a Box for all horizontal edges
+        //   of a Box unless it was one of the poles (edge degenerated to a point) or equator and
+        //   longitude difference was lesser than 360 (otherwise depending on the CS there would be
+        //   no solution or there would be two possible solutions - segment going through one of
+        //   the poles, at least in case of oblate spheroid, either way the answer would probably
+        //   be "false").
```
vissarion, line 177: +1

vissarion (approved):
> Todo (probably in a different PR or in this one if you insist):
> 
>     * Removal of strategy includes in tests when they are not needed (since the defaults are included with algorithms)
> 
>     * There is still one umbrella strategy missing for `remove_spikes` (`convex_hull` strategy could be reused for that, e.g. as an alias)
> 
>     * Aliases of umbrella strategies for all algorithms so they are not confusing and the user knows that e.g. `within` algorithm expects `within` strategy (right now `relate` strategy is used for all relops and setops).
> 
>     * Relocation of legacy strategies into `strategy/` directory

I am OK with leaving them for separate PR. 

> Question:
> 
>     * should we use `side_robust` strategy in all algorithms by default, e.g. also in relops and setops?

We should at least try. I can prepare a PR.

include/boost/geometry/algorithms/length.hpp
```diff
@@ -45,9 +45,8 @@
 #include <boost/geometry/geometries/concepts/check.hpp>
 
 #include <boost/geometry/strategies/default_strategy.hpp>
-#include <boost/geometry/strategies/distance.hpp>
-#include <boost/geometry/strategies/default_length_result.hpp>
-
+#include <boost/geometry/strategies/default_length_result.hpp> // Move to algorithms
```
barendgehrels, line 48: Can you mark this with `TODO` ?

include/boost/geometry/strategies/buffer/cartesian.hpp
```diff
@@ -32,15 +33,7 @@ struct cartesian
...
-                         std::enable_if_t
-                            <
-                                util::is_pointlike<Geometry1>::value
-                                    && util::is_segmental<Geometry2>::value
-                             || util::is_segmental<Geometry1>::value
-                                    && util::is_pointlike<Geometry2>::value
-                             || util::is_segmental<Geometry1>::value
-                                    && util::is_segmental<Geometry2>::value
-                            > * = nullptr)
+                         distance::detail::enable_if_ps_t<Geometry1, Geometry2> * = nullptr)
```
barendgehrels, line 36: :+1: Looks better than the 3 double conditions!

include/boost/geometry/strategies/is_convex/cartesian.hpp
```diff
@@ -0,0 +1,77 @@
...
+template <typename CalculationType = void>
+class cartesian : public strategies::convex_hull::cartesian<CalculationType>
+{
+public:
+    template <typename Geometry1, typename Geometry2>
+    static auto relate(Geometry1 const&, Geometry2 const&,
+                       std::enable_if_t
+                            <
+                                util::is_pointlike<Geometry1>::value
+                             && util::is_pointlike<Geometry2>::value
```
barendgehrels, line 37: This should probably be replaced by
`distance::detail::enable_if_pp_t<Geometry1, Geometry2>` ?

barendgehrels: But that is in `distance`, maybe it's more generic than that?

barendgehrels (approved):
Thanks for the huge PR!

include/boost/geometry/strategies/is_convex/cartesian.hpp
awulkiew: The problem is that the naming correspond to the strategy, not geometries, e.g. enable_if_ps is true also for "segmental" geometries so two linestrings or two polygons. This is why I put it in `distance::detail`. But I agree that in general a utility like that could be useful. I'd only have to review all of the strategies and come up with a proper names. But I'd prefer to do this in a different PR.

barendgehrels: In another PR is fine.

awulkiew:
Thanks for the reviews.

PR #857, opened 2021-05-26
Fix/buffer cluster order

barendgehrels:
This fixes the last 88 errors from the `recursive_polygons_buffer` test. Result is now:
`geometries: 310000 errors: 0 type: d time: 33.439`

There were  `2286` errors when I started to work on this part (many of them had of course the same cause) so I'm glad it's now better than the rescaled version (8 errors).

With this, the version without rescaling is behaving as good, or better, as the rescaling version. See this grep below, where the first value are the detected errors using rescaling. The last 3 values are without rescaling (double, float, long double). There is only one test better now using rescaling, then with double.

So I propose that we change the default to not-rescale this release, and we can deprecate and remove rescaling in one of the coming releases. But I've still to check the tests other than set_operations (area) / buffer. However they should be less influenced.

What are your opinions?

```
set_operations/union/union_multi.cpp:    BoostGeometryWriteExpectedFailures(9, 0, 1, 0);
set_operations/union/union.cpp:    BoostGeometryWriteExpectedFailures(3, 1, 2, 0);
set_operations/difference/difference.cpp:    BoostGeometryWriteExpectedFailures(12, 5, 11, 6);
set_operations/difference/difference_multi.cpp:    BoostGeometryWriteExpectedFailures(22, 12, 16, 7);
set_operations/intersection/intersection.cpp:    BoostGeometryWriteExpectedFailures(5, 2, 6, 1);
set_operations/intersection/intersection_multi.cpp:    BoostGeometryWriteExpectedFailures(9, 1, 2, 1);
buffer/buffer_point.cpp:    BoostGeometryWriteExpectedFailures(0);
buffer/buffer_countries.cpp:    BoostGeometryWriteExpectedFailures(1, 0, 2, 0);
buffer/buffer_multi_polygon.cpp:    BoostGeometryWriteExpectedFailures(5, 1, 3, 4);
buffer/buffer_linestring.cpp:    BoostGeometryWriteExpectedFailures(2, 4, 9, 4); **
buffer/buffer_linestring_aimes.cpp:    BoostGeometryWriteExpectedFailures(0);
buffer/buffer_multi_point.cpp:    BoostGeometryWriteExpectedFailures(0);
buffer/buffer_polygon.cpp:    BoostGeometryWriteExpectedFailures(2, 1, 9, 1);
buffer/buffer_multi_linestring.cpp:    BoostGeometryWriteExpectedFailures(9, 6, 9, 3);
```

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/traversal.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/traversal.hpp
@@ -569,28 +569,24 @@ public :
         return true;
     }
 
-
-    template <typename RankedPoint>
-    inline turn_operation_type const& operation_from_rank(RankedPoint const& rp) const
-    {
-        return m_turns[rp.turn_index].operations[rp.operation_index];
-    }
-
-    inline int select_turn_in_cluster_union(sort_by_side::rank_type selected_rank,
+    inline int priority_of_turn_in_cluster_union(sort_by_side::rank_type selected_rank,
             typename sbs_type::rp const& ranked_point,
+            std::set<signed_size_type> const& cluster_indices,
             signed_size_type start_turn_index, int start_op_index) const
     {
-        // Returns 0 if it not OK
-        // Returns 1 if it OK
-        // Returns 2 if it OK and start turn matches
-        // Returns 3 if it OK and start turn and start op both match
+        // Returns 0: not OK
+        // Returns 1: OK but next turn is in same cluster
+        // Returns 2: OK
+        // Returns 3: OK and start turn matches
+        // Returns 4: OK and start turn and start op both match
         if (ranked_point.rank != selected_rank
             || ranked_point.direction != sort_by_side::dir_to)
         {
             return 0;
         }
 
-        turn_operation_type const& op = operation_from_rank(ranked_point);
+        auto const& turn = m_turns[ranked_point.turn_index];
+        auto const& op = turn.operations[ranked_point.operation_index];
 
         // Check finalized: TODO: this should be finetuned, it is not necessary
         if (op.visited.finalized())
@@ -606,11 +602,23 @@ public :
             return 0;
         }
 
-        return ranked_point.turn_index == start_turn_index
-                && ranked_point.operation_index == start_op_index ? 3
-            : ranked_point.turn_index == start_turn_index ? 2
-            : 1
+        bool const to_start = ranked_point.turn_index == start_turn_index;
+        bool const to_start_index = ranked_point.operation_index == start_op_index;
+
+        bool const next_in_same_cluster
+                = cluster_indices.count(op.enriched.get_next_turn_index()) > 0;
+
+        return to_start && to_start_index ? 4
+            : to_start ? 3
+            : next_in_same_cluster ? 1
+            : 2
             ;
+    }
+
+    template <typename RankedPoint>
+    inline turn_operation_type const& operation_from_rank(RankedPoint const& rp) const
+    {
+        return m_turns[rp.turn_index].operations[rp.operation_index];
     }
 
     inline sort_by_side::rank_type select_rank(sbs_type const& sbs,
@@ -648,37 +656,33 @@ public :
     }
 
     inline bool select_from_cluster_union(signed_size_type& turn_index,
+        std::set<signed_size_type> const& cluster_indices,
         int& op_index, sbs_type const& sbs,
         signed_size_type start_turn_index, int start_op_index) const
     {
         sort_by_side::rank_type const selected_rank = select_rank(sbs, false);
 
-        int best_code = 0;
-        bool result = false;
+        int current_priority = 0;
         for (std::size_t i = 1; i < sbs.m_ranked_points.size(); i++)
         {
             typename sbs_type::rp const& ranked_point = sbs.m_ranked_points[i];
 
             if (ranked_point.rank > selected_rank)
             {
-                // Sorted on rank, so it makes no sense to continue
                 break;
             }
 
-            int const code
-                = select_turn_in_cluster_union(selected_rank, ranked_point,
-                    start_turn_index, start_op_index);
+            int const priority = priority_of_turn_in_cluster_union(selected_rank,
+                ranked_point, cluster_indices, start_turn_index, start_op_index);
 
-            if (code > best_code)
+            if (priority > current_priority)
             {
-                // It is 1 or higher and matching better than previous
-                best_code = code;
+                current_priority = priority;
                 turn_index = ranked_point.turn_index;
                 op_index = ranked_point.operation_index;
-                result = true;
             }
         }
-        return result;
+        return current_priority > 0;
     }
 
     inline bool analyze_cluster_intersection(signed_size_type& turn_index,
@@ -731,15 +735,13 @@ public :
 
     inline bool fill_sbs(sbs_type& sbs,
                          signed_size_type turn_index,
-                         std::set<signed_size_type> const& ids,
+                         std::set<signed_size_type> const& cluster_indices,
                          segment_identifier const& previous_seg_id) const
     {
-        for (typename std::set<signed_size_type>::const_iterator sit = ids.begin();
-             sit != ids.end(); ++sit)
+
+        for (auto cluster_turn_index : cluster_indices)
         {
-            signed_size_type cluster_turn_index = *sit;
             turn_type const& cluster_turn = m_turns[cluster_turn_index];
-            bool const departure_turn = cluster_turn_index == turn_index;
             if (cluster_turn.discarded)
             {
                 // Defensive check, discarded turns should not be in cluster
@@ -752,7 +754,7 @@ public :
                         cluster_turn.operations[i],
                         cluster_turn_index, i, previous_seg_id,
                         m_geometry1, m_geometry2,
-                        departure_turn);
+                        cluster_turn_index == turn_index);
             }
         }
 
@@ -780,16 +782,17 @@ public :
         BOOST_ASSERT(mit != m_clusters.end());
 
         cluster_info const& cinfo = mit->second;
-        std::set<signed_size_type> const& ids = cinfo.turn_indices;
+        std::set<signed_size_type> const& cluster_indices = cinfo.turn_indices;
 
         sbs_type sbs(m_strategy);
 
-        if (! fill_sbs(sbs, turn_index, ids, previous_seg_id))
+
+        if (! fill_sbs(sbs, turn_index, cluster_indices, previous_seg_id))
         {
             return false;
         }
 
-        cluster_exits<OverlayType, Turns, sbs_type> exits(m_turns, ids, sbs);
+        cluster_exits<OverlayType, Turns, sbs_type> exits(m_turns, cluster_indices, sbs);
 
         if (exits.apply(turn_index, op_index))
         {
@@ -800,8 +803,9 @@ public :
 
         if (is_union)
         {
-            result = select_from_cluster_union(turn_index, op_index, sbs,
-                start_turn_index, start_op_index);
+            result = select_from_cluster_union(turn_index, cluster_indices,
+                                               op_index, sbs,
+                                               start_turn_index, start_op_index);
             if (! result)
             {
                // There no way out found, try second pass in collected cluster exits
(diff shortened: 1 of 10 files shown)


include/boost/geometry/algorithms/detail/overlay/traversal.hpp
```diff
@@ -606,13 +602,25 @@ public :
...
 
-        return ranked_point.turn_index == start_turn_index
-                && ranked_point.operation_index == start_op_index ? 3
-            : ranked_point.turn_index == start_turn_index ? 2
-            : 1
+        bool const to_start = ranked_point.turn_index == start_turn_index;
+        bool const to_start_index = ranked_point.operation_index == start_op_index;
+
+        bool const next_in_same_cluster
+                = cluster_indices.count(op.enriched.get_next_turn_index()) > 0;
```
barendgehrels, line 609: This is the fix. In a cluster, it should prefer a way out of the cluster.

test/algorithms/set_operations/difference/difference.cpp
```diff
@@ -636,7 +629,7 @@ int test_main(int, char* [])
 #if defined(BOOST_GEOMETRY_TEST_FAILURES)
     // Not yet fully tested for float and long double.
     // The difference algorithm can generate (additional) slivers
-    BoostGeometryWriteExpectedFailures(10, 7, 22, 8);
+    BoostGeometryWriteExpectedFailures(12, 5, 11, 6);
```
barendgehrels, line 632: (the 2 extra for rescaled is not a regression, one test now also tests symmetric difference which fails for rescaling but succeeds otherwise)

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
```diff
@@ -569,28 +569,24 @@ public :
...
     {
-        // Returns 0 if it not OK
-        // Returns 1 if it OK
-        // Returns 2 if it OK and start turn matches
-        // Returns 3 if it OK and start turn and start op both match
+        // Returns 0: not OK
+        // Returns 1: OK but next turn is in same cluster
+        // Returns 2: OK
+        // Returns 3: OK and start turn matches
+        // Returns 4: OK and start turn and start op both match
```
awulkiew, line 581: Why not make an enum describing this? These values are not used to calculate something are they?

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
awulkiew: `cluster_indices` is of type `std::set<signed_size_type>` which means that `count` returns either `0` or `1`. Is that what was desired or did you mean to use `std::multi_set` or `std::map<signed_size_type, size_t>` to count the indexes or maybe even `std::unordered_map<signed_size_type, size_t>` ?

awulkiew:
>What are your opinions?

Thanks! I'm all for removing rescaling.

Are all of the MySQL test cases in various algorithms passing?

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
```diff
@@ -606,13 +602,25 @@ public :
...
-                && ranked_point.operation_index == start_op_index ? 3
-            : ranked_point.turn_index == start_turn_index ? 2
-            : 1
+        bool const to_start = ranked_point.turn_index == start_turn_index;
+        bool const to_start_index = ranked_point.operation_index == start_op_index;
+
+        bool const next_in_same_cluster
+                = cluster_indices.count(op.enriched.get_next_turn_index()) > 0;
+
+        return to_start && to_start_index ? 4
```
vissarion, line 611: +1 for using `enum` here

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
```diff
@@ -667,6 +676,7 @@ public :
 
             int const code
                 = select_turn_in_cluster_union(selected_rank, ranked_point,
+                                               cluster_indices,
                     start_turn_index, start_op_index);
```
vissarion, line 680: indentation issue here

test/algorithms/set_operations/difference/difference.cpp
```diff
@@ -345,17 +345,18 @@ void test_all()
 #endif
 
     {
-        //geos_2_a #area expected: 138.69238279999999008 detected: 138.53125 type: f
```
vissarion, line 348: is this replaced by `expectation_limits` below? Should we keep stating the library and version  (geos?) used to produce the expected results?

test/algorithms/set_operations/difference/difference.cpp
```diff
@@ -345,17 +345,18 @@ void test_all()
...
-        //geos_2_b #area expected: 211.859375 detected: 210.53125 type: f
+        ut_settings settings;
+        settings.set_test_validity(BG_IF_RESCALED(false, true));
 
-        // MSVC 14 expects 138.69214 and 211.85913: increase percentage
         // Output polygons for sym difference might be combined
+        expectation_limits a{138.5312, 138.6924};
```
vissarion, line 352: where are these numbers come from?

test/algorithms/set_operations/difference/difference.cpp
vissarion: sorry, I didn't get where the extra '2' comes from

test/robustness/overlay/buffer/recursive_polygons_buffer.cpp
```diff
@@ -133,27 +129,34 @@ bool verify(std::string const& caseid, MultiPolygon const& mp, MultiPolygon cons
...
     {
-        std::ostringstream filename;
-        filename << caseid << "_"
-            << typeid(typename bg::coordinate_type<MultiPolygon>::type).name()
-            << ".wkt";
-        std::ofstream stream(filename.str().c_str());
+        std::ofstream stream(filename + "wkt");
         stream << bg::wkt(mp) << std::endl;
-        stream << bg::wkt(buffer) << std::endl;
+        if (false)
```
vissarion, line 155: is this part of code needed?

test/robustness/overlay/buffer/recursive_polygons_buffer.cpp
```diff
@@ -213,7 +216,7 @@ bool test_buffer(MultiPolygon& result, int& index,
     bg::strategy::buffer::end_round end_strategy;
     bg::strategy::buffer::point_circle point_strategy;
     bg::strategy::buffer::side_straight side_strategy;
-    bg::strategy::buffer::join_round join_round_strategy(32); // Compatible with MySQL
+    bg::strategy::buffer::join_round join_round_strategy(settings.points_per_circle);
```
vissarion, line 219: what does "compatible with MySQL" means and why it is removed ?

vissarion (approved):
Thanks, I have a few questions but in general I am OK for merging.

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
barendgehrels: Indeed, `count` returns `1` if it is in the set, and that is what I'm using (I like it more than checking `find` and `end`)
So it should not be a map or a multiset

test/algorithms/set_operations/difference/difference.cpp
barendgehrels: From symmetric difference which is now also tested for that case (resulting in 2 errors)

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
barendgehrels: I'm not sure about that. We just need an order here, and we don't have to name the states.
The comments are to explain the ordering and only relevant for this place, and no other place only.
If it would be an enum (possible of course) we would only assign it and never use it, and it would be more error-prone if people accidentally change the order later (or we should assign values to them)

barendgehrels:
> Are all of the MySQL test cases in various algorithms passing?

I'll check

test/algorithms/set_operations/difference/difference.cpp
barendgehrels: From latest tests.
Especially if there are nearly collinear segments, the calculated areas can differ.
If they differ I always check them visually (with SVG's)

test/robustness/overlay/buffer/recursive_polygons_buffer.cpp
barendgehrels: It's not removed, and 32 still the default.  But it's flexible now.
It means that MySQL also uses 32 points per circle - but actually for this test that is not very relevant so I removed that comment here.

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
barendgehrels: Thanks, conventions changed...

test/algorithms/set_operations/difference/difference.cpp
barendgehrels: In the past I incidentally mentioned used libraries or tools, but it's too much work to do that everywhere.
And now that we have a range of allowed values it's kind of redundant.

test/robustness/overlay/buffer/recursive_polygons_buffer.cpp
barendgehrels: Not really, I can omit it. If it is wished it's trivial to add again.

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
awulkiew: Ok, you probably want to check whether or not an id was found, and how many times is not important. Correct?

Then I'd only suggest to think about `unordered_set<>` because the complexity of the lookup is O(1) on average, instead of O(logN). And AFAIU you only care to find if the id exists and sorting is not important. On the other hand it has higher constant cost than `set<>` and will become faster above some number of ids. So it depends on how many clusters do you expect to deal with.

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
awulkiew: Ah ok, I see now that this function is called below in `select_from_cluster_union` and what is returned here is some kind of priority. So indeed an enum is not needed because the caller is not concerned about specific values, only about relative priority.

But AFAIU this means that even though this function is called `select_turn_in_cluster_union` it doesn't select anything. It calculates this priority value instead. If my understanding is correct I propose to rename it accordingly. To e.g. `priority_of_turn_in_cluster_union()` or whatever you think reflects values that are returned the best.

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
barendgehrels: Sure, correct.

The number of clusters is usually very low. It’s where turns are colocated and that does not happen often in nature. (In our testsets it happens more because they are constructed for exactly that reason, these cases are harder.)

So just `set` is probably fine then. Thanks for your view.

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
barendgehrels: Thanks, good idea, `priority...` is good, I will rename the function

barendgehrels: :heavy_check_mark:

barendgehrels:
I updated the code but this PR doesn't reflect it yet (though the commits are on the server, you can see that in the branch it wants to merge). Normally it's reflected immediately.

barendgehrels:
> I'll check
I see one MySQL cases which is still failing (`mysql_23023665_6` for `difference`), and, on the other hand, one case (`touching1_1` in `buffer_multi_linestring` which is derived from `mysql_23023665_1`) which is now succeeding

So it looks in balance. But I will much do more extensive tests first, before really changing the default behavior, and keep these cases in mind.

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
barendgehrels: It's now a `priority`, see conversation with Adam. I assume that's OK for you too.

barendgehrels:
> @barendgehrels Yes, I'm ok with merging.
> 
> Before disabling rescaling please make sure that this test is passing as well.

Sure, I will investigate it now. Thanks.

PR #860, opened 2021-06-02
[strategies] avoid warnings

barendgehrels:


vissarion (approved):

barendgehrels:
I'm now reviewing #855 and see many (or all) are replaced now (good!).
@awulkiew I can decline this PR then to avoid conflicts.

awulkiew:
This is already done here in a different way: https://github.com/boostorg/geometry/pull/855
See e.g.: https://github.com/boostorg/geometry/blob/b90a3116a5f4f3cadaf12a30f59ed41c166f330f/include/boost/geometry/strategies/buffer/spherical.hpp#L50

PR #862, opened 2021-06-07
Ranges, views and iterators improvements

awulkiew:
This PR makes changes allowing passing rvalue, temporary ranges and proxies without the need to explicitly create variables into `range::` utilities. Views and iterators are refactored to either allow to create them as temporaries or to simplify the code.

The initial motivation was to replace this:
```
typename traits::interior_mutable_type<Polygon>::type
    rings = geometry::interior_rings(poly);
range::push_back(rings, std::move(ring));
```
or C++11 equivalent:
```
auto&& rings = geometry::interior_rings(poly);
range::push_back(rings, std::move(ring));
```
with this:
```
range::push_back(geometry::interior_rings(poly), std::move(ring));
```
in case when `geometry::interior_rings(poly)` returns a temporary object like a proxy which is the case e.g. with Boost.Polygon adaptation.

The list of the changes:
- rvalue references are now allowed in `range::` utilities
- pointers/references to ranges were removed from some iterators (`closing_iterator`) and views (`detail::closing_view`, `identity_view`) because the ranges pointed to could be destroyed temporaries and the pointers would be dangling
- iterators conversion constructors were implemented using `enable_if` v.s. `static_assert`
- `has_one_element` was moved to `detail::` namespace and moved to `detail/` directory
- `base.hpp` file defining `detail::iterator_base` was moved to `detail/iterator_base.hpp` file
- `detail::normalized_view` was renamed to `detail::reverse_close_view` and its implementation simplified
- `detail::close_reverse_view` was added as a counterpart performing closing and reversing in the opposite order. In some of the functions we use one order and in the other functions the opposite. We should probably replace all places with the new `detail::close_reverse_view` because this one preserves the first point of the range.
- `convert()` now preserves the first point of the range after the change from `reversible_view` followed by `closeable_view` to `detail::close_reverse_view`
- `box_view` and `segment_view` were reimplemented (see the comments), `detail::point_view` removed and replaced with `std::array`.
- `detail::range_type` was removed and its usage replaced with `ring_type`, the reason behind this is that it is not really needed and is not consistent semantically. It was the same as `ring_type` with the exception that it could define `box_view` which was semantically different. This is because `box_view` is not a view in the C++20 sense since it owns the points, it's a container or a geometry. Furthermore `detail::range_type` for Box was not used anywhere in the library.
- `detail::as_range()` now can return proxy ranges by value as defined by the users in traits
- added several aliases for views taking various arguments (see the "For the future" below):
  - `detail::clockwise_close_view`
  - `detail::clockwise_view`
  - `detail::close_clockwise_view`
  - `detail::close_view`
  - `detail::reverse_view`
- the `reverse` views takes Direction, the `clockwise` views takes Order, the `close` views takes Closure, the defaults for all arguments are taken from Range
- added utility `detail::minimum_ring_size` taking Geometry instead of Closure
- `closeable_view` and `reversible_view` usage replaced in algorithms, the utilities left for backward compatibility

Currently we have three kinds of "views":
- proper views in C++20 sense, constant time copy, not owning the base range, can be temporaries, etc. (e.g. `identity_view` or `detail::closing_view`)
- geometries that doesn't own the base geometry (e.g. `detail::indexed_point_view`, `detail::two_dimensional_view`, probably `detail::boundary_view` but I have to look into it, it may be in yet another category corresponding to C++20 range factory). These are proxy geometries and probably should be named as such, e.g. `detail::point_of_indexed_proxy` or something similar.
- geometries that owns the undelying geometries, which are copied from the source, (e.g. `box_view`, `segment_view`), these are proper geometries, of constant size like `array` or `static_vector` but just like them are still containers these are still geometries. So they should probably be moved entirely from `views/` and renamed to e.g. `static_ring` and `static_linestring` respectively.

Do you have different ideas for the names?
How to call a proxy geometry with one word? Consider "container" vs "range".

I'm open to suggestions.

----

For the future:
(actually not anymore since some of the issues mentioned below are addressed in this PR)

Regarding the names, AFAIU the naming convention for C++20 range adaptors uses simple verbs or nouns in plural form (in one instance in past tense), e.g. `transform_view`, `reverse_view`, `elements_view`, `keys_view` (see: https://en.cppreference.com/w/cpp/ranges). So we could think about renaming our views and iterators according to this convention, or creating aliases to maintain backward compatibility or creating new ones. E.g. there would be `close_view` or `closed_view` but not `closing_view` or `closeable_view`. However C++20 `std::ranges::reverse_view` always reversed the range unlike the view defined in `reversible_view` so we could pick a different name for that, e.g. `conditionally_close_view` and `conditionally_reverse_view`. Note that the std metafunction allowing to pick a type at compile time is called `std::conditional`. Another alternative for `reverse_view` is a name with `clockwise` in it because this is what this view is used for in the library, to make rings clockwise. For that the parameter would have to be changed from Direction to PointOrder which would be semantically consistent with `close_view`, see below.

Another thing is that `closeable_view` and `reversible_view` utilities are not needed. Distinguishing wheter or not close or reverse a range should be handled in the view itself.

Yet another thing is that the arguments of `closeable_view` and `reversible_view` are not consistent semantically. `closeable_view` takes an atribute of a geometry, so whether or not a geometry is closed or opened and based on that doesn't close or does close respectively (so it does the opposite of the attribute passed as parameter). `reversible_view` on the other hand takes a parameter telling it directly what to do, so whether or not the range should be reversed or not.

So I think these issues could be a base for future changes.


include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
```diff
@@ -537,7 +537,7 @@ struct buffer_inserter_ring
 
         if (n >= min_points)
         {
-            detail::normalized_view<RingInput const> view(simplified);
+            detail::reverse_close_view<RingInput const> view(simplified);
```
barendgehrels, line 540: Why is this a `reverse` view? This looks wrong to me. But maybe it's the name changes I don't fully understand yet.

barendgehrels: I believe I get it now. It is a view that **can** reverse and it **can** close, both if necessary.

In the past this was the definition of **normalized** 

So it should be correct, but I actually prefer the old name (or another name)

include/boost/geometry/views/detail/normalized_view.hpp
```diff
@@ -36,72 +36,99 @@ namespace boost { namespace geometry {
...
 
 private:
-    reversible_type m_reversible;
-    closeable_type m_closeable;
+    view m_view;
 };
 
+
+// Other possible names:
+//   clockwise_closed_view
```
barendgehrels, line 101: Why do we need it actually? Because it looks identical to the `reverse_close_view`

And indeed, a `clockwise_closed_view` is clockwise and closed, and that looks intuitive to me.

include/boost/geometry/views/detail/normalized_view.hpp
```diff
@@ -36,72 +36,99 @@ namespace boost { namespace geometry {
...
+
+// Other possible names:
+//   closed_clockwise_view
+template
+<
+    typename Range,
+    closure_selector Closure = geometry::closure<Range>::value,
+    order_selector Order = geometry::point_order<Range>::value    
+>
+using close_clockwise_view = close_reverse_view
```
barendgehrels, line 124: Same here, why do we need it, and why do we need two variants in different orders? First reversing and then closing should have the same effect as closing and then reversing... or are there exceptions here?

include/boost/geometry/views/reversible_view.hpp
```diff
@@ -22,20 +22,57 @@
...
-\ingroup views
-*/
-enum iterate_direction { iterate_forward, iterate_reverse };
+
+#ifndef DOXYGEN_NO_DETAIL
+namespace detail
+{
+
+// Other possible names:
+//   conditionally_reverse_view
```
barendgehrels, line 43: I would keep `reversible_view`
Or maybe it's not the best name.
So we want to view a range it as clockwise, regardless of the underlying orientation.
`clockwise_view` as you have somewhere?

include/boost/geometry/views/reversible_view.hpp
```diff
@@ -53,19 +90,14 @@ struct reversible_view {};
 template <typename Range>
 struct reversible_view<Range, iterate_forward>
 {
-    typedef identity_view<Range> type;
+    using type = identity_view<Range>;
 };
 
 
 template <typename Range>
 struct reversible_view<Range, iterate_reverse>
```
barendgehrels, line 98: Ah, it's still here.

barendgehrels (approved):
Thanks for this, again, huge change and improvement. At many placesl it's becoming simpler, also because we can use C++14 now, and I like that.
I'm not sure about all the namings, and there are still some TODO's left, but the PR in general is approved.

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
awulkiew: This naming is taken from the C++20 convention. Have a look at this: https://en.cppreference.com/w/cpp/ranges
e.g.: `drop_view`, `take_view`, `transform_view`, `split_view`, `join_view`, `reverse_view`. AFAIU the names are the same as corresponding algorithms/functions, i.e. do X with this view. So `reverse_close_view` is short for `reverse_and_close_view`.

Boost.Range have been using a different convention naming views by what a range becomes after applying a view, e.g. `reversed_view` is a range that was reversed. Using this convention it'd be `reversed_and_closed_view` or `reversed_closed_view`.

I'm open to suggestion regarding the naming but I think that `normalized_view` is a wrong name because normalization may mean different things. So I'd prefer to name the views to mean precisely what they are doing or becoming, whatever this is.

include/boost/geometry/views/detail/normalized_view.hpp
awulkiew: I described the reasoning at the bottom of the PR description. There are both views because in some cases Direction is used and in other cases Order is used. With Order semantically corresponding to Closure. In practice some algorithms pass `Reverse` template argument which is then translated to `Direction` while in other algorithms the `Order` is taken from Geometry and this is translated to `Direction`. So to avoid translations in many algorithms I added these redundant views doing the translation internally.

Personally I think we need only one view. The one that first closes the geometry and then reverses it. There are two problems with this:
- in some algorithms the different order of operations is performed so we have to check if the algorithm fails or not, e.g. `buffer()`
- in many algorithms `Reverse` template argument is passed explicitly instead being taken from Geometry in place where it's needed
After both of these are changed and tested we could have only one view. Furthermore, Order->Direction translation will not be needed anymore.

include/boost/geometry/views/detail/normalized_view.hpp
awulkiew: I described this above, but to reiterate.

> Same here, why do we need it, and why do we need two variants in different orders?

Because in some algorithms `Reverse` parameter is passed and in other algorithms `Order` is taken from Geometry, so to avoid translating them to `Direction` in each algorithm this is done here once.

> First reversing and then closing should have the same effect as closing and then reversing... or are there exceptions here?

The difference is that when we first reverse and then close the first point is not the same as in the original range. Geometrically they are the same, but the results of some algorithms will be different. Currently we have algorithms performing these operations in two different orders and this PR is not intended to change that. It's because I didn't want to change the results of e.g. set operations or buffer. This might also interfere with your rescaling changes. I think that we should change that in the future though but this requires testing.

include/boost/geometry/views/reversible_view.hpp
awulkiew: I'd prefer `clockwise_view` too but this view takes `Direction` which either means "reverse" or "forward" and this has nothing to do with point order semantically. For this view to be replaced with `clockwise_view` we'd have to remove all instances of `Reverse` to `Direction` translation from the library and replace these with a view taking `Order` directly from Geometry.

Well, I guess we could replace `Reverse` to `Direction` with `Reverse` to `Order` but this would be confusing.

include/boost/geometry/views/reversible_view.hpp
awulkiew: Yes, these metafunctions are here only for backward compatibility.

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
awulkiew: I'd like to stress one more thing. The difference between these views and C++20 ones is that these are done conditionally unlike e.g. `std::ranges::reverse_view` which always reverses a range, just like `boost::reversed_view` or `std::reverse` or in fact `boost::geometry::reverse()`. This is why I added `conditionally` in the alternatives.

include/boost/geometry/views/detail/normalized_view.hpp
awulkiew: One more thing, regarding:

> It's because I didn't want to change the results of e.g. set operations or buffer for opened rings.

The exception is `convert()` which I modified in this PR to preserve the first point for opened rings which makes the result strictly equal to the one for closed rings.

awulkiew: On second thought. Correct me if I'm wrong. AFAIR we mainly test algorithms for clockwise, closed geometries anyway. So I think it would be quite safe to change the order of closing and reversing even now. If anything this would make the results more consistent to what we test against. So I think I could change it now and remove the redundant views (`reverse_close_view` and `clockwise_close_view`) in this PR. What do you think?

awulkiew:
@barendgehrels I decided to remove the redundant views, rename them and leave only the following:
- `closed_view`
- `clockwise_view`
- `closed_clockwise_view`

which I think better describes them (`closed` vs `close` and `reverse` not needed). My reasoning is that a verb could be used if a view always did something (like e.g. C++20 `reverse_view`), but here it's conditional and it's a representation of a different abstraction, i.e. closed range with clockwise order of points, a new thing, so here a noun is more suitable (like e.g. C++20 `elements_view`). Does it make sense?

I think that `Reverse` to `Order` conversion looks good and is not confusing as I initially thought it'd be.

So after this change the following are not used anymore:
- `closeable_view`
- `iterate_direction`
- `order_as_direction`
- `reversible_view`

but left for backward compatibility

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
barendgehrels: Thanks for your explanations. It's clear and I agree. Indeed, normalization is ambiguous.
`Conditionally` is OK, a bit long though.

include/boost/geometry/views/detail/normalized_view.hpp
barendgehrels: I would welcome one view. We most probably might remove the `Reverse` template argument in those cases.
So I used another order of operations in buffer, I was not aware of that. But it should result into the same source geometry, and therefore in the same buffer...

barendgehrels:
> I decided to remove the redundant views, rename them and leave only the following

This looks very clear to me! Agree!

barendgehrels:
> but left for backward compatibility

Can we clearly deprecate them? Then we can remove it earlier. I doubt if they are used from outside, but some people might make their own algorithms with our base functionality.

include/boost/geometry/algorithms/convert.hpp
```diff
@@ -175,11 +164,16 @@ struct range_to_range
     {
         geometry::clear(destination);
 
-        rview_type rview(source);
+        using view_type = detail::closed_clockwise_view
+            <
+                Range1 const,
+                geometry::closure<Range1>::value,
+                Reverse ? counterclockwise : clockwise
+            >;
```
barendgehrels, line 172: Clear, I'm glad with this.
One question, isn't it sufficient to just pass `Reverse` here? Do we need the ternary?
I'm fine either way, you can decide.

barendgehrels: OK, I just reviewed the new definition of `closed_clockwise_view`. It makes sense like this, no need to change.
And I now understand better that if we pass `Reverse`, it should be specified here; if not, it's retrieved automatically.

But sometimes we need `Reverse` as a separate template parameter (for example for `difference` and therefore in general for all overlay), because we walk those geometries in opposite direction, on purpose.

barendgehrels:
I approved it already and approve it again :smiley: 
I hope it's merged soon, to avoid conflicts.

include/boost/geometry/views/detail/normalized_view.hpp
awulkiew: The different starting point is generated only for open, ccw rings and AFAIK we mainly test with closed cw rings.

include/boost/geometry/algorithms/convert.hpp
awulkiew: Yes, this is why the Reverse to Order (and previously Reverse to Direction) translation is here.

Another possibility would be to pass the Order instead of Reverse but I'd have to check all of the places to see if this is more clear.

awulkiew:
> Can we clearly deprecate them? Then we can remove it earlier. I doubt if they are used from outside, but some people might make their own algorithms with our base functionality.

Yes, we could. In the docs and in the code with https://en.cppreference.com/w/cpp/language/attributes/deprecated
But I think I'd prefer to do this in a different PR, and maybe at the beginning of the next version cycle.
Is that reasonable?

What do you think about renaming `box_view` and `segment_view` to e.g. `static_ring` and `static_segment` (because they own the points)?
What do you think about renaming e.g. `indexed_point_view` to e.g. `indexed_point_proxy` (because it contains a reference to geometry)?
The reason would be to distinguish between views and other entities.

include/boost/geometry/algorithms/detail/convex_hull/interface.hpp
```diff
@@ -88,17 +86,18 @@ struct hull_to_geometry
     static inline void apply(Geometry const& geometry, OutputGeometry& out,
             Strategy const& strategy)
     {
+        // TODO: Why not handle multi-polygon here?
+        // TODO: detail::as_range() is only used in this place in the whole library
+        //       it should probably be located here.
+        // NOTE: A variable is created here because this can be a proxy range
+        //       and back_insert_iterator<> can store a pointer to it.
+        // Handle linestring, ring and polygon the same:
+        auto&& range = detail::as_range(out);
```
vissarion, line 95: I am not sure if `auto` is of good use here since it is difficult to derive the type of `range`

include/boost/geometry/algorithms/detail/equals/collect_vectors.hpp
```diff
@@ -327,17 +326,13 @@ struct range_collect_vectors
...
-
-        apply_impl(collection, normalized_range_type(range));
+        apply_impl(collection,
+                   detail::closed_clockwise_view<Range const>(range));
     }
 
 private:
-    template <typename NormalizedRange>
-    static inline void apply_impl(Collection& collection, NormalizedRange const& range)
+    template <typename RevClRange>
```
vissarion, line 334: What `RevClRange` means? Reverse closed or reversed clockwise range? `normalized_view` was renamed to  `closed_clockwise_view` should NormalizedRange be renamed accordingly, or I misunderstood something?

include/boost/geometry/algorithms/detail/is_valid/has_duplicates.hpp
```diff
@@ -62,10 +56,10 @@ struct has_duplicates
...
             > equal;
 
-        const_iterator it = boost::const_begin(view);
-        const_iterator next = it;
-        ++next;
-        for (; next != boost::const_end(view); ++it, ++next)
+        auto it = boost::begin(view);
+        auto const end = boost::end(view);
+        auto next = it;
+        for (++next; next != end; ++it, ++next)
```
vissarion, line 62: why not 
```
        auto it = boost::begin(view);
        auto next = it;
        for (++next; next != boost::end(view); ++it, ++next)
```
?

vissarion (approved):
Thanks I am OK with this PR.

include/boost/geometry/algorithms/detail/convex_hull/interface.hpp
awulkiew: What do you mean by "it is difficult to derive the type of `range`"?

Do you mean that you'd prefer the following? Or do you have something else in mind?
```
typename ring_return_type<Geometry>::type range = detail::as_range(out);
```

Btw, I'm working on a different PR in which this will be removed.

include/boost/geometry/algorithms/detail/equals/collect_vectors.hpp
awulkiew: Yes, I'll change the name.

include/boost/geometry/algorithms/detail/is_valid/has_duplicates.hpp
awulkiew: Because AFAIU otherwise the new end iterator has to be created in each iteration.

include/boost/geometry/algorithms/detail/convex_hull/interface.hpp
vissarion: > What do you mean by "it is difficult to derive the type of `range`"?
> 
> Do you mean that you'd prefer the following? Or do you have something else in mind?
> 
> ```
> typename ring_return_type<Geometry>::type range = detail::as_range(out);
> ```

yes this is what I mean, it is more clear 

> Btw, I'm working on a different PR in which this will be removed.

OK so there should not be any issue then.

include/boost/geometry/algorithms/detail/is_valid/has_duplicates.hpp
vissarion: OK, if there is a creation in `boost::end(view)` I agree.

vissarion:
> > Can we clearly deprecate them? Then we can remove it earlier. I doubt if they are used from outside, but some people might make their own algorithms with our base functionality.
> 
> Yes, we could. In the docs and in the code with https://en.cppreference.com/w/cpp/language/attributes/deprecated
> But I think I'd prefer to do this in a different PR, and maybe at the beginning of the next version cycle.
> Is that reasonable?
> 
> What do you think about renaming `box_view` and `segment_view` to e.g. `static_ring` and `static_segment` (because they own the points)?
> What do you think about renaming e.g. `indexed_point_view` to e.g. `indexed_point_proxy` (because it contains a reference to geometry)?
> The reason would be to distinguish between views and other entities.

I am OK will all, except `static_ring`, why not `static_box`?

awulkiew:
> I am OK will all, except `static_ring`, why not `static_box`?

@vissarion Because this geometry stores 4 or 5 points and is adapted to Ring concept. I also made a mistake in the name `static_segment`. It should've been `static_linestring` (I edited my question). The reason is the same. They both store copies of points and have size known at compile-time.

include/boost/geometry/algorithms/detail/convex_hull/interface.hpp
awulkiew: > OK so there should not be any issue then.

I'm using `auto&&` there too, e.g.:
```
auto&& ring = exterior_ring(polygon);
```
so it's possible that we'll have the same discussion there.

I'll try to explain how I see this and maybe you'll agree. If not please explain why do you think this this is harder to understand because I think otherwise.

In general `typename ring_return_type<Geometry>::type` is generic and can be of arbitrary type. In Boost.Geometry in majority of cases this is a reference but it can be a proxy type since AFAIR this is defined as a trait. However to a person not knowing how Boost.Geometry works on the first sight this looks like a value (not reference). At least I think this is what a random person would think. That this line would be a copy:
```
typename ring_return_type<Geometry>::type range = detail::as_range(out);
```

In the case of universal reference `auto&&` it's clear that range is of generic type and can be a reference or value. So I think that it's more clear than the alternative.

barendgehrels:
> > > Can we clearly deprecate them? Then we can remove it earlier. I doubt if they are used from outside, but some people might make their own algorithms with our base functionality.
> > 
> > 
> > Yes, we could. In the docs and in the code with https://en.cppreference.com/w/cpp/language/attributes/deprecated
> > But I think I'd prefer to do this in a different PR, and maybe at the beginning of the next version cycle.
> > Is that reasonable?
> > What do you think about renaming `box_view` and `segment_view` to e.g. `static_ring` and `static_segment` (because they own the points)?
> > What do you think about renaming e.g. `indexed_point_view` to e.g. `indexed_point_proxy` (because it contains a reference to geometry)?
> > The reason would be to distinguish between views and other entities.
> 
> I am OK will all, except `static_ring`, why not `static_box`?

I'm OK with these new names. And indeed I would prefer to have it in a different PR.

include/boost/geometry/views/detail/normalized_view.hpp
barendgehrels: Was this question still relevant?
I'm fine with it.

awulkiew: Not relevant, this was already done.

awulkiew:
Thanks for the reviews! I'll make other changes we talked about in a different PR.

Issue #865, opened 2021-06-11
serializer false error


maliberty:
In the attached program I try to serialize/deserialize a tree with a single element.  It fails on deserialize in:

        if ( elements_count < parameters.get_min_elements() || parameters.get_max_elements() < elements_count )
            BOOST_THROW_EXCEPTION(std::runtime_error("rtree loading error"));

since there is just one element the root node must violate the min_elements rule (min of 4 in this case).  This shouldn't be an error.   I expect this is easy to fix but I don't know the internals.  
[test.cpp.gz](https://github.com/boostorg/geometry/files/6635560/test.cpp.gz)

awulkiew:
Thanks for the report! Yes, root leaf node should be able to hold smaller number of elements than the minimum.

PR #866, opened 2021-06-11
Fix rtree serialization load for 1-level tree with number of elements lesser than min.

awulkiew:
This PR fixes https://github.com/boostorg/geometry/issues/865


vissarion (approved):
Looks OK to me

Issue #869, opened 2021-06-18
Sym difference may generate polygon with disconnected interior


kleunen:
Please see this example:
https://wandbox.org/permlink/Ib8ZrLmUlRI8UZEU

````C++
#include <iostream>

#include <boost/geometry.hpp>
#include <boost/geometry/geometries/point_xy.hpp>
#include <boost/geometry/geometries/polygon.hpp>
#include <boost/geometry/geometries/multi_polygon.hpp>

using point_t = boost::geometry::model::d2::point_xy<double>;
using polygon_t = boost::geometry::model::polygon<point_t>;
using ring_t = boost::geometry::model::ring<point_t>;
using multi_polygon_t = boost::geometry::model::multi_polygon<polygon_t>;


int main()
{
    multi_polygon_t a;
    multi_polygon_t b;
    multi_polygon_t c;
    
    boost::geometry::read_wkt("MULTIPOLYGON(((70 50,70 10,10 10,10 70,50 70,50 90,70 90,70 70,90 70,90 50,70 50)))", a);
    boost::geometry::read_wkt("MULTIPOLYGON(((50 50,50 70,70 70,70 50,50 50)))", b);
    boost::geometry::read_wkt("MULTIPOLYGON(((50 50,50 30,30 30,30 50,50 50)))", c);
    
    multi_polygon_t result;
    boost::geometry::sym_difference(b, c, result);
    std::cout << boost::geometry::wkt(result) << std::endl;
    
    multi_polygon_t result_2;
    boost::geometry::sym_difference(result, a, result_2);
    
    std::string message;
    if(!boost::geometry::is_valid(result_2, message))
        std::cout << "Geometry is not valid: " << message << std::endl;
    
    std::cout << boost::geometry::wkt(result_2) << std::endl;
    return 0;
}
````

Expected output is a valid polygon, output is: 
Geometry is not valid: Geometry has disconnected interior
MULTIPOLYGON(((50 70,50 90,70 90,70 70,90 70,90 50,70 50,70 10,10 10,10 70,50 70),(50 70,50 50,70 50,70 70,50 70),(50 50,30 50,30 30,50 30,50 50)))

**Comment by awulkiew** [comment]
Thanks! I confirm this is a bug.

So reduced issue is that the result of `difference()`:
```
MULTIPOLYGON(((70 50,70 10,10 10,10 70,50 70,50 90,70 90,70 70,90 70,90 50,70 50)))
\
[...]

PR #872, opened 2021-06-23
[traverse] modernize code and add documentation, make clusters const

barendgehrels:
This documents traversal switch detector for purpose of #869 but it doesn't solve it yet.
It's clean-up only.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/traversal_switch_detector.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/traversal_switch_detector.hpp
@@ -14,9 +14,6 @@
 #ifndef BOOST_GEOMETRY_ALGORITHMS_DETAIL_OVERLAY_TRAVERSAL_SWITCH_DETECTOR_HPP
 #define BOOST_GEOMETRY_ALGORITHMS_DETAIL_OVERLAY_TRAVERSAL_SWITCH_DETECTOR_HPP
 
-#include <cstddef>
-#include <map>
-
 #include <boost/range/value_type.hpp>
 
 #include <boost/geometry/algorithms/detail/ring_identifier.hpp>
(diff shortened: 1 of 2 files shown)


vissarion (approved):
This is fine with me. Thanks.

include/boost/geometry/algorithms/detail/overlay/traversal_switch_detector.hpp
```diff
@@ -64,59 +95,46 @@ struct traversal_switch_detector
...
+    using connection_map = std::map<signed_size_type, connection_properties>;
 
     // Per region, a set of properties is maintained, including its connections
     // to other regions
     struct region_properties
     {
-        signed_size_type region_id;
-        isolation_type isolated;
+        signed_size_type region_id{-1};
+        isolation_type isolated{isolation_unknown};
```
awulkiew, line 124: I'm just curious. These are native type and enum, correct? Why have you decided to use braces instead of `=`?
```
signed_size_type region_id = -1;
isolation_type isolated = isolation_unknown;
```

awulkiew (approved):

include/boost/geometry/algorithms/detail/overlay/traversal_switch_detector.hpp
barendgehrels: Ah, it's my new habit now. Because we do it at the office. But I now remember that you and @mloskot objected to it, and indeed it does not make sense. I'll change it.

barendgehrels: Updated :heavy_check_mark:

awulkiew: Right, I now remember that we talked about this. So this is a habit. I was curious that maybe there is rationale behind it that I'm not aware of. Anyway, thanks!

Issue #829, opened 2021-03-19
convex_hull discrepancy when points are collinear.


ayushgupta138:
When `boost::geometry::convex_hull()` function is called on collinear points, the returned polygon/linestring is incorrect and is inconsistent.
for example, the following code snippet,
` typedef boost::geometry::model::d2::point_xy<double> point2d;`
` using bg=boost::geometry;`
 `bg:model::polygon<point2d> poly1;`
 `bg::model::multi_point<point2d> p1;`
`bg::read_wkt("MULTIPOINT(2 1,3 2,4 3,7 6)", p1);`
`bg::convex_hull(p1,poly1);`
`cout<<bg::wkt(poly1)<<"\n";`
outputs,
`POLYGON(2 1,7 6,2 1,2 1)`
similar output is obtained when a linestring is used to store the hull.
This behaviour is caused by 
https://github.com/boostorg/geometry/blob/8b6929193db73f55b75563db1fdd4e86ded3d6fa/include/boost/geometry/algorithms/detail/convex_hull/graham_andrew.hpp#L322

What is the use of the above mentioned line that I am missing?

awulkiew:
I guess it's an edge case. It's impossible to create a convex hull that is valid polygon [1] of collinear points, or two points, or a single segment, or a single point, etc. The resulting polygon will have no interior.

The requirement of at least 4 points for closed polygon is correct. Open polygon should have at least 3 distinct points and closed polygon should have 4 with the last equal to the first one. So the result is correct (i.e. this is by design) but the polygon is invalid.

However I'm not saying that this behavior should stay. We could do something about it. I have several ideas. We could:
1. throw an exception that it's not possible to calculate the areal convex hull.
2. generate empty geometry if it is not possible to calculate the convex hull.
3. change/implement version of the algorithm for linear output and use it to hold linear result (collinear points)
4. change/implement version of the algorithm for pointlike output and use it to hold pointlike result (single point or equal points)
5. allow taking several kinds of geometries as output, e.g. geometry collection (not ready for that yet) or tuple of various multi-geometries (like we currently have in set operations, though it's undocummented, this should be simple).

2-4 would force the user to call the algorithm multiple times, once for each output geometry.

1-3(4?) would probably be a breaking change. Some users may currently rely on that in some cases an invalid polygon is returned by the algorithm so they can anylize it (check validity or area) and transform the output themselves without the need to call the algorithm several times for different output geometries.

[...]

PR #877, opened 2021-06-30
Support DG and GC in convex_hull, correct, correct_closure, is_convex, is_empty and reverse.

awulkiew:
In addition to the input GC and DG in convex_hull() also the output GC and DG is supported. Based on `traits::geometry_types` output type is chosen and then either emplaced back to GC or assigned to DG.

In case of GC and DG the output is always valid. Degenerated rings are not returned. Instead if convex hull is a segment then this segment is returned in a linear geometry and if the hull is a point then this is returned in pointlike geometry. Any geometry can be used for that with simpler ones chosen first. E.g. if GC can store rings, polygons, linestrings, segments and points then the output will either be ring, segment or point.

convex_hull is implemented in such a way that passes the GC into the graham_andrew algorithm and the algorithm works on the GC directly. In this case the convex hulls of all boxes are calculated first and passed together with the GC. The reason behind this is that boxes are treated separately.

This PR addresses https://github.com/boostorg/geometry/issues/829

-----

In the future we should probably implement strategies for boxes for all coordinate systems.

Note that calculating of hull for more than one box or a box and polygon in spherical/geographic will not be straightforward because AFAIU the problem can't be divided into calculating hulls for each of the geometries and then calculating the final hull from sub-hulls. Consider this case of box and polygon (blue) their hulls (red) and the hull of both (green).

rect833-6


include/boost/geometry/algorithms/detail/convex_hull/interface.hpp
```diff
@@ -21,86 +21,171 @@
...
+// TODO: This file is named interface.hpp but the code below is not the interface.
+//       It's the implementation of the algorithm.
 
 #ifndef DOXYGEN_NO_DETAIL
 namespace detail { namespace convex_hull
 {
 
-template <order_selector Order, closure_selector Closure>
-struct hull_insert
+// Abstraction representing ranges/rings of a geometry
```
barendgehrels, line 69: @awulkiew  can you explain the rationale of the `proxy` in more detail?
I glanced through the code but don't get it by that yet.

For example, is it just useful (shorter), or is it essential (we cannot make this improvement without). And why is it necessary for this function `convex_hull` but not (at least not yet) for other algorithms?

awulkiew: Short answer:
It's just shorter. It allows to support boxes without rewriting the `graham_andrew` algorithm or calling it twice.

Long answer:
GeometryCollection can store various kinds of geometries, e.g. Rings together with Boxes. We handle Boxes differently in a separate dispatch and generate a ring out of it. So now we have several options:
1. pass GC into `graham_andrew` algorithm and rewrite it to distinguish between various kinds of geometries
2. leave dispatches and `graham_andrew` as they are now and for GC separately calculate a hull for each Box, then the hull for the rest of the GC and then the resulting hull from all of the intermediate hulls
3. leave dispatches and `graham_andrew` as they are now and for GC separately calculate a hull for each Box, but then pass them together with GC into `graham_andrews` transparently, as if the Box hulls were contained in GC instead of Boxes (I've choosen this solution)

As I mentioned in the description this will work only in cartesian. In spherical/geographic we'll have to distinguish between geometries inside the algorithm. But the algorithm will probably be different than `graham_andrews`.

barendgehrels: Thanks! I'll finish my review next week (assuming there is no hurry)

awulkiew: No, take your time, thanks.

vissarion (approved):
Thanks, I am OK with merging! It would be useful to include some basic test cases for the new functionalities such as some convex hulls of geometry collections.

vissarion:
>Note that calculating of hull for more than one box or a box and polygon in spherical/geographic will not be straightforward because AFAIU the problem can't be divided into calculating hulls for each of the geometries and then calculating the final hull from sub-hulls. Consider this case of box and polygon (blue) their hulls (red) and the hull of both (green).

Just a side note here, the (convex) hull of hulls should always be the hull of the original geometries (due to convexity). That is, in the picture above the red shape is not the convex hull of a box while the green is the convex hull of their union. The issue here is that we cannot represent the convex hull of a box in non-cartesian system. We can discuss this in a separate issue I guess.

include/boost/geometry/algorithms/detail/convex_hull/interface.hpp
barendgehrels: Oops, this was forgotten. But I'm OK with merging.

barendgehrels (approved):

PR #883, opened 2021-07-05
Fix various warnings, e.g. remove unused variables, etc.

awulkiew:


test/geometries/initialization.cpp
```diff
@@ -237,7 +237,7 @@ void test_sanity_check()
 
         V v = boost::assign::list_of(P(1, 1))(P(2, 2))(P(3, 3));
         //v = boost::assign::list_of(P(1, 1))(P(2, 2))(P(3, 3));
-        v.empty();
+        //v.empty();
```
vissarion, line 240: what was the problem with `empty` here?

vissarion (approved):

test/geometries/initialization.cpp
awulkiew: Not using a return value of a function declared with `[[nodiscard]]`.

```
warning: ignoring return value of 'bool std::vector<...>::empty() const ...', declared with attribute 'nodiscard' [-Wunused-result]
```

https://en.cppreference.com/w/cpp/language/attributes/nodiscard

e.g. reported here: https://www.boost.org/development/tests/develop/output/teeks99-dkr-dg11-opt-geometry-gcc-11~c++20~O2-warnings.html#algorithms_area_geo

This could also be fixed by storing the return value in variable and checking it or passing to `ignore_unused()` but I don't think that this call is needed at all. But since I'm not sure about it, in case it was needed I commented it out for now instead of removing it.

awulkiew:
This fix is simple so I'll merge it now.

Issue #867, opened 2021-06-14
qbegin(idx::nearest(pt, 100)) fast but qbegin(idx::nearest(pt, tree_size)) really slow


karme:
If I do a ```t->qbegin(idx::nearest(pt, 100))``` it is really fast (<1ms) but
if I do a ```t->qbegin(idx::nearest(pt, tree_size))``` with tree_size=37133 it is really
slow (~500ms)

s.a. boost users mailing list thread:
"geometry: knn search of point to many linestrings?"
https://lists.boost.org/boost-users/2021/05/90908.php
especially:
https://lists.boost.org/boost-users/2021/06/90929.php

karme:
trying to reproduce with a simple test:
```
$ ./get-polylines.scm > polylines
$ ./test3.cpp < polylines
linestrings=29973
t.size()=29973
2850usec
14524956usec
```

karme:
the test files:
[rtree.zip](https://github.com/boostorg/geometry/files/6656799/rtree.zip)

awulkiew:
Thanks for the report! I'll look into this next week.

karme:
@awulkiew can you reproduce the problem?

awulkiew:
@karme Sorry that it takes so long. Yes, I was able to reproduce it with msvc. The time depends on `k` passed into `nearest()`.
```
linestrings=29973
t.size()=29973
k=100:   0.0001562sec
k=299:   0.001073sec
k=2997:  0.0636702sec
k=29973: 0.47566sec
```

There is also another issue, i.e. `nearest()` takes `k` as `unsigned int` which may be too small to hold value returned by `rtree.size()`.

awulkiew:
@karme I prepared a fix for this issue. The times decreased to:
```
k=100:   1.45e-05sec
k=299:   8.1e-06sec
k=2997:  6.7e-06sec
k=29973: 5.5e-06sec
```
Besides the two issues I mentioned above I found the third one. Details are in the PR. If you want to test the fix you can clone this branch: https://github.com/awulkiew/geometry/tree/fix/rtree_nearest

karme:
looks very good so far - thanks for looking into this!

karme:
in a bigger test (388521289 linestrings) the performance isn't good
enough:

qbegin(idx::nearest(pt, 100)) 398usec
qbegin(idx::nearest(pt, 388521289)) 11265usec
qbegin(idx::nearest(pt, 100)) 545usec
[...]

PR #886, opened 2021-07-07
Fix traversing of the rtree during incremental knn query.

awulkiew:
Before this PR unneeded nodes were traversed by the query iterator and by extension many unneeded neighbours were taken into consideration. This fixes: https://github.com/boostorg/geometry/issues/867

Additional improvement is that the neighbors that were already handled are not sorted again.

This PR also changes type `unsigned` to `size_t` in several places.


vissarion (approved):
Looks OK to me.

barendgehrels (commented):
Looks fine to me

barendgehrels (approved):

PR #891, opened 2021-07-21
[test] remove the non used KRAMER define and some other changes in testing is_valid

barendgehrels:
The objective of this PR is removing the define `BOOST_GEOMETRY_USE_KRAMER` because it's always used.
Last year I tried to change that by using the generic form instead. That's possible but didn't give much improvements and some regressions, and needed more and more conditions and tuning.

So it's redundant now, and I applied some other changes in tests, for example in difference where the symmetric difference was actually never tested without rescaling (I forgot that earlier).

### DIFF
--- a/test/algorithms/set_operations/difference/difference_multi.cpp
+++ b/test/algorithms/set_operations/difference/difference_multi.cpp
@@ -152,25 +152,19 @@ void test_areal()
         TEST_DIFFERENCE_WITH(0, 1, ggl_list_20120221_volker, 2, 7962.66, 2, 2775258.93, 4);
     }
 
+#if ! defined(BOOST_GEOMETRY_USE_RESCALING) || defined(BOOST_GEOMETRY_TEST_FAILURES)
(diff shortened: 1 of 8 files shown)


test/algorithms/set_operations/difference/difference_multi.cpp
```diff
@@ -525,7 +513,7 @@ int test_main(int, char* [])
 #if defined(BOOST_GEOMETRY_TEST_FAILURES)
     // Not yet fully tested for float.
     // The difference algorithm can generate (additional) slivers
-    BoostGeometryWriteExpectedFailures(22, 12, 16, 7);
+    BoostGeometryWriteExpectedFailures(28, 15, 19, 10);
```
barendgehrels, line 516: Some more tests are done in all cases. Therefore there are more failures.

awulkiew (approved):

vissarion (approved):
Thanks!

PR #894, opened 2021-07-30
Optimize rtree nearest queries, refactor spatial and nearest queries.

awulkiew:
This is an enhancement addressing the issue https://github.com/boostorg/geometry/issues/867 and more.

List of changes:
- added minmax heap and double-ended priority queue,
- sorting of branches in distance query replaced with priority queue,
- sorting of neighbors in iterative distance query replaced with d-e priority queue,
- branches are prioritized based on distance and rtree node's level to ensure that leafs are reached ASAP in cases when there are multiple nodes with the same distance to visit,
- nodes are no longer visited with `apply_visitor`, queries are no longer visitors, instead the rtree node's level is tracked and nodes are converted to `internal_node` or `leaf` with `get` (this is a change needed for the future modifications of the rtree).


awulkiew:
@barendgehrels Thanks! Though I'd appreciate more if you looked into this issue: https://github.com/boostorg/geometry/issues/900 in particular the case with robustness enabled.

vissarion (approved):
Thanks! Should the test case of https://github.com/boostorg/geometry/issues/867 be added in the test-suite together with a note on performance/issue?

barendgehrels (approved):

awulkiew:
Thanks for the reviews!

@vissarion I think it is not needed. Even if I added a test case we wouldn't detect potential performance regressions because we don't test that. Correctness is not the issue here and is covered by other tests. Besides the full test case from this issue has AFAIR tens of millions of linestrings stored in mapped memory file.

PR #897, opened 2021-08-11
fix: use balance between distance-to-end and length-of-segments

barendgehrels:
I indicated in a comment what the actual fix is.

**NOTE**: several `is_valid` checks in difference tests are false negatives. I noticed already earlier some of them, but now it was in the test I concentrated on (for mysql). **We should fix that.** I explicitly skip those validity checks now, otherwise we cannot detect good enough if we're improving.

This one for example:
uu_false_positive

And this one:
c_false_positive_mysql

### DIFF
--- a/include/boost/geometry/core/coordinate_promotion.hpp
+++ b/include/boost/geometry/core/coordinate_promotion.hpp
@@ -1,0 +1,111 @@
+// Boost.Geometry
+
+// Copyright (c) 2021 Barend Gehrels, Amsterdam, the Netherlands.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_CORE_COORDINATE_PROMOTION_HPP
+#define BOOST_GEOMETRY_CORE_COORDINATE_PROMOTION_HPP
+
+#include <boost/geometry/core/coordinate_type.hpp>
+
+// TODO: move this to a future headerfile implementing traits for these types
+#include <boost/rational.hpp>
+#include <boost/multiprecision/cpp_bin_float.hpp>
+#include <boost/multiprecision/cpp_int.hpp>
+
+namespace boost { namespace geometry
+{
+
+namespace traits
+{
+
+// todo
+
+} // namespace traits
+
+/*!
+    \brief Meta-function converting, if necessary, to "a floating point" type
+    \details
+        - if input type is integer, type is double
+        - else type is input type
+    \ingroup utility
+ */
+// TODO: replace with, or call, promoted_to_floating_point
+template <typename T, typename PromoteIntegerTo = double>
+struct promote_floating_point
+{
+    typedef std::conditional_t
+        <
+            std::is_integral<T>::value,
+            PromoteIntegerTo,
+            T
+        > type;
+};
+
+// TODO: replace with promoted_to_floating_point
+template <typename Geometry>
+struct fp_coordinate_type
+{
+    typedef typename promote_floating_point
+        <
+            typename coordinate_type<Geometry>::type
+        >::type type;
+};
+
+namespace detail
+{
+
+// Promote any integral type to double. Floating point
+// and other user defined types stay as they are, unless specialized.
+// TODO: we shold add a coordinate_promotion traits for promotion to
+// floating point or (larger) integer types.
+template <typename Type>
+struct promoted_to_floating_point
+{
+    using type = std::conditional_t
+        <
+            std::is_integral<Type>::value, double, Type
+        >;
+};
+
+// Boost.Rational goes to double
+template <typename T>
+struct promoted_to_floating_point<boost::rational<T>>
+{
+    using type = double;
+};
+
+// Any Boost.Multiprecision goes to double (for example int128_t),
+// unless specialized
+template <typename Backend>
+struct promoted_to_floating_point<boost::multiprecision::number<Backend>>
+{
+    using type = double;
+};
+
+// Boost.Multiprecision binary floating point numbers are used as FP.
+template <unsigned Digits>
+struct promoted_to_floating_point
+    <
+        boost::multiprecision::number
+            <
+                boost::multiprecision::cpp_bin_float<Digits>
+            >
+    >
+{
+    using type = boost::multiprecision::number
+        <
+            boost::multiprecision::cpp_bin_float<Digits>
+        >;
+};
+
+}
+
+
+}} // namespace boost::geometry
+
+
+#endif // BOOST_GEOMETRY_CORE_COORDINATE_PROMOTION_HPP
--- a/include/boost/geometry/policies/robustness/segment_ratio.hpp
+++ b/include/boost/geometry/policies/robustness/segment_ratio.hpp
@@ -135,11 +135,15 @@ struct possibly_collinear<Type, false>
 template <typename Type>
 class segment_ratio
 {
-public :
-    typedef Type numeric_type;
+    // Type used for the approximation (a helper value)
+    // and for the edge value (0..1) (a helper function).
+    using floating_point_type =
+        typename detail::promoted_to_floating_point<Type>::type;
 
     // Type-alias for the type itself
-    typedef segment_ratio<Type> thistype;
+    using thistype = segment_ratio<Type>;
+
+public :
 
     inline segment_ratio()
         : m_numerator(0)
@@ -221,8 +225,8 @@ public :
         m_approximation =
             m_denominator == 0 ? 0
             : (
-                boost::numeric_cast<fp_type>(m_numerator) * scale()
-                / boost::numeric_cast<fp_type>(m_denominator)
+                boost::numeric_cast<floating_point_type>(m_numerator) * scale()
+                / boost::numeric_cast<floating_point_type>(m_denominator)
             );
     }
 
@@ -254,21 +258,16 @@ public :
         return m_numerator > m_denominator;
     }
 
-    inline bool near_end() const
+    //! Returns a value between 0.0 and 1.0
+    //! 0.0 means: exactly in the middle
+    //! 1.0 means: exactly on one of the edges (or even over it)
+    inline floating_point_type edge_value() const
     {
-        if (left() || right())
-        {
-            return false;
-        }
-
-        static fp_type const small_part_of_scale = scale() / 100;
-        return m_approximation < small_part_of_scale
-            || m_approximation > scale() - small_part_of_scale;
-    }
-
-    inline bool close_to(thistype const& other) const
-    {
-        return geometry::math::abs(m_approximation - other.m_approximation) < 50;
+        using fp = floating_point_type;
+        fp const one{1.0};
+        floating_point_type const result
+                = fp(2) * geometry::math::abs(fp(0.5) - m_approximation / scale());
+        return result > one ? one : result;
     }
 
     template <typename Threshold>
@@ -313,19 +312,7 @@ public :
     }
 #endif
 
-
-
 private :
-    // NOTE: if this typedef is used then fp_type is non-fundamental type
-    // if Type is non-fundamental type
-    //typedef typename promote_floating_point<Type>::type fp_type;
-
-    // TODO: What with user-defined numeric types?
-    //       Shouldn't here is_integral be checked?
-    typedef std::conditional_t
-        <
-            std::is_floating_point<Type>::value, Type, double
-        > fp_type;
 
     Type m_numerator;
     Type m_denominator;
@@ -335,12 +322,19 @@ private :
     // Boost.Rational is used if the approximations are close.
     // Reason: performance, Boost.Rational does a GCD by default and also the
     // comparisons contain while-loops.
-    fp_type m_approximation;
+    floating_point_type m_approximation;
 
-
-    static inline fp_type scale()
+    inline bool close_to(thistype const& other) const
     {
-        return 1000000.0;
+        static floating_point_type const threshold{50.0};
+        return geometry::math::abs(m_approximation - other.m_approximation)
+                < threshold;
+    }
+
+    static inline floating_point_type scale()
+    {
+        static floating_point_type const fp_scale{1000000.0};
+        return fp_scale;
     }
 };
 
--- a/include/boost/geometry/policies/robustness/segment_ratio.hpp
+++ b/include/boost/geometry/policies/robustness/segment_ratio.hpp
@@ -19,8 +19,8 @@
 #include <boost/rational.hpp>
 
 #include <boost/geometry/core/assert.hpp>
+#include <boost/geometry/core/coordinate_promotion.hpp>
 #include <boost/geometry/util/math.hpp>
-#include <boost/geometry/util/promote_floating_point.hpp>
 
 namespace boost { namespace geometry
 {
--- a/include/boost/geometry/strategies/cartesian/intersection.hpp
+++ b/include/boost/geometry/strategies/cartesian/intersection.hpp
@@ -68,6 +68,61 @@ namespace boost { namespace geometry
 namespace strategy { namespace intersection
 {
 
+namespace detail_usage
+{
+
+// When calculating the intersection, the information of "a" or "b" can be used.
+// Theoretically this gives equal results, but due to floating point precision
+// there might be tiny differences. These are edge cases.
+// This structure is to determine if "a" or "b" should be used.
+// Prefer the segment closer to the endpoint.
+// If both are about equally close, then prefer the longer segment
+// To avoid hard thresholds, behavior is made fluent.
+// Calculate comparable length indications,
+// the longer the segment (relatively), the lower the value
+// such that the shorter lengths are evaluated higher and will
+// be preferred.
+template <bool IsArithmetic>
+struct use_a
+{
+  template <typename Ct, typename Ev>
+  static bool apply(Ct const& cla, Ct const& clb, Ev const& eva, Ev const& evb)
+  {
+      auto const clm = (std::max)(cla, clb);
+      if (clm <= 0)
+      {
+          return true;
+      }
+
+      // Relative comparible length
+      auto const rcla = Ct(1.0) - cla / clm;
+      auto const rclb = Ct(1.0) - clb / clm;
+
+      // Multipliers for edgevalue (ev) and relative comparible length (rcl)
+      // They determine the balance between edge value (should be larger)
+      // and segment length. In 99.9xx% of the cases there is no difference
+      // at all (if either a or b is used). Therefore the values of the
+      // constants are not sensitive for the majority of the situations.
+      // One known case is #mysql_23023665_6 (difference) which needs mev >= 2
+      Ev const mev = 5;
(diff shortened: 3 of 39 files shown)


include/boost/geometry/policies/robustness/segment_ratio.hpp
```diff
@@ -254,16 +254,13 @@ public :
         return m_numerator > m_denominator;
     }
 
-    inline bool near_end() const
+    //! Returns a value between 0.0 and 1.0
+    //! 0.0 means: exactly in the middle
+    //! 1.0 means: exactly on one of the edges (or even over it)
```
barendgehrels, line 259: This replaces `near_end` (which had a threshold). This better allows more fluent behavior without thresholds.

include/boost/geometry/strategies/cartesian/intersection.hpp
```diff
@@ -176,36 +176,43 @@ struct cartesian_segments
...
+            double const rclb = 1.0 - clb / clm;
+
+            double const eva = robust_ra.edge_value();
+            double const evb = robust_rb.edge_value();
+
+            // Multipliers for edgevalue (ev)
+            // and relative comparible length (rcl)
+            constexpr double mev = 5.0;
+            constexpr double mrcl = 1.0;
+            return mev * eva + mrcl * rcla > mev * evb + mrcl * rclb;
```
barendgehrels, line 208: This is the fix. Instead of just checking `near_end` of a and b, and if not select the smallest segment, we now have a balance: near the end gets priority (factor 5) but if they are equal, then the formula will select the smallest segment (as earlier).
Theoretically (mathematically) the result is the same, using either a or b. But in case of numeric precision, they might be just a bit different from each other, and it can (mainly for validity checks) be important which one is chosen. Though there is also randomness involved.

include/boost/geometry/strategies/cartesian/intersection.hpp
```diff
@@ -176,36 +176,43 @@ struct cartesian_segments
...
+            // If both are about equally close, then prefer the longer segment
+            // To avoid hard thresholds, behavior is made fluent:
+            // Calculate comparable length indications,
+            // the longer the segment (relatively), the lower the value
+            // such that the shorter lengths are evaluated higher and will
+            // be preferred.
+
+            auto const cla = comparable_length_a();
+            auto const clb = comparable_length_b();
+            auto const clm = std::max(cla, clb);
```
awulkiew, line 191: This line fails with msvc. It should've been:
```
auto const clm = (std::max)(cla, clb);
```

include/boost/geometry/policies/robustness/segment_ratio.hpp
awulkiew: Why is `double` returned instead of coordinate type or even better distance/comparable_distance result type (since it's compared with distances/lengths)?
What if coordinate type and/or distance result type is more precise than double?
Isn't `fp_type` defined for this purpose (for now ignoring the fact that it's definition is wrong, see TODO)?

barendgehrels: Indeed, `fp_type` was defined for this purpose, thanks. Updated.

include/boost/geometry/strategies/cartesian/intersection.hpp
barendgehrels: :heavy_check_mark:

include/boost/geometry/policies/robustness/segment_ratio.hpp
```diff
@@ -254,21 +262,14 @@ public :
...
-
-    inline bool close_to(thistype const& other) const
+    //! Returns a value between 0.0 and 1.0
+    //! 0.0 means: exactly in the middle
+    //! 1.0 means: exactly on one of the edges (or even over it)
+    inline floating_point_type edge_value() const
     {
-        return geometry::math::abs(m_approximation - other.m_approximation) < 50;
+        floating_point_type const result
+                = 2.0 * std::fabs(0.5 - m_approximation / scale());
```
awulkiew, line 271: `geometry::math::abs` should probably still be used here because `std::fabs` may be not specialized for user-defined FP type. Also because it is used below too, so for consistency.

include/boost/geometry/policies/robustness/segment_ratio.hpp
```diff
@@ -135,11 +135,19 @@ struct possibly_collinear<Type, false>
...
-public :
-    typedef Type numeric_type;
+    // Use the type (if it is already floating point), or a double
+    // (for integer types or user defined types).
+    // This is used for the approximation (a helper value)
+    // and for the edge value (0..1) (a helper function).
+    using floating_point_type = std::conditional_t
+        <
+            std::is_floating_point<Type>::value, Type, double
+        >;
```
awulkiew, line 145: Since you moved it we may address its problem now. The problem is that user-defined FP type is going to be degraded to `double` so this typedef should rather be:
```
    using floating_point_type = std::conditional_t
        <
            std::is_integral<Type>::value, double, Type
        >;
```
After this change it would not work properly for user-defined integral types but AFAIU we assume that user-defined types are FP types.

In fact we should probably add traits allowing the users to specify how the library should handle user-defined numeric types but that's outside of the scope of this PR.

include/boost/geometry/strategies/cartesian/intersection.hpp
```diff
@@ -68,6 +68,57 @@ namespace boost { namespace geometry
...
+// To avoid hard thresholds, behavior is made fluent.
+// Calculate comparable length indications,
+// the longer the segment (relatively), the lower the value
+// such that the shorter lengths are evaluated higher and will
+// be preferred.
+template <bool IsArithmetic>
+struct use_a
+{
+  template <typename Ct, typename Ev>
+  static bool apply(Ct cla, Ct clb, Ev eva, Ev evb)
```
awulkiew, line 89: Shouldn't these be `const&` in case these types were user-defined to avoid copies?

include/boost/geometry/strategies/cartesian/intersection.hpp
```diff
@@ -68,6 +68,57 @@ namespace boost { namespace geometry
...
+
+      // Relative comparible length
+      auto const rcla = 1.0 - cla / clm;
+      auto const rclb = 1.0 - clb / clm;
+
+      // Multipliers for edgevalue (ev)
+      // and relative comparible length (rcl)
+      using ctype = decltype(eva);
+      constexpr ctype mev = 5.0;
+      constexpr ctype mrcl = 1.0;
```
awulkiew, line 105: This will fail to compile if `Ev` is user-defined type without constexpr constructor implicitly taking double.

include/boost/geometry/strategies/cartesian/intersection.hpp
```diff
@@ -68,6 +68,57 @@ namespace boost { namespace geometry
...
+      return mev * eva + mrcl * rcla > mev * evb + mrcl * rclb;
+  }
+};
+
+// Specialization for non arithmetic types. They will always use "a"
+template <>
+struct use_a<false>
+{
+    template <typename Ct, typename Ev>
+    static bool apply(Ct , Ct , Ev , Ev )
```
awulkiew, line 115: `const&` in case these are user-defined types.

include/boost/geometry/policies/robustness/segment_ratio.hpp
barendgehrels: Thanks for this and for your other remarks, I'm processing them.

One thing: this proposed change does not work for `boost::rational` because it will set floating_point_type to rational. 

Your suggestion (`add traits allowing ...`)  is very good and I will use that approach (not yet making it a generic traits though)

barendgehrels: Or actually it will be generic. We also have `fp_coordinate_type` and `promote_floating_point` which I will combine.

awulkiew: Ok, though I think that they should be added consistently in the whole library which is outside of the scope of this PR. I'd be fine with a simple fix here.

barendgehrels: Agreed, I've something but it's growing. For now I'll apply a simple fix.

include/boost/geometry/policies/robustness/segment_ratio.hpp
barendgehrels: :heavy_check_mark:

include/boost/geometry/policies/robustness/segment_ratio.hpp
barendgehrels: :heavy_check_mark:

include/boost/geometry/strategies/cartesian/intersection.hpp
barendgehrels: :heavy_check_mark:

include/boost/geometry/strategies/cartesian/intersection.hpp
barendgehrels: :heavy_check_mark: 
Indeed, `decltype` comes from an earlier version and is not necessary at all now. Thanks.

include/boost/geometry/strategies/cartesian/intersection.hpp
barendgehrels: :heavy_check_mark:

include/boost/geometry/policies/robustness/segment_ratio.hpp
```diff
@@ -335,12 +342,17 @@ private :
...
+    {
+        return geometry::math::abs(m_approximation - other.m_approximation)
+                < 50;
+    }
 
-    static inline fp_type scale()
+    static inline floating_point_type scale()
     {
-        return 1000000.0;
+        return 1000000;
```
awulkiew, line 355: On platforms with 16 bit int this would be incorrect. Why did you change it from `double` anyway? I thought that we assume that user-defined numeric types has to be implicitly constructible from `double` (which btw also means that they are implicitly constructible from `int`).

awulkiew (approved):

include/boost/geometry/policies/robustness/segment_ratio.hpp
barendgehrels: > > I changed it because of your suggestion
> 
> Which one? :)

_"Since you moved it we may address its problem now. The problem is that user-defined FP type is going to be degraded to double so this typedef should rather be"_

barendgehrels: > Do you mean that this is a problem for user-defined floating point type which may not be implicitly constructible from `double`? Which type exactly this fails to compile for? I think `multiprecision::number` should be fine, the ctors are not marked as explicit. They may be disabled/enabled with type traits though. I tested your branch with `1000000.0` and the tests compile with clang-win for me.

I just retested, it is not the Boost multiprecision **float** but the Boost multi precision **int** which gave the problem. I test that in the other branch. But I now realize that it's not enough, also that type should go to floating point. I will fix that later in this PR. And you are right then, this could stay as it was before.

awulkiew: So you were talking about FP typedef, not the integral literal.

There are 4 issues at play here:
1. The type `floating_point_type`
2. whether or not it defines explicit ctor
3. whether or not it is constructible from double
4. the type of the literal

2. and 3. may be more complex because of ctors enabled with `enable_if`.

My initial remark was about the last point. I was wrong and the type is not going to be `int` but any integral type able to hold this number. So now I think that any literal you use here should be fine. This doesn't mean that this literal could be used in construction of the `floating_point_type`.

I agree we shouldn't assume that the UD type can be implicitly constructible. This means that in all places ctor should be explicitly called.

> I just retested, it is not the Boost multiprecision float but the Boost multi precision int which gave the problem.

Ok, so this was one of the multiprecision integral types which is not "promoted" to `double` because of the change of `floating_point_type`. For this we probably needs these traits.

Note that if it's not possible to create integral multipresion number from double or float then such numbers will not compile in other parts of the code where we create FP constants.

awulkiew: And once again, integer constant is fine here. So this may stay as it is now.

include/boost/geometry/strategies/cartesian/intersection.hpp
```diff
@@ -68,6 +68,56 @@ namespace boost { namespace geometry
...
+  {
+      auto const clm = (std::max)(cla, clb);
+      if (clm <= 0)
+      {
+          return true;
+      }
+
+      // Relative comparible length
+      auto const rcla = 1.0 - cla / clm;
+      auto const rclb = 1.0 - clb / clm;
```
awulkiew, line 99: So based on our discussion about the explicit creation `1.0`  should either be replaced with `Ct(1)` or a const variable as it is the case below with `mrcl`.

include/boost/geometry/strategies/cartesian/intersection.hpp
vissarion: Why using factor 5? Are similar values also tested? Maybe add a comment in the code about that choice for future reference.

include/boost/geometry/policies/robustness/segment_ratio.hpp
```diff
@@ -335,12 +342,17 @@ private :
     // Boost.Rational is used if the approximations are close.
     // Reason: performance, Boost.Rational does a GCD by default and also the
     // comparisons contain while-loops.
-    fp_type m_approximation;
+    floating_point_type m_approximation;
 
+    inline bool close_to(thistype const& other) const
+    {
+        return geometry::math::abs(m_approximation - other.m_approximation)
+                < 50;
```
vissarion, line 350: Is 50 here a magic number or is it related with some values declared in other parts of the file?

vissarion (approved):
I am ok to merge, I only have some questions related to some constant values in the code.

include/boost/geometry/strategies/cartesian/intersection.hpp
barendgehrels: It's basically already commented
```
// Prefer the segment closer to the endpoint.
// If both are about equally close, then prefer the longer segment
// To avoid hard thresholds, behavior is made fluent.
// Calculate comparable length indications,
// the longer the segment (relatively), the lower the value
// such that the shorter lengths are evaluated higher and will
// be preferred.
```
The essention is that the `mev` is much larger than `mrcl` . I'll add a comment.

include/boost/geometry/policies/robustness/segment_ratio.hpp
barendgehrels: It's for efficiency only, and related to the scale value 1000000.
If it is close_to, then `operator==` uses `geometry::math::equals`, otherwise it's just false.

This was already the case, the method is just moved.

include/boost/geometry/strategies/cartesian/intersection.hpp
barendgehrels: I was testing to try the effect but in the test program I used, there is no difference between `1` and `5`
I'll come back to this.

include/boost/geometry/core/coordinate_promotion.hpp
```diff
@@ -0,0 +1,111 @@
...
+#include <boost/multiprecision/cpp_bin_float.hpp>
+#include <boost/multiprecision/cpp_int.hpp>
+
+namespace boost { namespace geometry
+{
+
+namespace traits
+{
+
+// todo
```
barendgehrels, line 25: So this is the start of Adam's proposal - I didn't apply it everywhere, but the promotion is already here (below) and specialized for rational and multi precision. It should be worked out later, in another PR.
Two related types which should be replaced are moved in this file as well.

awulkiew: Ok, we can build on top of it in a different PR. In case you wanted to jump into it I'll write what I think about this.

I think we should add more general utilities for numeric types. So to not only cover the promotion but also type traits that could be generalized to non-fundamental types. At minimum we'd need `is_integral`/`is_floating_point` type traits and a trait allowing to define floating point type corresponding to input integer type.

A name `promote_to_floating_point` seems to be ok. Another possibility is `make_floating_point` because `std::make_unsigned` and `std::make_signed` exist but I think the semantics is different so `promote` is fine. But I'd avoid past tense for sure because type traits use present tense so `make`, `remove`, `add` and therefore `promote`, `select`, `transcribe`, etc.

I'm not sure about the directory. On one hand this is important and these are traits which would make `core/` the correct directory. On the other hand type selection/promotion and type_traits are already in `util/`.

Anyway, these are my quick thoughts about it.

barendgehrels: Thanks, we'll come back to this.

include/boost/geometry/strategies/cartesian/intersection.hpp
barendgehrels: Checking this with other robustness tests, there is something wrong, there are regressions.
Those are not regularly tested (because they are not unit tests, they work in a different way).

I'm working on it and update later - it's not the balance, but it's something in the other changes for `approximately_equals`

barendgehrels: So the fixes are unrelated now. This PR (as at the start) fixes the balance, and the promotion type. But not yet the other-types which will follow in another commit.

include/boost/geometry/strategies/cartesian/intersection.hpp
```diff
@@ -68,6 +68,61 @@ namespace boost { namespace geometry
...
+      // Relative comparible length
+      auto const rcla = Ct(1.0) - cla / clm;
+      auto const rclb = Ct(1.0) - clb / clm;
+
+      // Multipliers for edgevalue (ev) and relative comparible length (rcl)
+      // They determine the balance between edge value (should be larger)
+      // and segment length. In 99.9xx% of the cases there is no difference
+      // at all (if either a or b is used). Therefore the values of the
+      // constants are not sensitive for the majority of the situations.
+      // One known case is #mysql_23023665_6 (difference) which needs mev >= 2
```
barendgehrels, line 106: @vissarion this should address your concern.

It tested many other robustness too (therefore you see some changes in this PR) but they are all neutral to this balance

test/robustness/overlay/areal_areal/intersection_pies.cpp
```diff
@@ -1,13 +1,15 @@
...
-// Copyright (c) 2009-2020 Barend Gehrels, Amsterdam, the Netherlands.
+// Copyright (c) 2009-2021 Barend Gehrels, Amsterdam, the Netherlands.
 
 // Use, modification and distribution is subject to the Boost Software License,
 // Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
 // http://www.boost.org/LICENSE_1_0.txt)
 
 #define BOOST_GEOMETRY_NO_BOOST_TEST
+#define BOOST_GEOMETRY_NO_ROBUSTNESS
+#define BOOST_GEOMETRY_TEST_ONLY_ONE_TYPE
```
barendgehrels, line 12: The robustness tests now all test without rescaling, preceding coming change to the default

include/boost/geometry/strategies/cartesian/intersection.hpp
vissarion: Thanks! Useful to have a specific sensitive case.

PR #899, opened 2021-08-13
Various small fixes and improvements.

awulkiew:
- fix warnings in tests
- add workaround (fix?) gcc-11 c++20 compilation error
- rename some utils


barendgehrels (approved):
Thanks

PR #901, opened 2021-08-18
fix: small fixes to support boost multiprecision and rational

barendgehrels:
Because of the first version of previous PR (errors in rational), I added a small unit test for this type and Boost multi precision. There were some small fixes necessary, mainly in new functions.

The new test case looks like this and is supported and correct (even using boost rational).

rectangular

The first test (the simplex) was incorrect for boost rational, and give different results with or without my other PR (because of diagonals), so I left that out for now).

multi_simplex

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
@@ -316,7 +316,8 @@ public :
         // then take a point (or more) further back.
         // The limit of offset avoids theoretical infinite loops.
         // In practice it currently walks max 1 point back in all cases.
-        double const tolerance = 1.0e9;
+        using ct_type = typename geometry::coordinate_type<Point>::type;
+        ct_type const tolerance = 1000 * 1000 * 1000;
         int offset = 0;
         while (approximately_equals(point_from, turn.point, tolerance)
                && offset > -10)
--- a/include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
+++ b/include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
@@ -783,22 +783,24 @@ inline void enlarge_sections(Sections& sections, Strategy const&)
 
     // It makes section a tiny bit too large, which might cause (a small number)
     // of more comparisons
-    for (typename boost::range_iterator<Sections>::type it = boost::begin(sections);
-        it != boost::end(sections);
-        ++it)
+    for (auto& section : sections)
     {
 #if defined(BOOST_GEOMETRY_USE_RESCALING)
         detail::sectionalize::expand_by_epsilon
             <
                 typename Strategy::cs_tag
-            >::apply(it->bounding_box);
+            >::apply(section.bounding_box);
 
 #else
         // Expand the box to avoid missing any intersection. The amount is
         // should be larger than epsilon. About the value itself: the smaller
         // it is, the higher the risk to miss intersections. The larger it is,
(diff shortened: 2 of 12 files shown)


include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
```diff
@@ -316,7 +316,8 @@ public :
         // then take a point (or more) further back.
         // The limit of offset avoids theoretical infinite loops.
         // In practice it currently walks max 1 point back in all cases.
-        double const tolerance = 1.0e9;
+        using ct_type = typename geometry::coordinate_type<Point>::type;
+        ct_type const tolerance = 1000 * 1000 * 1000;
```
awulkiew, line 320: I this this will not compile properly on platforms with 16-bit `int` becasue there will be an overflow. We probably don't target such platforms but still. What was the reason behind this change?

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
```diff
@@ -783,22 +783,24 @@ inline void enlarge_sections(Sections& sections, Strategy const&)
 
     // It makes section a tiny bit too large, which might cause (a small number)
     // of more comparisons
-    for (typename boost::range_iterator<Sections>::type it = boost::begin(sections);
-        it != boost::end(sections);
-        ++it)
+    for (auto& section : sections)
```
awulkiew, line 786: Can we be certain that `sections` defines `begin()` and `end()` member functions? Is this a standard container? Becasue if this is a type adapted to Boost.Range concept then range-based for loop may not compile for it.

test/string_from_type.hpp
```diff
@@ -47,7 +47,8 @@ template <> struct string_from_type<int>
 template <> struct string_from_type<long>
 { static std::string name() { return "l"; }  };
 
-template <> struct string_from_type<boost::multiprecision::cpp_bin_float_100>
+template <typename Backend>
+struct string_from_type<boost::multiprecision::number<Backend>>
```
awulkiew, line 51: `Backend` can define various floating point and integral types. Is this specialization correct for all of them or should it be narrowed to floating point types only?

awulkiew (approved):

include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
vissarion: I also not understand this change. Is this to support multiprecision? Why not `ct_type const tolerance = 1000000000` ?

vissarion (approved):
OK with merging, thanks!

include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
barendgehrels: I'll fix it.

Because I'm now working in two branches which are quite related to each other, and they are both approved, I will decline this PR and move this commit to the other PR.

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
barendgehrels: I'll check.

test/string_from_type.hpp
barendgehrels: Yes, it's for debug. It's used for display on screen and in svg filenames. It's OK if they all are indicated with `m`

barendgehrels:
Closing, the code will appear in #897

PR #907, opened 2021-09-22
fix: small fixes to support boost multiprecision and rational

barendgehrels:
This PR is a continuation of  #901 which was first to be merged into the other branch, and it's pr declined, but later split off

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
@@ -316,7 +316,15 @@ public :
         // then take a point (or more) further back.
         // The limit of offset avoids theoretical infinite loops.
         // In practice it currently walks max 1 point back in all cases.
-        double const tolerance = 1.0e9;
+        // Use the coordinate type, but if it is too small (e.g. std::int16), use a double
+        using ct_type = typename geometry::select_most_precise
+            <
+                typename geometry::coordinate_type<Point>::type,
+                double
+            >::type;
+
+        ct_type const tolerance = 1000000000;
+
         int offset = 0;
         while (approximately_equals(point_from, turn.point, tolerance)
                && offset > -10)
--- a/include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
+++ b/include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
@@ -783,22 +783,24 @@ inline void enlarge_sections(Sections& sections, Strategy const&)
 
     // It makes section a tiny bit too large, which might cause (a small number)
     // of more comparisons
-    for (typename boost::range_iterator<Sections>::type it = boost::begin(sections);
-        it != boost::end(sections);
-        ++it)
+    for (auto& section : sections)
     {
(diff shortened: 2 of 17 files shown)


include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
```diff
@@ -316,7 +316,15 @@ public :
         // then take a point (or more) further back.
         // The limit of offset avoids theoretical infinite loops.
         // In practice it currently walks max 1 point back in all cases.
-        double const tolerance = 1.0e9;
+        // Use the coordinate type, but if it is too small (e.g. std::int16), use a double
+        using ct_type = typename geometry::select_most_precise
+            <
+                typename geometry::coordinate_type<Point>::type,
+                double
+            >::type;
```
barendgehrels, line 324: Required for types as `std::int16_t` to avoid overflow
(therefore it's tested in `union_other_types.cpp` now too)

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
```diff
@@ -783,22 +783,24 @@ inline void enlarge_sections(Sections& sections, Strategy const&)
...
 
 #else
         // Expand the box to avoid missing any intersection. The amount is
         // should be larger than epsilon. About the value itself: the smaller
         // it is, the higher the risk to miss intersections. The larger it is,
         // the more comparisons are made. So it should be on the high side.
-        detail::buffer::buffer_box(it->bounding_box, 0.001, it->bounding_box);
+        using gt = decltype(section.bounding_box);
+        using ct = typename geometry::coordinate_type<gt>::type;
+        ct const tolerance = ct(1) / ct(1000);
```
barendgehrels, line 801: Required for `Boost.Rational`

barendgehrels:
Note, I get errors in `algorithms_convex_hull_robust`, unrelated to this PR.
This might be related to the merging of `use_side_robust_predicate`

vissarion:
> Note, I get errors in `algorithms_convex_hull_robust`, unrelated to this PR.
> This might be related to the merging of `use_side_robust_predicate`

Thanks for noticing! https://github.com/boostorg/geometry/pull/908 should fix that.

vissarion (approved):
Thanks! It would be nice to have a list of all Boost multi-precision types that are supported. Are they all?

barendgehrels:
> Thanks! It would be nice to have a list of all Boost multi-precision types that are supported. Are they all?

Good point. Supported are the ones tested here. But there are more, I'll check.

awulkiew (approved):

barendgehrels:
> Thanks! It would be nice to have a list of all Boost multi-precision types that are supported. Are they all?

All types supported by Boost multi-precision are now included in the unit test. Those are (basically) all types but not types using other libraries (GMP, tom, ...) and not float128 (compiler dependent). Also finished the intended testcases, it was actually hard finding a test going wrong without rescaling (a good sign).

Also added some additional small `zero` constants for boost::rational<short> (requires somehow more than other types)

I squashed it, and will merge.

PR #913, opened 2021-10-03
Support DG/GC in densify, envelope, expand, is_valid, num_points, num_segments, simplify.

awulkiew:
- specialize `point_type` using the point of the first geometry of DG/GC.
- update tests.


vissarion (approved):
Looks OK to me.

PR #917, opened 2021-10-08
[test] exclude most of the types, keep only one per family

barendgehrels:
as discussed


vissarion:
+1 for excluding `gcc-6.3` but keep the issue open. 
Out of curiosity, does other MP types give wrong results with this compiler ?

awulkiew:
So it fails for `rational_adaptor<cpp_int_backend<>>` but not for `boost::rational<>`. Maybe this will be enough to guide them to find the issue. Another simple thing would be to pinpoint the exact commit of MultiPrecision (or at least merge commit) introducing this issue.

PR #919, opened 2021-10-08
fix: example of wxwidgets

barendgehrels:
Side project, I needed the sample for something else, and it needed some updates.

It's still nice and useful, now showing this window
image


awulkiew:
@barendgehrels should I mark this as a bug or an enhancement?

barendgehrels:
> @barendgehrels should I mark this as a bug or an enhancement?

It is in the examples, so I think it's part of the documentation.

PR #931, opened 2021-10-30
[equals] Define collect_vectors type in use_collect_vectors

awulkiew:
After PR https://github.com/boostorg/geometry/pull/926

I propose to define collect_vectors type in `use_collect_vectors`, so at the same place where this version is is enabled for specific side strategy and `cs_tag`.

Furthermore there is no need to instantiate both versions of collect_vectors with `std::conditional` in `equals_by_collection_or_relate` so this PR changes that.

### DIFF
--- a/include/boost/geometry/algorithms/detail/equals/implementation.hpp
+++ b/include/boost/geometry/algorithms/detail/equals/implementation.hpp
@@ -159,6 +159,33 @@ struct length_check
 };
 */
 
+
+// Small helper structure do decide to use collect_vectors, or not
+template <typename Strategy, typename CsTag>
+struct use_collect_vectors
+{
+    static constexpr bool value = false;
+};
+
+template <typename Strategy>
+struct use_collect_vectors<Strategy, cartesian_tag>
+{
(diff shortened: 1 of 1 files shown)


include/boost/geometry/algorithms/detail/equals/implementation.hpp
```diff
@@ -226,50 +246,42 @@ struct equals_by_relate
...
+    <
+        typename Geometry1, typename Geometry2, typename Strategy,
+        std::enable_if_t<! use_vectors<Strategy>::value, int> = 0
+    >
+    static inline bool apply(Geometry1 const& geometry1,
+                             Geometry2 const& geometry2,
+                             Strategy const& strategy)
+    {
+        return equals_by_relate<Geometry1, Geometry2>::apply(geometry1, geometry2, strategy);
     }
```
barendgehrels, line 285: Thanks for this! I was already thinking how the two structures could be combined, but didn’t get this yet. Looks good!

barendgehrels (approved):

PR #934, opened 2021-11-09
Replace check_iterator_range by std::any_of, std::all_of or std::none_of

vissarion:
This PR proposes the replacement of `check_iterator_range` utility with the standard `find_if` function. 

In the cases where the predicate was only used by `check_iterator_range` I just renamed the predicate's `apply()` method to `operator()` but when the predicate is used by other functions too I used lambda functions to pass `apply()` as a predicate operator.  

Note that `find_if` return `false` if the predicate fails in all members of the input range contrary to the implementation of `check_iterator_range` so now the predicate is returning the negation of the return value of the current implementation.


include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
```diff
@@ -75,12 +76,12 @@ class multipoint_multipoint
...
+        inline bool operator()(Point const& point) const
         {
-            return !std::binary_search(m_first,
-                                       m_last,
-                                       point,
-                                       static_cast<base_type const&>(*this));
+            return std::binary_search(m_first,
+                                      m_last,
+                                      point,
+                                      static_cast<base_type const&>(*this));
```
awulkiew, line 84: You have changed the behavior of predicates and left the old names. E.g. in this case this predicate no longer is `unary_disjoint_predicate` it's `unary_intersects_predicate`.

include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
```diff
@@ -93,30 +94,27 @@ class multipoint_multipoint
...
+        using less_type = geometry::less<void, -1, cs_tag>;
+        using point1_type = typename boost::range_value<MultiPoint1>::type;
 
         std::vector<point1_type> points1(boost::begin(multipoint1),
                                          boost::end(multipoint1));
 
         std::sort(points1.begin(), points1.end(), less_type());
 
-        typedef unary_disjoint_predicate
+        using predicate_type = unary_disjoint_predicate
```
awulkiew, line 108: So here we have `intersects_predicate`.

include/boost/geometry/algorithms/detail/check_iterator_range.hpp
```diff
@@ -1,71 +0,0 @@
...
-    static inline bool apply(InputIterator first, InputIterator beyond)
-    {
-        for (InputIterator it = first; it != beyond; ++it)
-        {
-            if (! Predicate::apply(*it))
-            {
-                return false;
-            }
-        }
-        return AllowEmptyRange || first != beyond;
```
awulkiew, line 41: After removing this and replacing with `find_if` we have to take care about this special condition because in some operations empty ranges are permitted and in other operations they're not.

include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
```diff
@@ -93,30 +94,27 @@ class multipoint_multipoint
...
-        return check_iterator_range
-            <
-                predicate_type
-            >::apply(boost::begin(multipoint2),
-                     boost::end(multipoint2),
-                     predicate_type(points1.begin(), points1.end()));
+        return boost::end(multipoint2) == find_if(
+            boost::begin(multipoint2),
+            boost::end(multipoint2),
+            predicate_type(points1.begin(), points1.end()));
```
awulkiew, line 117: Here we search for intersecting points and if they are not found then the result of disjoint is `true`. If multipoint is empty they are disjoint so ok.

include/boost/geometry/algorithms/detail/disjoint/multirange_geometry.hpp
```diff
@@ -40,9 +40,9 @@ class unary_disjoint_geometry_to_query_geometry
     {}
 
     template <typename QueryGeometry>
-    inline bool apply(QueryGeometry const& query_geometry) const
+    inline bool operator()(QueryGeometry const& query_geometry) const
     {
-        return BinaryPredicate::apply(query_geometry, m_geometry, m_strategy);
+        return ! BinaryPredicate::apply(query_geometry, m_geometry, m_strategy);
```
awulkiew, line 45: Again this is now `intersects` predicate. Assuming that `BinaryPredicate` is `disjoint` predicate because technically this predicate is negating whatever `BinaryPredicate` ism but that's another issue. So the name should be changed as well.

awulkiew: Regarding the names. Note that if we decided to name these predicate as `intersects` then we'd end up with confusing types, i.e. disjoint predicates passed into intersects predicates, e.g.:
`unary_intersects_geometry_to_query_geometry` taking `disjoint_range_segment_or_box`. So I guess it'd be better to use name containing e.g. `unary_not`, `unary_not_disjoint` etc.

include/boost/geometry/algorithms/detail/disjoint/linear_areal.hpp
```diff
@@ -141,28 +141,26 @@ struct disjoint_segment_areal
...
-                unary_predicate_type
-            >::apply(boost::begin(interior_rings),
-                     boost::end(interior_rings),
-                     unary_predicate_type(segment, strategy));
+            >;
+
+        return boost::end(interior_rings) == 
+            std::find_if(boost::begin(interior_rings),
+                         boost::end(interior_rings), 
+                         unary_predicate_type(segment, strategy));
```
awulkiew, line 163: Find if a segment intersects any interior ring. If not found and also for empty range return `true`. Ok.

----

`end` is created 2 times and makes the formatting harder. What do you think about replacing it with the code below?
```
auto const end = boost::end(interior_rings);
return std::find_if(boost::begin(interior_rings), end, 
                    unary_predicate_type(segment, strategy)) == end;
```

include/boost/geometry/algorithms/detail/disjoint/multirange_geometry.hpp
```diff
@@ -68,13 +68,12 @@ struct multirange_constant_size_geometry
...
 
-        return detail::check_iterator_range
-            <
-                unary_predicate_type
-            >::apply(boost::begin(multirange), boost::end(multirange),
-                     unary_predicate_type(constant_size_geometry, strategy));
+        return boost::end(multirange) == 
+            std::find_if(boost::begin(multirange),
+                         boost::end(multirange), 
+                         unary_predicate_type(constant_size_geometry, strategy));
```
awulkiew, line 76: Find if a point/segment/box intersects a geometry in a range and if not then return true. For empty range the result is true as well, ok.

include/boost/geometry/algorithms/detail/distance/multipoint_to_geometry.hpp
```diff
@@ -126,14 +126,11 @@ class multipoint_to_areal
...
 
-        if (check_iterator_range
-                <
-                    not_covered_by_areal, false
-                >::apply(boost::begin(multipoint),
-                         boost::end(multipoint),
-                         predicate))
+        if (boost::end(multipoint) == std::find_if(boost::begin(multipoint),
+                                                   boost::end(multipoint),
+                                                   predicate))
```
awulkiew, line 133: For empty multipoint the result was `false` here before, now it would be `true`. There is no need to call whatever is below for empty geometry while calculating distance.

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
```diff
@@ -285,19 +285,12 @@ struct is_simple_multilinestring
...
-                 >::apply(boost::begin(multilinestring),
-                          boost::end(multilinestring),
-                          per_ls(strategy))
-             )
+        using per_ls = per_linestring<Strategy>;
+
+        auto first = boost::begin(multilinestring);
+        auto last = boost::end(multilinestring);
+
+        if ( ! (last == std::find_if(first, last, per_ls(strategy))))
```
awulkiew, line 293: I'd write `auto const` and `!= last`.

awulkiew: This will return `true` for empty multilinestring so ok.

include/boost/geometry/algorithms/detail/is_valid/has_invalid_coordinate.hpp
```diff
@@ -80,26 +80,26 @@ struct indexed_has_invalid_coordinate
...
 {
+private:
     struct point_has_valid_coordinates
     {
         template <typename Point>
-        static inline bool apply(Point const& point)
+        inline bool operator()(Point const& point)
         {
-            return ! point_has_invalid_coordinate::apply(point);
+            return point_has_invalid_coordinate::apply(point);
```
awulkiew, line 89: This predicate is no longer `point_has_valid_coordinates`. I haven't checked but since there already is `point_has_invalid_coordinate` it's possible that this one is not needed.

include/boost/geometry/algorithms/detail/is_valid/has_invalid_coordinate.hpp
```diff
@@ -80,26 +80,26 @@ struct indexed_has_invalid_coordinate
...
-        bool const has_valid_coordinates = detail::check_iterator_range
-            <
-                point_has_valid_coordinates,
-                true // do not consider an empty range as problematic
-            >::apply(geometry::points_begin(geometry),
-                     geometry::points_end(geometry));
+        bool const has_valid_coordinates = geometry::points_end(geometry) == 
+            std::find_if(geometry::points_begin(geometry),
+                         geometry::points_end(geometry),
+                         point_has_valid_coordinates());
```
awulkiew, line 102: This is a search for points with invalid coordinates.

If not found then `true` is returned, also for empty range, ok.

include/boost/geometry/algorithms/detail/is_valid/linear.hpp
```diff
@@ -165,15 +164,12 @@ class is_valid
...
-            <
-                per_ls,
-                false // do not check for empty multilinestring (done above)
-            >::apply(boost::begin(multilinestring),
-                     boost::end(multilinestring),
-                     per_ls(visitor, strategy));
+        return boost::end(multilinestring) == std::find_if(
+            boost::begin(multilinestring),
+            boost::end(multilinestring),
+            per_ls(visitor, strategy));
```
awulkiew, line 172: The behavior is changed here. `false` was returned for empty range. But in this case this is probably irrelevant, right?

include/boost/geometry/algorithms/detail/is_valid/multipolygon.hpp
```diff
@@ -265,21 +265,22 @@ class is_valid_multipolygon
         {}
 
         template <typename Polygon>
-        inline bool apply(Polygon const& polygon) const
+        inline bool operator()(Polygon const& polygon) const
         {
-            return base::apply(polygon, m_policy, m_strategy);
+            return ! base::apply(polygon, m_policy, m_strategy);
```
awulkiew, line 270: I'd probably also change the name from `per_polygon` to `is_invalid_polygon` or `is_polygon_invalid`.

include/boost/geometry/algorithms/detail/is_valid/multipolygon.hpp
```diff
@@ -290,22 +291,22 @@ class is_valid_multipolygon
...
-                  <
-                      per_polygon<VisitPolicy, Strategy>,
-                      false // do not check for empty multipolygon (done above)
-                  >::apply(boost::begin(multipolygon),
-                           boost::end(multipolygon),
-                           per_polygon<VisitPolicy, Strategy>(visitor, strategy)))
+        if (! (boost::end(multipolygon) == std::find_if(
+            boost::begin(multipolygon),
+            boost::end(multipolygon),
+            per_polygon<VisitPolicy, Strategy>(visitor, strategy))))
```
awulkiew, line 297: Is preserving the old behavior for empty multipolygons important here? Now it's different.

I'd think about calculating `end` once and replacing `!` and `==` with `!=`.

include/boost/geometry/algorithms/detail/is_valid/polygon.hpp
```diff
@@ -94,9 +94,9 @@ class is_valid_polygon
         {}
 
         template <typename Ring>
-        inline bool apply(Ring const& ring) const
+        inline bool operator()(Ring const& ring) const
         {
-            return detail::is_valid::is_valid_ring
+            return ! detail::is_valid::is_valid_ring
```
awulkiew, line 99: Same story, the class defining this operator could be called `is_invalid_ring`.

include/boost/geometry/algorithms/detail/is_valid/polygon.hpp
```diff
@@ -111,14 +111,10 @@ class is_valid_polygon
...
-                <
-                    per_ring<VisitPolicy, Strategy>,
-                    true // allow for empty interior ring range
-                >::apply(boost::begin(interior_rings),
-                         boost::end(interior_rings),
-                         per_ring<VisitPolicy, Strategy>(visitor, strategy));
+        return boost::end(interior_rings) == std::find_if(
+            boost::begin(interior_rings),
+            boost::end(interior_rings),
+            per_ring<VisitPolicy, Strategy>(visitor, strategy));
```
awulkiew, line 117: The behavior is the same. I'd only think about creating `end` once and replacing `per_ring` to `is_invalid_ring`.

include/boost/geometry/policies/predicate_based_interrupt_policy.hpp
```diff
@@ -43,12 +43,12 @@ struct stateless_predicate_based_interrupt_policy
     template <typename Range>
     inline bool apply(Range const& range)
     {
-        // if there is at least one unacceptable turn in the range, return false
```
awulkiew, line 46: I'd probably keep the comment.

include/boost/geometry/policies/predicate_based_interrupt_policy.hpp
```diff
@@ -43,12 +43,12 @@ struct stateless_predicate_based_interrupt_policy
...
-            <
-                IsAcceptableTurnPredicate,
-                AllowEmptyTurnRange
-            >::apply(boost::begin(range), boost::end(range));
+        has_intersections = ! (boost::end(range) == std::find_if(
+            boost::begin(range), 
+            boost::end(range),
+            []( auto const& turn ){ 
+                return ! IsAcceptableTurnPredicate::apply(turn); }
+        ) || (AllowEmptyTurnRange && boost::empty(range)));
```
awulkiew, line 51: This is quite complex. How about:
```
auto const end = boost::end(range);
bool const has_unacceptable_turn
    = std::find_if(boost::begin(range), end,
                   []( auto const& turn ) { 
                       return ! IsAcceptableTurnPredicate::apply(turn);
                   }) != end;
has_intersections = ! has_unacceptable_turn
                 || (AllowEmptyTurnRange && boost::empty(range)));
```

or something more clear like that.

Also the comparison operator wasn't correct, right? With `==` true would be returned if all turns were acceptable or for empty range. So `has_intersection` would be true for unacceptable turns and non-empty or `AllowEmptyTurnRange` and empty. Or am I missing something? Does all of the tests pass?

include/boost/geometry/policies/predicate_based_interrupt_policy.hpp
```diff
@@ -79,11 +79,12 @@ struct predicate_based_interrupt_policy
...
-            <
-                IsAcceptableTurnPredicate,
-                AllowEmptyTurnRange
-            >::apply(boost::begin(range), boost::end(range), m_predicate);
+        has_intersections = ! (boost::end(range) == std::find_if(
+            boost::begin(range), 
+            boost::end(range),
+            [&]( auto const& turn ){ 
+                return ! m_predicate.apply(turn); }
+        ) || (AllowEmptyTurnRange && boost::empty(range)));
```
awulkiew, line 87: See above.

include/boost/geometry/policies/predicate_based_interrupt_policy.hpp
awulkiew: Ok, I know what I missed, parenthesis.

test/algorithms/test_is_valid.hpp
```diff
@@ -98,17 +97,18 @@ struct is_convertible_to_closed<Ring, bg::ring_tag, bg::open>
...
-        return bg::detail::check_iterator_range
-            <
-                is_convertible_to_closed<ring_type>
-            >::apply(boost::begin(interior_rings),
-                     boost::end(interior_rings));
+        return boost::end(interior_rings) == std::find_if(
+            boost::begin(interior_rings),
+            boost::end(interior_rings),
+            []( auto const& ring ){ return ! is_convertible_to_closed
+                <decltype(ring)>::apply(ring); }
```
awulkiew, line 110: `decltype(ring)` is const reference is this what you wanted?

Also here different formatting of lambda expression is used.

test/algorithms/test_is_valid.hpp
```diff
@@ -121,16 +121,16 @@ struct is_convertible_to_closed<Polygon, bg::polygon_tag, bg::open>
...
-                is_convertible_to_closed<polygon>,
-                false // do not allow empty multi-polygon
-            >::apply(boost::begin(multi_polygon),
-                     boost::end(multi_polygon));
+        return boost::end(multi_polygon) == std::find_if(
+            boost::begin(multi_polygon),
+            boost::end(multi_polygon),
+            []( auto const& polygon ){ return ! is_convertible_to_closed
+                <polygon_type>::apply(polygon); }
+        ) && !boost::empty(multi_polygon); // do not allow empty multi-polygon
```
awulkiew, line 133: Emptiness could be checked first because it's cheap.

awulkiew:
I think this is a good change. I've found some issues though.

include/boost/geometry/algorithms/detail/is_valid/has_invalid_coordinate.hpp
vissarion: I changed it to lambda and removed `point_has_valid_coordinates`

include/boost/geometry/algorithms/detail/is_valid/linear.hpp
vissarion: Yes, here is irrelevant.

include/boost/geometry/algorithms/is_empty.hpp
```diff
@@ -62,10 +62,10 @@ class polygon_is_empty
...
     static inline bool check_interior_rings(InteriorRings const& interior_rings)
     {
-        return check_iterator_range
-            <
-                range_is_empty, true // allow empty range
-            >::apply(boost::begin(interior_rings), boost::end(interior_rings));
+        return boost::end(interior_rings) == std::find_if(
+            boost::begin(interior_rings), 
+            boost::end(interior_rings),
+            []( auto const& range ){ return ! range_is_empty::apply(range); });
```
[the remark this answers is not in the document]
vissarion: > This could probably be replaced with:
> 
> ```
> []( auto const& range ){ return ! boost::empty(range); }
> ```

+1

> You could also check if `range_is_empty` is needed because it's possible that it's not used anywhere else.

It is used as default policy to `multi_is_empty`

include/boost/geometry/policies/predicate_based_interrupt_policy.hpp
awulkiew: There was one negation missing, so maybe it's better top keep these parenthesis and negate the whole expression as you did. For that `has_unacceptable_turn` would have to be renamed to something like `has_acceptable_turns_or_empty`.

vissarion: I think there is one more wrong negation still in the your proposed code. Indeed it is complicated. The correct should be
```
bool const has_unacceptable_turn
    = std::find_if(boost::begin(range), end,
                   []( auto const& turn ) { 
                       return ! IsAcceptableTurnPredicate::apply(turn);
                   }) != end;
has_intersections = has_unacceptable_turn
    && !(AllowEmptyTurnRange && boost::empty(range));
```
with only one negation as in the original but less complicated as you propose.

test/algorithms/test_is_valid.hpp
vissarion: This is a mistake, should be `ring_type`

vissarion:
Thanks for the review @awulkiew 
I addressed all your comments.

include/boost/geometry/policies/predicate_based_interrupt_policy.hpp
awulkiew: Well, I don't know, the original comment states:
```
// if there is at least one unacceptable turn in the range, return false
```
So AFAIU if `has_unacceptable_turn` is `true` then `has_intersections` should be `false`, right? So either something is wrong with the naming or with `find_if`. Or I'm missing something.

vissarion: See the original code too, it seems that the original comment is wrong it should state 
```
// if there is at least one unacceptable turn in the range, return false
```

awulkiew: There are two booleans here:
- `has_intersections` which is set to `true` if there are unacceptable turns or if the ring is empty when it shouldn't, this is used in `in_valid` and may have a different meaning than in other parts of the library, here it means that there the turns that were found are invalid [1].
- return value which indicates if processing of turns should end [2], AFAIU `true` means stop [3]

So if there are unacceptable turns found (or empty ring when it shouldn't) `has_intersections` should be `true` and this should also be the return value. So you're correct and I was missing that here `has_intersections` is used to return error condition.

[1] https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/is_valid/has_valid_self_turns.hpp#L92-L95
[2] https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/is_valid/has_valid_self_turns.hpp#L80-L83
[3] https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/self_turn_points.hpp#L48-L59

include/boost/geometry/algorithms/detail/disjoint/multirange_geometry.hpp
```diff
@@ -68,13 +68,11 @@ struct multirange_constant_size_geometry
...
+            >;
 
-        return detail::check_iterator_range
-            <
-                unary_predicate_type
-            >::apply(boost::begin(multirange), boost::end(multirange),
-                     unary_predicate_type(constant_size_geometry, strategy));
+        auto const end = boost::end(multirange);
+        return std::find_if(boost::begin(multirange), end, 
+            unary_predicate_type(constant_size_geometry, strategy)) == end;
```
awulkiew, line 75: I know that this is how this type was called before but don't you think that it'd be more readable if it instead of generic `predicate` we'd call it `not_disjoint`? Then this expression/sentence would be: find if not disjoint and the reader wouldn't have to check what the "predicate" is.

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
```diff
@@ -285,19 +285,17 @@ struct is_simple_multilinestring
...
-                 <
-                     per_ls, // do not compute self-intersections
-                     true // allow empty multilinestring
-                 >::apply(boost::begin(multilinestring),
-                          boost::end(multilinestring),
-                          per_ls(strategy))
-             )
+        // return true for empty multilinestring
+
+        using per_ls = per_linestring<Strategy>; // do not compute self-intersections
```
awulkiew, line 293: Here too, it's unclear what is checked below. I know that this is how it was implemented so I'd be ok with leaving it for now as it is but we could think about changing it.

include/boost/geometry/algorithms/detail/is_valid/has_invalid_coordinate.hpp
```diff
@@ -80,26 +80,18 @@ struct indexed_has_invalid_coordinate
...
-                point_has_valid_coordinates,
-                true // do not consider an empty range as problematic
-            >::apply(geometry::points_begin(geometry),
-                     geometry::points_end(geometry));
+        auto const points_end = geometry::points_end(geometry);
+        bool const has_valid_coordinates = std::find_if
+            (
+                geometry::points_begin(geometry), points_end,
+                []( auto const& point ){ return point_has_invalid_coordinate::
+                    apply(point); }
```
awulkiew, line 93: Inconsistent formatting of lambdas:
```
[]( auto const& point ) {
    return point_has_invalid_coordinate::apply(point);
}
```

Or you could add `point_has_invalid_coordinate::operator()` and simply pass it here.

include/boost/geometry/algorithms/detail/is_valid/multipolygon.hpp
```diff
@@ -257,29 +257,30 @@ class is_valid_multipolygon
 
 
     template <typename VisitPolicy, typename Strategy>
-    struct per_polygon
+    struct is_invalid_polygon
```
awulkiew, line 260: Other places could be changed like this. Now it's clear what's the purpose of this struct.

include/boost/geometry/policies/predicate_based_interrupt_policy.hpp
```diff
@@ -43,12 +43,18 @@ struct stateless_predicate_based_interrupt_policy
...
-            >::apply(boost::begin(range), boost::end(range));
+        // if there is at least one unacceptable turn in the range, return true
+        auto const end = boost::end(range);
+        bool const has_unacceptable_turn = std::find_if
+            (
+                boost::begin(range), end,
+                []( auto const& turn ) { 
+                    return ! IsAcceptableTurnPredicate::apply(turn);
+                }
+            ) != end;
```
awulkiew, line 54: I don't think that this formatting is correct.

First of all we do accept up to 100 characters in a line so you don't have to put everything in subsequent lines.

Another thing is that function name could be placed in the next line (maybe, do we allow this?):
```
bool const has_unacceptable_turn
    = std::find_if(boost::begin(range), end,
                   [](auto const& turn) { 
                       return ! IsAcceptableTurnPredicate::apply(turn);
                   }) != end;
```

Another solution would be to create lambda expression before the call:
```
auto is_unacceptable = [](auto const& turn) { return ! IsAcceptableTurnPredicate::apply(turn); };
```

include/boost/geometry/policies/predicate_based_interrupt_policy.hpp
```diff
@@ -78,12 +84,18 @@ struct predicate_based_interrupt_policy
...
-            >::apply(boost::begin(range), boost::end(range), m_predicate);
+        // if there is at least one unacceptable turn in the range, return true
+        auto const end = boost::end(range); 
+        bool const has_unacceptable_turn = std::find_if
+            (
+                boost::begin(range), end,
+                [&]( auto const& turn ) { 
+                    return ! m_predicate.apply(turn);
+                }
+            ) != end;
```
awulkiew, line 95: Same here.

test/algorithms/test_is_valid.hpp
```diff
@@ -98,17 +97,19 @@ struct is_convertible_to_closed<Ring, bg::ring_tag, bg::open>
...
-                is_convertible_to_closed<ring_type>
-            >::apply(boost::begin(interior_rings),
-                     boost::end(interior_rings));
+        auto const end = boost::end(interior_rings);
+        return std::find_if(
+            boost::begin(interior_rings), end, 
+            []( auto const& ring ){ 
+                return ! is_convertible_to_closed<ring_type>::apply(ring);
+            }
+        ) == end;
```
awulkiew, line 112: Is this style your personal preference or is this formatting caused by the 80 characters limit?

awulkiew (approved):

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
vissarion: I agree, renamed it to `not_simple`.

include/boost/geometry/algorithms/detail/is_valid/has_invalid_coordinate.hpp
vissarion: Thanks, I changed the format.

include/boost/geometry/policies/predicate_based_interrupt_policy.hpp
vissarion: I think I prefer 
```
bool const has_unacceptable_turn = std::find_if(boost::begin(range), end,
    [](auto const& turn) { 
        return ! IsAcceptableTurnPredicate::apply(turn);
    }) != end;
```
what do you think?

test/algorithms/test_is_valid.hpp
vissarion: personal preference but see in the comment above for an alternative

vissarion: what about 
```
return std::find_if(boost::begin(interior_rings), end, 
    []( auto const& ring ){ 
        return ! is_convertible_to_closed<ring_type>::apply(ring);
    }) == end;
```

include/boost/geometry/policies/predicate_based_interrupt_policy.hpp
awulkiew: It looks good to me.

include/boost/geometry/policies/predicate_based_interrupt_policy.hpp
vissarion: changed as above

vissarion:
:wave: @barendgehrels are you OK with merging this PR?

awulkiew:
Since we're waiting for @barendgehrels with the approval I have a question/proposal.

In the cases changed by PR the same pattern is used, i.e. `std::find_if(...) == last`. What do you think about replacing it with `std::none_of(...)`, which would be easier to understand?

E.g. something like this:
```
std::find_if(boost::begin(multi), boost::end(multi), covered_by_pred) == end
```
could be replaced by:
```
std::none_of(boost::begin(multi), boost::end(multi), covered_by_pred)
```

vissarion:
> Since we're waiting for @barendgehrels with the approval I have a question/proposal.
> 
> In the cases changed by PR the same pattern is used, i.e. `std::find_if(...) == end`. What do you think about replacing it with `std::none_of(...)`, which would be easier to understand?
> 
> E.g. something like this:
> 
> ```
> auto const end = boost::end(multi);
> std::find_if(boost::begin(multi), end, covered_by_pred) == end;
> ```
> 
> could be replaced with:
> 
> ```
> std::none_of(boost::begin(multi), boost::end(multi), covered_by_pred);
> ```

Thanks @awulkiew that's a very good idea, I have forgotten the existence of those (`std::none_of` and `std::any_of`). They are quite a fit for this case.

barendgehrels:
> Since we're waiting for @barendgehrels with the approval I have a question/proposal.
> 
> In the cases changed by PR the same pattern is used, i.e. `std::find_if(...) == end`. What do you think about replacing it with `std::none_of(...)`, which would be easier to understand?
> 
> E.g. something like this:
> 
> ```
> auto const end = boost::end(multi);
> std::find_if(boost::begin(multi), end, covered_by_pred) == end;
> ```
> 
> could be replaced with:
> 
> ```
> std::none_of(boost::begin(multi), boost::end(multi), covered_by_pred);
> ```

Yes, this is nice. I didn't approve yet because I had some questions, and they are not answered yet.

My main question/issue was that there were many small structures containing `not_...` which, I think, would be more reusable if they were implemented in a positive way, and then a `not` around it. 
Anyway, that does not prevent merging, I'm fine with it.

barendgehrels (approved):
As just mentioned, you can decide to merge it like this or think if my comment makes sense to you.
Thanks.

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
```diff
@@ -285,19 +285,17 @@ struct is_simple_multilinestring
...
-                 <
-                     per_ls, // do not compute self-intersections
-                     true // allow empty multilinestring
-                 >::apply(boost::begin(multilinestring),
-                          boost::end(multilinestring),
-                          per_ls(strategy))
-             )
+        // return true for empty multilinestring
+
+        using not_simple = not_simple<Strategy>; // do not compute self-intersections
```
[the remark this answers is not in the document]
awulkiew: This would be `std::unary_negate<is_simple<Strategy>>` and it'd have to be created with: `std::unary_negate<is_simple<Strategy>>(is_simple<Strategy>(strategy))` making it more complex than the function call below.

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
```diff
@@ -261,16 +261,16 @@ struct is_simple_multilinestring
...
+        not_simple(Strategy const& strategy)
             : m_strategy(strategy)
         {}
 
         template <typename Linestring>
-        inline bool apply(Linestring const& linestring) const
+        inline bool operator()(Linestring const& linestring) const
         {
-            return detail::is_simple::is_simple_linestring
+            return ! detail::is_simple::is_simple_linestring
```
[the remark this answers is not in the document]
awulkiew: `is_not_simple`? ;)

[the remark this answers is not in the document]
vissarion: > `is_not_simple`? ;)

why not, @barendgehrels ?

include/boost/geometry/algorithms/detail/is_simple/areal.hpp
```diff
@@ -116,12 +116,11 @@ struct is_simple<MultiPolygon, multi_polygon_tag>
...
     {
-        auto const end = boost::end(multipolygon);
-        return std::find_if(boost::begin(multipolygon), end,
-                [&](auto const& po) {
-                    return ! detail::is_simple::is_simple_polygon(po, strategy);
-                }) == end; // non-simple polygon not found
-                // allow empty multi-polygon
+        return std::none_of(boost::begin(multipolygon), boost::end(multipolygon),
+                            [&](auto const& po) {
+                                return ! detail::is_simple::is_simple_polygon(po, strategy);
```
barendgehrels, line 121: So this one looks good.

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
```diff
@@ -261,16 +261,16 @@ struct is_simple_multilinestring
 {
 private:
     template <typename Strategy>
-    struct per_linestring
+    struct not_simple
     {
-        per_linestring(Strategy const& strategy)
+        not_simple(Strategy const& strategy)
```
barendgehrels, line 266: But this one should be: `simple`
We shouldn't add many structures with `not_XXX`. They should be called `XXX` then they are reusable.

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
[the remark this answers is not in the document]
awulkiew: If we don't intend to use it anywhere else it doesn't have to be flexible. It makes no sense to add complexity only because we might some day in the future do something with it.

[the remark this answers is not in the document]
awulkiew: Besides is_simple + negate to represent is_not_simple is as flexible as is_not_simple + negate to represent is_simple. It is just a matter of preference, not flexibility.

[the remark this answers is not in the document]
awulkiew: It wouldn't be that complex if we used `std::not1`, i.e. `std::not1(is_simple<Strategy>(strategy))`, but this function is deprecated in C++17. In C++17 `std::not_fn` was introduced but we can't use that. So we'd be forced to write our own utility if we wanted this to look nice.

barendgehrels:
Hi, since this PR is meant as an improvement, I'm still in favor of the `all_of` combined with a positive predicate (`all_of(... is_valid)`). It's as simple as it is now, but clearer. @vissarion what do you think.

(So this is instead of one my earlier remarks (with the `find_if` / `negate` it is indeed more verbose, but that is not necessary)

Sorry that this takes so long then, I think I raised this issue already in November (but can't find it now). 
Anyway, if you both think it's fine like this, I'm fine with merging or course.

vissarion:
> Hi, since this PR is meant as an improvement, I'm still in favor of the `all_of` combined with a positive predicate (`all_of(... is_valid)`). It's as simple as it is now, but clearer. @vissarion what do you think?
> 
> (So this is instead of one my earlier remarks (with the `find_if` / `negate` it is indeed more verbose, but that is not necessary)
> 
> Sorry that this takes so long then, I think I raised this issue already in November (but can't find it now). Anyway, if you both think it's fine like this, I'm fine with merging or course.

I agree. I changed all the double negations (combinations of `none_of` with negative predicates). If you are ok with the current version I will merge it. Thanks for your comments.

awulkiew:
I'm ok with any approach as long as it doesn't require to make the code more complex. I also agree with @barendgehrels that it looks better this way.

barendgehrels (approved):

barendgehrels:
> > Hi, since this PR is meant as an improvement, I'm still in favor of the `all_of` combined with a positive predicate (`all_of(... is_valid)`). It's as simple as it is now, but clearer. @vissarion what do you think?
> > (So this is instead of one my earlier remarks (with the `find_if` / `negate` it is indeed more verbose, but that is not necessary)
> > Sorry that this takes so long then, I think I raised this issue already in November (but can't find it now). Anyway, if you both think it's fine like this, I'm fine with merging or course.
> 
> I agree. I changed all the double negations (combinations of `none_of` with negative predicates). If you are ok with the current version I will merge it. Thanks for your comments.

Thank you!
I'm OK now!

PR #938, opened 2021-11-17
test: add and use test settings

barendgehrels:
This
* excludes cases failing for rescaling
* removes duplicate testcases (all was already tested in all directions)
* adds settings to explicitly just skip some of the cases
* fixes svg and other small changes

There are still some cases failing without rescaling, to be investigated further

### DIFF
--- a/test/algorithms/set_operations/difference/difference_linear_linear.cpp
+++ b/test/algorithms/set_operations/difference/difference_linear_linear.cpp
@@ -271,12 +271,6 @@ BOOST_AUTO_TEST_CASE( test_difference_linestring_linestring )
          "lldf16");
 
     tester::apply
-        (from_wkt<L>("LINESTRING(0 0,1 0,2 1,3 2)"),
-         from_wkt<L>("LINESTRING(4 5,3 2,1 0,0.5 0)"),
-         from_wkt<ML>("MULTILINESTRING((0 0,0.5 0))"),
-         "lldf16-r");
-
-    tester::apply
         (from_wkt<L>("LINESTRING(0 0,10 0,20 1,30 1)"),
          from_wkt<L>("LINESTRING(1 1,2 0,3 1,20 1,25 1)"),
          from_wkt<ML>("MULTILINESTRING((0 0,10 0,20 1),(25 1,30 1))"),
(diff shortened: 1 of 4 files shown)


test/algorithms/set_operations/difference/difference_linear_linear.cpp
```diff
@@ -270,12 +270,6 @@ BOOST_AUTO_TEST_CASE( test_difference_linestring_linestring )
          from_wkt<ML>("MULTILINESTRING((0 0,0.5 0))"),
          "lldf16");
 
-    tester::apply
```
vissarion, line 273: I understand that all the reverse tests `-r` are removed as duplicates and the rest are disactivated for now and fixed in the future. Are the `-r` tests failing now too? Are they going to be fixed in the future as well?

barendgehrels: - the `-r` cases are indeed removed as duplicate
- some of the `-r` cases are now failing, and only they are disabled
- yes, this was the first step to see what's going wrong there and I will continue next Wednesday
- so this PR is a first step - it was the last step in letting all tests pass, without rescaling, in set_operations and buffer

test/algorithms/set_operations/difference/difference_linear_linear.cpp
```diff
@@ -548,61 +511,45 @@ BOOST_AUTO_TEST_CASE( test_difference_linestring_linestring )
...
-         from_wkt<ML>("MULTILINESTRING((10.0002 2,9 -1032.34324,.3 8,0 5,8 1),(4 .4,2 8))"),
-         "lldf31z",
-         1e-10
-         );
+    {
+        ut_settings settings{1e-10};
+#if ! defined(BOOST_GEOMETRY_USE_RESCALING) && ! defined(BOOST_GEOMETRY_TEST_FAILURES)
+        // Cases failing without rescaling when first linestring is reversed
+        settings.test_reverse_normal = false;
+        settings.test_reverse_reverse = false;
```
barendgehrels, line 519: @vissarion for example you can see here that only two of the cases are excluded from the test, and only for if rescaling is not used. So in the default case, everything is still tested as it was before (now without duplicates)

vissarion (approved):
Thanks for the explanations.

test/algorithms/set_operations/difference/difference_linear_linear.cpp
awulkiew: I don't understand what is the rationale behind removing them? These are not duplicates since the inputs are different. The fact that some of them fails now is a good reason why they should stay. It's because they might fail in the future after changing something else. This is what tests are for, not to pass but to fail if something is wrong.

awulkiew: @vissarion Ok thanks. So they are not removed but instead various `settings` like `test_normal_reverse` are used inside.

awulkiew (approved):

test/algorithms/set_operations/difference/difference_linear_linear.cpp
barendgehrels: Sure, conform description

The reverse cases where already tested inside. So they were always duplicate. But now they are configurable.

PR #933, opened 2021-11-05
Closest points algorithm with new strategies for all cartesian geometries except boxes

vissarion:
This PR builds on top of https://github.com/boostorg/geometry/pull/923

PR #707, opened 2020-04-29
New algorithm closest_points

vissarion:
## The problem 
This PR proposes an implementation of closest points algorithm. That is given two geometries G1, G2 compute the pair of points p1 \in G1, p2 \in G2 such that the distance between p1, p2 is minimized. 
In case that G1, G2 have an non empty intersection the function returns an (arbitrary) common point. See also issue  https://github.com/boostorg/geometry/issues/633
A similar function in `PostGIS` is the `ST_ClosestPoint` https://postgis.net/docs/ST_ClosestPoint.html

### Input: 
POINT(4.3 1.9);
POLYGON((2 1.3,2.4 1.7,2.8 1.8,3.4 1.2,3.7 1.6,3.4 2,4.1 3,5.3 2.6,5.4 1.2,4.9 0.8,2.9 0.7,2 1.3)(4.0 2.0, 4.2 1.4, 4.8 1.9, 4.4 2.2, 4.0 2.0))
LINESTRING(5 0, 2 0));
MULTIPOINT((2 0.5,2 3))
### Output: 
Point-Poly: LINESTRING(4.3 1.9,4.2 2.1)
Poly-Line: LINESTRING(2.9 0.7,2.9 0)
Poly-MultiPoint: LINESTRING(2.36923 1.05385,2 0.5)

closest_points_example

 ## Solution remarks

* Largely reuse the code of `distance` and `disjoint` algorithms wherever possible. 
* Created 5 strategies for each coordinate system (15 in total) namely pt-pt, pt-seg, pt-box, box-box and seg-box similarly to distance strategies.
* New formulas for computing the longitude of a point on a geodesic segment given starting point, azimuth and latitude of given point. In other words, compute where a given segment crosses a parallel. For geographic CS this is a variant of direct geodesic problem. For spherical this can be computed by a trigonometric formula.
* For all CSs algorithms and formulas have moved from `strategies` to `formulas`.
* All services of closest_point strategies are in `distance::service` namespace expect from default strategies that are in `closest_points::service`

PR #939, opened 2021-11-18
Closest points algorithm with new strategies for spherical and geographic geometries except boxes

vissarion:
This PR build on top of https://github.com/boostorg/geometry/pull/933


vissarion:
This PR includes documentation for the `closest_points` algorithm. Even if the documentation is build without errors (both locally and on CI) the generated algorithm is empty (i.e. it appears in the html list of algorithms but the algorithm's html page only contains its name). Also, the `doc/generated/closest_points.qbk` is almost empty, it contains only the copyright info and comments ` [/ Generated by doxygen_xml2qbk...]`. 

Note that I followed the instructions from `https://github.com/boostorg/geometry/blob/develop/doc/readme.txt`, I am using `gcc-9` and `python 3.8` in `ubuntu 20.04.3 LTS`. 

Interestingly the branch https://github.com/boostorg/geometry/pull/707 generates the correct documentation. 

I am missing something here that I have to investigate more or is it a more general problem with the doc system?

doc/src/examples/examples_utils/create_svg_closest_points.hpp
```diff
@@ -0,0 +1,66 @@
...
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+// Code to create SVG for closest_points examples
+
+#ifndef CREATE_SVG_CLOSEST_POINTS_HPP
+#define CREATE_SVG_CLOSEST_POINTS_HPP
+
+#include <fstream>
+#if defined(HAVE_SVG)
```
barendgehrels, line 17: Very minor, using this define can go here :-)

barendgehrels (approved):
Thanks!

vissarion:
> Thanks!

Thanks for the review. Since this PR is on top of https://github.com/boostorg/geometry/pull/933 you have also reviewed and approved https://github.com/boostorg/geometry/pull/933 implicitly. 

@awulkiew it is up to you to review those PRs all in once or separately, e.g. if you have many comments it would be useful to separate otherwise not.

doc/doxy/Doxyfile
```diff
@@ -1,187 +1,919 @@
...
+#
+# All text after a double hash (##) is considered a comment and is placed in
+# front of the TAG it is preceding.
+#
+# All text after a single hash (#) is considered a comment and will be ignored.
+# The format is:
+# TAG = value [value, ...]
+# For lists, items can also be appended using:
+# TAG += value [value, ...]
+# Values that contain spaces should be placed between quotes (\" \").
```
awulkiew: There are many additional comments in this file. Is this intentional?

doc/index/Doxyfile
```diff
@@ -1,17 +1,131 @@
...
+# All text after a double hash (##) is considered a comment and is placed in
+# front of the TAG it is preceding.
+#
+# All text after a single hash (#) is considered a comment and will be ignored.
+# The format is:
+# TAG = value [value, ...]
+# For lists, items can also be appended using:
+# TAG += value [value, ...]
+# Values that contain spaces should be placed between quotes (\" \").
+
```
awulkiew: The modifications in this file are not needed. Or am I missing something?

doc/readme.txt
```diff
@@ -31,7 +31,8 @@ reference: generated documentation QBK files (by doxygen_xml2qbk)
 src: examples used in documentation and tools (doxygen_xml2qbk)
 
 Per new algorithm (e.g. foo), one should add:
-1) in file boost/geometry/algorithms/foo.hpp, include a "\ingroup foo" in the doxygen comments
+1) in file boost/geometry/algorithms/foo.hpp, include a "\ingroup foo" in the doxygen comments 
+   (if a directory is created e.g. boost/geometry/algorithms/detail/foo the path should be added in doc/doxy/Doxyfile)
```
awulkiew, line 35: +1 though this is a different issue than the one mentioned in 1) Shouldn't it be a separate point, e.g. 3) ?

include/boost/geometry/algorithms/detail/closest_points/linear_or_areal_to_areal.hpp
```diff
@@ -0,0 +1,269 @@
...
+    {
+        using most_precise_type = typename select_coordinate_type
+                <
+                    Linear,
+                    Areal
+                >::type;
+
+        using point_type = typename std::conditional
+        <
+            std::is_same<coordinate_type<Linear>, most_precise_type>::value,
```
awulkiew, line 45: `geometry::coordinate_type` is not a type alias so here it should be `typename coordinate_type<Linear>::type`.

Side note. We could think about adding type aliases and variable templates for traits: `coordinate_t`, `point_t`, `dimension_v`, etc.

include/boost/geometry/algorithms/detail/closest_points/linear_or_areal_to_areal.hpp
```diff
@@ -0,0 +1,269 @@
...
+         *       this should be implemented directly with dynamic geometries
+        using polygon_type = geometry::model::polygon<point_type>;
+        std::tuple
+        <
+            geometry::model::multi_point<point_type>, 
+            geometry::model::multi_linestring<linestring_type>,
+            geometry::model::multi_polygon<polygon_type>
+        > tp;
+        bool intersect_tp = geometry::intersection(linear, areal, tp, strategies);
+        */
```
awulkiew, line 62: It should be possible to pass tupled geometry as output geometry of `intersection()`.
https://github.com/boostorg/geometry/blob/develop/test/algorithms/set_operations/intersection/intersection_tupled.cpp

include/boost/geometry/algorithms/detail/closest_points/linear_or_areal_to_areal.hpp
```diff
@@ -0,0 +1,269 @@
...
+    {
+        using most_precise_type = typename select_coordinate_type
+                <
+                    Areal1,
+                    Areal2
+                >::type;
+
+        using point_type = typename std::conditional
+        <
+            std::is_same<coordinate_type<Areal1>, most_precise_type>::value,
```
awulkiew, line 152: `typename coordinate_type<Areal1>::type`

include/boost/geometry/algorithms/detail/closest_points/linear_or_areal_to_areal.hpp
```diff
@@ -0,0 +1,269 @@
...
+        /* TODO: currently intersection does not support tupled input
+         *       this should be implemented directly with dynamic geometries
+        */
+            
+        geometry::model::multi_point<point_type> mp_out;
+        geometry::intersection(areal1, areal2, mp_out, strategies);
+        
+        if (! boost::empty(mp_out))
+        {
+            set_segment_from_points::apply(*boost::begin(mp_out),
```
awulkiew, line 169: In case you wanted to make this shorter we have `util::front()`.

include/boost/geometry/algorithms/detail/closest_points/multipoint_to_geometry.hpp
```diff
@@ -0,0 +1,326 @@
...
+                             OutSegment& shortest_seg,
+                             Strategies const& strategies)
+    {
+        using linestring_type = geometry::model::linestring
+            <
+                typename point_type<Segment>::type
+            >;
+        linestring_type linestring;
+        convert(segment, linestring);
+        multipoint_to_linear::apply(multipoint, linestring, shortest_seg, strategies);
```
awulkiew, line 131: Converting segment to linestring seems to be unnecessary but if it simplifies the implementation then I'm fine with it.

include/boost/geometry/algorithms/detail/closest_points/linear_to_linear.hpp
```diff
@@ -0,0 +1,155 @@
...
+                    typename point_type<Linear2>::type,
+                    Linear1
+                >::apply(*points_begin(linear2), linear1, shortest_seg, strategies);
+            detail::closest_points::swap_segment_points::apply(shortest_seg);
+            return;
+        }
+
+        if (geometry::num_segments(linear1) < geometry::num_segments(linear2))
+        {
+            point_or_segment_range_to_geometry_rtree::apply(
```
awulkiew, line 65: This can be called for empty geometry which will cause assertion failure inside `point_or_segment_range_to_geometry_rtree`. Or am I missing something?

include/boost/geometry/algorithms/detail/closest_points/linear_to_linear.hpp
```diff
@@ -0,0 +1,155 @@
...
+                geometry::segments_begin(linear2),
+                geometry::segments_end(linear2),
+                linear1,
+                shortest_seg,
+                strategies);
+            detail::closest_points::swap_segment_points::apply(shortest_seg);
+            return;
+        }
+
+        point_or_segment_range_to_geometry_rtree::apply(
```
awulkiew, line 75: Here also empty geometry may be passed and `point_or_segment_range_to_geometry_rtree` is unable to handle that.

include/boost/geometry/algorithms/detail/closest_points/segment_to_segment.hpp
```diff
@@ -0,0 +1,154 @@
...
+                decltype(strategies.distance(cp1, q[1]))
+            >::apply(strategies.distance(cp1, q[1])).apply(cp1, q[1]);
+        d[2] = strategy::distance::services::get_comparable
+            <
+                decltype(strategies.distance(p[0], cp2))
+            >::apply(strategies.distance(p[0], cp2)).apply(p[0], cp2);
+        d[3] = strategy::distance::services::get_comparable
+            <
+                decltype(strategies.distance(p[1], cp3))
+            >::apply(strategies.distance(p[1], cp3)).apply(p[1], cp3);
```
awulkiew, line 106: All of these calculate point to point distances right?

If that's the case you only need to create one strategy, e.g. using the first pair of points. But there is actually no need to pass the actual points. You could pass `detail::dummy_point`s [1]. This way you could create the strategy somewhere above where the points are unavailable. Furthermore instead of `get_comparable` you could use `strategies::distance::detail::make_comparable()` [2] to make umbrella strategy returning comparable legacy strategies from `distance()` member function. Though this would be an additional dependency so you might want to choose to stick with `get_comparable` but using it only once. Anyway, this is how this code could be reimplemented:
```
auto const cds = strategies::distance::detail::make_comparable(strategies)
                     .distance(detail::dummy_point(), detail::dummy_point());
d[0] = cds.apply(cp0, q[0]);
d[1] = cds.apply(cp1, q[1]);
// etc.
```

[1] https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/dummy_geometries.hpp#L26
[2] https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/strategies/distance/comparable.hpp#L51

include/boost/geometry/strategies/closest_points/geographic.hpp
```diff
@@ -0,0 +1,121 @@
...
+} // namespace detail
+#endif // DOXYGEN_NO_DETAIL
+
+template
+<
+    typename FormulaPolicy = strategy::andoyer,
+    typename Spheroid = srs::spheroid<double>,
+    typename CalculationType = void
+>
+class geographic
```
awulkiew, line 76: Out of curiosity, why didn't you just define `closest_points()` member function here?

include/boost/geometry/strategies/closest_points/spherical.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+} // namespace detail
+#endif // DOXYGEN_NO_DETAIL
+
+template
+<
+    typename RadiusTypeOrSphere = double,
+    typename CalculationType = void
+>
+class spherical
```
awulkiew, line 76: Here as well, why is base class in `detail` needed instead of defining membre functions here?

include/boost/geometry/strategies/spherical/distance_cross_track.hpp
```diff
@@ -508,22 +549,22 @@ template
...
     {
         return relate_segment_segment_strategy_type();
     }
 
-    typedef within::spherical_winding
+    using point_in_geometry_strategy_type = within::spherical_winding
         <
             void, void, CalculationType
-        > point_in_geometry_strategy_type;
+        >;
```
awulkiew, line 567: These are probably not used at all after the introduction of umbrella strategies.

include/boost/geometry/strategies/geographic/closest_points_pt_seg.hpp
```diff
@@ -0,0 +1,121 @@
...
+                <
+                    typename base_t::template return_type<Point, PointOfSegment>::type,
+                    dimension<PointOfSegment>::value,
+                    typename coordinate_system<PointOfSegment>::type
+                > cp;
+            
+            geometry::set_from_radian<0>(cp, result.lon);
+            geometry::set_from_radian<1>(cp, result.lat);
+
+            return cp;
```
awulkiew, line 114: So the result is a point using the coordinate type being the result of distance? If that's the case here then for consistency it has to be the case in cartesian CS too. So technically it's possible to pass points with integral coordinates and get a point with double coordinates as the output. Is that what we want?

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
```diff
@@ -62,9 +62,37 @@ namespace boost { namespace geometry
 
 namespace strategy { namespace distance
 {
-
 namespace detail
 {
+
+template <bool>
```
awulkiew, line 68: I propose to add a tparam name here to indicate what's happening. This would be `EnableClosestPoints` right?

Btw, if this utility alwasy has to take distance and coordinates of the closest point then it means that all of these always exist. If they exist they has to be calculated. So what is the purpose of assigning them conditionally if they are always there anyway? Or is this only the case in geographic CS?

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
```diff
@@ -254,17 +273,22 @@ private :
                                                       spheroid).distance
                       << std::endl;
 #endif
-            if (inverse_distance_type::apply(lon3, lat3, pl_lon, pl_lat, spheroid).distance
-                < inverse_distance_type::apply(lon3, lat3, pr_lon, pr_lat, spheroid).distance)
+            CT const dist_l = inverse_distance_type::apply
+                (lon3, lat3, pl_lon, pl_lat, spheroid).distance;
+            CT const dist_r = inverse_distance_type::apply
+                (lon3, lat3, pr_lon, pr_lat, spheroid).distance;
```
awulkiew, line 279: Accroding to our guidelines lines can have up to 100 characters. I don't know if this would fit into 100 characters but in general putting parts of the expression in multiple lines makes it harder to read.

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
```diff
@@ -103,128 +144,106 @@ class geographic_cross_track
...
-        typedef typename FormulaPolicy::template inverse<CT, true, false, false, false, false>
-                inverse_distance_type;
+        using direct_distance_type = typename FormulaPolicy::template direct
+            <
+                CT,
+                true,
+                false,
+                false,
+                false
+            >;
```
awulkiew, line 237: I propose to put all of this in one line. I can't see any benefits of formatting it like this.

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
```diff
@@ -307,17 +337,18 @@ private :
...
+                       result_type<CT>& result)
     {
-        typedef typename FormulaPolicy::template inverse<CT, true, true, false, true, true>
-                inverse_distance_azimuth_quantities_type;
-        typedef typename FormulaPolicy::template inverse<CT, false, true, false, false, false>
-                inverse_dist_azimuth_type;
-        typedef typename FormulaPolicy::template direct<CT, true, false, false, false>
-                direct_distance_type;
+        using inverse_distance_azimuth_quantities_type = typename 
+            FormulaPolicy::template inverse<CT, true, true, false, true, true>;
```
awulkiew, line 343: So here the same thing is formatted differently.

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
```diff
@@ -328,22 +359,25 @@ private :
...
-            // g4 is the angle between segment (p1,p2) and segment (p3,p4) that meet on p4 (GEO)
+            // g4 is the angle between segment (p1,p2) and segment (p3,p4)
+            // that meet on p4 (GEO)
 
-            CT a4 = inverse_dist_azimuth_type::apply(res14.lon2, res14.lat2,
-                                                lon2, lat2, spheroid).azimuth;
-            res34 = inverse_distance_azimuth_quantities_type::apply(res14.lon2, res14.lat2,
-                                                                    lon3, lat3, spheroid);
+            CT a4 = inverse_dist_azimuth_type
+                    ::apply(res14.lon2, res14.lat2, lon2, lat2, spheroid).azimuth;
```
awulkiew, line 374: I can see three types of formatting for such call in this file. In addition to this one there are also:
```
            CT const dist_l = inverse_distance_type::apply
                (lon3, lat3, pl_lon, pl_lat, spheroid).distance;
```
and
```
                s14 -= inverse_distance_type::apply(res14.lon2, res14.lat2,
                                                    pl_lon, pl_lat,
                                                    spheroid).distance/2;
```

I don't know which one is the best but we could pick one of them preferably with the least number of lines. Do we have guidelines for this?

I also insist on not breaking lines shorter than 101 characters.

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
```diff
@@ -431,21 +486,59 @@ private :
...
+    static inline auto apply(CT const& lo1, CT const& la1, //p1
+                             CT const& lo2, CT const& la2, //p2
+                             CT const& lo3, CT const& la3, //query point p3
+                             Spheroid const& spheroid)
+    {
+        using inverse_dist_azimuth_type = typename FormulaPolicy::template 
+            inverse<CT, true, true, false, false, false>;
+
+        using inverse_dist_azimuth_reverse_type = typename FormulaPolicy::
+            template inverse<CT, true, true, true, false, false>;
```
awulkiew, line 529: Should we pick one formatting here or is this ok?

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
```diff
@@ -479,19 +572,19 @@ private :
         //segment on equator
         //Note: antipodal points on equator does not define segment on equator
         //but pass by the pole
-        CT diff = geometry::math::longitude_distance_signed<geometry::radian>(lon1, lon2);
+        CT diff = geometry::math::longitude_distance_signed<geometry
+            ::radian>(lon1, lon2);
```
awulkiew, line 576: This line break is interesting. Are you using some kind of automatic fomatter?

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
```diff
@@ -527,40 +621,34 @@ private :
...
-            {
-                return res13;
-            }
-            else
-            {
-                return res23;
-            }
+            
+            auto res13 = apply(lon1, lat1, lon1, half_pi * sign_non_zero, lon3, 
+                lat3, spheroid);
```
awulkiew, line 626: Yet another version of function call formatting

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
```diff
@@ -569,14 +657,15 @@ private :
         // TODO: meridian case optimization
         if (geometry::math::equals(a312, c0) && meridian_not_crossing_pole)
         {
-            boost::tuple<CT,CT> minmax_elem = boost::minmax(lat1, lat2);
-            if (lat3 >= minmax_elem.template get<0>() &&
-                lat3 <= minmax_elem.template get<1>())
+            std::tuple<CT,CT> minmax_elem = std::minmax(lat1, lat2);
```
awulkiew, line 660: `std::minmax()` returns `std::pair<>`. There is no need to convert it to `std::tuple<>`.

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
```diff
@@ -669,21 +760,25 @@ private :
 
         if (Bisection)
         {
-            bisection<CT>(lon1, lat1, lon2, lat2, lon3, lat3,
-                          spheroid, res12.distance/2, res12.azimuth, result);
+            bisection(lon1, lat1, lon2, lat2, lon3, lat3, spheroid, 
+                res12.distance/2, res12.azimuth, result);
```
awulkiew, line 764: Ok so I see that in this file the formatting is consistent and it's the last version. I prefer the one that was here used before the change but this is ok too. The question is which one should we use in general?

include/boost/geometry/strategies/spherical/closest_points_pt_seg.hpp
```diff
@@ -0,0 +1,200 @@
...
+            
+            // Note: this is similar to spherical computation in geographic
+            // point_segment_distance formula
+            CT earth_radius = m_strategy.radius();
+            CT cos_frac = cos(dist_d1 / earth_radius) / cos(dist / earth_radius);
+            CT s14_sph = cos_frac >= 1 
+                ? CT(0) : cos_frac <= -1 ? math::pi<CT>() * earth_radius
+                                         : acos(cos_frac) * earth_radius;
+
+            CT a12 = geometry::formula::spherical_azimuth<>(lon1, lat1, lon2, lat2);
```
awulkiew, line 155: `const`

include/boost/geometry/strategies/spherical/distance_cross_track.hpp
```diff
@@ -50,6 +50,82 @@ namespace boost { namespace geometry
...
+                                 PointOfSegment const& sp2)
+        {            
+            CalculationType lon1 = geometry::get_as_radian<0>(sp1);
+            CalculationType lat1 = geometry::get_as_radian<1>(sp1);
+            CalculationType lon2 = geometry::get_as_radian<0>(sp2);
+            CalculationType lat2 = geometry::get_as_radian<1>(sp2);
+            CalculationType lon = geometry::get_as_radian<0>(p);
+            CalculationType lat = geometry::get_as_radian<1>(p);
+
+            CalculationType crs_AD = geometry::formula::spherical_azimuth
```
awulkiew, line 71: `const`

test/algorithms/closest_points/empty_geometry.hpp
```diff
@@ -0,0 +1,267 @@
...
+    try
+    {
+        bg::closest_points(geometry1, geometry2, out_seg);
+    }
+    catch(bg::empty_input_exception const& )
+    {
+        return;
+    }
+    BOOST_CHECK_MESSAGE(false,
+                        "A empty_input_exception should have been thrown");
```
awulkiew, line 31: Shouldn't exceptions be checked with `BOOST_CHECK_EXCEPTION`?

awulkiew:
@vissarion Thanks for the PR! I apologize for the late review.

The vast majority of issues I mentioned are minor issues like formatting. The most important I think is the coordinate type of the result of pt-seg strategy which AFAIU was also merged with cartesian strategies before? Right now the coordinate type is the result type of distance strategy. I'm not sure if that's correct.

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
barendgehrels: The official boost line length is 80 characters. I don't remember that we changed it.

I would actually be in favor of making it larger - but then maybe even more than 100 (for example 120). Years ago Bruno objected to make it larger than 80 characters, because he sometimes edits sources in a terminal. That might still be valid for some users. Also for showing a side-by-side diff the max line length is important.

But most probably for this one, we're just still restricted to Boost Guideline which is 80.
https://www.boost.org/development/requirements.html

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
barendgehrels: you can use `auto const` (and it will be a pair)

include/boost/geometry/strategies/closest_points/spherical.hpp
```diff
@@ -0,0 +1,112 @@
...
+#define BOOST_GEOMETRY_STRATEGIES_CLOSEST_POINTS_SPHERICAL_HPP
+
+
+//#include <boost/geometry/strategies/cartesian/azimuth.hpp>
+
+//#include <boost/geometry/strategies/cartesian/distance_projected_point.hpp>
+//#include <boost/geometry/strategies/cartesian/distance_pythagoras.hpp>
+//#include <boost/geometry/strategies/spherical/distance_pythagoras_box_box.hpp>
+//#include <boost/geometry/strategies/spherical/distance_pythagoras_point_box.hpp>
+//#include <boost/geometry/strategies/spherical/distance_segment_box.hpp>
```
barendgehrels, line 20: Can we remove these commented includes?

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
awulkiew: @barendgehrels from https://github.com/boostorg/geometry/wiki/Guidelines-for-Developers

> The preferred line length is 80 characters, with maximum length of 100.

It's like that since guidelines were added to the [Wiki in 2014](https://github.com/boostorg/geometry/wiki/Guidelines-for-Developers/9d366b7a52005b3b995e79501c07adeb5d0d75a5). And I'm pretty sure that we talked about this then.

I'm ok with increasing it to 120.

Isn't line length mainly defined by the size of the screen? Nowadays we have bigger screens or multiple screens. This also affects the size of the terminal as well as widths of diff columns.

awulkiew: It's in the original version from the docs, since 2013:
https://github.com/boostorg/geometry/commit/10dd010ecc189276de1584187de847c4b4239e3c#diff-f7aee8e109b4ecd10dec03a3010cc6c110436140b709fa3632a5053c57d1bb84

doc/src/examples/examples_utils/create_svg_closest_points.hpp
vissarion: sorry, I don't get it

doc/doxy/Doxyfile
vissarion: This is automatic, this file is generated after upgrading to Doxyfile version 1.8.17.
The additions are almost all comments (probably manually removed from previous versions ?) but there are some extra options added and some old ones removed.

doc/index/Doxyfile
vissarion: This is again automatic, this file is generated after upgrading to Doxyfile version 1.8.17.
The additions are almost all comments (probably manually removed from previous versions ?) but there are some extra options added and some old ones removed. 
For example the [`PERL_PATH`](https://github.com/boostorg/geometry/blob/develop/doc/index/Doxyfile#L295) which seems to be deprecated; see https://github.com/doxygen/doxygen/commit/6d1535c38fe6bdaa2a00fff0e7e43774a740a4ce

doc/readme.txt
vissarion: OK I added as a separate point, 2) and update the enumeration.

include/boost/geometry/algorithms/detail/closest_points/linear_or_areal_to_areal.hpp
vissarion: > `geometry::coordinate_type` is not a type alias so here it should've been `typename coordinate_type<Linear>::type`.

sure, thanks for noticing 

> Side note. We could think about adding type aliases and variable templates for traits: `coordinate_t`, `point_t`, `dimension_v`, etc.

+1

include/boost/geometry/algorithms/detail/closest_points/linear_or_areal_to_areal.hpp
vissarion: sure, but some cases are not supported, e.g. linestring - multipolygon, I updated my comment to explain that.

include/boost/geometry/algorithms/detail/closest_points/linear_or_areal_to_areal.hpp
vissarion: sure, but I do not want to assert on empty `mp_out`

include/boost/geometry/algorithms/detail/closest_points/linear_to_linear.hpp
vissarion: Empty input is checked by the interface here: https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/closest_points/interface.hpp#L178
Several empty input combinations are also tested in unit tests.

include/boost/geometry/algorithms/detail/closest_points/linear_to_linear.hpp
vissarion: Empty input is checked by the interface here: https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/closest_points/interface.hpp#L178

include/boost/geometry/strategies/closest_points/geographic.hpp
vissarion: The definition of base class in detail namespace is not needed, probably leftover from some previous implementation. Thanks for noticing.

include/boost/geometry/strategies/closest_points/spherical.hpp
vissarion: The definition of base class in detail namespace is not needed.

include/boost/geometry/strategies/geographic/closest_points_pt_seg.hpp
vissarion: No, this should be just like in spherical. Changed.

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
vissarion: Exactly, in geographic (unlike in spherical) computing distance and closest points is the same computation so `EnableClosestPoints` is only to avoid some (few) extra assignments. I do not mind removing it if you prefer.

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
vissarion: > According to our guidelines lines can have up to 100 characters. I don't know if this would fit into 100 characters but in general putting parts of the expression in multiple lines makes it harder to read.

It doesn't fit. 
Is 
```
CT const dist_l = 
    inverse_distance_type::apply(lon3, lat3, pl_lon, pl_lat, spheroid).distance;
CT const dist_r = 
    inverse_distance_type::apply(lon3, lat3, pr_lon, pr_lat, spheroid).distance;
```
preferable?

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
vissarion: changed to 
```
using inverse_distance_azimuth_quantities_type = 
            typename FormulaPolicy::template inverse<CT, true, true, false, true, true>;
    
using inverse_dist_azimuth_type = 
            typename FormulaPolicy::template inverse<CT, false, true, false, false, false>;

using direct_distance_type = 
            typename FormulaPolicy::template direct<CT, true, false, false, false>;
```

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
vissarion: what about this ?
```
s14 -= inverse_distance_type::apply(
    res14.lon2, res14.lat2, pl_lon, pl_lat, spheroid).distance/2;
```

vissarion: for the rest see comments above

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
vissarion: what about 
```
using inverse_dist_azimuth_type = 
            typename FormulaPolicy::template inverse<CT, true, true, false, false, false>;

using inverse_dist_azimuth_reverse_type = 
            typename FormulaPolicy::template inverse<CT, true, true, true, false, false>;
```
to be consistent with above?

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
vissarion: could be from the previous IDE, cutting the minimal part that exceeds 80 lines.
I put it in one line now

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
vissarion: one line now

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
vissarion: So the question is align below `(` or change line and use tab, right?

test/algorithms/closest_points/empty_geometry.hpp
vissarion: probably yes, but in the whole library they are checked like this so I just follow that.

include/boost/geometry/strategies/geographic/closest_points_pt_seg.hpp
vissarion: Rethinking of it, we probably should always promote to floating point since even in the case of input integral coordinates the closest point on the segment is possibly not be able to be represented by integral coordinates. This should happen in all CS.

vissarion:
Thanks for the reviews! I addressed all the comments. The only issue remaining is the promotion to fp in the case of pt-seg strategy (discussion above).

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
barendgehrels: > @barendgehrels from https://github.com/boostorg/geometry/wiki/Guidelines-for-Developers
> 
> > The preferred line length is 80 characters, with maximum length of 100.
> 
> It's like that since guidelines were added to the [Wiki in 2014](https://github.com/boostorg/geometry/wiki/Guidelines-for-Developers/9d366b7a52005b3b995e79501c07adeb5d0d75a5). And I'm pretty sure that we talked about this then.

:+1: thanks for the reminder!

So I'm already setting the right margin in the IDE to 80. Anyway: "preferred length is 80" but you may exceed it is actually not such a clear guideline... But anyway, I think we all do.

> 
> I'm ok with increasing it to 120.
> 
> Isn't line length mainly defined by the size of the screen? Nowadays we have bigger screens or multiple screens. This also affects the size of the terminal as well as widths of diff columns.

Sorry, I've to retract my own statement about 120, it's probably too much. Yes, the length is of course determined by the screen width - but also by the used font. I just did a `meld` (side by side view) in some adapted code and for me 107 is the personal limit. And of course for others the limit might be higher, but also smaller, all depending on personal settings.

So maybe better to just keep it as it is written there.

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
barendgehrels: We do both in practice. In the past I often used a tab (= spaces) but many people and IDE's now align at the `(`

doc/doxy/Doxyfile
awulkiew: Do we really need these additional comments? The file is long enought without them.

doc/src/examples/examples_utils/create_svg_closest_points.hpp
awulkiew: Probably that you can remove the `#ifdef`. Note also that since below `boost::geometry::svg_mapper` is created technically this `#include` is needed anyway. Unless you put the whole code behind the `#ifdef`. But AFAIU is some helper utility so there is no need for that.

include/boost/geometry/algorithms/detail/closest_points/linear_or_areal_to_areal.hpp
awulkiew: `mp_out` is checked for emptiness above so the assertion will not fail. Or do you mean that you don't want to perform an additional unnecessary check? Anyway I'm ok with leaving it as it is.

awulkiew (approved):

include/boost/geometry/strategies/geographic/closest_points_pt_seg.hpp
awulkiew: Yes, it makes sense. I don't know if this is correct since a user will typically pass the same point type as the in the input and output. Still internally an FP type can be used for calculation purposes even if at the end it will be converted to user-defined type.

If you take the coordinate type from the distance result in practice it is going to be FP in all coordinate systems.

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
awulkiew: Whatever works for you, just be consistent. I prefer the second one with all arguments at the same place but I'm ok with a different one if you prefer it.

awulkiew:
@vissarion Since this is built on top of https://github.com/boostorg/geometry/pull/933 do I have to review the other PR as well?

vissarion:
> @vissarion Since this is built on top of #933 do I have to review the other PR as well?

Now, you have already reviewed here.

doc/doxy/Doxyfile
vissarion: We do not need them but could be useful as a documentation. Also to my understanding one has to remove extra comments manually after every Doxyfile upgrade and that could be error prone. But I am OK removing those if you insist because I guess a similar decision had already been taken in the past.

include/boost/geometry/algorithms/detail/closest_points/linear_or_areal_to_areal.hpp
vissarion: > Or do you mean that you don't want to perform an additional unnecessary check?

yes

include/boost/geometry/strategies/geographic/closest_points_pt_seg.hpp
vissarion: OK, I left it as is and added a unit test with integral coordinates.

doc/doxy/Doxyfile
awulkiew: > remove extra comments manually after every Doxyfile upgrade

I thought you made this on purpose. AFAIU `doxygen` can generate a warning if something is wrong with the input file but will not modify it unless you pass `-u` option. This option is not passed in `make_qbk.py` so you must have passed it by yourself.

For me doxygen 1.9.1 reports:
```
warning: Tag 'TCL_SUBST' at line 39 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'COLS_IN_ALPHA_INDEX' at line 172 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'PERL_PATH' at line 295 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
warning: Tag 'MSCGEN_PATH' at line 300 of file 'Doxyfile' has become obsolete.
         To avoid this warning please remove this line from your configuration file or upgrade it using "doxygen -u"
```

So I'd simply remove these variables from the file if I wanted to silence this warning or leave the doxyfile unmodified, as it is now. Because this is not related to `closest_points` in any way.

addy90:
I just noticed that in the version info on the website for Boost 1.79.0 the info about this PR and the new closest_points algorithm is missing: https://www.boost.org/users/history/version_1_79_0.html
As I follow this feature since more than a year already and also helped in finding a small bug [here](https://github.com/boostorg/geometry/pull/707#issuecomment-786650747) I would allow myself to suggest that this huge and incredible work from @vissarion should indeed be listed in the version docs on the website for the changes in boost::geometry.
This is just my opinion and I hope it is still possible to add the kudos. Also thanks again from my side for finalizing this feature in Boost 1.79.0!

awulkiew:
@addy90 Thanks for bringing it up. Of course it should be listed. I must have missed it.

It is possible to add it on the website. It is however not possible to add it in the documentation of geometry ver 1.79 (https://www.boost.org/doc/libs/1_79_0/libs/geometry/doc/html/geometry/release_notes.html). But it will be there in the future versions.

awulkiew:
@addy90 @vissarion 
https://github.com/boostorg/website/pull/681
https://github.com/boostorg/geometry/commit/e2496e2fedbc0c4468427cfb973a94afee896ca7

PR #949, opened 2021-12-08
fix: use policy instead of define

barendgehrels:
This PR removes the define `BOOST_GEOMETRY_USE_RESCALING` from `get_turn_info.hpp` and replaces it with a verification policy, which is defined for areal/areal (if rescaling is removed) but not for areal/linear and linear/linear because they did not do these verifications before.

After this, all tests from set_operations pass when rescaling is turned off.


test/algorithms/set_operations/intersection/intersection_linear_linear.cpp
```diff
@@ -1272,13 +1272,9 @@ BOOST_AUTO_TEST_CASE( test_intersection_ml_ml_degenerate )
...
...8),(4 3.8,4 3.8),(4 3.6,4 3.6),(4 3.5,4 3.5),(4 3.03179,4 3.03179),(4 3,4 3),(4 2.9435,4 2.9435),(4 2.82162,4 2.82162),(4 2.47965,4 2.47965),(4 1.72377,4 1.72377),(4 1.38014,4 1.38014),(3.43534 0.441146,3.43534 0.441146),(2.06005 1.27206,2.06005 1.27206),(1.88383 1.37852,1.88383 1.37852),(1.8713 1.38609,1.8713 1.38609),(1.01183 1.90535,1.01183 1.90535),(0.72677 2.07758,0.72677 2.07758),(0.619143 2.1426,0.619143 2.1426),(-0.532243 2.83823,-0.532243 2.83823),(-0.789427 2.99361,-0.789427 2.99361),(-0.756651 3.30964,-0.756651 3.30964),(-0.706503 3.66784,-0.706503 3.66784),(-0.705343 3.67612,-0.705343 3.67612),(0.502294 7.64221,0.502294 7.64221),(0.601362 7.58336,0.601362 7.58336),(0.641699 7.5594,0.641699 7.5594),(1.43457 7.08839,1.43457 7.08839),(3.26673 6,3.26673 6),(5.09496 4.91395,5.09496 4.91395),(5.26567 4.81254,5.26567 4.81254),(6.47672 4.09311,6.47672 4.09311),(8.18862 3.07616,8.18862 3.07616),(8.49103 2.89652,8.49103 2.89652),(8.5655 2.85228,8.5655 2.85228),(9.98265 0.00543606,9.98265 0.00543606),(9.09826 -100.516,9.09826 -100.516),(7.08745 -329.067,7.08745 -329.067),(5.06428 -559.024,5.06428 -559.024),(3.23365 -767.097,3.23365 -767.097),(3.16036 -775.427,3.16036 -775.427))"),
```
vissarion, line 1279: why is this case removed?

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -66,6 +62,36 @@ class turn_info_exception : public geometry::exception
 namespace detail { namespace overlay
 {
 
+
+struct policy_verify_nothing
+{
+    static bool const use_side_verification = false;
+    static bool const use_start_turn = false;
```
vissarion, line 69: should `use_start_turn`, `use_handle_as_touch`, `use_handle_as_equal` and `use_handle_imperfect_touch` be merged to one?

vissarion (approved):
Thanks! I am OK with merging modulo some minor remarks.

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
barendgehrels: Good question. I was not sure yet. Yes, we could merge them all - but basically linestrings could profit from these too (to be investigated), and these settings make it easy to gradually change them

barendgehrels: I'll keep it for now, it's internal, we can revisit it later.

barendgehrels:
Merging it, @awulkiew if you have concerns they can be addressed later

barendgehrels:
I get mails from `circleci` regarding a failure in `cs_undefined`
Now checking.

awulkiew:
@barendgehrels Is this: https://github.com/boostorg/geometry/commit/ca42385d3b5a38b4b459af5b22d20013c72a89de the fix for circleci and cs_undefined? If that's the case then the fix is incorrect.

The CSTag should be taken from the strategy rather than from the geometries inside `get_distance_measure` here:
https://github.com/boostorg/geometry/blob/ca42385d3b5a38b4b459af5b22d20013c72a89de/include/boost/geometry/algorithms/detail/overlay/get_distance_measure.hpp#L111

In general this utility is CS specific and technically should be made into strategy. But if it's convenient for you passing `cs_tag` from umbrella strategy shoudl be enough.

PR #952, opened 2021-12-15
Use side_by_triangle by default

barendgehrels:
This PR (edited):

- removes the define such that the new `side_triangle` is now always the default strategy ( (to be researched further why the robust fails for some obvious cases, as discussed with @vissarion ).
- instead, for the convex_hull, the umbrella_strategy defines still `side_robust`

After this PR:
- convex hull passes again, when `NO_ROBUSTNESS` is defined
- get_turns_linear_linear passes again, when `NO_ROBUSTNESS` is defined
- there is less dependency on the define `BOOST_GEOMETRY_USE_RESCALING`

### DIFF
--- a/include/boost/geometry/algorithms/sym_difference.hpp
+++ b/include/boost/geometry/algorithms/sym_difference.hpp
@@ -567,7 +567,7 @@ struct sym_difference<default_strategy, false>
                              Collection & output_collection,
                              default_strategy)
     {
-        typedef typename strategy::relate::services::default_strategy
+        typedef typename strategies::relate::services::default_strategy
             <
                 Geometry1, Geometry2
             >::type strategy_type;
--- a/include/boost/geometry/strategies/convex_hull/cartesian.hpp
+++ b/include/boost/geometry/strategies/convex_hull/cartesian.hpp
@@ -12,6 +12,7 @@
 #define BOOST_GEOMETRY_STRATEGIES_CONVEX_HULL_CARTESIAN_HPP
 
 #include <boost/geometry/strategies/side.hpp>
+#include <boost/geometry/strategy/cartesian/side_robust.hpp>
 
 #include <boost/geometry/strategies/cartesian/point_in_point.hpp>
 #include <boost/geometry/strategies/convex_hull/services.hpp>
@@ -43,8 +44,7 @@ public:
     static auto side()
     {
         using side_strategy_type
-            = typename strategy::side::services::default_strategy
-                <cartesian_tag, CalculationType>::type;
+            = strategy::side::side_robust<CalculationType>;
         return side_strategy_type();
     }
 };
--- a/include/boost/geometry/strategy/cartesian/side_by_triangle.hpp
+++ b/include/boost/geometry/strategy/cartesian/side_by_triangle.hpp
@@ -252,7 +252,6 @@ private:
     }
 };
 
-#if ! defined(BOOST_GEOMETRY_USE_RESCALING)
 #ifndef DOXYGEN_NO_STRATEGY_SPECIALIZATIONS
 
 namespace services
@@ -266,7 +265,6 @@ struct default_strategy<cartesian_tag, CalculationType>
 
 }
 
-#endif
 #endif
 
 }} // namespace strategy::side
--- a/include/boost/geometry/strategy/cartesian/side_robust.hpp
+++ b/include/boost/geometry/strategy/cartesian/side_robust.hpp
@@ -178,23 +178,6 @@ public:
 
 };
 
-#ifdef BOOST_GEOMETRY_USE_RESCALING
-#ifndef DOXYGEN_NO_STRATEGY_SPECIALIZATIONS
-
-namespace services
-{
-
-template <typename CalculationType>
-struct default_strategy<cartesian_tag, CalculationType>
-{
-    typedef side_robust<CalculationType> type;
-};
-
-}
-
-#endif
-#endif
-
 }} // namespace strategy::side
 
 }} // namespace boost::geometry
--- a/test/algorithms/convex_hull/Jamfile
+++ b/test/algorithms/convex_hull/Jamfile
@@ -14,4 +14,8 @@ test-suite boost-geometry-algorithms-convex_hull
     [ run convex_hull_multi.cpp        : : : : algorithms_convex_hull_multi ]
     [ run convex_hull_robust.cpp       : : : : algorithms_convex_hull_robust ]
     [ run convex_hull_sph_geo.cpp      : : : : algorithms_convex_hull_sph_geo ]
+    [ run convex_hull.cpp              : : : <define>BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE : algorithms_convex_hull_alternative ]
+    [ run convex_hull_multi.cpp        : : : <define>BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE : algorithms_convex_hull_multi_alternative ]
+    [ run convex_hull_robust.cpp       : : : <define>BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE : algorithms_convex_hull_robust_alternative ]
+    [ run convex_hull_sph_geo.cpp      : : : <define>BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE : algorithms_convex_hull_sph_geo_alternative ]
     ;
--- a/test/algorithms/overlay/Jamfile
+++ b/test/algorithms/overlay/Jamfile
@@ -26,6 +26,7 @@ test-suite boost-geometry-algorithms-overlay
     [ run get_turns_linear_areal.cpp       : : : : algorithms_get_turns_linear_areal ]
     [ run get_turns_linear_areal_sph.cpp   : : : : algorithms_get_turns_linear_areal_sph ]
     [ run get_turns_linear_linear.cpp      : : : : algorithms_get_turns_linear_linear ]
+    [ run get_turns_linear_linear.cpp      : : : <define>BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE : algorithms_get_turns_linear_linear_alternative ]
     [ run get_turns_linear_linear_geo.cpp  : : : : algorithms_get_turns_linear_linear_geo ]
     [ run get_turns_linear_linear_sph.cpp  : : : : algorithms_get_turns_linear_linear_sph ]
     [ run overlay.cpp                      : : : : algorithms_overlay ]
--- a/test/algorithms/overlay/get_turns_linear_linear.cpp
+++ b/test/algorithms/overlay/get_turns_linear_linear.cpp
@@ -282,12 +282,14 @@ void test_all()
                               "LINESTRING(2 8,4 0.4,8 1,0 5)",
                               expected("iuu++")("mui=+")("tiu+="));
 
+#ifdef BOOST_GEOMETRY_TEST_FAILURES
         test_geometry<ls, ls>("LINESTRING(-2305843009213693956 4611686018427387906, -33 -92, 78 83)",
                               "LINESTRING(31 -97, -46 57, -20 -4)",
                               expected("iuu++"));
         test_geometry<ls, ls>("LINESTRING(31 -97, -46 57, -20 -4)",
                               "LINESTRING(-2305843009213693956 4611686018427387906, -33 -92, 78 83)",
                               expected("iuu++"));
+#endif
     }
 
     // In 1.57 the results of those combinations was different for MinGW
--- a/test/algorithms/overlay/relative_order.cpp
+++ b/test/algorithms/overlay/relative_order.cpp
@@ -28,9 +28,7 @@
 #endif
 
 #include <boost/geometry/strategy/cartesian/side_robust.hpp>
-#include <boost/geometry/strategies/cartesian/side_by_triangle.hpp>
-#include <boost/geometry/strategies/side.hpp>
-
+#include <boost/geometry/strategy/cartesian/side_by_triangle.hpp>
 #include <boost/geometry/strategies/side.hpp>
 
 template <typename P, typename T>
--- a/test/algorithms/set_operations/alternative_robustness_strategy.hpp
+++ b/test/algorithms/set_operations/alternative_robustness_strategy.hpp
@@ -1,0 +1,27 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+
+// Copyright (c) 2021 Barend Gehrels, Amsterdam, the Netherlands.
+
+// Licensed under the Boost Software License version 1.0.
+// http://www.boost.org/users/license.html
+
+
+#ifndef BOOST_GEOMETRY_TEST_INTERSECTION_ALTERNATIVE_ROBUSTNESS_STRATEGY_HPP
+#define BOOST_GEOMETRY_TEST_INTERSECTION_ALTERNATIVE_ROBUSTNESS_STRATEGY_HPP
+
+#include <boost/geometry/strategies/relate/cartesian.hpp>
+
+
+template <typename SideStrategy>
+class alternative_robustness_strategy
+    : public bg::strategies::relate::cartesian<>
+{
+public:
+    static auto side()
+    {
+        return SideStrategy();
+    }
+};
+
+
+#endif // BOOST_GEOMETRY_TEST_INTERSECTION_ALTERNATIVE_ROBUSTNESS_STRATEGY_HPP
--- a/test/algorithms/set_operations/intersection/intersection_linear_linear.cpp
+++ b/test/algorithms/set_operations/intersection/intersection_linear_linear.cpp
@@ -477,6 +477,28 @@ BOOST_AUTO_TEST_CASE( test_intersection_linestring_linestring )
          from_wkt<ML>("MULTILINESTRING((1 0,1 1))"),
          "lli25"
          );
+
+#if defined(BOOST_GEOMETRY_TEST_FAILURES)
+    {
+        // This test fails if side_by_triangle is used in intersection
+        ut_settings settings;
+        settings.tolerance = 1.0e-9;
+        settings.test_alternative_side_strategy = true;
+        settings.test_default_strategy = BG_IF_TEST_FAILURES;
+        settings.test_explicit_strategy = BG_IF_TEST_FAILURES;
+        settings.test_invariance = BG_IF_TEST_FAILURES;
+
+        tester::apply
+                (from_wkt<L>("LINESTRING(-2305843009213693956 4611686018427387906, -33 -92, 78 83)"),
+                 from_wkt<L>("LINESTRING(31 -97, -46 57, -20 -4)"),
+                 from_wkt<ML>("MULTILINESTRING((1.39042821159 -37.7808564232,1.39042821159 -37.7808564232))"),
+                 "lli26",
+                 settings
+                 );
+    }
+#endif
+
+
 }
 
 
--- a/test/algorithms/set_operations/intersection/test_intersection_linear_linear.hpp
+++ b/test/algorithms/set_operations/intersection/test_intersection_linear_linear.hpp
@@ -17,15 +17,40 @@
 
 #include <boost/geometry/geometry.hpp>
 #include "../test_set_ops_linear_linear.hpp"
+#include "../alternative_robustness_strategy.hpp"
 #include <from_wkt.hpp>
 #include <to_svg.hpp>
 
+
+//! Contains (optional) settings such as tolerance
+//! and to skip some test configurations
+struct ut_settings
+{
+    ut_settings() = default;
+
+    // Make it backwards compatible by a non-explicit constructor
+    // such that just tolerance is also accepted
+    ut_settings(const double t) : tolerance(t) {}
+
+    double tolerance{std::numeric_limits<double>::epsilon()};
+
+    // By default everything is tested.
+    // Some rare cases might fail for some strategies, in that case
+    // some strategies can be turned off
+    bool test_invariance{true};
+    bool test_reverse{true};
+
+    bool test_default_strategy{true};
+    bool test_explicit_strategy{true};
+    bool test_alternative_side_strategy{true};
+};
 
 //==================================================================
 //==================================================================
 // intersection of (linear) geometries
 //==================================================================
 //==================================================================
+
 
 template <typename Geometry1, typename Geometry2, typename MultiLineString>
 inline void check_result(Geometry1 const& geometry1,
@@ -42,7 +67,7 @@ inline void check_result(Geometry1 const& geometry1,
                          << " Expected: len=" << bg::length(mls_int1) << " count=" << bg::num_points(mls_int1)
                          << " or: len=" << bg::length(mls_int2) << " count=" << bg::num_points(mls_int2)
                          << " Detected: len=" << bg::length(mls_output) << " count=" << bg::num_points(mls_output)
-                         << " wkt=" << bg::wkt(mls_output)
+                         << " wkt=" << std::setprecision(12) << bg::wkt(mls_output)
                          );
 }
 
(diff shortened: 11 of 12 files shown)


include/boost/geometry/strategies/relate/cartesian.hpp
```diff
@@ -39,7 +39,7 @@ namespace boost { namespace geometry
 namespace strategies { namespace relate
 {
 
-template <typename CalculationType = void>
+template <typename CalculationType = void, bool IsSetOperation = false>
```
awulkiew, line 42: I'm against this change. If you think that set operations need different umbrella strategy then you should create another one instead of adding template argument that doesn't exist in any other umbrella strategy.

One of the reasons for umbrella strategies to exist is to have intuitive and consistent interface. This unfortunately also means: not adding arguments because it's convenient.

awulkiew: Btw, what with the general one umbrella strategy `strategies::cartesian<>` that can be passed into all algorithms?
Which side strategy should be used there?
Are relational and set operations different so they need different side strategies?
Does it mean that the results of relational and set operations are not consistent?

include/boost/geometry/strategies/spherical/intersection.hpp
```diff
@@ -145,11 +145,12 @@ struct ecef_segments
...
         typename UniqueSubRange1,
         typename UniqueSubRange2,
-        typename Policy
+        typename Policy,
+        typename SideStrategy
     >
     static inline typename Policy::return_type
         apply(UniqueSubRange1 const& range_p, UniqueSubRange2 const& range_q,
-              Policy const&)
+              Policy const&, SideStrategy const&)
```
awulkiew, line 153: So side strategy is not used in all intersection strategies. Only in cartesian one? So side strategy is passed here it is ignored? This is inconsistent. But I don't understand why is this needed at all (see other comment).

include/boost/geometry/strategies/cartesian/intersection.hpp
```diff
@@ -295,15 +295,17 @@ struct cartesian_segments
...
-        typename Policy
+        typename Policy,
+        typename SideStrategy
     >
     static inline typename Policy::return_type
         apply(UniqueSubRange1 const& range_p,
               UniqueSubRange2 const& range_q,
-              Policy const& policy)
+              Policy const& policy,
+              SideStrategy const& side)
```
awulkiew, line 305: Is the intention to allow to pass various side strategies in the same algorithm? Or is this changed to allow defining side strategy in one place only, i.e. in umbrella strategy and pass it into all strategies using it? If it's the former then this can lead to inconsistencies. If it's the latter then I don't understand why isn't it done for other coordinate systems too?

If the purpose is to "play" with the cartesian strategy then why not passing the side strategy as class template argument? This way the responsibility for the consistency of the strategies will rest only on the umbrella strategy.

include/boost/geometry/strategies/relate/services.hpp
```diff
@@ -52,6 +52,30 @@ struct strategy_converter
...
+{
+
+template
+<
+    typename Geometry1,
+    typename Geometry2,
+    typename CSTag1 = typename geometry::cs_tag<Geometry1>::type,
+    typename CSTag2 = typename geometry::cs_tag<Geometry2>::type
+>
+struct default_strategy
```
awulkiew, line 67: I haven't chacked this out but I think it is not required to define this struct and its specializations. AFAIU the only thing you need is template alias for `strategies::relate::services::default_strategy`:
```
template
<
    typename Geometry1,
    typename Geometry2,
    typename CSTag1 = typename geometry::cs_tag<Geometry1>::type,
    typename CSTag2 = typename geometry::cs_tag<Geometry2>::type
>
using default_strategy = strategies::relate::default_strategy<Geometry1, Geometry2, CSTag1, CSTag2>;
```

or rather four aliases, one for each set operation (see other comment).

include/boost/geometry/strategies/relate/spherical.hpp
```diff
@@ -369,6 +369,29 @@ struct strategy_converter<strategy::side::spherical_side_formula<CalculationType
 
 }} // namespace strategies::relate
 
+namespace strategies { namespace set_operations { namespace services {
+
+template <typename Geometry1, typename Geometry2>
+struct default_strategy<Geometry1, Geometry2, spherical_tag, spherical_tag>
```
awulkiew, line 375: AFAIU these are not needed (see above).

include/boost/geometry/strategies/relate/geographic.hpp
```diff
@@ -348,6 +348,16 @@ struct strategy_converter<strategy::covered_by::geographic_point_box_by_side<Cal
 
 }} // namespace strategies::relate
 
+namespace strategies { namespace set_operations { namespace services {
+
+template <typename Geometry1, typename Geometry2>
+struct default_strategy<Geometry1, Geometry2, geographic_tag, geographic_tag>
```
awulkiew, line 354: AFAIU this is not needed (see below).

barendgehrels:
Answering (in general) to the questions of @awulkiew
- no, I don’t think that set operations need a different umbrella strategy.  However, it’s better than the define. The point is: the `side_robust` side strategy isn’t as good as I expected. As soon as rescaling is turned off, it breaks even the most simple cases. I don’t know yet how exactly.@vissarion  is investigating that. For that I first added that define (making `side_by_triangle` the default if rescaling is turned off). But that breaks other cases (if rescaling is turned off), for example convex hull. Therefore I changed this, to make the change more limited. And I think it’s good, but it’s temporary.
- yes, I think passing the side strategy to the intersection strategy is a good change. Because before, it just got it from the default_strategy service. But it should be passed from the umbrella strategy, because that one specifies it. And yes, we pass all strategies via methods because they might have parameters (such as earth radius etc). So I think this change is exactly as it should be.
- and yes, the geographic intersection and spherical don’t use that argument, but it should be passed for consistency, of course. Otherwise things don’t compile :-)

awulkiew:
> no, I don’t think that set operations need a different umbrella strategy. However, it’s better than the define.

Ok, but this define was added by you before right? If `side_robust` causes problems then I propose to simply go back to `side_by_triangle` in both set and relational operations. Of course this can be investigated. And only after we know the results of this investigation we can change the strategy.

If some cases work with one side strategy and other cases with other one are you sure that they can be simply used interchangeably inside this particular intersection strategy? My guess would rather be that `side_robust` needs corresponding `intersection_robust`.

> Because before, it just got it from the default_strategy service.

Yes, but this was added by you before wasn't it? Originally side strategy was defined inside intersection strategy.

> geographic intersection and spherical don’t use that argument, but it should be passed for consistency

If this is only used by cartesian strategy it could also be taken as template argument. Or there could be a separate intersection strategy. I don't understand why do we assume that it's fine to pass arbitrary side strategies to intersection strategies.

vissarion:
I agree to switch back to `side_by_triangle` in both set and relational operations and investigate more. 

What I currently understand as the issue of `side_robust` is the following: if the input is a set of rational numbers and we compute the result with arbitrary precision then this result will defer from the one we get if you transform the input in floating point numbers and the compute the side with `side_robust`. Note that the error comes from the transformation to floating points not by the computation of side.

barendgehrels:
OK, I'll create another PR to go back to `side_by_triangle` - because that will have the same effect.
Thanks for your input.

Still, some of these changes I want to keep, for example passing side-strategy to intersection. That's really needed and I don't understand why it's not done before. But not problem, we can do it now. Even if it's not really used (by spherical/geographical), it should be passed because it's used by cartesian, and all the strategies should look the same (not from esthetic viewpoint, but otherwise it doesn't compile :-) )

barendgehrels:
> > no, I don’t think that set operations need a different umbrella strategy. However, it’s better than the define.
> 
> Ok, but this define was added by you before right? If `side_robust` causes problems then I propose to simply go back to `side_by_triangle` in both set and relational operations. Of course this can be investigated. And only after we know the results of this investigation we can change the strategy.
> 
> If some cases work with one side strategy and other cases with other one are you sure that they can be simply used interchangeably inside this particular intersection strategy? My guess would rather be that `side_robust` needs corresponding `intersection_robust`.
> 
> > Because before, it just got it from the default_strategy service.
> 
> Yes, but this was added by you before wasn't it? Originally side strategy was defined inside intersection strategy.
> 
> > geographic intersection and spherical don’t use that argument, but it should be passed for consistency
> 
> If this is only used by cartesian strategy it could also be taken as template argument. Or there could be a separate intersection strategy. I don't understand why do we assume that it's fine to pass arbitrary side strategies to intersection strategies.

I don't understand why it's not fine... Of course it should take the side-strategy from the umbrella-strategy, that's crystal-clear to me...
Ever, indeed, the Cartesian used side-by-triangle (hard-coded), I don't remember when it was changed and by who. It's probably not so important.
I don't know the geographic intersection so well - but I think, and I just looked, that it should use the passed strategy too, instead of this for example: `azimuth_side_value`. But that can be fixed later if necessary.

barendgehrels:
Or, actually, most changes will be the same so I reopen this and change the name.

Changes still to be pushed.

awulkiew:
> Still, some of these changes I want to keep, for example passing side-strategy to intersection.

Changes should be made to solve a problem, not because they can be made. What problem does this one solve? Because I think it only introduces complexity and doesn't solve any problem.

> That's really needed and I don't understand why it's not done before.

Why is it needed? Side strategy was a member of intersection strategy only because someone made this implementation choice. It might not have been used. A formula might have been used instead. There might have been no need for it at all. Is the use of side strategy a feature of IntersectionStrategy concept? Is it mathematically guaranteed that intersection of two segments has to use point-segment side computation? @vissarion you looked into that. I don't think so (unless I'm wrong) and therefore side strategy should not be a part of the interface of IntersectionStrategy. Even a cartesian intersection strategy could probably have been implemented differently, without any side strategy used internally. A good candidate would probably be one using precise arithmetic. Who said that a side strategy would be needed for that?

If you need different intersection strategy you could create it. Passing side strategy to all existing strategies is not required for that.

If side strategy really has to be passed into one implementation of intersection strategy it should be passed a template argument as it is already done in the library e.g. for distance strategies.

Furthermore practice shows that there are places in the library where various strategies are not correctly taken from umbrella strategy but created by hand (probably becasue that's convenient at the moment). What you're proposing would require to pass 2 strategies instead of one, a good place to take shortcuts for convenience.

> it should be passed because it's used by cartesian, and all the strategies should look the same (not from esthetic viewpoint, but otherwise it doesn't compile

It doesn't compile only because of the way how you designed it. It's obvious that if you pass a dummy parameter somewhere then a function has to take it. Without the parameter it would compile just fine.

If something is needed by one specific implementation of intersection strategy (this is not even a matter of CS but one specific implementation) then it should be a part of this implementation, not the whole Concept. It should not be propagated into all other implementations.

> Of course it should take the side-strategy from the umbrella-strategy, that's crystal-clear to me...

Not really. The strategies defined in umbrella strategy has to be consistent with each other, that's it. There is no other requirement. They don't have to take other strategies. If we at some point choose to use some strategy in another specific implementation of another strategy then it is not required to change the interfaces of all existing strategies, etc.

> Ever, indeed, the Cartesian used side-by-triangle (hard-coded), I don't remember when it was changed and by who. It's probably not so important.

Probably recently by @vissarion .

This is not a matter of cartesian vs other CSes. It's a matter of this specific implementation of intersection strategy. It could be implemented differently without side strategy at all.

> I don't know the geographic intersection so well - but I think, and I just looked, that it should use the passed strategy too, instead of this for example: azimuth_side_value. But that can be fixed later if necessary.

I don't remember but let's assume that sides are calculated there. The answer is still no. It's because sides can be calculated differently and they are not interchangeable. E.g. a spherical intersection strategy using ECEF 3d vectors should internally calculate sides using the same method. Another intersection strategy could use spherical formulas and sides should've been calculated the same way (and probably reuse some of the variables). In geographic you could model segments with great ellipses, loxodromes or geodesics. They are not interchangeable. The correctness of the intersection strategy should not depend on the caller's ability to pass the correct side strategy. This way we can be sure that the intersection strategy gives correct results without thinking about sides.

There is a matter of consistency of intersection strategy and side strategy in algorithms that can use both of them. But this is a different story and umbrella strategy is for this. Not only side and intersection has to be consistent with each other but all strategies used in an algorithm.

We could of course rethink and redesign all of the strategies. Because various strategies use other strategies internally. I'm strongly against doing it partially only in one place. Using one approach for one set of strategies and then another approach for other set, and then yet another one, etc. because that's convenient at the moment.

barendgehrels:
> > > Still, some of these changes I want to keep, for example passing side-strategy to intersection.
> 
> > Changes should be made to solve a problem, not because they can be made. What problem does this one solve? Because I think it only introduces complexity and doesn't solve any problem.
> 
> The `intersection` from `cartesian` should **NOT** use the `side::default_strategy` because the user can pass an umbrella strategy containing another strategy. It's crystal clear, and I actually don't understand why it's not done yet by adding the robustness strategies.
> 
> However, because you're not convinced yet, and because it (now we reverted the rest) is not necessarily part of this PR anymore, I'll take it out.
> 
> > If side strategy really has to be passed into one implementation of intersection strategy it should be passed a template argument as it is already done in the library e.g. for distance strategies.
> 
> I already answered this :-) The side strategy can contain member variables such as earth radius.

Oops, excuse me, I edited your comment (didn't know it was possible) instead of replying it...

barendgehrels:
:heavy_check_mark: 
So passing the side-strategy is undone, it's not part of this PR anymore.
Small side-effect: the test I added in this PR now fails exactly because of that reason. That one is still there (now only tested for failures)

awulkiew:
> The intersection from cartesian should NOT use the side::default_strategy

Yes.

> because the user can pass an umbrella strategy containing another strategy. It's crystal clear, and I actually don't understand why it's not done yet by adding the robustness strategies.

The user can define inconsistent strategies in umbrella strategy, yes. The user may also define wrong side strategy in umbrella strategy which would then be passed into the inconsistent intersection strategy. Furthermore this side strategy may be inconsitent with the one used in winding strategy. The problem stays the same.

I understand what you wanted to do and indeed this is an issue but the solution you proposed is a partial one.

In general any strategy may use any number of other strategies which in turn may use any number of yet other strategies. It's the responsibility of umbrella strategy to define consistent strategies. An alternative would be to pass umbrella strategy into all legacy strategies as it is the case with algorithms. And even with that the strategies defined in umbrella strategies would have to be consistent with each other.

> However, because you're not convinced yet, and because it (now we reverted the rest) is not necessarily part of this PR anymore, I'll take it out.

We could think about redesigning strategies, but all of them in the whole library. Until now if a strategy had to be passed into another strategy this was done in the constructor with type defined as class template argument. Here it's passed as function argument. AFAIU this was not required since only one side strategy is used in cartesian intersection strategy.

vissarion:
Regarding the connection between intersection and side strategies, they could use similar computations like determinant computations but they should be independent. That is, you should not, in my opinion, pass the one strategy to another.

barendgehrels:
> Regarding the connection between intersection and side strategies, they could use similar computations like determinant computations but they should be independent. That is, you should not, in my opinion, pass the one strategy to another.

And use one strategy from another? This is done now.

barendgehrels:
> > The intersection from cartesian should NOT use the side::default_strategy
> 
> Yes.
> 
> > because the user can pass an umbrella strategy containing another strategy. It's crystal clear, and I actually don't understand why it's not done yet by adding the robustness strategies.
> 
> The user can define inconsistent strategies in umbrella strategy, yes. The user may also define wrong side strategy in umbrella strategy which would then be passed into the inconsistent intersection strategy. Furthermore this side strategy may be inconsitent with the one used in winding strategy. The problem stays the same.

To be honest, I don't see why in this case this creates an inconsistency... IMO using two different side strategies (as can be done now) is inconsistent.

> 
> I understand what you wanted to do and indeed this is an issue but the solution you proposed is a partial one.
> 
> In general any strategy may use any number of other strategies which in turn may use any number of yet other strategies. It's the responsibility of umbrella strategy to define consistent strategies. An alternative would be to pass umbrella strategy into all legacy strategies as it is the case with algorithms. And even with that the strategies defined in umbrella strategies would have to be consistent with each other.
> 
> > However, because you're not convinced yet, and because it (now we reverted the rest) is not necessarily part of this PR anymore, I'll take it out.
> 
> We could think about redesigning strategies, but all of them in the whole library. Until now if a strategy had to be passed into another strategy this was done in the constructor with type defined as class template argument. Here it's passed as function argument. AFAIU this was not required since only one side strategy is used in cartesian intersection strategy.

Actually I was not aware that, at this moment, still, the `oriented_winding` strategy (and `cartesian_winding` too) does the same thing indeed. It takes the side from default strategy. Wrong, it should take it from the umbrella. I thought these things were already done.

I'm not in favor of redesigning strategies again - we just finished, or nearly finished, that process (umbrella). So if I understand you correctly, I should add it as a class argument and data (if any) can be passed via the constructor, instead of specifying it via `apply`. I will think about that.

Anyway, thanks for your input, let's close the discussion here because it's out of this PR, you can review.

awulkiew:
> And use one strategy from another? This is done now.

Various strategies can internally call other strategies if that's what allows us to avoid duplicating code. That's only an implementation detail. It could be replaced with formula called in both strategies, etc. This doesn't mean that these internally used strategies should be exposed in the interface of other strategies.

> To be honest, I don't see why in this case this creates an inconsistency... IMO using two different side strategies (as can be done now) is inconsistent.

If you define a strategy and pass it into a different strategy which was not designed around the former strategy then the result may still be wrong. Consider the spherical example I gave before. Even if you used spherical side formula (3d vectors) to calculate side everywhere (in intersection, winding, etc.) but used spherical math to calculate intersection these methods wouldn't be consistent with each other. The burden of defining consistent strategies would still rely on umbrella strategy. The proposed change wouldn't have any effect regarding this. It'd only make things more complex i.e. it'd add another parameter changing the behavior of a strategy that might be incompatible with some parts of this strategy. E.g. if we wanted to test such intersection strategy we'd have to know which side strategies it is able to take to produce correct results.

> Actually I was not aware that, at this moment, still, the oriented_winding strategy (and cartesian_winding too) does the same thing indeed. It takes the side from default strategy.

AFAIR the default strategy mechanism was added by you to avoid using `#ifdefs` and was a part of the experiment of replacing `side_by_triangle` with `side_robust` which AFAIU improved some cases and break other ones. This was not the original solution. Originally only one `side_by_triangle` strategy was used in the cartesian intersection strategy. There are many tweaks in intersection strategy to ensure that both of them work well with each other.

> Wrong, it should take it from the umbrella. I thought these things were already done.

No they were not and I think this shouldn't be done. In principle non-precise intersection strategy doesn't guarantee to work well with precise side strategy. Just like intersection strategy modeling segment as geodesics will not work well with side strategy modeling segments as loxodromes. Just like intersection strategy using spherical math will not work well with side strategy using ECEF vectors. AFAIU in principle intersection strategy doesn't have anything to do with side computation. The fact that in Boost.Geometry side strategy is used in cartesian intersection strategy is only an implementation detail.

I think a strategy shouldn't have to guarantee to take any other strategy. The interfaces in general should be the least complex as possible. A strategy should only calculate CS-specific result for whatever input is passed into it.

Consider the fix I proposed for non-cartesian envelope of a polygon: https://github.com/boostorg/geometry/pull/936 where winding strategy is called internally in envelope strategy to check if a pole is contained in a polygon. This winding strategy use special spherical/geographic side strategy (the same for both CSes) capable only to tell the relation WRT a pole assuming other conditions are met (e.g. there is no segment going through the pole). This makes sense only in this case and shouldn't be exposed anywhere. This is only an implementation detail. It could potentially be replaced with some other method. The fact that winding strategy with custom side strategy is used internally doesn't mean that we have to expose it in the interface.

> So if I understand you correctly, I should add it as a class argument and data (if any) can be passed via the constructor, instead of specifying it via apply.

I don't know if you should. You could. If cartesian intersection strategy can work correctly with various side strategies. There are so many tweaks in this strategy that my guess would rahter be that it can properly work only with one side strategy. But if you want to do this then just look at winding or distance strategies. AFAIR they were implemented like this from the beginning.

tinko92:
Because I wrote some of the code used by side_robust, I would like to offer my perspective on the different side strategies, I hope the following can help with the decision. In short, I don't think a side-strategy can be at the same tolerant of near-collinearity (e.g. having an epsilon policy), simple and robust (as in not having substantial bugs for at least some inputs).

The orientation predicate side_robust (with fp_equals_policy and a binary floating-point calculation type) is exact on double/float-coordinates (ignoring overflow/underflow issues for very small/large numbers) and thereby guarantees logical consistency across multiple calls, e.g. (assuming no duplicate points) if p1, p2, p3 and p1, p2, p4 are collinear, then side_robust will also see p1, p3, p4 as collinear / if p1, p2, p3 are not collinear but p1, p2, p4 are, then side_robust will not see p1, p3, p4 as collinear. So, for an algorithm like e.g. a Triangulation algorithm, which does not construct any new points and is correct under the real RAM model for its predicate, this correctness-guarantee of side_robust should guarantee valid, correct triangulations. That is the motivation for side_robust.

I looked into some of the failing tests in Boost.Geometry because @barendgehrels  mentioned that side_robust causes regressions without rescaling. For example, the difference test for LINESTRING(2 8,4 0.4,8 1) minus LINESTRING(0 -0.2,8 1) failed in my test. The correct result in decimal arithmetic is MULTILINESTRING((2 8,4 0.4)). In double arithmetic, this problem cannot be expressed correctly with point_xy<double> because decimal 0.4 is the infinite binary fraction 0.01100110..._2, which is rounded at the 53rd binary digit when stored in double, similarly for -0.2. So, as @vissarion  mentioned, the inaccuracy is already there before the geometries are given to bg::difference and the side_strategy.

A) For these double approximations of decimal numbers side_robust correctly says that POINT(0 -0.2)	POINT(8 1)	POINT(4 0.4) are not collinear and the difference algorithm computes MULTILINESTRING((2 8,4 0.4,8 1)) if it uses side_robust, which, I think, is correct for the nearest double-approximation of the problem.
B) side_by_triangle with its epsilon policy will declare POINT(0 -0.2), POINT(8 1), POINT(4 0.4) to be collinear and produce the result that was perhaps expected by the user.
C) If the problem is rescaled to integers with any factor divisible by 10, this will cause a rounding error that happens to exactly compensate the original error introduced from converting "0.4" and "-0.2" to double and probably also deliver the expected result.

B) and C) will produce the expected result but they do so by making subtle errors themselves that happen to often mask or compensate the subtle degeneracies in test cases that are written in particular ways, e.g. reading decimal numbers with few significant digits from text-input, whose magnitude is not too much larger or too much smaller than 1.

But programs can be written that will make the non-robust predicates behave in unexpected ways, too, e.g. consider the line strings a=LINESTRING(0.01 68.39,100 94.39) and b=LINESTRING(77.63 0.01,80.32 100). If I compute their intersection with boost::geometry (an inexact construction), I get c=POINT(80.029373406089661 89.197117797363049). side_by_triangle (compile with g++ -O3 -march=native) will return a non-zero result for a.first, a.second and c, so its epsilon tolerance can be easily exceeded with simple examples that intuitively would be collinear by construction, and side_by_triangle also returns the same value for (a.first, a.second, c), (c, a.first, a.second) and (a.first, c, a.second) which also defeats the lexicographic sorting that was probably intended to make it consistent for permutations of three points.

So, in summary, I would say:
naive computation ( det = (ax-cx)*(by-cy) - (ay-cy)*(bx-cx); return (det > 0) - (det < 0); ):
+ potentially fastest
- least robust
- can easily produce counterintuitive results

side_robust with fp_equals:
+ robust (never contradicts itself, exact results for float/double-coordinates)
+ relatively fast
- will produce counterintuitive results for many almost-collinear points, like points from decimal inputs that were read to double

side_robust with epsilon policy:
+ relatively fast
+ will not give the opposite of the correct sign (it will not say left when it is right, but it might say collinear)
+ probably produces fewer counterintuitive results for decimal inputs
- not robust (can contradict itself)

side_by_triangle with epsilon policy
+ probably produces fewer counterintuitive results for decimal inputs
- not robust (can contradict itself, depending on compilation options, even for permutations of three points)
- lots of unpredictable branches for the implicit lexicographic sorting (performance penalty)

I don't think there exists a solution that is simultaneously fast, simple and produces neither counterintuitive results nor invalid results for at least some cases (non-robustness). A solution that is simultaneously robust and produces expected results for decimal inputs and constructed points, would have to read the WKT as decimal, convert it to an arbitrary precision rational type and perform all constructions and predicate calls with that. I tested that and it produces MULTILINESTRING((2 8,4 2/5)) for the example problem, as expected. 

I think for exact, arbitrary precision number types (I think I saw tests for rational), the naive computation without epsilon policy is always a good choice. And I think for integers (including rescaled coordinates) that fit into the mantissa of doubles (<2^53), converting to double and using side_robust with fp_equals policy is always a good choice because it will guarantee a correct result in terms of the integer input and avoid the potential integer overflow for integers that are larger than the square-root of the largest allowed integer value.

In case, it could be a useful solution, I implemented a (branchless) side-strategy that returns collinear if the orientation is unclear either due to rounding-error in the computation or due to rounding error in the input coordinates, see here: https://gist.github.com/tinko92/5f242339ae312f66206f2be69184be72 .

barendgehrels:
Thanks for all input. I'll come back to it.

But we're talking now about things not in this PR. I'm curious if I can merge it.

barendgehrels:
> In case, it could be a useful solution, I implemented a (branchless) side-strategy that returns collinear if the orientation is unclear either due to rounding-error in the computation or due to rounding error in the input coordinates, see here: https://gist.github.com/tinko92/5f242339ae312f66206f2be69184be72 .

@tinko92 I tested that one and it looks promising! A quick run shows that almost all tests succeed, and the few tests which don't succeed are usually the really fragile tests. Plus a few in buffer to be investigated further (but these aren't trivial either). So it that sense it's much more compatible to the current implementation. This unlike the robust version which let some trivial testcases fail (for double coordinates, no rescaling).

The implementation is concise, are the `5` and `56` a kind of tunable constants or what is the reasoning behind them?

test/algorithms/set_operations/intersection/alternative_side_strategy.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+
+#ifndef BOOST_GEOMETRY_TEST_INTERSECTION_ALTERNATIVE_SIDE_STRATEGY_HPP
+#define BOOST_GEOMETRY_TEST_INTERSECTION_ALTERNATIVE_SIDE_STRATEGY_HPP
+
+#include <boost/geometry/strategies/relate/cartesian.hpp>
+
+
+template <typename SideStrategy>
+class alternate_side_strategy
```
awulkiew, line 17: Technically this is not side strategy but umbrella strategy using non-standard side strategy.

test/algorithms/set_operations/intersection/alternative_side_strategy.hpp
```diff
@@ -0,0 +1,48 @@
...
+    template <typename Geometry1, typename Geometry2>
+    static auto relate(Geometry1 const& g1, Geometry2 const& g2)
+    {
+        return default_umbrella::relate(g1, g2);
+    }
+
+    static auto relate()
+    {
+        return default_umbrella::relate();
+    }
```
awulkiew, line 39: This is in the tests so ok but there are some technicall issues I'd like to point out in case you wanted to build on top of it in the future.

1. If these functions are needed then this strategy should have been derived from `bg::strategies::relate::cartesian<>`. Then you could ommit these functions. Later if you needed to overload one of them you could bring the rest to the scope with e.g. `using bg::strategies::relate::cartesian<>::relate`.

2. `SideStrategy` may be incompatible with intersection and winding strategies. I'd implement base strategies for them taking `SideStrategy` as template argument and derive the public versions from these bases to hide this additional `SideStrategy` argument. Like it was done for spherical winding here:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/strategies/spherical/point_in_poly_winding.hpp#L52

awulkiew (approved):

awulkiew:
@barendgehrels Yes, I'm ok with merging. Thanks for the discussion!

I commented about some issues that might be a problem in future PRs but they don't prevent this one to be merged.

tinko92:
> The implementation is concise, are the `5` and `56` a kind of tunable constants or what is the reasoning behind them?

@barendgehrels They are the result of numerical forward error analysis but the 56 could be 32 (I think made a mistake yesterday, I corrected the gist. It shouldn't matter much, though. the impact of the coefficient for epsilon² is almost negligible). 

## Reasoning
They are computed by starting with an error of |x| * eps for each input and iteratively computing and simplifying error bounds for each +,-,*-operation. The final error bound is (5*eps + 25*eps²) and the least upper bound in floating-point numbers for that is (5*eps + 32*eps²). If the determinant is larger than this bound then it should be guaranteed to be correct, otherwise, it's uncertain due to either rounded inputs or rounding errors in the computation of the determinant.

## Tunability
5 and 32 should be the smallest (or very close to the smallest) coefficients if one wants to account for rounded inputs and computation error in the side computation. 3 and 16 would only account for the computation error in the side computation and assume precise inputs.
One could also increase the epsilon to account for even worse rounding in the inputs rather than rounding to the nearest float E.g. set epsilon to 5*10^-(x+1) to only consider inputs precise up to the x-th digit, e.g. to consider something collinear, if it was collinear in some program, written out with to_wkt with limited digits and read back. But at that point the bound would be too pessimistic, IMHO, and it would be to the detriment of someone who purposefully enters more than x digits into points.
I don't think it can be tuned to be consistent with all constructions because one cannot really compute a rigorous error bound/epsilon tolerance for a point that was computed e.g. as the intersection of two segments without knowing the original coordinates of the two segments.

barendgehrels:
> @barendgehrels Yes, I'm ok with merging. Thanks for the discussion!
> 
> I commented about some issues that might be a problem in future PRs but they don't prevent this one to be merged.

Thanks @awulkiew , I applied the renaming and inheritance and I'll merge it.

barendgehrels:
> 5 and 32 should be the smallest (or very close to the smallest) coefficients if one wants to account for rounded inputs and computation error in the side computation. 
@tinko92 thanks again, great

Indeed the behavior looks the same with 5/56

Next week (=year) I'll look further and probably create a PR with this new strategy.

barendgehrels:
> They are computed by starting with an error of |x| * eps for each input and iteratively computing and simplifying error bounds for each +,-,*-operation. The final error bound is (5_eps + 25_eps²) and the least upper bound in floating-point numbers for that is (5_eps + 32_eps²). Edit: [sketch proof.pdf](https://github.com/boostorg/geometry/files/7789179/sketch.proof.pdf) If the determinant is larger than this bound then it should be guaranteed to be correct, otherwise, it's uncertain due to either rounded inputs or rounding errors in the computation of the determinant.

Great explanations and document! Thanks!

It's merged now.
@awulkiew @vissarion @mloskot @tinko92 Happy New Year!

PR #958, opened 2022-01-05
[sectionalize] remove dependency on define

barendgehrels:
This PR removes the `#if defined(BOOST_GEOMETRY_USE_RESCALING)` condition from sectionalize, and makes the implementation shorter anyway.
This is the last real dependency on the define `BOOST_GEOMETRY_USE_RESCALING` and `BOOST_GEOMETRY_NO_ROBUSTNESS`, which means we can change the default somewhere (hopefully very soon) after this PR is merged.

### DIFF
--- a/include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
+++ b/include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
@@ -42,7 +42,6 @@
 #include <boost/geometry/algorithms/detail/recalculate.hpp>
 #include <boost/geometry/algorithms/detail/ring_identifier.hpp>
 #include <boost/geometry/algorithms/detail/signed_size_type.hpp>
-#include <boost/geometry/algorithms/detail/buffer/buffer_box.hpp>
 
 #include <boost/geometry/core/access.hpp>
 #include <boost/geometry/core/closure.hpp>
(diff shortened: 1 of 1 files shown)


include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
```diff
@@ -729,79 +728,28 @@ struct sectionalize_multi
...
-template <>
-struct expand_by_epsilon<spherical_tag>
-{
-    template <typename Box>
-    static inline void apply(Box & box)
-    {
-        typedef typename coordinate_type<Box>::type coord_t;
-        static const coord_t eps = std::is_same<coord_t, float>::value
-            ? coord_t(1e-6)
-            : coord_t(1e-12);
```
barendgehrels, line 762: These values correspond with the new values

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
```diff
@@ -729,79 +728,28 @@ struct sectionalize_multi
...
-        // should be larger than epsilon. About the value itself: the smaller
-        // it is, the higher the risk to miss intersections. The larger it is,
-        // the more comparisons are made. So it should be on the high side.
         using gt = decltype(section.bounding_box);
         using ct = typename geometry::coordinate_type<gt>::type;
-        ct const tolerance = ct(1) / ct(1000);
-        detail::buffer::buffer_box(section.bounding_box, tolerance,
-                                   section.bounding_box);
-#endif
+        static ct const eps = math::scaled_epsilon<ct>(1000);
```
barendgehrels, line 751: This is similar to `numeric_limits<T>::epsilon() * 1000` and we could replace it if you prefer that.

awulkiew (approved):

vissarion (approved):

PR #951, opened 2021-12-15
fix: define get_distance_measure by default

barendgehrels:
This fixes circleci
https://app.circleci.com/pipelines/github/boostorg/geometry/257/workflows/00234789-5f80-4b23-9075-4c359ec83d11/jobs/15405

PR #959, opened 2022-01-05
Propagate umbrella strategy in buffer and overlay.

awulkiew:
Furthermore:
- pass umbrella strategy to some algorithms which were using default strategy.
- derive buffer umbrella strategies from distance strategies.

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
@@ -3,8 +3,8 @@
 // Copyright (c) 2012-2020 Barend Gehrels, Amsterdam, the Netherlands.
 // Copyright (c) 2017 Adam Wulkiewicz, Lodz, Poland.
 
-// This file was modified by Oracle on 2016, 2018.
-// Modifications copyright (c) 2016-2018 Oracle and/or its affiliates.
+// This file was modified by Oracle on 2016-2022.
+// Modifications copyright (c) 2016-2022 Oracle and/or its affiliates.
 // Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
 
 // Use, modification and distribution is subject to the Boost Software License,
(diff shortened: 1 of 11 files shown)


awulkiew:
This PR also fixes an issue introduced by this change: https://github.com/boostorg/geometry/pull/951

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
```diff
@@ -21,6 +21,7 @@
 #include <boost/geometry/algorithms/covered_by.hpp>
 #include <boost/geometry/algorithms/detail/disjoint/point_box.hpp>
 #include <boost/geometry/algorithms/detail/disjoint/box_box.hpp>
+#include <boost/geometry/algorithms/detail/dummy_geometries.hpp>
```
vissarion, line 24: it is unclear to me whether dummy geometries are used in this file

vissarion (approved):
Thanks!

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
awulkiew: Because they are used by algorithms and they shouldn't be used by the users, hence `detail`. If you think they should be moved somewhere. I'd be ok with it.

barendgehrels (approved):
Thanks! Looks all good to me

barendgehrels:
> This PR also fixes an issue introduced by this change: #951 as described here: #949

I don't know exactly what the issue was (I thought #951 fixes #949 but it's unclear to me what was wrong later). Anyway, sorry for the inconvenience and fix for fixing it and adding umbrella to buffer.

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
vissarion: Maybe it is cleaner to be included by the details of any algorithm using them, but I am also OK as is for now since this is a bit unrelated with this PR.

PR #961, opened 2022-01-13
Add Web Mercator projection

vissarion:
This PR adds the Web Mercator / Pseudo Mercator projection a variant of the regular Mercator projection. For computations the Web Mercator projection uses the formulas of Mercator projection.

More info:
- https://proj.org/operations/projections/merc.html
- https://proj.org/operations/projections/webmerc.html


include/boost/geometry/srs/projections/proj/webmerc.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <boost/geometry/srs/projections/impl/base_static.hpp>
+#include <boost/geometry/srs/projections/impl/base_dynamic.hpp>
+#include <boost/geometry/srs/projections/impl/factory_entry.hpp>
+#include <boost/geometry/srs/projections/impl/pj_msfn.hpp>
+#include <boost/geometry/srs/projections/impl/pj_param.hpp>
+#include <boost/geometry/srs/projections/impl/pj_phi2.hpp>
+#include <boost/geometry/srs/projections/impl/pj_tsfn.hpp>
+#include <boost/geometry/srs/projections/impl/projects.hpp>
+
+#include <boost/geometry/srs/projections/proj/merc.hpp>
```
awulkiew, line 23: Note that if this file is included then the includes above and below are not needed.

But I'd propose to define this projection in the same file as merc projection. This is how it's done with projections sharing the same bases and this is how it's done in proj.

awulkiew:
Thanks!
Are there any EPSG codes using this projection?
If there are then you could add them here:
- [dynamic](https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/srs/projections/epsg.hpp)
- [static](https://raw.githubusercontent.com/boostorg/geometry/develop/include/boost/geometry/srs/projections/epsg_traits.hpp)

I only don't know what to do with the old ones where `merc` is used while `webmrec` probably should. Should they be modified/replaced?
@mloskot what do you think?

include/boost/geometry/srs/projections/proj/webmerc.hpp
awulkiew: See e.g. [sts.hpp](https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/srs/projections/proj/sts.hpp#L245-L263) where 4 projections are defined.

vissarion:
> Thanks! Are there any EPSG codes using this projection? If there are then you could add them here:
> 
>     * [dynamic](https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/srs/projections/epsg.hpp)
> 
>     * [static](https://raw.githubusercontent.com/boostorg/geometry/develop/include/boost/geometry/srs/projections/epsg_traits.hpp)

I only know EPSG:3857 which is Web Mercator with WGS84 as a base ellipsoid. 
https://epsg.org/crs_3857/WGS-84-Pseudo-Mercator.html
I added that. 
 
> 
> I only don't know what to do with the old ones where `merc` is used while `webmrec` probably should. Should they be modified/replaced? @mloskot what do you think?

Do you have an example of an old one?

awulkiew:
> Do you have an example of an old one?

Deprecated EPSG 3785, see: https://github.com/boostorg/geometry/issues/657

vissarion:
> EPSG 3785

I changed it to be the same as in EPSG: 3857 see https://github.com/boostorg/geometry/pull/961/commits/6d22df0850668504878af1b62b824230264bbc63

awulkiew (approved):

awulkiew:
Ok thanks. Defining it as the same projection is on of the possible solutions. I don't know if this is the correct approach but I'm ok with it.

Minor detail: the last commit message is using a different format. It should've been `[srs] Set ...` or `[projections] Set ...`, not 'Setting ...'.

mloskot:
Sorry guys, I was taking a week off.

Yes, the `EPSG:3857` is the official code, but there is a collection of historical 'aliases', here is complete list https://wiki.openstreetmap.org/wiki/EPSG:3857

vissarion:
> Sorry guys, I was taking a week off.
> 
> Yes, the `EPSG:3857` is the official code, but there is a collection of historical 'aliases', here is complete list https://wiki.openstreetmap.org/wiki/EPSG:3857

thanks @mloskot !

I also replaced the deprecated 3587 but the rest of the list (esri and osgeo) are not supported by boost geometry. Does it make sense to support them and bind them to the same behavior are webmerc, or we simply keep avoiding them. I am in favor of the latter.

mloskot:
@vissarion I'd also suggest to avoid them. The sooner users stop using them, globally, the better :)

barendgehrels (approved):
Cool!
Is it 100% compatible with Web Mercator google uses?
Can we just combine it and overlay our maps over it?

barendgehrels:
> > Cool! Is it 100% compatible with Web Mercator google uses?
> 
> It is not clear what Google uses in its several map services and apps. According to [wikipedia](https://en.wikipedia.org/wiki/Google_Maps) in `Google maps` a variant of the mercator projection was used in the past while the `Google maps platform` [uses](https://developers.google.com/maps/documentation/javascript/examples/map-coordinates) the web mercator projection similar to the one implemented in Boost Geometry up to some constants and rounding.
> 
> > Can we just combine it and overlay our maps over it?
> 
> That sounds cool, do you have a specific project in mind?

Yes, I think it would be cool! But I didn't plan something until now... Maybe we can make an example, like we've more of them.

Issue #956, opened 2021-12-31
boost::geometry::simplify() returns out with incorrect closure


Zeratul-Aiur:
I create a ring with counterclockwise and open points.When calculating the simplify I found that the result did not detect closure_selector.

example code:
```
#include <boost/geometry.hpp>
#include <iostream>

using namespace boost::geometry;

int main() {
  using Point = model::point<float, 2, cs::cartesian>;
  using Ring = model::ring<Point, false, false>;
  Ring ring{{0, 0}, {0.4, 0}, {0.4, 0.4}};
  Ring result;
  simplify(ring, result, 0);
  std::cout << wkt(result) << std::endl;
}
```
Output : POLYGON((0.4 0.4,0 0,0.4 0,0.4 0.4))
I would expect to get : POLYGON((0.4 0.4,0 0,0.4 0))

I think the problem is here:
```
// In simplify.hpp

std::rotate_copy(boost::begin(ring), range::pos(ring, index),
                        boost::end(ring), rotated.begin());

// Close the rotated copy
rotated.push_back(range::at(ring, index));

simplify_range<0>::apply(rotated, out, max_distance, impl, strategies);
```

yashdabhade1:
A polygon is defined in aw ay where the first and last points are same
So i feel that the output is correct

awulkiew:
Thanks for the report! I was able to reproduce it and prepare a fix.

@yashdabhade1 the Ring/Polygon in the example is defined as counter-clockwise and opened so there should be no closing point in the output.

Side note. With `bg::wkt(result)` the text representation for Rings will have 3 points but 4 points for Polygons. This doesn't mean that the Polygon is closed. Current implementation of `bg::wkt()` works this way. So if you want to get the number of points that are in the geometry consider using e.g. `bg::wkt_manipulator<Geometry>(result, false)` or `bg::dsv(result)` instead.

awulkiew:
Should the behavior of `bg::wkt()` be changed for rings and made consistent with polygons @barendgehrels @mloskot @vissarion ?

vissarion:
> @barendgehrels @mloskot @vissarion Should the behavior of `bg::wkt()` be changed for rings and made consistent with polygons?

In general I am in favor of consistency since a ring is semantically a polygon and the current behavior of `wkt()` is confusing. But I have no clue about the effect to current users since this is a braking change. 

In conclusion +1 from me to change the behavior.

mloskot:
[...]

PR #966, opened 2022-01-18
Fix simplify for open rings and improve it

awulkiew:
Fix for https://github.com/boostorg/geometry/issues/956

Furthermore:
- avoid duplicating closing point for closed rings while rotating the range
- use more efficient strategy while calculating index of opposite point

### DIFF
--- a/include/boost/geometry/algorithms/simplify.hpp
+++ b/include/boost/geometry/algorithms/simplify.hpp
@@ -4,8 +4,8 @@
 // Copyright (c) 2008-2015 Bruno Lalande, Paris, France.
 // Copyright (c) 2009-2015 Mateusz Loskot, London, UK.
 
-// This file was modified by Oracle on 2018-2021.
-// Modifications copyright (c) 2018-2021 Oracle and/or its affiliates.
+// This file was modified by Oracle on 2018-2022.
+// Modifications copyright (c) 2018-2022 Oracle and/or its affiliates.
 // Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
 
 // Parts of Boost.Geometry are redesigned from Geodan's Geographic Library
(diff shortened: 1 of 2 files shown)


vissarion (approved):
Thanks, I am OK with changing this.

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -472,14 +474,29 @@ public :
...
-            std::rotate_copy(boost::begin(ring), range::pos(ring, index),
-                             boost::end(ring), rotated.begin());
+            // Do not duplicate the closing point
+            auto rot_end = boost::end(ring);
+            std::size_t rot_index = index;
+            if (is_closed && size > 1)
+            {
+                --rot_end;
+                if (rot_index == size - 1) { rot_index = 0; }
+            }
```
barendgehrels, line 484: Right, thanks!

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -472,14 +474,29 @@ public :
...
-            rotated.push_back(range::at(ring, index));
+            rotated.push_back(range::at(ring, rot_index));
 
             simplify_range<0>::apply(rotated, out, max_distance, impl, strategies);
 
+            // Open output if needed
+            if (! is_closed && boost::size(out) > 1)
+            {
+                range::pop_back(out);
+            }
```
barendgehrels, line 498: Maybe I get it wrongly, but above you rotate with one point less (fine) - and now you pop another one. Is that not one too much?

barendgehrels (approved):
Still a question but the tests look good, I so it will work.
Maybe a small extra comment would be useful.

include/boost/geometry/algorithms/simplify.hpp
awulkiew: These are two different issues.

The point above is removed only for closed rings. Otherwise when they are rotated then two equal points (beginning and closing) are moved somewhere into the middle of the ring.

After the rotation an additional closing point is pushed back, always, even for open rings. This is why here, for open rings this additional point is removed.

awulkiew: An alternative to `pop_back()` would be not copying the last point for open output ring inside the `simplify_range<0>::apply()`.

PR #967, opened 2022-01-18
Fix closest points test

awulkiew:
- remove #define
- alter error thresholds, otherwise the tests are failing with msvc-14.1


vissarion (approved):
Thanks!

Issue #930, opened 2021-10-27
Intersection does not behave as expected


Zilleplus:
I created a square of width 10, and a flat triangle of height 1 that crosses the square. When calculating the intersection I get the whole plane above the triangle. I would expect to get small triangle back.

example code:

```
#include<iostream>
#include<boost/geometry.hpp>

using namespace boost::geometry;

int main(){
    using point = model::point<int, 2, cs::cartesian>;
    using poly = model::polygon<point>;

    poly r = {{{0,0}, {0,10}, {10,10}, {10,0},{0,0}}};
    poly cross = {{{-10,2},{5,3},{20,2},{-10,2}}};

    std::vector<poly> out;
    intersection(r, cross, out);

    for(auto x : out)
    {
        std::cout << wkt(x) << std::endl;
    }

}
```
compiler explorer link: https://compiler-explorer.com/z/GhfhTGz7q

Output: POLYGON((0 2,0 10,10 10,10 2,0 2))
I would expect to get POLYGON((0 2, 5 3, 10 2, 0 2))

Zilleplus:
the never version 1.78.0 still contains this bug.

barendgehrels:
Did you try already with the define `BOOST_GEOMETRY_NO_ROBUSTNESS`?

Zilleplus:
@barendgehrels thank you for the advice I did not.  I just tried and unfortunately the result is exactly the same.

Kissh12:
can I work on this?

barendgehrels:
> can I work on this?

Yes, you can try. But it won't be easy...
I'll have a look myself next Wednesday.

Kissh12:
> > can I work on this?
> 
> Yes, you can try. But it won't be easy... I'll have a look myself next Wednesday.

can I have some help over this as I am new to this

barendgehrels:
> > > can I work on this?
> > 
> > 
> > Yes, you can try. But it won't be easy... I'll have a look myself next Wednesday.
> 
> can I have some help over this as I am new to this

I would suggest then to start with a simpler issue. This is one of the more complex things in the library.

The best starting point is the documentation for contributors.

barendgehrels:
It's because of the integer coordinates, there are integer divisions (no rounding, no conversion to floating point). This causes the intersection point to casted to its lower value.

See this picture:

![image](https://user-images.githubusercontent.com/334849/150117914-1dbb0cc2-0cc8-4996-9567-fed3de3baadc.png)

[...]

PR #972, opened 2022-01-26
[fix] for integer coordinates, segment intersection now rounds to nearest point

barendgehrels:
This fixes issue #930

For some integer cases there were slight variations, I inspected them visual. In general it seems a bit better, and there are no real regressions (spikes might be located on different places).

The best improvement is this one:

old behaviour:
image

and it is now:
image


include/boost/geometry/util/math.hpp
```diff
@@ -796,6 +820,17 @@ inline Result rounding_cast(T const& v)
...
+/*
+\brief Short utility to divide. If the division is integer, it rounds the division
+       to the nearest value, without using floating point calculations
+\ingroup utility
+*/
+template <typename T>
+inline T divide(T const& n, T const& d)
+{
+    return detail::divide<T>::apply(n, d);
+}
```
awulkiew, line 832: I'd like to propose to move away from this implementation technique (function calling struct member function) and use more modern one instead. This way there is one less function call per utility, which means smaller call trees and therefore simpler compilation error messages, maybe shorter compilation times? I'm proposing to do this in details, utilities such as this one, etc. The tags dispatching of algorithms would stay as it is. So this implementation could be replaced with:
```
template <typename T, std::enable_if_t<! std::is_integral<T>::value, int> = 0>
inline T divide(T const& n, T const& d)
{
    return ...
}

template <typename T, std::enable_if_t<std::is_integral<T>::value, int> = 0>
inline T divide(T const& n, T const& d)
{
    return ...
}
```

barendgehrels: > I'd like to propose to move away from this implementation technique (function calling struct member function) and use more modern one instead. This way there is one less function call per utility, which means smaller call trees and therefore simpler compilation error messages, maybe shorter compilation times? I'm proposing to do this in details, utilities such as this one, etc. The tags dispatching of algorithms would stay as it is. So this implementation could be replaced with:
> 
> ```
> template <typename T, std::enable_if_t<! std::is_integral<T>::value, int> = 0>
> inline T divide(T const& n, T const& d)
> {
>     return ...
> }
> 
> template <typename T, std::enable_if_t<std::is_integral<T>::value, int> = 0>
> inline T divide(T const& n, T const& d)
> {
>     return ...
> }
> ```

I don't agree with this. Our architecture is tag dispatching and specialization, we've that all over the place, and I don't feel to rebuild that all.
This is made consistent with the other code, in the math unit - but basically everywhere.

Maybe, if C++20 is there, we can make use of concepts, and we might make another choice, but that will first need some research (especially about error messages in case of errors).

barendgehrels:
@vissarion OK to merge?

include/boost/geometry/util/math.hpp
```diff
@@ -525,6 +525,30 @@ struct rounding_cast<Result, Source, true, false>
     }
 };
 
+template <typename T, bool IsIntegral = std::is_integral<T>::value>
+struct divide
+{
+    static inline T apply(T const& n, T const& d)
+    {
+        return n / d;
```
vissarion, line 533: what if `d==0` ?

awulkiew: Then the result is as it was before so +/-Inf or NaN. But AFAIU `denominator` in segment ratio is never 0.

include/boost/geometry/util/math.hpp
```diff
@@ -525,6 +525,30 @@ struct rounding_cast<Result, Source, true, false>
...
+struct divide<T, true>
+{
+    static inline T apply(T const& n, T const& d)
+    {
+        return n == 0 ? 0
+          : n < 0
+          ? (d < 0 ? (n + (-d + 1) / 2) / d + 1
+                   : (n + ( d + 1) / 2) / d - 1  )
+          : (d < 0 ? (n - (-d + 1) / 2) / d - 1
+                   : (n - ( d + 1) / 2) / d + 1  )
```
awulkiew, line 547: More interesting is what would happen here if d was 0. The answer is the same as above, i.e.: the same as before this change was made, i.e. undefined behavior.

include/boost/geometry/util/math.hpp
barendgehrels: Indeed, the behavior is as before.

include/boost/geometry/util/math.hpp
barendgehrels: I thought about it too but since the integer division lines up with the floating point division, the behavior should be the same. In other words, if we check integer divisor 0, we should also check floating point divisor 0...

PR #976, opened 2022-02-07
[test] fix coordinates to make the polygon invalid as designed

barendgehrels:
for both rescaled and without rescaling


test/algorithms/is_valid.cpp
```diff
@@ -948,13 +1020,18 @@ inline void test_open_multipolygons()
...
...9187183650888029,3.2446420208654896 7.0415020265598844,3.1807360092909640 7.1691699739962251,3.1261005500004773 7.3010716408018013,3.0810140527710059 7.4365348863171379,3.0457062680576819 7.5748694208940437,3.0203571162381349 7.7153703234534268,3.0050957707794934 7.8573216336015328,3.0000000000000000 7.9999999999999973,3.0050957707794925 8.1426783663984619,3.0203571162381344 8.2846296765465670,3.0457062680576801 8.4251305791059501,3.0810140527710042 8.5634651136828559,3.1261005500004755 8.6989283591981934,3.1807360092909618 8.8308300260037704,3.2446420208654869 8.9584979734401102,3.3174929343376358 9.0812816349111927,3.3989175181512770 9.1985553330226910,3.4885008512914810 9.3097214678905669,3.5857864376269024 9.4142135623730923,3.6902785321094269 9.5114991487085145,3.8014446669773028 9.6010824818487190,3.9187183650888020 9.6825070656623602,4.0415020265598844 9.7553579791345104,4.1691699739962260 9.8192639907090360,4.3010716408018030 9.8738994499995236,4.4365348863171405 9.9189859472289950,4.5748694208940472 9.9542937319423199,4.7153703234534312 9.9796428837618656,4.8573216336015381 9.9949042292205075,5.0000000000000036 10.0000000000000000,5.1426783663984690 9.9949042292205075)))";
```
barendgehrels, line 1028: This is the fix, of one of the points the coordinates were `7.0 7.9999999999...3`, which succeeded with rescaling (because coordinates are rounded to the integer grid) but failed without rescaling (which is actually correct here).
Now the two points are exactly the same such that the validity fails in the same way

barendgehrels:
After this PR, all tests run fine without rescaling.
So we can think about changing the default now.
I will give (but not today0 an overview on which tests perform how in each configuration.

The deadline for major changes (and this is one) is March 2

test/algorithms/is_valid.cpp
awulkiew: Isn't it the whole point of rescaling to handle FP coordinates differently?

This test case looks like some real-life example. E.g. it might have been added after a user had reported an issue. Or it might have been a result of some operation which should not generate invalid geometries. When was this case added?

Is the result the same with different compilers?

barendgehrels: It's my own test. And I explained what was wrong, and why it was wrong, and why it is right now.

awulkiew (approved):
LGTM

vissarion (approved):

PR #977, opened 2022-02-16
[major change] change the default, disable rescaling by default

barendgehrels:
This will disable rescaling by default. Current behavior on most testsuites, using `BOOST_GEOMETRY_TEST_FAILURES` (so it checks how many failures there are left) is shown below.

It's in almost all cases better now then with rescaling (but it took me quite some time...)

Here is the list (occasionally the expectations are not updated and off, but it's the **detected** which is important)

**Set operations**, without rescaling:
```
===  union using double
NOT RESCALED - Expected: 1 error(s) ** 1 failure detected
===  union_multi using double
NOT RESCALED - Expected: 0 error(s) ** No errors detected
===  intersection using double
NOT RESCALED - Expected: 2 error(s) ** 2 failures detected
===  intersection_multi using double
NOT RESCALED - Expected: 1 error(s) ** 1 failure detected
===  difference using double
NOT RESCALED - Expected: 5 error(s) ** 5 failures detected
===  difference_multi using double
NOT RESCALED - Expected: 11 error(s) ** 12 failures detected
```

**Set operations**, with rescaling:
```
===  union using double
RESCALED - Expected: 4 error(s) ** 4 failures detected
===  union_multi using double
RESCALED - Expected: 9 error(s) ** 9 failures detected
===  intersection using double
RESCALED - Expected: 6 error(s) ** 6 failures detected
===  intersection_multi using double
RESCALED - Expected: 9 error(s) ** 9 failures detected
===  difference using double
RESCALED - Expected: 15 error(s) ** 15 failures detected
===  difference_multi using double
RESCALED - Expected: 24 error(s) ** 25 failures detected
```

**Buffer**, without rescaling:
```
=== RUNNING buffer_polygon using double
NOT RESCALED - Expected: 1 error(s) ** 1 failure detected
== RUNNING buffer_multi_polygon using double
NOT RESCALED - Expected: 1 error(s) ** 1 failure detected
== RUNNING buffer_linestring using double
NOT RESCALED - Expected: 4 error(s) ** 4 failures detected
== RUNNING buffer_multi_linestring using double
NOT RESCALED - Expected: 6 error(s) ** 6 failures detected
== RUNNING buffer_linestring_aimes using double
NOT RESCALED - Expected: 0 error(s) ** No errors detected
```

**Buffer**, with rescaling:
```
== RUNNING buffer_polygon using double
RESCALED - Expected: 2 error(s) ** 2 failures detected
== RUNNING buffer_multi_polygon using double
RESCALED - Expected: 3 error(s) ** 3 failures detected
== RUNNING buffer_linestring using double
RESCALED - Expected: 2 error(s) ** 2 failures detected
== RUNNING buffer_multi_linestring using double
RESCALED - Expected: 9 error(s) ** 8 failures detected
== RUNNING buffer_linestring_aimes using double
RESCALED - Expected: 0 error(s) ** No errors detected
```


awulkiew:
That's great!

What do you think about removing robust policy from all of the functions?

mloskot (approved):
LGTM

barendgehrels:
> That's great!
> 
> What do you think about removing robust policy from all of the functions?

Yes, that should be done at some point. But that disables all rescaling - which we indeed want, but I think we should not do that in this release yet. In next, or using the common deprecation period (was that 3 releases)?

Because this change influences behavior, customers might rely on it. If it unfortunately breaks something, they can easily mitigate that by setting the RESCALING flag explicitly and report the bug.

So my proposal is to wait with this for at least one release, or maybe even 3. (We can see in after next release what we get). Then we can remove all the robustness policies and the rescaling policy itself.

vissarion (approved):
Great, thanks!

barendgehrels:
> @barendgehrels Ok, I agree. Should we deprecate it now i.e. generate a `#warning` if rescaling is enabled?

Good idea, I’ll do that. And I’ve to update some of the docs too.

Great that we will switch now! Really cool.

include/boost/geometry/core/config.hpp
```diff
@@ -13,17 +13,27 @@
...
     #define BOOST_GEOMETRY_NO_ROBUSTNESS
   #endif
 #endif
 
+#if defined(BOOST_GEOMETRY_USE_RESCALING) && ! defined(BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE)
+    BOOST_PRAGMA_MESSAGE("Rescaling is deprecated and its functionality will be removed in Boost 1.82")
+#endif
+
+#if defined(BOOST_GEOMETRY_USE_RESCALING) && defined(BOOST_GEOMETRY_NO_ROBUSTNESS)
+    #error "Define either BOOST_GEOMETRY_NO_ROBUSTNESS (default) or BOOST_GEOMETRY_USE_RESCALING"
```
barendgehrels, line 34: We didn't have it before. It's similar to, for example:
https://github.com/boostorg/chrono/blob/develop/include/boost/chrono/config.hpp#L22

include/boost/geometry/strategy/spherical/envelope_range.hpp
```diff
@@ -165,9 +165,6 @@ inline bool pole_within(T const& lat_pole, Ring const& ring,
     }
 
     using point_t = typename geometry::point_type<Ring>::type;
-    using coord_t = typename geometry::coordinate_type<point_t>::type;
-    using units_t = typename geometry::detail::cs_angular_units<point_t>::type;
-    using constants_t = math::detail::constants_on_spheroid<coord_t, units_t>;
```
barendgehrels, line 170: This is unrelated, but it gave me warnings in many files I compiled.
So I thought it's better to get this in too on short term.

include/boost/geometry/core/config.hpp
```diff
@@ -13,17 +13,27 @@
...
+#if !defined(BOOST_GEOMETRY_NO_ROBUSTNESS) && !defined(BOOST_GEOMETRY_USE_RESCALING)
+  #if defined(BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE)
+    #define BOOST_GEOMETRY_USE_RESCALING
   #else
     #define BOOST_GEOMETRY_NO_ROBUSTNESS
   #endif
 #endif
 
+#if defined(BOOST_GEOMETRY_USE_RESCALING) && ! defined(BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE)
+    BOOST_PRAGMA_MESSAGE("Rescaling is deprecated and its functionality will be removed in Boost 1.82")
```
barendgehrels, line 30: Is this right? `1.79` is next, `1.80` is second, `1.81` is third so we can remove it in `1.82`?
Which is around a year.

awulkiew: I think it's ok.

barendgehrels:
Merged, thanks for your quick responses!

PR #981, opened 2022-02-23
[fix] sectionalize with const geometry

### LINKED ISSUE #980: sectionalize does not work for const points
Sectionalize uses a box internally. That box uses the `point_type` of the passed geometry.

If that geometry is `const` (no `set` methods in traits), it won't work.

For example: `typedef typename point_type<Polygon>::type point_type;` [here](https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/sections/sectionalize.hpp#L631)
This is passed to `sectionalize_range` and a box is kept.

There is more code where similar things happen, and which is why `boost::geometry::disjoint(a, b)` does not work for cases where `a` or `b` are correctly defined const geometries.

I'm working at sectionalize (because it's the first step), which can be fixed relatively quickly by passing extra a writable point type.

barendgehrels:
Fixes #980

For this algorithm, I use the `helper_geometry` in the test only, because the box-type should be writable anyway (it writes section boxes) and we can conveniently use that point type


include/boost/geometry/geometries/point_xy.hpp
```diff
@@ -48,7 +48,7 @@ template<typename CoordinateType, typename CoordinateSystem = cs::cartesian>
 class point_xy : public model::point<CoordinateType, 2, CoordinateSystem>
 {
 public:
-    /// \constructor_default_no_init
+    /// Constructor_default_no_init
```
barendgehrels, line 51: Unrelated

awulkiew: This was doxygen alias I think:
https://github.com/boostorg/geometry/blob/develop/doc/doxy/Doxyfile#L93

barendgehrels: > This was doxygen alias I think: https://github.com/boostorg/geometry/blob/develop/doc/doxy/Doxyfile#L93

Sorry. Thanks, it's fixed.

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
```diff
@@ -935,12 +942,16 @@ inline void sectionalize(Geometry const& geometry,
...
         <
             typename tag<Geometry>::type,
             Geometry,
             Reverse,
-            DimensionVector
+            DimensionVector,
+            point_type
         >::apply(geometry, robust_policy, sections,
                  strategy,
                  ring_id, max_count);
```
awulkiew, line 957: If `sections` are passed as function argument then you don't have to pass `point_type` as template parameter. `box_type` should rather be taken from the type of `sections` inside `apply()`. The number of template parameters should be minimal that is required for the dispatching to work. `point_type` is not needed for that. It's an additional information needed only in `apply()`.

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
```diff
@@ -628,10 +629,9 @@ struct sectionalize_polygon
                 ring_identifier ring_id,
                 std::size_t max_count)
     {
-        typedef typename point_type<Polygon>::type point_type;
         typedef sectionalize_range
             <
-                closure<Polygon>::value, Reverse, point_type, DimensionVector
+                closure<Polygon>::value, Reverse, Point, DimensionVector
```
awulkiew, line 634: So here instead of passing `Point` into `secionalize_range` you could remove this template parameter and instead inside `secionalize_range::apply()` take it from `Sections` as:
```
typename geometry::point_type<typename Sections::box_type>::type
```

awulkiew (changes requested):

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
barendgehrels: Yes, it’s true. I started in another way, which is why I added them.
But you are right, I can remove them now. I’ll do that. Thanks.

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
barendgehrels: :heavy_check_mark: done

I moved it to in `sectionalize_part` which makes it now simpler than it was before.

(though the diff might be now larger because I changed some `typedef` with `using`)

Thanks for your comment.

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
```diff
@@ -380,11 +378,7 @@ struct box_next_in_section<cartesian_tag>
 };
 
 /// @brief Helper class to create sections of a part of a range, on the fly
-template
-<
-    typename Point,
```
barendgehrels, line 385: So this `Point` is gone.

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
```diff
@@ -628,23 +624,21 @@ struct sectionalize_polygon
...
-        per_range::apply(exterior_ring(poly), robust_policy, sections,
+        sectionalizer::apply(exterior_ring(poly), robust_policy, sections,
                          strategy, ring_id, max_count);
 
         ring_id.ring_index++;
-        typename interior_return_type<Polygon const>::type
-            rings = interior_rings(poly);
-        for (typename detail::interior_iterator<Polygon const>::type
-                it = boost::begin(rings); it != boost::end(rings); ++it, ++ring_id.ring_index)
+        auto const& rings = interior_rings(poly);
```
barendgehrels, line 637: I think this is right, but I'm not 100% sure. 
Should it stay `interior_return_type` ?

awulkiew: `const&` is fine because even if `interior_rings()` returned proxy reference like `reference_wrapper` this object would live as long as the const reference `rings`.

You could also use `auto&&` which would be able to represent both true references and proxy references exactly as they were returned from `interior_rings()`. But since we don't need perfect forwarding here enforcing constness is probably better.

awulkiew (approved):

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
```diff
@@ -160,16 +160,14 @@ template
 >
 struct get_direction_loop
 {
-    typedef typename util::sequence_element<Index, DimensionVector>::type dimension;
+    using dimension = typename util::sequence_element<Index, DimensionVector>::type;
 
     template <typename Segment>
     static inline void apply(Segment const& seg,
                 int directions[Count])
```
vissarion, line 167: this is unrelated but you can also fix the indentation here.

vissarion (approved):
LGTM thanks

include/boost/geometry/algorithms/detail/sections/sectionalize.hpp
barendgehrels: Yes, 1-tab it should have been I think.

`clang-format` is for 90% nice indeed. But the last 10% is awful, from all what I've seen, and it destroys readability which we carefully wrote everywhere (nearly).

I'm quite sensible to code layout and we try to do it carefully with optimum readability. Such as this one

https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/srs/projections/impl/dms_parser.hpp#L249-L257

If you handle this with clang-format, it's destroyed. The same for streaming:

https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/simplify.hpp#L133-L151

This is also destroyed by clang-format.

But clang-format has many options, and I'm not an expert. If you have a configuration which works well in all cases, we can consider that.

PR #986, opened 2022-03-16
Fix const disjoint

### LINKED ISSUE #985: disjoint does not work for const points
A disjoint of two geometries of which both points are const (or one of them) does not work, because internally those point types are used to calculate turns

barendgehrels:
Fixes #985 

Some pieces of the touched code become a bit simpler by this (for example avoiding an extra conversion).

There are some renamings as well, such as `get_turns` (this was really inconvenient, there is a namespace and a free function as well - and this does something different - so I call it `turn_retriever` now) and `define_turn_operation_type` i/o `turn_operation_type` (because that is there as well).

Also there are some white space changes (at end of lines), fixed by my editor.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
@@ -926,24 +926,24 @@ struct get_turn_info_type<Geometry1, Geometry2, AssignPolicy, Tag1, Tag2, linear
     : overlay::get_turn_info_linear_areal<AssignPolicy>
 {};
 
-template <typename Geometry1, typename Geometry2, typename SegmentRatio,
+template <typename Geometry1, typename Geometry2, typename Point, typename SegmentRatio,
           typename Tag1 = typename tag<Geometry1>::type, typename Tag2 = typename tag<Geometry2>::type,
           typename TagBase1 = typename topological_tag_base<Geometry1>::type, typename TagBase2 = typename topological_tag_base<Geometry2>::type>
 struct turn_operation_type
 {
-    typedef overlay::turn_operation<typename point_type<Geometry1>::type, SegmentRatio> type;
+    using type = overlay::turn_operation<Point, SegmentRatio>;
 };
 
-template <typename Geometry1, typename Geometry2, typename SegmentRatio, typename Tag1, typename Tag2>
-struct turn_operation_type<Geometry1, Geometry2, SegmentRatio, Tag1, Tag2, linear_tag, linear_tag>
+template <typename Geometry1, typename Geometry2, typename Point, typename SegmentRatio, typename Tag1, typename Tag2>
+struct turn_operation_type<Geometry1, Geometry2, Point, SegmentRatio, Tag1, Tag2, linear_tag, linear_tag>
 {
-    typedef overlay::turn_operation_linear<typename point_type<Geometry1>::type, SegmentRatio> type;
+    using type = overlay::turn_operation_linear<Point, SegmentRatio>;
 };
 
-template <typename Geometry1, typename Geometry2, typename SegmentRatio, typename Tag1, typename Tag2>
-struct turn_operation_type<Geometry1, Geometry2, SegmentRatio, Tag1, Tag2, linear_tag, areal_tag>
+template <typename Geometry1, typename Geometry2, typename Point, typename SegmentRatio, typename Tag1, typename Tag2>
+struct turn_operation_type<Geometry1, Geometry2, Point, SegmentRatio, Tag1, Tag2, linear_tag, areal_tag>
 {
-    typedef overlay::turn_operation_linear<typename point_type<Geometry1>::type, SegmentRatio> type;
+    using type = overlay::turn_operation_linear<Point, SegmentRatio>;
 };
 
 }} // namespace detail::get_turns
--- a/include/boost/geometry/algorithms/detail/relate/boundary_checker.hpp
+++ b/include/boost/geometry/algorithms/detail/relate/boundary_checker.hpp
@@ -17,6 +17,8 @@
 #include <boost/geometry/algorithms/detail/sub_range.hpp>
 #include <boost/geometry/algorithms/num_points.hpp>
 
+#include <boost/geometry/geometries/helper_geometry.hpp>
+
 #include <boost/geometry/policies/compare.hpp>
 
 #include <boost/geometry/util/has_nan_coordinate.hpp>
@@ -41,7 +43,8 @@ class boundary_checker {};
 template <typename Geometry, typename Strategy>
 class boundary_checker<Geometry, Strategy, linestring_tag>
 {
-    typedef typename point_type<Geometry>::type point_type;
+    using mutable_point_type
+        = typename helper_geometry<typename point_type<Geometry>::type>::type;
 
 public:
     boundary_checker(Geometry const& g, Strategy const& s)
@@ -55,8 +58,8 @@ public:
         , m_strategy(s)
     {}
 
-    template <boundary_query BoundaryQuery>
-    bool is_endpoint_boundary(point_type const& pt) const
+    template <boundary_query BoundaryQuery, typename Point>
+    bool is_endpoint_boundary(Point const& pt) const
     {
         boost::ignore_unused(pt);
 #ifdef BOOST_GEOMETRY_DEBUG_RELATE_BOUNDARY_CHECKER
@@ -85,7 +88,8 @@ private:
 template <typename Geometry, typename Strategy>
 class boundary_checker<Geometry, Strategy, multi_linestring_tag>
 {
-    typedef typename point_type<Geometry>::type point_type;
+    using point_type = typename point_type<Geometry>::type;
+    using mutable_point_type = typename helper_geometry<point_type>::type;
 
 public:
     boundary_checker(Geometry const& g, Strategy const& s)
@@ -94,42 +98,35 @@ public:
 
     // First call O(NlogN)
     // Each next call O(logN)
-    template <boundary_query BoundaryQuery>
-    bool is_endpoint_boundary(point_type const& pt) const
+    template <boundary_query BoundaryQuery, typename Point>
+    bool is_endpoint_boundary(Point const& pt) const
     {
-        typedef geometry::less<point_type, -1, typename Strategy::cs_tag> less_type;
+        using less_type = geometry::less<mutable_point_type, -1, typename Strategy::cs_tag>;
 
-        typedef typename boost::range_size<Geometry>::type size_type;
-        size_type multi_count = boost::size(m_geometry);
+        auto const multi_count = boost::size(m_geometry);
 
         if ( multi_count < 1 )
+        {
             return false;
+        }
 
         if ( ! m_is_filled )
         {
             //boundary_points.clear();
             m_boundary_points.reserve(multi_count * 2);
 
-            typedef typename boost::range_iterator<Geometry const>::type multi_iterator;
-            for ( multi_iterator it = boost::begin(m_geometry) ;
-                  it != boost::end(m_geometry) ; ++ it )
+            for (auto it = boost::begin(m_geometry); it != boost::end(m_geometry); ++it)
             {
-                typename boost::range_reference<Geometry const>::type
-                    ls = *it;
+                auto const& ls = *it;
 
-                // empty or point - no boundary
+                // empty or only one point - no boundary
                 if (boost::size(ls) < 2)
                 {
                     continue;
                 }
 
-                typedef typename boost::range_reference
-                    <
-                        typename boost::range_value<Geometry const>::type const
-                    >::type point_reference;
-
-                point_reference front_pt = range::front(ls);
-                point_reference back_pt = range::back(ls);
+                auto const& front_pt = range::front(ls);
+                auto const& back_pt = range::back(ls);
 
                 // linear ring or point - no boundary
                 if (! equals::equals_point_point(front_pt, back_pt, m_strategy))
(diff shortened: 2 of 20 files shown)


include/boost/geometry/algorithms/detail/sections/section_functions.hpp
```diff
@@ -148,11 +149,11 @@ template
     typename Box,
     typename RobustPolicy
 >
-static inline bool exceeding(int dir,
-                             Point const& point,
-                             Box const& point_box,
-                             Box const& other_box,
-                             RobustPolicy const& robust_policy)
+inline bool exceeding(int dir,
```
barendgehrels, line 152: These two are unrelated but should not be declared as `static`

test/algorithms/relate/relate_const.cpp
```diff
@@ -0,0 +1,63 @@
...
+    namespace bg = boost::geometry;
+
+    BOOST_CHECK_EQUAL(exp_cov, bg::covered_by(geometry1, geometry2));
+    BOOST_CHECK_EQUAL(exp_crosses, bg::crosses(geometry1, geometry2));
+    BOOST_CHECK_EQUAL(exp_disjoint, bg::disjoint(geometry1, geometry2));
+    BOOST_CHECK_EQUAL(exp_intersects, bg::intersects(geometry1, geometry2));
+    BOOST_CHECK_EQUAL(exp_overlaps, bg::overlaps(geometry1, geometry2));
+    BOOST_CHECK_EQUAL(exp_touches, bg::touches(geometry1, geometry2));
+    BOOST_CHECK_EQUAL(exp_relate, bg::relate(geometry1, geometry2, bg::de9im::mask("F0F******")));
+    BOOST_CHECK_EQUAL(exp_within, bg::within(geometry1, geometry2));
```
barendgehrels, line 39: So it fixes not only disjoint, but many boolean relation operators

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
```diff
@@ -926,24 +926,24 @@ struct get_turn_info_type<Geometry1, Geometry2, AssignPolicy, Tag1, Tag2, linear
     : overlay::get_turn_info_linear_areal<AssignPolicy>
 {};
 
-template <typename Geometry1, typename Geometry2, typename SegmentRatio,
+template <typename Geometry1, typename Geometry2, typename Point, typename SegmentRatio,
           typename Tag1 = typename tag<Geometry1>::type, typename Tag2 = typename tag<Geometry2>::type,
           typename TagBase1 = typename topological_tag_base<Geometry1>::type, typename TagBase2 = typename topological_tag_base<Geometry2>::type>
-struct turn_operation_type
+struct define_turn_operation_type
```
awulkiew, line 932: What is the reasoning behind this change? Verbs are typically used for functions or metafunctions semantically modifying a type. Consider that `point_type` is not named `define_point_type`.

include/boost/geometry/algorithms/detail/relate/turns.hpp
```diff
@@ -40,21 +40,20 @@ struct assign_policy
...
     typename Geometry1,
     typename Geometry2,
+    typename Point,
     typename GetTurnPolicy = detail::get_turns::get_turn_info_type
         <
             Geometry1, Geometry2, assign_policy<>
         >
 >
-struct get_turns
+struct turn_retriever
```
awulkiew, line 55: Why was this name changed? What was the original problem? That it's confused with the original `get_turns`?

Anyway this is semantically not a class which object is created somewhere like a functor/predicate whatever (it's not a retriever created and used somewhere) but implementation of an algorithm with static member function `apply()`. We use verbs for these in the library right?

BTW, if I was implementing this now it'd be free function `detail::relate::get_turns()`.

awulkiew:
It looks good to me in general but you mixed fixing the "const" issue with renaming of interfaces. This is out of scope of this PR potentially interferes with my work on the support of GCs in relops (I haven't checked how much). Not to mention I think the names are incorrect as I mentioned in the comments.

I also don't understand the confusion. `turns::get_turns::apply()` calls `get_turns` algorithm with some of the parameters defined. Would it be more intuitive for you if it was a free function `relate::get_turns()`?

include/boost/geometry/algorithms/detail/relate/boundary_checker.hpp
```diff
@@ -94,42 +98,33 @@ class boundary_checker<Geometry, Strategy, multi_linestring_tag>
...
 
         if ( ! m_is_filled )
         {
             //boundary_points.clear();
             m_boundary_points.reserve(multi_count * 2);
 
-            typedef typename boost::range_iterator<Geometry const>::type multi_iterator;
-            for ( multi_iterator it = boost::begin(m_geometry) ;
-                  it != boost::end(m_geometry) ; ++ it )
+            for (auto const& ls : m_geometry)
```
awulkiew, line 118: This is incorrect. MultiLinestring has to be Boost.Range not STD range.

In general we cannot use range based for loop for a geometry that can be passed by the user, only for geometries and containers that we create ourselves internally.

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
```diff
@@ -1299,9 +1289,21 @@ struct linear_areal
...
+        bool calculate_from_inside_sides(Pi const& pi, Pj const& pj, Pk const& pk,
+                                         Qi const& qi, Qj const& qj, Qk const& qk,
+                                         Strategy const& strategy)
         {
+            la_side_calculator
+            <
+                Strategy,
+                Pi, Pj, Pk,
+                Qi, Qj, Qk
+            > side_calc(pi, pj, pk, qi, qj, qk, strategy);
```
awulkiew, line 1305: If that's the preferred way then I suggest to remove `la_side_calculator` entirely and replace function calls like `pk_wrt_p()` with corresponding strategy calls.

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
```diff
@@ -1255,21 +1252,14 @@ struct linear_areal
...
             {
-                point2_type new_qj;
-                geometry::convert(turn.point, new_qj);
-
-                la_side_calculator<cs_tag, point1_type, point2_type, Strategy>
-                    side_calc(qi_conv, new_pj, pi, qi, new_qj, qj, strategy);
-
-                return calculate_from_inside_sides(side_calc);
+                // Calculate sides with different points for P and Q
+                return calculate_from_inside_sides(qi, turn.point, pi, qi, turn.point, qj, strategy);
```
awulkiew, line 1262: AFAIR the conversion was done because sometimes point type of one geometry is used to create a segment with the point type of another geometry. But AFAIS side strategies are able to take three different points so I think it should work but is it tested?

awulkiew:
As I mentioned in one of the previous PRs related to const points this issue is confusing to me. I'd like to do some testing myself because I didn't get an answer then. I suspect that the cause of this issue is somewhere else. The changes made in this PR should be irrelevant.

Furthermore this is not about const points but about points defining only `get()`. You test point can be assigned to, it defines copy ctor and assignment operator, etc. It's not about copying. It is possible to copy it or store in a container. You couldn't use it in a linestring if it was otherwise.

There is also a separate issue which is that your test point doesn't define default ctor. This also has nothing to do with constness. It's a different issue.

include/boost/geometry/algorithms/detail/relate/turns.hpp
barendgehrels: See the PR's description: `get_turns` is a free function which does something else.

barendgehrels:
> It looks good to me in general but you mixed fixing the "const" issue with renaming of interfaces. This is out of scope of this PR potentially interferes with my work on the support of GCs in relops (I haven't checked how much). Not to mention I think the names are incorrect as I mentioned in the comments.
> 

Yes, I could have split it - but then you would see a name change and right after a PR with a contents change. I thought there was no problem in one go for lines touched anyway.

I don't think it will interfere (a lot), this is not specific for any geometry type. Apart from that: with splitted PR's it interfers even more.

include/boost/geometry/algorithms/detail/relate/boundary_checker.hpp
barendgehrels: All right I will revert this. Thanks.

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
barendgehrels: I don't want to make the PR bigger. I don't say it's the preferred way - it is a more convenient way than before, because it does not need conversions.

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
barendgehrels: Yes, side strategies should take three different point types.

barendgehrels:
> 

I answered your point in the previous PR. See the file `const_point.hpp` defining such a point type, and indeed: with no setters and no default constructor. It's exactly this type that I encountered in practice. And you could implement first one (with default constructor) and then this one (without default constructor) but what's the point... It touches nearly the same code.

Maybe the name is not perfect, I mean a point type which no setters and no default constructor (effectively it can't be changed after construction, therefore a `const_point` is not that bad. It is a `non_mutable_point`

include/boost/geometry/algorithms/detail/relate/boundary_checker.hpp
barendgehrels: :heavy_check_mark:

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
awulkiew: What different `turn_operation_type` are you talking about? I tried to find it but I only see typedefs. AFAIU this is the only `struct` of this name. So if my understanding is correct this is the same case as it is with `geometry::point_type<>::type` and `typedef ... point_type`. Or am I missing something?

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
awulkiew: Ok got it.

include/boost/geometry/algorithms/detail/relate/turns.hpp
awulkiew: Yes, `geometry::get_turns()` is a free function, `dispatch::get_turns` is a struct and `detail::get_turns_generic` is a struct as well (the implementation). `detail::relate::turns::get_turns` is an implementation as well. The only difference between `geometry::get_turns()` and `relate::turns::get_turns::apply()` is that parameters common across all relops and L/L setops are set here internally so the caller is not forced to pass them each time the same way. As I mentioned it could as well be implemented as a free function. It's neither a retriever nor a getter it's an implementation of an algorithm `get_turns()` specific for relops.

awulkiew:
I get your point now. So there are several issues addressed by your recent PRs which may be confusing:
1. support for point types defining only getters
2. support for point types without default constructor
3. refactoring

1 and 2 are often used together in the library because of the semantics of some utilities used in the library, e.g. `point_on_border` or `envelope` which requires to create an object with default ctor and then to set the coordinates. So it indeed makes sense to address these issues together.

I'd only refrain from doing refatoring at the same time because this would limit the scope of the PR. E.g. now instead of simply talking about the important issues 1 and 2 we have to also talk about naming etc.

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
barendgehrels: Yes, it’s probably the only struct indeed. All right, I will rename it back, it’s easier than it was to make it unique.

include/boost/geometry/algorithms/detail/relate/turns.hpp
barendgehrels: I will also rename this back.

barendgehrels:
> I get your point now. So there are several issues addressed by your recent PRs which may be confusing:
> 
> 1. support for point types defining only getters
> 2. support for point types without default constructor
> 3. refactoring
> 
> 1 and 2 are often used together in the library because of the semantics of some utilities used in the library, e.g. `point_on_border` or `envelope` which requires to create an object with default ctor and then to set the coordinates. So it indeed makes sense to address these issues together.
> 
> I'd only refrain from doing refatoring at the same time because this would limit the scope of the PR. E.g. now instead of simply talking about the important issues 1 and 2 we have to also talk about naming etc.

Thank you. So we go for 1 and 2. About 3: I understand and will rename it back. And I will not create a follow-up, because apparently the new names are not so easy to choose, so it’s fine then. It was still useful for me in the process of adding the extra template parameter(s) there.

barendgehrels:
:heavy_check_mark: renamed back go `turn_operation_type` and `get_turns`

awulkiew (approved):

awulkiew:
Thanks for the fix and the discussion!

Btw, would you be ok with changing `struct get_turns` to free function `detail::relate::get_turns()`? Would it be more intuitive or would you still prefer a different name to distinguish it from `geometry::get_turns()`?

barendgehrels:
> Thanks for the fix and the discussion!
> 
> Btw, would you be ok with changing `struct get_turns` to free function `detail::relate::get_turns()`? Would it be more intuitive or would you still prefer a different name to distinguish it from `geometry::get_turns()`?

Actually the names are fine, you convinced me. ;-)

A free function would be fine too, for me, unless it will depend on SFINAE.

vissarion:
Sorry I was super late here! It look ok to me, though. Thanks!

Issue #982, opened 2022-02-26
disjoint of ring/box can give errors


barendgehrels:
See next code:
```
void test_disjoint_box_ring()
{
  using Point = boost::geometry::model::point<double, 2, boost::geometry::cs::cartesian>;
  using Ring = boost::geometry::model::ring<Point, true, false>;
  using Box = boost::geometry::model::box<Point>;

  Ring ring;
  boost::geometry::append(ring, Point{139.655, 35.346});
  boost::geometry::append(ring, Point{139.618, 35.342});
  boost::geometry::append(ring, Point{139.623, 35.358});
  boost::geometry::correct(ring);

  Box box{{139.62, 35.34}, {139.63, 35.35}};

  std::cout << "disjoint ring/box: " << std::boolalpha << boost::geometry::disjoint(ring, box) << std::endl;

  Ring ring_of_box;
  {
    // Convert box-to-ring

    // This code does compile but does not work:
    //boost::geometry::convert(box, ring);

    // So use WKT, which works
    std::ostringstream out;
    out << boost::geometry::wkt(box);
    boost::geometry::read_wkt(out.str(), ring_of_box);
    boost::geometry::correct(ring_of_box);
    std::cout << boost::geometry::area(box) << " " << boost::geometry::area(ring_of_box) << std::endl;
  }
  
  std::cout << "ring-of-box: " << boost::geometry::wkt(ring_of_box) << std::endl;

  std::cout << "disjoint ring/ring: " << boost::geometry::disjoint(ring, ring_of_box) << std::endl;

  {
    std::vector<Ring> intersections;
    boost::geometry::intersection(ring, box, intersections);
    std::cout << "intersections ring/box size: " << intersections.size() << std::endl;
    for (const auto &p : intersections)
    {
      std::cout << "Intersection ring/box: " << boost::geometry::area(p) << std::endl;
    }
  }  
  {
    std::vector<Ring> intersections;
    boost::geometry::intersection(ring, ring_of_box, intersections);
    std::cout << "intersections ring/ring size: " << intersections.size() << std::endl;
    for (const auto &p : intersections)
    {
      std::cout << "Intersection ring/ring: " << boost::geometry::area(p) << std::endl;
    }
  }  
}
```
Delivering:
```
disjoint ring/box: true
0.0001 0.0001
ring-of-box: POLYGON((139.62 35.34,139.62 35.35,139.63 35.35,139.63 35.34))
disjoint ring/ring: false
intersections ring/box size: 1
[...]

Issue #837, opened 2021-03-29
intersects(polygon, box) returns the wrong result


vschoech:
This problem has been reported earlier and it **was fixed in 1.50** already:
https://lists.boost.org/geometry/2012/06/1957.php

It has now returned in boost.geometry 1.75.

My `tc::geo::polygon` type is actually a multi-polygon, using a polygon type that is based on **int**, oriented **counter-clockwise** and **open** (not closed). I am using **boost 1.75.0**. Please consider the following example: 

```C++
tc::geo::polygon<int> polygonA;
boost::geometry::read_wkt("MULTIPOLYGON(((1992 3240,1992 1440,3792 1800,3792 3240,1992 3240)))", polygonA); // does not throw
boost::geometry::is_valid(polygonA); // returns true

tc::geo::polygon<int> polygonB;
boost::geometry::read_wkt("MULTIPOLYGON(((1941 2066,2055 2066,2055 2166,1941 2166,1941 2066)))", polygonB); // does not throw
boost::geometry::is_valid(polygonB); // returns true

bool const bPolyPoly = boost::geometry::intersects(polygonA, polygonB); // returns true (correct)

// the same shape as polygonB, but type box/rect
tc::geo::rect<int> const rectB = Box(1941;2066;2055;2166); // pseudo code; not using WKT here b/c WKT represents boxes as polygons which is confusing in this context 

bool const bPolyBox = boost::geometry::intersects(polygonA, rectB); // returns false (WRONG)
```

vschoech:
P.S. Please add my reproduction as a unit test to your project to ensure that it doesn't break again! :)

ayushgupta138:
I can reproduce the error mentioned above. 
After some code reading, I found that the function `boost::geometry::all_segments_of()` 
https://github.com/boostorg/geometry/blob/e098a85f5adb5bbf28422a6415c6bf5ad601840a/include/boost/geometry/algorithms/detail/disjoint/areal_areal.hpp#L115
is not performing in the desired way.
https://github.com/boostorg/geometry/blob/e098a85f5adb5bbf28422a6415c6bf5ad601840a/include/boost/geometry/algorithms/for_each.hpp#L275
The function `fe_segment_range_with_closure<open>::apply()` invokes the functor once instead of invoking the functor for all possible segments in an open polygon.
I think adding a return statement before
[...]

PR #990, opened 2022-04-08
Check return value from fe_segment_range_with_closure in for_each.hpp

Mitsuhiko-Matsukawa:
#982


Mitsuhiko-Matsukawa:
#837

awulkiew:
Thanks for the fix!

awulkiew (approved):

PR #991, opened 2022-04-11
Fix c++20 compilation errors related to conditional expression

stefan301:
Fix c++20 compilation errors related to conditional expression with int and floating_point_type

A floating_point_type with implicit conversions can lead to this C++20 compilation error:

boost\geometry\policies\robustness\segment_ratio.hpp(232,1): error C2445: result type of conditional expression is ambiguous: types 'int' and '<floating_point_type>' can be converted to multiple common types

### DIFF
--- a/include/boost/geometry/policies/robustness/segment_ratio.hpp
+++ b/include/boost/geometry/policies/robustness/segment_ratio.hpp
@@ -225,7 +225,7 @@ public:
         }
 
         m_approximation =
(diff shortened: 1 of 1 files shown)


awulkiew:
Thanks!

awulkiew (approved):

barendgehrels (approved):
Thanks!

Issue #988, opened 2022-04-05
Buffer Issue


danston:
When running `boost::geometry::buffer` on the given input set (see the Figure below), the result is not valid. The input seems to be quite reasonable without complex parts. Moreover, other similar inputs succeed. I have created a simple minimal example in [this gist](https://gist.github.com/danston/fbdcb04e9f92114995a62c31937dce82) to illustrate the issue. The last two points in the input are duplicates but even when removing them and having an open polyline, the code still fails to produce the valid result. Could you please clarify if this is a bug or I do something wrong here? Thank you in advance!

Tested with: Boost 1.77 and Boost 1.78.

![input](https://user-images.githubusercontent.com/8123586/161758623-21cd4b41-a25c-43c2-8de1-3926285915b9.png)

barendgehrels:
Thanks, I can reproduce it and it is indeed wrong (also in 1.79).
Will work on it.

![image](https://user-images.githubusercontent.com/334849/162464694-7a8ee6b0-60d6-41bb-916d-939ad69342ce.png)

barendgehrels:
> Could you please clarify if this is a bug or I do something wrong here? 

Your code is right. It's a bug. It internally classifies to many intersection points as acceptable and therefore creates an additional (wrong) path.

danston:
@barendgehrels Thanks a lot for looking into this issue!

PR #993, opened 2022-04-13
[buffer] fix min distance boundary

barendgehrels:
This fixes issue #988

There is one regression, an invalid case:

image

I explain it in more detail here.

Turns 5 and 6 (inside the yellow circle, denoted as `5 1` and `6 1`) are tricky.

The calculated intersection point is **right of** the line from which the intersection is calculated. This can happen occasionally. Therefore both points are inside the "other piece" (it is about pieces 3 and 12 marked by the green circle, and these pieces are identical because the linestring is invalid and has a duplicate part).
I verified it using the robust-side of @tinko92  and indeed, it gives "right of" in both cases. Later it's calculated by the distance-measure and that one gives a measure of `8.88178e-16`.

For this reason I had that threshold of `1.0e-4` (now gone, in this PR, for non-rescaling). But that value is way too much. A value of for example `1.0e-15` works for all cases. But I like to get rid of that threshold at all.

In more general terms: we calculate an intersection P on two lines A and B. Due to FP-precision, P can be located left, on top or right of A and left, on top or right of B. Ideally P is on top of both A and B, of course. But it can be a little-bit off. If we later need the side-information, we get left (1), right (-1) or on-segment (0). Also there is a small gray area for which we have different approaches. Ideally, the two strategies intersection and side should correspond and cooperate, such that if the intersection is calculated, and the side is verified later, it is always 0 for both of the lines. As far as I know we never researched it. Probably a nice challenge (@tinko92 ?)

Because of this reason, this problem will happen more frequently for invalid cases (with duplicate pieces and intersection points).

There is another solution as well: if the intersection points of intersections piece 0/3 and 0/12 are both inside the other (12 and 3), we can detect that and set both of them as "traversable". That would require keeping additional information and using this for detection.

Solutions:
* accept this regression, it's an invalid case anyway
* use a threshold but smaller
* revise intersection/side such that they cooperate (in terms of thresholds) and this never occurs (I don't know if that is possible)
* detect mutual turn-in-piece occurances with more administration
* revise buffer

(Revise buffer would be huge: an alternative would be to generate buffers for smaller geometry-parts and union their results - I tried it long ago but maybe things have been changed - and maybe it might be convenient for the geographic buffer as well - the final code would probably be simpler)

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
@@ -180,7 +180,12 @@ public:
             // unit tests of hard cases start to fail (5 in multi_polygon)
             // But it is acknowlegded that such a threshold depends on the
             // scale of the input.
+#if defined(BOOST_GEOMETRY_USE_RESCALING)
             if (state.m_min_distance > 1.0e-4 || ! state.m_close_to_offset)
+#else
+            constexpr double zero = 0;
+            if (math::larger(state.m_min_distance, zero) || ! state.m_close_to_offset)
+#endif
             {
                 Turn& mutable_turn = m_turns[turn.turn_index];
                 mutable_turn.is_turn_traversable = false;
--- a/test/algorithms/buffer/buffer_linestring.cpp
+++ b/test/algorithms/buffer/buffer_linestring.cpp
@@ -108,6 +108,7 @@ static std::string const mysql_25662426a = "LINESTRING(170 4756, 168 4756, 168 4
 
 static std::string const issue_596 = "LINESTRING(292979.660 6688731.370, 292979.600 6688733.420, 292979.540 6688735.440, 292979.540 6688735.500)";
 static std::string const issue_803 = "LINESTRING(2773.6899360413681 -17.493356405074767,2767.9084041267429 -9.0288291122330797,2765.7403296587586 -1.6929054585683376,2765.3273630934282 6.6613381477509392e-16,2762.8495637014462 10.157432751410026,2759.9587977441333 24.829280059133669,2756.3453402974928 38.372523727680374,2752.0091913615238 45.708447381345117)";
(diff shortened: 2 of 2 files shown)


barendgehrels:
The reported case, a linestring buffer

**With fix:**

image

**Without fix:**

image

You see that the case without the fix has many green (traversable) turns because the scale of the reported linstring was quite small and therefore the threshold of `1.0e-4` was unfortunate.

tinko92:
> In more general terms: we calculate an intersection P on two lines A and B. Due to FP-precision, P can be located left, on top or right of A and left, on top or right of B. Ideally P is on top of both A and B, of course. But it can be a little-bit off. If we later need the side-information, we get left (1), right (-1) or on-segment (0). Also there is a small gray area for which we have different approaches. Ideally, the two strategies intersection and side should correspond and cooperate, such that if the intersection is calculated, and the side is verified later, it is always 0 for both of the lines. As far as I know we never researched it. Probably a nice challenge (@tinko92 ?)

Thanks for the mention, @barendgehrels .

This might be a little tricky but I think there are two practical solutions for what you ask. I am no expert, though, so there might be something better. The following ignores floating-point underflow/overflow which are probably very uncommon:

* Solution 1: The side strategy https://gist.github.com/tinko92/5f242339ae312f66206f2be69184be72 with an intersection strategy that computes P's coordinates up to the machine ε using arbitrary precision arithmetic.
    * Pro: Most precise for a given coordinate type
    * Con: every intersection computation will be expensive (I think at least an order of magnitude slower than normally, possibly dominating the runtime of the algorithm that uses intersections even if they didn't before)
* Variation: Solution 1 but replace the machine ε with a tunable relative error 𝛿 and use a fast error bound test in the intersection strategy to determine, whether the results of the usual intersection formula satisfy this relative error. If they don't, fall back to arbitrary precision arithmetic. Intuitively I think something like 10^4 * ε could be a good choice for 𝛿.
    * Pro: probably only slightly slower than the simple intersection strategy for non-degenerate inputs (I would expect something like 10-20% slower intersection computations on average)
    * Con: Less precision for P, slightly more complicated to implement

Limitations for both suggestions:
* The side strategy is not exact so there are still other inconsistencies possible, like saying some a,b,c and a,b,d are collinear but a,c,d aren't.
* If there are 2nd order constructions, e.g. intersection point P is then used to calculate another point P2 that should also be collinear with A, the relative error will increase and then it might be inconsistent again, e.g. P2 could be found not collinear with A.

Both solutions are easy to implement if an added dependency on some arbitrary-precision library is acceptable. Solution 2 is slightly more involved because a good error bound test needs to be derived. If no added dependency is acceptable, the implementation would be more complicated.

For completeness and in case I made a mistake, here are the options that imho don't work with reasoning: https://gist.github.com/tinko92/d6c5f966bcdbe8f2b7f4ea4e0331ecf0

barendgehrels:
> Both solutions are easy to implement if an added dependency on some arbitrary-precision library is acceptable. Solution 2 is slightly more involved because a good error bound test needs to be derived. If no added dependency is acceptable, the implementation would be more complicated.
> 

Super, thanks a lot @tinko92 !! I will look at it in more detail and try your suggestions next Tuesday.

barendgehrels:
>     * Solution 1: The side strategy https://gist.github.com/tinko92/5f242339ae312f66206f2be69184be72 with an intersection strategy that computes P's coordinates up to the machine ε using arbitrary precision arithmetic.
>       


For the reference I add a link to where you explained the side strategy related to this before: #952

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
```diff
@@ -180,7 +180,12 @@ class turn_in_piece_visitor
             // unit tests of hard cases start to fail (5 in multi_polygon)
             // But it is acknowlegded that such a threshold depends on the
             // scale of the input.
+#if defined(BOOST_GEOMETRY_USE_RESCALING)
             if (state.m_min_distance > 1.0e-4 || ! state.m_close_to_offset)
+#else
+            constexpr double zero = 0;
```
awulkiew, line 186: Shouldn't this be of the same type as `state.m_min_distance`?

barendgehrels: Indeed. I'm considering an alternative fix (that takes a bit longer), but otherwise I'll change it.

vissarion:
@barendgehrels do you need a review of the code in this PR? I understand that you are about to do some changes (as proposed by Tinko) so in that case it is better to do a review afterwards. Sorry if I misunderstood and this PR is waiting for me...

barendgehrels:
> @barendgehrels do you need a review of the code in this PR? I understand that you are about to do some changes (as proposed by Tinko) so in that case it is better to do a review afterwards. Sorry if I misunderstood and this PR is waiting for me...

Indeed, it's fine to wait. No problem.

PR #994, opened 2022-04-19
[side] added side_rounded_input and robustness test

barendgehrels:
Add algorithm created by @tinko92 , including a unit test.

The robustness test is not committed anymore in this PR (wip).

The next PR (for slopes) will use this side algorithm.

Old description:
---------------------------------------------------------------------------------------------------

Add robustness test for comparing intersection with sides, for the problem described in #993 

This test generates random segments and calculate the intersection points between them (if any).

If there is one, it uses a side calculation to verify if the intersection point is determined as on-both-segments.

Like this:
image

It uses many side strategies we have (for cartesian) plus the new one of @tinko92 

The new one seems to be the best, by far:
* it returns nearly always 0 (correct) for the intersection point being on the two lines
* if an additional point is generated near the intersection point (bisecting_point), at a distance of several epsilons, it is determined as left or right (correct) of the two lines. That of course depends on the epsilon.

So this side strategy seems the best to be used in the `buffer` `turn_in_ring_winding` algorithm.

Herewith a sample output:

```
$ ./strategies/intersection_points_and_sides --count 1000000  --size 100 --multiplier 1.0e6 --type d

Test configuration:
  - Debug mode
  - No rescaling
  - Default test type: d

Results  type: d
  intersections: 231133
  errors (triangle): 223186 0
  errors (non robust): 224080 0
  errors (side robust fp 3): 231133 0
  errors (side robust eq 3): 230716 0
  errors (rounded): 48 2
  distance (avg): 2.6233714731400201749454688750541e-13
  distance (avg, rnd): 1.4256091165024009451598202213168e-16
  time: 3.558
```

As you can see, the errors using all the strategies we have are nearly as large as the samples. That means it is too precise (for this purpose) and always determines the intersection point either left or right of one (or both) of the two input lines. Which is inconvenient.
For the extra point they determine it as correct (related to previous conclusion). But if that point goes closer, the results are still not good:
```
  intersections: 231171
  errors (triangle): 223193 71469
  errors (non robust): 224073 71801
  errors (side robust fp 3): 231171 0
  errors (side robust eq 3): 230788 14626
  errors (rounded): 44 231171
```
So the strategy of @tinko92 is the only one where the intersection point is considered as ON both segments (nearly always) and if the point moves a bit away (about 1.0e5 epsilons) from it, usually considered as NOT on both segments.

I think this will fix at least buffer, and it explains why my previous (short) attempt to just replace it was not yet successful (I don't think I reported that yet), it is currently too different from the other strategies.

Also, it's less sensible to calculation type or field size.

To be continued.

### DIFF
--- a/include/boost/geometry/strategies/cartesian/side_rounded_input.hpp
+++ b/include/boost/geometry/strategies/cartesian/side_rounded_input.hpp
@@ -1,0 +1,57 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+
+// Copyright (c) 2021 Tinko Bartels, Berlin, Germany.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_STRATEGY_CARTESIAN_SIDE_ROUNDED_INPUT_HPP
+#define BOOST_GEOMETRY_STRATEGY_CARTESIAN_SIDE_ROUNDED_INPUT_HPP
+
+
+#include <boost/geometry/core/config.hpp>
+
+#include <boost/geometry/strategies/side.hpp>
+
+#include <boost/geometry/util/select_calculation_type.hpp>
+
+namespace boost { namespace geometry
+{
+
+namespace strategy { namespace side
+{
+
+template <typename CalculationType = void, int Coeff1 = 5, int Coeff2 = 32>
+struct side_rounded_input
+{
+    using cs_tag = cartesian_tag;
+
+    template <typename P1, typename P2, typename P>
+    static inline int apply(P1 const& p1, P2 const& p2, P const& p)
+    {
+        using coor_t = typename select_calculation_type_alt<CalculationType, P1, P2, P>::type;
+
+        coor_t const p1_x = geometry::get<0>(p1);
+        coor_t const p1_y = geometry::get<1>(p1);
+        coor_t const p2_x = geometry::get<0>(p2);
+        coor_t const p2_y = geometry::get<1>(p2);
+        coor_t const p_x = geometry::get<0>(p);
+        coor_t const p_y = geometry::get<1>(p);
+
+        constexpr coor_t eps = std::numeric_limits<coor_t>::epsilon() / 2;
+        coor_t const det = (p1_x - p_x) * (p2_y - p_y) - (p1_y - p_y) * (p2_x - p_x);
+        coor_t const err_bound = (Coeff1 * eps + Coeff2 * eps * eps) *
+            (  (geometry::math::abs(p1_x) + geometry::math::abs(p_x))
+             * (geometry::math::abs(p2_y) + geometry::math::abs(p_y))
+             + (geometry::math::abs(p2_x) + geometry::math::abs(p_x))
+             * (geometry::math::abs(p1_y) + geometry::math::abs(p_y)));
+        return (det > err_bound) - (det < -err_bound);
+    }
+};
+
+}} // namespace strategy::side
+
+}} // namespace boost::geometry
+
+#endif // BOOST_GEOMETRY_STRATEGY_CARTESIAN_SIDE_ROUNDED_INPUT_HPP
--- a/include/boost/geometry/strategy/cartesian/side_by_triangle.hpp
+++ b/include/boost/geometry/strategy/cartesian/side_by_triangle.hpp
@@ -33,6 +33,7 @@
 #include <boost/geometry/strategies/compare.hpp>
 #include <boost/geometry/strategies/side.hpp>
 
+#include <boost/geometry/util/select_calculation_type.hpp>
 #include <boost/geometry/util/select_most_precise.hpp>
 
 
@@ -70,7 +71,7 @@ class side_by_triangle
     };
 
 public :
-    typedef cartesian_tag cs_tag;
+    using cs_tag = cartesian_tag;
 
     // Template member function, because it is not always trivial
     // or convenient to explicitly mention the typenames in the
@@ -171,7 +172,7 @@ public :
             // arguments, we cyclically permute them so that the first
             // argument is always the lexicographically smallest point.
 
-            typedef compare::cartesian<compare::less> less;
+            using less = compare::cartesian<compare::less>;
 
             if (less::apply(p, p1))
             {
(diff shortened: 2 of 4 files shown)


tinko92:
Hi @barendgehrels 

I have some suggestions for the test. First, I would suggest allowing negative coordinates, e.g. by allowing a lower limit for the uniform_real_distribution. For uniform_real on [-1, 1], I get the following results:
```
  intersections: 231394
  errors (triangle): 10239 0
  errors (non robust): 213640 0
  errors (side robust fp 3): 231394 0
  errors (side robust eq 3): 15273 0
  errors (rounded): 1718 0
```

I would also test points on a grid with rounded coordinates, e.g. by sampling from a uniform_int_distribution on [-1000, 1000] and dividing by 1000.0, because this might produce more pathological cases. This gives me the following results:

```
  intersections: 231557
  errors (triangle): 11893 0
  errors (non robust): 215074 0
  errors (side robust fp 3): 231555 0
  errors (side robust eq 3): 17545 0
  errors (rounded): 1953 0
```

The errors for the rounded side strategy can be entirely eliminated (assuming no overflow/underflow) by doing the intersection calculation with increased precision (result computed with full precision, then rounded to target precision). You can find a commented implementation here: https://github.com/tinko92/geometry/blob/cc3666f8412f8443140ec5edd3b30c1bab67d059/include/boost/geometry/util/precise_cartesian_intersection.hpp#L255

Using that gives me:
```
  intersections: 231794
  errors (triangle): 1241 0
  errors (non robust): 194766 0
  errors (side robust fp 3): 231794 0
  errors (side robust eq 3): 0 0
  errors (rounded): 0 0
```

However, the performance penalty is very noticeable. Computing a million intersections takes
robustly: 2.2 seconds (1.3 if an error of 2*eps instead of 1*eps is permitted)
with the BG strategy (which also does other things, like classifying the points): 0.11 seconds
only intersection computation directly with the formula on https://mathworld.wolfram.com/Line-LineIntersection.html: 0.006 seconds.

The performance penalty can be mitigated by only doing the precise computation when a fast error bound check cannot guarantee a relative error of less than e.g. 200 * eps (filtered).
filtered: 0.18 seconds

For such an error tolerance, a corresponding rounding side strategy (which would always find these points to be collinear) can be given, in this case, the coefficient for eps in the error bound would not be 5 but 404. This tradeoff of speed/precision is tunable.

I understand that there is more to the intersection strategy than the line-line intersection formula. You can find all mentioned above and some further primitives for intersection calculations in https://github.com/tinko92/geometry/blob/cc3666f8412f8443140ec5edd3b30c1bab67d059/include/boost/geometry/util/precise_cartesian_intersection.hpp .

This is mostly debugged code but some of the used error bounds are definitely not optimal and there might be room for performance optimization in the algorithms.

barendgehrels:
> I have some suggestions for the test. First, I would suggest allowing negative coordinates, e.g. by allowing a lower limit for the uniform_real_distribution. For uniform_real on [-1, 1], I get the following results:
> 

Hi @tinko92 , thanks a lot for your comments, improvements and algorithms!

I'm busy studying and testing them for my purpose and will come back to it.
(I made a few minor comments in your commits in your fork)

include/boost/geometry/strategies/cartesian/side_rounded_input.hpp
```diff
@@ -0,0 +1,60 @@
...
+    template <typename P1, typename P2, typename P>
+    static inline int apply(P1 const& p1, P2 const& p2, P const& p)
+    {
+
+        using CT = std::conditional_t
+            <
+                std::is_void<CalculationType>::value,
+                double,
+                CalculationType
+            >;
```
awulkiew, line 37: Typically if `CT` is `void` we use the most precise coordinate type. Or is there a reason why `double` is used here?

include/boost/geometry/strategies/cartesian/side_rounded_input.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+        using namespace boost::geometry;
+        CT _1 = get<0>(p1);
+        CT _2 = get<1>(p1);
+        CT _3 = get<0>(p2);
+        CT _4 = get<1>(p2);
+        CT _5 = get<0>(p);
+        CT _6 = get<1>(p);
```
awulkiew, line 45: There are two problems with these names. They are the same as Boost.Bind placeholders. Furthermore typically identifiers starting with `_` are reserved for predefined preprocesor macros. So I propose to change them.

awulkiew: One more. They are confusing because these are numbers but they doesnt represents these specific numbers neither a position in some order etc. Better would be e.g. `x` or `p0`, etc.

test/robustness/overlay/strategies/intersection_points_and_sides.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+    T sum_distance_error = 0;
+    T sum_distance_for_rounded_error = 0;
+
+    for (std::size_t index = 0; index < settings.count; index++)
+    {
+        const point_t p1 = random_point<point_t>(gen, dis);
+        const point_t p2 = random_point<point_t>(gen, dis);
+        const point_t q1 = random_point<point_t>(gen, dis);
+        const point_t q2 = random_point<point_t>(gen, dis);
```
awulkiew, line 162: Random points probably shouldn't be usen in tests because various libraries may implement random generator differently so we may be potentially testing agsinst different points.

include/boost/geometry/strategies/cartesian/side_rounded_input.hpp
```diff
@@ -0,0 +1,70 @@
...
+            <
+                std::is_void<CalculationType>::value,
+                typename select_most_precise
+                    <
+                        coordinate_type1,
+                        coordinate_type2,
+                        coordinate_type3
+                    >::type,
+                CalculationType
+            >;
```
awulkiew, line 46: All of these lines could be replaced with:
```
using coor_t = typename select_calculation_type_alt<CalculationType, P1, P2, P>::type;
```
from here:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/util/select_calculation_type.hpp#L54-L70

awulkiew (approved):
LGTM

barendgehrels:
@vissarion OK?
@tinko92 this is the new version, OK?

tinko92:
> @tinko92 this is the new version, OK?
Sorry, I missed the mention. I think it is ok!

vissarion:
OK from me too and sorry for the delays.

PR #997, opened 2022-05-07
Add support for GeometryCollection in difference()

awulkiew:
Add also:
- `make_rtree()` utility making an rtree of GC elements
- `geometry_collection_view<>` allowing to represent any geometry as GC
- tests

This PR also adds support for difference(g1, g2) where topological dimension of g1 is greater than of g2 so the result is always g1, e.g. Polygon \ MultiPoint.

### DIFF
--- a/include/boost/geometry/algorithms/detail/make_rtree.hpp
+++ b/include/boost/geometry/algorithms/detail/make_rtree.hpp
@@ -1,0 +1,113 @@
+// Boost.Geometry
+
+// Copyright (c) 2022, Oracle and/or its affiliates.
+
+// Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
+
+// Licensed under the Boost Software License version 1.0.
+// http://www.boost.org/users/license.html
+
+#ifndef BOOST_GEOMETRY_ALGORITHMS_DETAIL_MAKE_RTREE_HPP
+#define BOOST_GEOMETRY_ALGORITHMS_DETAIL_MAKE_RTREE_HPP
+
+#include <vector>
+
+#include <boost/range/begin.hpp>
+#include <boost/range/end.hpp>
+#include <boost/range/size.hpp>
+
+#include <boost/geometry/algorithms/detail/expand_by_epsilon.hpp>
+#include <boost/geometry/algorithms/detail/visit.hpp>
+#include <boost/geometry/algorithms/envelope.hpp>
+#include <boost/geometry/index/rtree.hpp>
+#include <boost/geometry/strategies/index/services.hpp>
+
+namespace boost { namespace geometry
+{
+
+
+#ifndef DOXYGEN_NO_DETAIL
+namespace detail
+{
+
+
+template <typename GC>
+using make_rtree_box_t = geometry::model::box
+    <
+        geometry::model::point
+            <
+                typename geometry::coordinate_type<GC>::type,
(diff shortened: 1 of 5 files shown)


include/boost/geometry/algorithms/difference.hpp
```diff
@@ -15,19 +15,21 @@
 #define BOOST_GEOMETRY_ALGORITHMS_DIFFERENCE_HPP
 
 
-#include <boost/variant/apply_visitor.hpp>
-#include <boost/variant/static_visitor.hpp>
-#include <boost/variant/variant_fwd.hpp>
-
 #include <boost/geometry/algorithms/detail/intersection/multi.hpp>
+#include <boost/geometry/algorithms/detail/make_rtree.hpp>
```
barendgehrels, line 19: Did we have a dependency on `rtree` in the non-index part of the library before?

Usually all the overlay algorithms use partition. Why do we need an rtree now for the same (didn't review all yet, but it's probably the same) purpose?

include/boost/geometry/algorithms/detail/make_rtree.hpp
```diff
@@ -0,0 +1,113 @@
...
+{
+    using box_t = make_rtree_box_t<GC>;
+    using iter_t = typename boost::range_iterator<GC>::type;
+
+    using rtree_param_t = index::rstar<4>;
+    using rtree_parameters_t = index::parameters<rtree_param_t, Strategy>;
+    using value_t = std::pair<box_t, iter_t>;
+    using rtree_t = index::rtree<value_t, rtree_parameters_t>;
+
+    // TODO: get rid of the temporary vector
```
vissarion, line 57: what is the obstacle of doing it in this PR?

include/boost/geometry/algorithms/difference.hpp
```diff
@@ -268,10 +299,291 @@ inline OutputIterator difference_insert(Geometry1 const& geometry1,
...
+                    range::emplace_back(output_collection, std::move(out));
+                }
+            }
+
+            return true;
+        }, geometry1);
+    }
+
+private:
+    // Implemented as separate function because msvc is unable to do nested lambda capture
```
vissarion, line 425: what version of MSVC does not support this?

vissarion (approved):
Thanks Adam! Look OK to me.

include/boost/geometry/algorithms/difference.hpp
awulkiew: I tested with msvc-14.1 i.e. VS2017.

vissarion: OK thanks! Side note, it could have been fixed in VS2019 
https://developercommunity.visualstudio.com/t/this-captured-in-nested-lambda-when-used-in-captur/292128

include/boost/geometry/algorithms/detail/make_rtree.hpp
vissarion: Cool. Thanks for both versions!

include/boost/geometry/algorithms/difference.hpp
barendgehrels: Thanks for your explanation! I agree with you that code using partition is less readable.
I'm fine.

barendgehrels (approved):

PR #1003, opened 2022-05-16
Fix line_interpolate termination condition for single points.

### LINKED ISSUE #1000: The function boost::geometry::line_interpolate provides wrong results.
What I found is that , when the input geometry type is not exactly match with the output point like type, this function can not provide correct results for some input.  It works for certain data but not all.

Here is the code for reproducing the bug.

I tested the code using both boost 1.71 and 1.77 on ubuntu20.04 gcc 9.4.0. and clang10.0.0. All the combinations give me the same result.

``` c++


#include <boost/fusion/include/adapt_struct_named.hpp>
#include <boost/geometry.hpp>
#include <boost/geometry/geometries/adapted/boost_fusion.hpp>
#include <iostream>
#include <iomanip>
struct Point {
    double x;
    double y;
    Point() = default;
    Point(double x, double y)
        : x(x), y(y)
    {
    }
};

BOOST_FUSION_ADAPT_STRUCT(Point, (double, x)(double, y))
BOOST_GEOMETRY_REGISTER_BOOST_FUSION_CS(cs::cartesian)
using Line = boost::geometry::model::linestring<Point>;

int main()
{
    using point_type = boost::geometry::model::d2::point_xy<double>;
    using linestring_type = boost::geometry::model::linestring<point_type>;
    Line l1;

    boost::geometry::read_wkt("LINESTRING("
                              "13534187.97 4710766.12,"
                              "13534197.43 4710790.00,"
                              "13534224.15 4710830.21,"
                              "13534230.07 4710836.83,"
                              "13534259.87 4710864.74,"
                              "13534308.82 4710902.11)",
                              l1);

    double s = 10.0;

    std::cout << std::fixed << std::setprecision(2);

    Point p1;
    boost::geometry::line_interpolate(l1, s, p1);
    std::cout << "p1: " << boost::geometry::wkt(p1) << std::endl;

    point_type p2;
    boost::geometry::line_interpolate(l1, s, p2);
    std::cout << "p2: " << boost::geometry::wkt(p2) << std::endl;



    return 0;
}

```

Out put of the code:
```
p1: POINT(13534191.65 4710775.42)
p2: POINT(13534304.64 4710898.92)
```
p1 and p2 should be the same.

tinko92:
If I understood line_interpolate correctly, the early termination condition is intended to apply to all output geometry types that are single points rather than just the point type of the input range. I think this fixes #1000 .

### DIFF
--- a/include/boost/geometry/algorithms/line_interpolate.hpp
+++ b/include/boost/geometry/algorithms/line_interpolate.hpp
@@ -27,7 +27,6 @@
 #include <boost/geometry/core/cs.hpp>
 #include <boost/geometry/core/closure.hpp>
 #include <boost/geometry/core/static_assert.hpp>
-#include <boost/geometry/core/tag.hpp>
 #include <boost/geometry/core/tags.hpp>
 
 #include <boost/geometry/geometries/concepts/check.hpp>
@@ -37,6 +36,9 @@
 #include <boost/geometry/strategies/line_interpolate/cartesian.hpp>
 #include <boost/geometry/strategies/line_interpolate/geographic.hpp>
 #include <boost/geometry/strategies/line_interpolate/spherical.hpp>
+
+#include <boost/geometry/util/condition.hpp>
+#include <boost/geometry/util/type_traits.hpp>
 
 namespace boost { namespace geometry
 {
@@ -131,7 +133,7 @@ struct interpolate_range
                                p,
                                diff_distance);
                 Policy::apply(p, pointlike);
-                if (std::is_same<typename tag<PointLike>::type, point_tag>::value)
+                if ( BOOST_GEOMETRY_CONDITION(util::is_point<PointLike>::value) )
                 {
                     return;
                 }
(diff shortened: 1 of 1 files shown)


vissarion (approved):
Great thanks!

include/boost/geometry/algorithms/line_interpolate.hpp
```diff
@@ -130,7 +131,7 @@ struct interpolate_range
                                p,
                                diff_distance);
                 Policy::apply(p, pointlike);
-                if (std::is_same<PointLike, point_t>::value)
+                if (std::is_same<typename tag<PointLike>::type, point_tag>::value)
```
awulkiew, line 134: FYI, we have type traits for this now, i.e. `util::is_point<PointLike>::value` here:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/util/type_traits.hpp#L38

Some compilers will complain about this `constant condition` but it was here before.

tinko92: Thanks for the hint, I will update the PR tomorrow to use is_point.

With regard to the constant condition: This was last solved with the BOOST_GEOMETRY_CONDITION macro, should that be used here as well?
Alternatively, there is the BOOST_IF_CONSTEXPR macro from boost.config, which expands to "if" for <=c++14 and "if constexpr" for >c++17 (which was apparently available in the compiler that generated the warning: https://github.com/boostorg/geometry/issues/727)

barendgehrels (approved):
Thanks!

PR #1007, opened 2022-05-22
[strategies] make cartesian_winding similar to spherical_winding

barendgehrels:
This prepares a next phase where this will be used from `turn_in_ring_winding`


include/boost/geometry/strategies/cartesian/point_in_poly_winding.hpp
```diff
@@ -229,6 +217,29 @@ class cartesian_winding
...
+}
+ */
+template
+<
+    typename Point_ = void, // for backward compatibility
+    typename PointOfSegment_ = Point_, // for backward compatibility
+    typename CalculationType = void
+>
+class cartesian_winding
+  : public detail::cartesian_winding_base<typename side::services::default_strategy<cartesian_tag, CalculationType>::type, CalculationType> {};
```
vissarion, line 242: isn't it too long? If not skip my comment.

vissarion (approved):
Looks good to me.

include/boost/geometry/strategies/cartesian/point_in_poly_winding.hpp
barendgehrels: It is indeed. Will fix.

PR #1008, opened 2022-05-22
fix compilation error (svg), warnings (unused), style (return), examp…

barendgehrels:
…les and cmake (C++14)

### DIFF
--- a/example/05_b_overlay_linestring_polygon_example.cpp
+++ b/example/05_b_overlay_linestring_polygon_example.cpp
@@ -19,9 +19,6 @@
 #include <string>
 #include <vector>
 
-#include <boost/foreach.hpp>
-
-
 #include <boost/geometry/geometry.hpp>
 #include <boost/geometry/geometries/linestring.hpp>
 #include <boost/geometry/geometries/point_xy.hpp>
@@ -39,7 +36,7 @@ int main(void)
 {
     namespace bg = boost::geometry;
 
-    typedef bg::model::d2::point_xy<double> point_2d;
+    using point_2d = bg::model::d2::point_xy<double>;
 
     bg::model::linestring<point_2d> ls;
     {
@@ -67,16 +64,16 @@ int main(void)
 #endif
 
     // Calculate intersection points (turn points)
-    typedef bg::detail::segment_ratio_type<point_2d, bg::detail::no_rescale_policy>::type segment_ratio;
-    typedef bg::detail::overlay::turn_info<point_2d, segment_ratio> turn_info;
+    using segment_ratio = bg::detail::segment_ratio_type<point_2d, bg::detail::no_rescale_policy>::type;
+    using turn_info = bg::detail::overlay::turn_info<point_2d, segment_ratio>;
     std::vector<turn_info> turns;
     bg::detail::get_turns::no_interrupt_policy policy;
     bg::detail::no_rescale_policy rescale_policy;
-    bg::strategy::intersection::services::default_strategy<typename bg::cs_tag<point_2d>::type>::type intersection_strategy;
-    bg::get_turns<false, false, bg::detail::overlay::assign_null_policy>(ls, p, intersection_strategy, rescale_policy, turns, policy);
+    bg::strategies::relate::services::default_strategy<bg::model::linestring<point_2d>, bg::model::polygon<point_2d>>::type strategy;
+    bg::get_turns<false, false, bg::detail::overlay::assign_null_policy>(ls, p, strategy, rescale_policy, turns, policy);
 
     std::cout << "Intersection of linestring/polygon" << std::endl;
-    BOOST_FOREACH(turn_info const& turn, turns)
+    for (turn_info const& turn : turns)
     {
         std::string action = "intersecting";
         if (turn.operations[0].operation
--- a/example/c10_custom_cs_example.cpp
+++ b/example/c10_custom_cs_example.cpp
@@ -21,7 +21,7 @@
 template<typename DegreeOrRadian>
 struct martian
 {
-    typedef DegreeOrRadian units;
+    using units = DegreeOrRadian;
 };
 
 // 2: give it also a family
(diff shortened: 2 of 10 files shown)


example/c10_custom_cs_example.cpp
```diff
@@ -67,10 +51,10 @@ struct default_strategy<point_tag, point_tag, Point1, Point2, martian_tag, marti
 
 int main()
 {
-    typedef boost::geometry::model::point
+    using mars_point = boost::geometry::model::point
```
vissarion, line 54: Should we make this consistently in all examples of the library?

vissarion (approved):
Thanks! OK by me.

example/c10_custom_cs_example.cpp
barendgehrels: In the end, yes. I usually update them now if they are at or close to touched code.

example/05_b_overlay_linestring_polygon_example.cpp
```diff
@@ -72,8 +72,8 @@ int main(void)
     std::vector<turn_info> turns;
     bg::detail::get_turns::no_interrupt_policy policy;
     bg::detail::no_rescale_policy rescale_policy;
-    bg::strategy::intersection::services::default_strategy<typename bg::cs_tag<point_2d>::type>::type intersection_strategy;
-    bg::get_turns<false, false, bg::detail::overlay::assign_null_policy>(ls, p, intersection_strategy, rescale_policy, turns, policy);
+    typename bg::strategies::relate::services::default_strategy<bg::model::linestring<point_2d>, bg::model::polygon<point_2d>>::type strategy;
```
awulkiew, line 75: None of the types are dependant so `typename` shouldn't be used. AFAIR some compilers may report error here. Or am I missing something?

example/c10_custom_cs_example.cpp
```diff
@@ -79,24 +63,26 @@ int main()
...
-    // Using the Mars mean radius
+    // To calculate distance, declare and construct a strategy with Mars mean radius, in KM
     // (Source: http://nssdc.gsfc.nasa.gov/planetary/factsheet/marsfact.html)
+    boost::geometry::strategy::distance::haversine<double> const haversine(3389.5);
+
+    double d = boost::geometry::distance(viking1, pathfinder, haversine);
+
     std::cout << "Distance between Viking1 and Pathfinder landing sites: "
-        << d * 3389.5 << " km" << std::endl;
+        << d * 1.0  << " km" << std::endl;
```
awulkiew, line 73: So this `1.0` is not needed anymore. :)

example/c10_custom_cs_example.cpp
```diff
@@ -79,24 +63,26 @@ int main()
     mars_point viking1(-48.23, 22.54); // Viking 1 landing site in Chryse Planitia
     mars_point pathfinder(-33.55, 19.33); // Pathfinder landing site in Ares Vallis
 
-    double d = boost::geometry::distance(viking1, pathfinder); // Distance in radians on unit-sphere
-
-    // Using the Mars mean radius
+    // To calculate distance, declare and construct a strategy with Mars mean radius, in KM
     // (Source: http://nssdc.gsfc.nasa.gov/planetary/factsheet/marsfact.html)
+    boost::geometry::strategy::distance::haversine<double> const haversine(3389.5);
```
awulkiew, line 68: We have umbrella strategies now. Shouldn't we use them in the example? In this case `strategies::distance::spherical<>` ?

example/c10_custom_cs_example.cpp
```diff
@@ -79,24 +63,26 @@ int main()
...
-#ifdef OPTIONALLY_ELLIPSOIDAL
-    // Optionally the distance can be calculated more accurate by an Ellipsoidal approach,
+    // The distance can be calculated more accurately by an Ellipsoidal approach,
     // giving 834.444 km
-    d = boost::geometry::distance(viking1, pathfinder,
-        boost::geometry::strategy::distance::andoyer<mars_point>
-            (boost::geometry::srs::spheroid<double>(3396.2, 3376.2)));
+    using spheroid_type = boost::geometry::srs::spheroid<double>;
+    spheroid_type spheroid(3396.2, 3376.2);
+    boost::geometry::strategy::distance::andoyer<spheroid_type> const andoyer(spheroid);
```
awulkiew, line 82: And here `strategies::distance::geographic` ?

awulkiew (approved):

test/to_svg.hpp
```diff
@@ -190,7 +190,7 @@ inline void to_svg(G1 const& g1, G2 const& g2, G3 const& g3,
 
     bg::detail::relate::turns::get_turns
         <
-            G1, G2, point_type, turn_policy
+            G1, G2, turn_policy
```
awulkiew, line 193: This is the fix for error similar to this one, correct?
```
In file included from libs/geometry/test/algorithms/set_operations/sym_difference/test_sym_difference_linear_linear.hpp:21:0,
                 from libs/geometry/test/algorithms/set_operations/sym_difference/sym_difference_linear_linear.cpp:23:
libs/geometry/test/to_svg.hpp: In function ‘void to_svg(const G1&, const G2&, const G3&, const string&)’:
libs/geometry/test/to_svg.hpp:194:9: error: wrong number of template arguments (4, should be at least 2)
         >::apply(turns, g1, g2, interrupt_policy, strategy);
```

Should `to_svg.hpp` be included if `TEST_WITH_SVG` is not defined?
It is in the following files:
https://github.com/boostorg/geometry/blob/develop/test/algorithms/set_operations/intersection/test_intersection_linear_linear.hpp#L22
https://github.com/boostorg/geometry/blob/develop/test/algorithms/set_operations/difference/test_difference_linear_linear.hpp#L22
https://github.com/boostorg/geometry/blob/develop/test/algorithms/set_operations/sym_difference/test_sym_difference_linear_linear.hpp#L21
https://github.com/boostorg/geometry/blob/develop/test/algorithms/set_operations/union/test_union_linear_linear.hpp#L21

barendgehrels: That makes sense, I will change it.

example/c10_custom_cs_example.cpp
barendgehrels: Ah, yes, thanks, will fix.

example/c10_custom_cs_example.cpp
barendgehrels: OK I'll try.
The "problem" was that these original examples didn't work anymore. So somewhere we were not backwards compatible, the strategy converters did not work for these cases.
Anyway, I was a bit short on time and fixed it with the least amount of time at that moment. I'll look at it again, but anyway, you're welcome to update them again if you think it's better or better illustrates the possibilities.

example/05_b_overlay_linestring_polygon_example.cpp
barendgehrels: Indeed, will fix.

example/c10_custom_cs_example.cpp
awulkiew: Yes, we should probably find out why this is not backward compatible.

barendgehrels: > We have umbrella strategies now. Shouldn't we use them in the example? In this case `strategies::distance::spherical<>` ?

That works

example/c10_custom_cs_example.cpp
barendgehrels: :heavy_check_mark:

example/05_b_overlay_linestring_polygon_example.cpp
barendgehrels: :heavy_check_mark:

test/to_svg.hpp
barendgehrels: :heavy_check_mark:

Issue #900, opened 2021-08-14
Wrong result of sym_difference.


awulkiew:
To reproduce (polygon's point order doesn't matter):
```
namespace bg = boost::geometry;
namespace bgm = boost::geometry::model;

using cpt_t = bgm::point<double, 2, bg::cs::cartesian>;
using cpo_t = bgm::polygon<cpt_t, false, true>;
using cmpo_t = bgm::multi_polygon<cpo_t>;

int main()
{
    cpo_t a, b, c, d;
    bg::read_wkt("POLYGON((25 0,0 15,30 15,22 10,25 0))", a);
    bg::read_wkt("POLYGON((5 0,15 25,25 0,15 5,5 0))", b);
    bg::read_wkt("POLYGON((15 0,17 10,10 15,20 15,25 25,30 15,40 15,32 10,35 0,25 5,15 0))", c);
    bg::read_wkt("POLYGON((5 0,7 10,0 15,10 15,15 25,20 15,30 15,22 10,25 0,15 5,5 0))", d);
    bg::correct(a);
    bg::correct(b);
    bg::correct(c);
    bg::correct(d);

    cmpo_t ab, abc, dc, abcdc;
    bg::sym_difference(a, b, ab);       // ok
    bg::sym_difference(ab, c, abc);     // ok
    bg::sym_difference(d, c, dc);       // ok
    bg::sym_difference(abc, dc, abcdc); // wrong!
}
```

Wrong result for geometries:
```
[...]

PR #1010, opened 2022-05-29
Allow start for self-turns for difference overlay operations

tinko92:
This changes the handling of colocations in overlay for difference operations such that self-turns are not marked as not startable. As far as I can see, this fixes test case case_recursive_boxes_62 and, I believe, this also solves #900 without breaking any other cases.

Here is a reduced version of that test case, transformed to integer coordinates, in case it needs further analysis:

```cpp
#include <iostream>

#include <boost/geometry.hpp>

using point = bg::model::d2::point_xy<int>;
using multi_polygon = bg::model::multi_polygon<polygon>;

int main()
{
    auto a = bg::from_wkt<multi_polygon>("MULTIPOLYGON(((5 3,6 4,5 2,4 4,5 3)),((4 4,4 6,6 4,4 4)),((6 4,6 5,8 4,6 4)),((4 4,2 4,2 5,4 4)))");
    auto b = bg::from_wkt<multi_polygon>("MULTIPOLYGON(((1 1,1 6,9 6,9 1,1 1)))");
    multi_polygon diff;
    bg::difference(b, a, diff);
    std::cout << bg::to_wkt(diff);
    return 0;
}
```

I am not certain that I completely understand the overlay algorithm, though. Is it based on a specific polygon overlay algorithm that can be found in the literature?

With this change applied, I no longer find any error in the set operation results in Boost.Geometry compared to CGAL for any of the test cases, see https://tinko92.github.io/bg_cgal_set_operations_comparison/ (**Edit**: The results on the page are based on the branch in this PR and sorting the column "(BG Real Δ CGAL Real area) / (CGAL Real area)" in descending order shows that there are no examples in which BG and CGAL disagree if both use an infinite-precision number type.) and https://github.com/tinko92/boost-geometry-cgal-set-operation-comparison for the full list and the code used to do the comparison. Maybe the remaining issues (with BOOST_GEOMETRY_TEST_FAILURES)  can be solved with suitable strategies.


barendgehrels:
Can you add the fixed testcase to the unit testsuite? I checked if (but it's more convenient in this PR):

In file `multi_overlay_cases.hpp`: after `issue_888_53` add :
```
static std::string issue_900[2] =
{
    "MULTIPOLYGON(((5 3,6 4,5 2,4 4,5 3)),((4 4,4 6,6 4,4 4)),((6 4,6 5,8 4,6 4)),((4 4,2 4,2 5,4 4)))",
    "MULTIPOLYGON(((1 1,1 6,9 6,9 1,1 1)))"
};

```

And in difference_multi.cpp, after the same case, add
```
TEST_DIFFERENCE(issue_900, 0, 0.0, 2, 35, 2);
```

Thanks!

barendgehrels:
The algorithm is Graph Traversal and is based on Weiler-Atherton and Greiner–Hormann  (so *not* on Vatti).

So the basic algorithm follows exactly that graph traversal. But things as colocated intersections make it more difficult, those implementations are basically added by myself.

See for example https://www.sciencedirect.com/science/article/pii/S259014861930007X (but I didn't use that article, it didn't exist yet).

barendgehrels (approved):

tinko92:
Thanks for the link and the feedback. I hope I didn't forget anything in the updated commit. With regard to the comment, out of your suggestions I chose the one of keeping the TODO.

vissarion (approved):
Looks good! Thanks.

ps. was that a typo?

awulkiew (approved):

barendgehrels:
> Thanks for the link and the feedback. I hope I didn't forget anything in the updated commit. With regard to the comment, out of your suggestions I chose the one of keeping the TODO.

Looks good, thanks for the testcase, I'll merge it.

PR #1011, opened 2022-06-02
Add support for GC in intersection, sym_difference and union.

awulkiew:
These algorithms calculate the intermediate results based on the bounding boxes of elements of GCs and these results are merged together based on topological dimension. So e.g. parts of linestrings of GC1 that are intersecting any linear or areal geometry from GC2 are merged into multilinestring(s).

The next step could be to subtract the results from each other to prevent them occupying the same space. It's not clear to me what the result of set operations should be for GCs because they are not well defined by the standard. I decided to leave these geometries as they are, so potentially intersecting each other, in order not to loose information.

If merging of elements of the result is desired one may call `merge_elements()` afterwards.

For example the intersection of GC1:

gc1

and GC2:

gc2

is:

result

Note that some of the linestrings are results of touching polygons, so intersecting boundaries.

PR #1016, opened 2022-06-14
Add support for GC in covered_by, crosses, equals, overlaps, relate, relation, touches and within

awulkiew:
All of these algorithms use `relate`.

The `relate` algorithm merges parts of geometry collections with set operations and calculates partial results.

Some of the GC utilities are moved, their names changed to use the same pattern, i.e. `detail::gc_xxx`. There are also some fixes and small refactoring.

I've also found a case of `relate` L/L which result is wrong but it's unrelated to this PR.

This PR is built on top of https://github.com/boostorg/geometry/pull/1011 so I think we should merge the other one before reviewing this one.

----

For future reference below is a set of combinations of Areal and Linear parts of GCs that I used while developing the algorithm. The combinations of these parts are tricky to handle because they may touch and therefore can have common intersection points. Unlike PointLike part which is filtered-out by merging.
rect184
For X/Y, X is part of GC1 (red), Y is part of GC2 (green).
A is Areal, L is Linear, 0 is empty geometry.
Parts here are after merging so Linear part is always outside of the Areal part of the same GC.
Li is the intersection of Linear part of a GC and the Areal part of the other GC, so part of Linear that is inside other Areal.
Ld is the difference of Linear part of a GC and Areal part of the other GC, so part of Linear that is outside of other Areal.


test/algorithms/covered_by/test_covered_by.hpp
```diff
@@ -76,7 +76,7 @@ void test_geometry(std::string const& wkt1,
     boost::variant<Geometry1> v1(geometry1);
     boost::variant<Geometry2> v2(geometry2);
 
-    typedef typename bg::strategy::covered_by::services::default_strategy
+    typedef typename bg::strategies::relate::services::default_strategy
```
vissarion, line 79: since you changed it maybe use `using` ?

test/algorithms/relate/test_relate.hpp
```diff
@@ -120,7 +120,7 @@ void check_geometry(Geometry1 const& geometry1,
             << " -> Expected: " << matrix_format(expected1, expected2)
             << " detected: " << res_str);
 
-        typedef typename bg::strategy::relate::services::default_strategy
+        typedef typename bg::strategies::relate::services::default_strategy
```
vissarion, line 123: as in overlaps, do it with `using`?

vissarion (approved):
I skipped the setops and review the rest of the code that looks OK. Thanks Adam!

include/boost/geometry/algorithms/detail/gc_group_elements.hpp
```diff
@@ -0,0 +1,196 @@
...
+                    if (! visited[n.source_id][n.gc_id])
+                    {
+                        queue.push_back(n);
+                        visited[n.source_id][n.gc_id] = true;
+                        group.push_back(n);
+                    }
+                }
+            }
+        }
+        if (! group.empty())
```
vissarion, line 158: or
```
if (! group.empty() && ! intersecting_fun(group) ) return;
```

awulkiew: Yes, this would be shorter but IDK really that it'd be better. The original makes it more clear that this is a call of a function passed by the user and not another condition check. It is also more similar to the second call of the other function below. @barendgehrels what's your opinion?

test/algorithms/covered_by/test_covered_by.hpp
awulkiew: Done

test/algorithms/relate/test_relate.hpp
awulkiew: Done

include/boost/geometry/core/topological_dimension.hpp
```diff
@@ -78,6 +78,10 @@ template <>
 struct top_dim<multi_polygon_tag> : std::integral_constant<int, 2> {};
 
 
+template <>
+struct top_dim<geometry_collection_tag> : std::integral_constant<int, -1> {};
```
barendgehrels, line 82: Okay for me.

include/boost/geometry/algorithms/crosses.hpp
```diff
@@ -62,6 +61,82 @@ struct crosses
...
 
+template <typename Geometry1, typename Geometry2>
+struct crosses<Geometry1, Geometry2, geometry_collection_tag, geometry_collection_tag>
+{
+    template <typename Strategy>
+    static inline bool apply(Geometry1 const& geometry1, Geometry2 const& geometry2,
+                             Strategy const& strategy)
+    {
+        int dimension1 = detail::gc_topological_dimension(geometry1);
+        int dimension2 = detail::gc_topological_dimension(geometry2);
```
barendgehrels, line 72: Minor: can it be const, or even static maybe?

barendgehrels (approved):
Looks like great work!
I reviewed some of it and it looks good to me, and things are consistent and looks the same way in different algorithms.
Thanks!

include/boost/geometry/algorithms/crosses.hpp
awulkiew: `const` yes, `static` no because the dimension depends on what is actually stored in a GC.

PR #1017, opened 2022-06-14
Add support for GC and DG in buffer().

awulkiew:
The algorithm calculates partial results for each element of a GC and merges them iteratively.

I was forced to move parts of the code around and create the "proper" structure for the implementation with parts in `resolve_dynamic` and `dispatch`. I omitted `resolve_strategy`.

One of the changes was dropping the support for vector of polygons as the output geometry since this is not a geometry. It is not tested anywhere. All of the tests pass. If you think we should keep the support for this I could bring it back.

Side effect of the above is that now multipolygons are properly handled because before they were cleared with `clear()` member function which is not in line with the concept. I replaced it with `bg::clear()` so it works for GC too.

This PR is built on top of https://github.com/boostorg/geometry/pull/1016 so I think we should merge the other one before reviewing this one.

The result:

obraz


include/boost/geometry/algorithms/detail/buffer/implementation.hpp
```diff
@@ -0,0 +1,210 @@
...
+                             Output& geometry_out,
+                             DistanceStrategy const& distance_strategy,
+                             SideStrategy const& side_strategy,
+                             JoinStrategy const& join_strategy,
+                             EndStrategy const& end_strategy,
+                             PointStrategy const& point_strategy,
+                             Strategies const& strategies)
+    {
+        // NOTE: We could also allow returning GC containing only polygons.
+        //   We'd have to wrapp them in model::multi_polygon and then
```
barendgehrels, line 181: `wrap`

include/boost/geometry/algorithms/detail/buffer/implementation.hpp
```diff
@@ -0,0 +1,210 @@
...
+                             DistanceStrategy const& distance_strategy,
+                             SideStrategy const& side_strategy,
+                             JoinStrategy const& join_strategy,
+                             EndStrategy const& end_strategy,
+                             PointStrategy const& point_strategy,
+                             Strategies const& strategies)
+    {
+        // NOTE: This algorithm merges partial results iteratively.
+        //   We could first gather all of the results and after that
+        //   use some more optimal method like merge_elements().
```
barendgehrels, line 137: The buffer normally calculates everything at once (by pieces) and traverses all of them to apply the union operation. Not even by merging elements.
But that is complex and has led to issues as well. So I understand that you now do it like this. Maybe forever indeed. OK for me,
but maybe you can include it in the comment.

barendgehrels (approved):
OK, let's go for it!
Thanks for all these efforts!

vissarion (approved):
I am OK with buffer gc too. Thanks!

include/boost/geometry/algorithms/detail/buffer/implementation.hpp
awulkiew: I included your comment.

PR #1024, opened 2022-06-24
Allow different input and output types in simplify

awulkiew:
Requested in this comment: https://github.com/boostorg/geometry/commit/0f72af95f0ac94f94022c6e4b7831895c2291dde#commitcomment-76745077

### DIFF
--- a/include/boost/geometry/algorithms/simplify.hpp
+++ b/include/boost/geometry/algorithms/simplify.hpp
@@ -660,7 +660,7 @@ struct static_geometry_type
     using type = typename util::sequence_find_if
         <
             typename traits::geometry_types<DynamicGeometryOut>::type,
-            detail::simplify::has_same_tag_as<StaticGeometryIn>::pred
+            detail::simplify::has_same_tag_as<StaticGeometryIn>::template pred
         >::type;
 
     BOOST_GEOMETRY_STATIC_ASSERT(
--- a/include/boost/geometry/algorithms/simplify.hpp
+++ b/include/boost/geometry/algorithms/simplify.hpp
@@ -98,10 +98,8 @@ struct douglas_peucker_point
 \brief Implements the simplify algorithm.
 \details The douglas_peucker policy simplifies a linestring, ring or
     vector of points using the well-known Douglas-Peucker algorithm.
-\tparam Point the point type
-\tparam PointDistanceStrategy point-segment distance strategy to be used
-\note This strategy uses itself a point-segment-distance strategy which
-    can be specified
+\note This strategy uses itself a point-segment potentially comparable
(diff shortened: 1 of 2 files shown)


icehuli:
Hi @awulkiew,

I'm sorry, I think it is my fault that I didn't specified the problem more clearly. In the latest version of `buffer_inserter` `simplify_range` is used instead of `simplify`. `simplify_range` can produce a different output Type.
https://github.com/boostorg/geometry/blob/6d57aedf7f2b883ed1ac0ffdc59d27c9342d3637/include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp#L73

So what needs to be changed is `simplify_input`:
https://github.com/boostorg/geometry/blob/6d57aedf7f2b883ed1ac0ffdc59d27c9342d3637/include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp#L58

I'll try to create a PR for this.

awulkiew:
@icehuli No, it's ok. I stated that `simplify` should be able to take different input and output types and this PR is about that. This is more or less unrelated to `buffer`.

Regarding `buffer`, I wanted to wait for @barendgehrels to give his opinion but you could create a PR addressing it so the discussion would be moved there.

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -624,6 +641,35 @@ struct simplify_multi
...
+template <typename StaticGeometryIn, typename DynamicGeometryOut>
+struct static_geometry_type
+{
+    using type = typename util::sequence_find_if
+        <
+            typename traits::geometry_types<DynamicGeometryOut>::type,
+            detail::simplify::has_same_tag_as<StaticGeometryIn>::template pred
+        >::type;
+
+    BOOST_GEOMETRY_STATIC_ASSERT(
```
barendgehrels, line 666: We keep using this? Or would new code use just `static_assert` ?

barendgehrels (approved):
I understood that the reported problem was inside buffer itself, but this PR on itself looks also useful to me. Thanks!

include/boost/geometry/algorithms/simplify.hpp
awulkiew: For consistency and also because this macro has variadic number of arguments [1] so you can pass types after condition and message and these types will typically be displayed in the error message. Furthermore if you pass `false` as condition it will compile as long as a struct inside is not instantiated while `static_assert` will never compile. This is why we can't use `static_assert(false)` in all cases, e.g. inside `not_implemented`.

[1] https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/core/static_assert.hpp#L24

Issue #1025, opened 2022-06-27
boost::buffer doesn't work on const span/view like polygon input. 


icehuli:
I tried to use boost::buffer on `span`-like const type Polygon input. But it doesn't work due to
https://github.com/boostorg/geometry/blob/6d57aedf7f2b883ed1ac0ffdc59d27c9342d3637/include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp#L527-L528

Since my Polygon only provides a const RingType, `simplify` doesn't work.  
In current version, is it possible to apply boost::buffer on `span`/`view`-like input?

PR #1027, opened 2022-06-28
Allow const inputs in buffer()

awulkiew:
I decided to use helper geometry using the same coordinate type instead of output ring type to avoid result changes in case different coordinate type was used in the output.

This PR also fixes range emptiness check where member function of a range was called instead of `boost::empty()`.

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
@@ -2,8 +2,8 @@
 
 // Copyright (c) 2012-2020 Barend Gehrels, Amsterdam, the Netherlands.
 
-// This file was modified by Oracle on 2017-2021.
-// Modifications copyright (c) 2017-2021 Oracle and/or its affiliates.
+// This file was modified by Oracle on 2017-2022.
+// Modifications copyright (c) 2017-2022 Oracle and/or its affiliates.
 // Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
 
 // Use, modification and distribution is subject to the Boost Software License,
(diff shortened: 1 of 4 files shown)


awulkiew:
This is a fix for https://github.com/boostorg/geometry/issues/1025

vissarion (approved):
LGMT, thanks!

barendgehrels (approved):
Agree with all! Thanks!

awulkiew:
This PR caused buffer to stop compiling for input linestrings which `point_type` is different than `model::point`:
https://github.com/boostorg/geometry/runs/7120358803

The reason is that `buffer_inserter<linestring_tag, ...>` defines points of type taken from input Linestring:
https://github.com/awulkiew/geometry/blob/develop/include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp#L644-L645
but `helper_geometry` linestring always uses `model::point`:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/geometries/helper_geometry.hpp#L47

There are two fixes possible:
1. Make these points `auto const&`, however then `end_strategy.apply()` has to take various point types, now it's capable of taking only one:
https://github.com/awulkiew/geometry/blob/develop/include/boost/geometry/strategies/cartesian/buffer_end_round.hpp#L112
2. Convert these points to `output_point_type`

@barendgehrels which one do you prefer?

awulkiew:
Note that `segment_strategy` is capable of taking only one point type as well but since it takes both input point types it compiles:
https://github.com/awulkiew/geometry/blob/develop/include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp#L660

It is possible that there are parts of the code that also should fail for different input and output point types but are not tested.

awulkiew:
4. Using `point_type<Output>::type` with helper linestring/ring solves the issue and at the same time doesn't force us to refactor buffer. The reason is that with the current implementation output point type is used inside and expected as parameters in various places in `buffer_inserter` and `buffered_piece_collection`.

Issue #998, opened 2022-05-12
buffering features with integer coordinates may result in numeric overflow in segment length calculation


MaartenHilferink:
in "boost/geometry/strategies/cartesian/buffer_side_straight.hpp"

lines 78..83:

        // Simulate a vector d (dx,dy)
        coordinate_type const dx = get<0>(input_p2) - get<0>(input_p1);
        coordinate_type const dy = get<1>(input_p2) - get<1>(input_p1);

       // For normalization [0,1] (=dot product d.d, sqrt)
        promoted_type const length = geometry::math::sqrt(dx * dx + dy * dy);

fail to calculate a correct length when coordinate_type is int and dx * dx + dy * dy should be bigger than INT_MAX, which it technically never is as the result is also an int, which becomes negative when for example dx = dy = 47000.

When the calculation of dx and dy are promoted directly to promoted_type, this problem is fixed:

        // Simulate a vector d (dx,dy)
        promoted_type const dx = get<0>(input_p2) - get<0>(input_p1);
        promoted_type const dy = get<1>(input_p2) - get<1>(input_p1);

       // For normalization [0,1] (=dot product d.d, sqrt)
        promoted_type const length = geometry::math::sqrt(dx * dx + dy * dy);

MaartenHilferink:
Tested with: Boost 1.78.

MaartenHilferink:
seen with:
	boost::geometry::strategy::buffer::distance_symmetric<Float64> distStrategy(10.0);
	boost::geometry::strategy::buffer::join_round          joinStrategy(8);
	boost::geometry::strategy::buffer::end_round          endStrategy(8);
	boost::geometry::strategy::buffer::point_circle         circleStrategy(8);
	boost::geometry::strategy::buffer::side_straight       sideStrategy;
 
	using PointType = Point<int>; // for example: rd-coordinates in millimeters
	boost::geometry::model::ring<PointType> tmpRing;
	boost::geometry::model::polygon<PointType> tmpPolygon;
	boost::geometry::model::multi_polygon<decltype(tmpPolygon)> currMP, resMP;

	boost::geometry::buffer(currMP, resMP, distStrategy, sideStrategy, joinStrategy, endStrategy, circleStrategy);

where the following traits were defined for Point:

  namespace boost::geometry::traits
  {
	// define Point<T> as model for concept boost::geometry::point
	template <typename T> struct tag<Point<T>> { using type = point_tag; };
[...]

PR #1028, opened 2022-06-29
[buffer] fix for integer overflow

barendgehrels:
This fixes issue #998 
A similar issue was found in two other strategies.
Unit tests are added (apart from the fix for miter)

### DIFF
--- a/include/boost/geometry/strategies/cartesian/buffer_join_miter.hpp
+++ b/include/boost/geometry/strategies/cartesian/buffer_join_miter.hpp
@@ -89,8 +89,8 @@ public:
         // Check the distance ip-vertex (= miter distance)
         // (We calculate it manually (not using Pythagoras strategy) to reuse
         //  dx and dy)
-        coordinate_type const dx = get<0>(p) - get<0>(vertex);
-        coordinate_type const dy = get<1>(p) - get<1>(vertex);
+        promoted_type const dx = get<0>(p) - get<0>(vertex);
+        promoted_type const dy = get<1>(p) - get<1>(vertex);
 
         promoted_type const distance = geometry::math::sqrt(dx * dx + dy * dy);
 
--- a/include/boost/geometry/strategies/cartesian/buffer_join_round.hpp
+++ b/include/boost/geometry/strategies/cartesian/buffer_join_round.hpp
@@ -149,20 +149,8 @@ public :
             return false;
         }
 
-        // Generate 'vectors'
-        coordinate_type vix = (get<0>(ip) - get<0>(vertex));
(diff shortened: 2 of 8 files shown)


include/boost/geometry/strategies/cartesian/buffer_join_round.hpp
```diff
@@ -149,20 +149,8 @@ public :
...
-        coordinate_type vix = (get<0>(ip) - get<0>(vertex));
-        coordinate_type viy = (get<1>(ip) - get<1>(vertex));
-
-        promoted_type length_i = geometry::math::sqrt(vix * vix + viy * viy);
-        DistanceType const bd = geometry::math::abs(buffer_distance);
-        promoted_type prop = bd / length_i;
-
-        Point bp;
-        set<0>(bp, get<0>(vertex) + vix * prop);
-        set<1>(bp, get<1>(vertex) + viy * prop);
```
barendgehrels, line 162: This wasn't used...
We get this performance increase for free :-)

vissarion (approved):
LGTM, thanks.

awulkiew (approved):

include/boost/geometry/strategies/cartesian/buffer_join_miter.hpp
```diff
@@ -89,8 +89,8 @@ class join_miter
         // Check the distance ip-vertex (= miter distance)
         // (We calculate it manually (not using Pythagoras strategy) to reuse
         //  dx and dy)
-        coordinate_type const dx = get<0>(p) - get<0>(vertex);
-        coordinate_type const dy = get<1>(p) - get<1>(vertex);
+        promoted_type const dx = get<0>(p) - get<0>(vertex);
+        promoted_type const dy = get<1>(p) - get<1>(vertex);
```
awulkiew, line 93: Even though the probability is a lot smaller this could still cause overflow. It's ok for now but we could think about `determinant()` function which would internally perform integral computation using bigger types, potentially some simplified form of multiprecision.

awulkiew: Or you could first convert to `promoted_type` and subtract afterwards. With this the precision would be lesser but there wouldn't be an overflow in subtraction.

awulkiew: Btw, there would be a warning here about implicit converson from `int` to `double`.

barendgehrels: Thanks for your comments. Yes, there could be an overflow in subtraction if coordinates are huge and both plus and minus. I'll create an issue but will not handle it immediately. We probably have many of these kind of places in the library.

PR #1029, opened 2022-07-01
Fix compilation of touches for geometry collections with gcc-10

awulkiew:
Avoid -1 to size_t conversion while instantiating static mask. Relate is now called with proper mask based on run-time topological dimension of GCs just like it's done with other relational operations which masks depend on dimensions of arguments.

Also fix warning in relate.


barendgehrels (approved):

vissarion (approved):
thanks

PR #1030, opened 2022-07-02
[buffer] Use output_point_type with helper geometries instead of model::point

awulkiew:
This fixes compilation errors when output point type is different.

The issue was caused by this PR: https://github.com/boostorg/geometry/pull/1027


barendgehrels (approved):
Looks good, thanks

vissarion (approved):
LGTM thanks!

Issue #1012, opened 2022-06-05
C++03 deprecation message outdated


Flamefire:
There is a deprecation warning for C++03 targetting versions prior to Boost 1.75:

 https://github.com/boostorg/geometry/blob/d31cd27488a236ad4f4c4bd4023dc99fbe279d66/include/boost/geometry/geometry.hpp#L27

That should/can now be enforced as we will be at 1.80.

barendgehrels:
Thanks for the report! It's fixed and merged.

mloskot:
Yes, it's only feasible for tests and examples.
We have similar situation in GIL and the root `Jamfile` only declares projects for tests and examples
https://github.com/boostorg/gil/blob/develop/Jamfile

but the actual requirements like `cxx14_constexpr` go to `test/Jamfile`, e.g.
https://github.com/boostorg/gil/blob/develop/test/Jamfile#L26-L37

PR #1043, opened 2022-07-21
build: Add cxx14 requirements to root Jamfile

mloskot:
The library requires C++14 so it should not even be tried to build with any older `cxxstd`

IOW, `b2 cxxstd=11 libs/geometry/test` will build nothing.

Related to discussion in #1012


vissarion (approved):
Thanks!

Issue #604, opened 2019-06-21
covered_by not implemented for (Box, MultiPolygon)


vschoech:
Any chance this gets implemented any time soon?
See our conversation here:
https://lists.boost.org/geometry/2012/02/1838.php

barendgehrels:
Right, thanks for the reminder, that is a long time ago indeed.
From my side I still have to postpone this until rescaling removal is finished.
But thanks for opening the issue, it has a greater visibility than the mailing list.

digu-007:
[covered_by](https://www.boost.org/doc/libs/1_72_0/libs/geometry/doc/html/geometry/reference/algorithms/covered_by/covered_by_2.html) is implemented for (Polygon, Multipolygon), can't we just make use of this and convert Box into the rectangle (which is a Polygon) out of the given two corners of Box, and then solve this query for (Box, MultiPolygon).

If that's the case I would like to contribute towards the issue. :)

vissarion:
>covered_by is implemented for (Polygon, Multipolygon), can't we just make use of this and convert Box into the rectangle (which is a Polygon) out of the given two corners of Box, and then solve this query for (Box, MultiPolygon).

`Box` is a `Polygon` only in `cartesian` for other coordinate systems it cannot represented as a polygon since polygon's segments are geodesics (e.g. great circles for spherical) but box's segments are not.

digu-007:
> `Box` is a `Polygon` only in `cartesian` for other coordinate systems it cannot represented as a polygon since polygon's segments are geodesics (e.g. great circles for spherical) but box's segments are not.

Oh ok, I just now realized that we have four different kinds of coordinate systems here, I'll study them.

vschoech:
Still missing implementations for covered_by in boost 1.72.0.

vschoech:
Still missing implementations for covered_by in boost 1.73.0.

vschoech:
Still missing implementations for covered_by in boost 1.75.0.

awulkiew:
@vschoech out of curiosity, why do you need this combination for? Wouldn't checking
```
covered_by(box, return_envelope<Box>(multi_polygon));
```
be enough?

vschoech:
> ```
> covered_by(box, return_envelope<Box>(multi_polygon));
> ```

[...]

PR #1046, opened 2022-07-28
Support covered_by() for box, geometry combinations

vissarion:
This PR implements support for `covered_by()` for the following combinations of a box and a geometry:
- `convered_by(box, areal)` (only cartesian CS) 
- `convered_by(geometry, box)`

The first case resolves issue https://github.com/boostorg/geometry/issues/604


include/boost/geometry/algorithms/detail/covered_by/implementation.hpp
```diff
@@ -57,34 +60,26 @@ struct use_relate
...
-struct covered_by<Box1, Box2, box_tag, box_tag>
-{
-    template <typename Strategy>
-    static inline bool apply(Box1 const& box1, Box2 const& box2, Strategy const& strategy)
-    {
-        assert_dimension_equal<Box1, Box2>();
-        return strategy.covered_by(box1, box2).apply(box1, box2);
+        box_type box_areal;
+        geometry::envelope(geometry, box_areal, strategy);
+        return strategy.covered_by(box_areal, box).apply(box_areal, box);
```
awulkiew, line 73: This approach while giving the correct result is suboptimal because it cannot be interrupted  if e.g. one of the points/segments of a geometry is outside the box. The whole envelope is calculated each time. If that's acceptable then ok. But if it isn't I'd at least put here big TODO.

include/boost/geometry/algorithms/detail/relate/box_areal.hpp
```diff
@@ -0,0 +1,71 @@
...
+    template <typename Result, typename Strategy>
+    static inline void apply(Box const& box, Areal const& areal,
+                             Result & result,
+                             Strategy const& strategy)
+    {
+        using is_cartesian = std::is_same
+            <
+                typename traits::coordinate_system<typename traits::point_type<Box>::type>::type,
+                cs::cartesian
+            >;
```
awulkiew, line 38: You should probably use `Strategy::cs_tag` to distinguish between CSes.

include/boost/geometry/algorithms/detail/relate/box_areal.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+    template <typename Result, typename Strategy>
+    static inline void apply(Box const& box, Areal const& areal,
+                             Result & result,
+                             Strategy const& strategy,
+                             std::false_type /*is_cartesian*/)
+    {
+        BOOST_GEOMETRY_STATIC_ASSERT_FALSE(
+            "Not implemented for this coordinate system.",
+            typename traits::coordinate_system<typename traits::point_type<Box>::type>::type);
```
awulkiew, line 61: Is this a breaking change? AFAIU previously box was silently converted to ring in non-cartesian CSes, correct?

vissarion: I think we are OK since this operation was not supported in the past, i.e. `covered_by(box, polygon)` do not compile in any  cs before this PR.

vissarion:
@barendgehrels are you OK with this PR?

barendgehrels (approved):
I made some comments, still pending, but basically I'm OK. Thanks!

barendgehrels:
> @barendgehrels are you OK with this PR?

See my review remark, basically yes, but can you look at my comments I made earlier?

vissarion:
> > @barendgehrels are you OK with this PR?
> 
> See my review remark, basically yes, but can you look at my comments I made earlier?

Thanks for the review! Somehow your comments appeared to me yesterday. I addressed them and now I can merge this PR.

PR #1056, opened 2022-08-20
Fix c++20 compilation errors related to conditional expression

stefan301:
Fix c++20 compilation errors related to conditional expression with int and floating_point_type 

A floating_point_type with implicit conversions can lead to this C++20 compilation error:

boost\boost\geometry\algorithms\detail\overlay\get_turn_info.hpp(233,1): error C2445: result type of conditional expression is ambiguous: types '<floating_point_type>' and 'int' can be converted to multiple common types


barendgehrels (approved):
Thanks!

vissarion (approved):
Thanks

PR #1069, opened 2022-10-19
[buffer] harmonize geographic strategy code and share code

barendgehrels:
As discussed in last PR

### DIFF
--- a/include/boost/geometry/strategies/geographic/buffer_join_round.hpp
+++ b/include/boost/geometry/strategies/geographic/buffer_join_round.hpp
@@ -38,10 +38,9 @@ class geographic_join_round
 public :
 
     //! \brief Constructs the strategy
-    //! \param points_per_circle Number of points which would be used for a full circle
-    //! (if points_per_circle is smaller than 4, it is internally set to 4)
-    explicit inline geographic_join_round(std::size_t points_per_circle = 90)
-        : m_points_per_circle((points_per_circle < 4u) ? 4u : points_per_circle)
+    //! \param points_per_circle Number of points (minimum 4) that would be used for a full circle
+    explicit inline geographic_join_round(std::size_t points_per_circle = default_points_per_circle)
+        : m_points_per_circle(get_point_count_for_join(points_per_circle))
     {}
 
 #ifndef DOXYGEN_SHOULD_SKIP_THIS
--- a/include/boost/geometry/strategies/geographic/buffer_join_round.hpp
+++ b/include/boost/geometry/strategies/geographic/buffer_join_round.hpp
@@ -15,6 +15,7 @@
 
 #include <boost/geometry/srs/spheroid.hpp>
 #include <boost/geometry/strategies/buffer.hpp>
+#include <boost/geometry/strategies/geographic/buffer_helper.hpp>
 #include <boost/geometry/strategies/geographic/parameters.hpp>
 #include <boost/geometry/util/math.hpp>
 #include <boost/geometry/util/select_calculation_type.hpp>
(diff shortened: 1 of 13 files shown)


include/boost/geometry/strategies/geographic/buffer_join_round.hpp
```diff
@@ -71,52 +59,34 @@ public :
...
+        if (! helper::calculate_angles(lon_rad, lat_rad, perp1, perp2, m_spheroid,
+                                       angle_diff, first_azimuth))
         {
-            // Defensive check with asserts
-            BOOST_GEOMETRY_ASSERT(angle_diff >= 0);
-            BOOST_GEOMETRY_ASSERT(angle_diff <= geometry::math::pi<calc_t>());
             return false;
         }
 
+        static calc_t const two_pi = geometry::math::two_pi<calc_t>();
```
awulkiew, line 75: This is probably out of scope of this PR but we could replace `static` variables with `constexpr` variables. For this to work math constants would have to be `constexpr` functions.

include/boost/geometry/strategies/geographic/buffer_side_straight.hpp
```diff
@@ -52,21 +54,23 @@ public :
     template
     <
         typename Point,
-        typename OutputRange,
+        typename RangeOut,
```
awulkiew, line 57: Is there a reason why template arguments are in different order than function arguments?

awulkiew (approved):

include/boost/geometry/strategies/geographic/buffer_point_circle.hpp
```diff
@@ -55,77 +59,67 @@ template
...
+
     //! \brief Constructs the strategy
-    //! \param count number of points for the created circle (if count
-    //! is smaller than 3, count is internally set to 3)
-    explicit geographic_point_circle(std::size_t count = 90)
-        : m_count((count < 3u) ? 3u : count)
+    //! \param points_per_circle Number of points for a full circle
+    //! (if points_per_circle is smaller than 4, it is internally set to 4)
+    explicit geographic_point_circle(std::size_t points_per_circle = 90)
+        : m_points_per_circle((points_per_circle < 4u) ? 4u : points_per_circle)
```
vissarion, line 67: what is the point of changing this from 3 to 4? Is this for robustness?

vissarion (approved):

include/boost/geometry/strategies/geographic/buffer_join_round.hpp
barendgehrels: Indeed, I'll do that later in a separate PR

include/boost/geometry/strategies/geographic/buffer_side_straight.hpp
barendgehrels: Thanks, no, I'll update it.

include/boost/geometry/strategies/geographic/buffer_point_circle.hpp
barendgehrels: One had 3, the other had 4. I liked to harmonize it. I can make both `3` too. Either choice is fine, it's very low anyway, not really circle-like.

include/boost/geometry/strategies/geographic/buffer_side_straight.hpp
barendgehrels: :heavy_check_mark:

include/boost/geometry/strategies/geographic/buffer_point_circle.hpp
barendgehrels: For the join it was also 3 but that was erroneous, I made it minimum 4 to get well-shaped symmetric joins.
I don't think it was ever used with 3 (default = 90)

Issue #1063, opened 2022-10-08
Bug in intersection of simple spherical polygons


rconde01:
The following code shows the intersection of two simple spherical polygons. The first is a ring around the z-axis defining a sphere truncated from the bottom. The other is a ring defining a half-sphere. When the latter is rotated about y by 90 degrees, you get a null result where you would expect it to cut the first polygon in half. A 270 degree rotation works as expected, as does a 89 and 91 degree rotation.

I'm not sure, but the problem appears to be in [add_rings.hpp](https://github.com/boostorg/geometry/blob/b18db001b813f486b014744e9dd920e894bc3523/include/boost/geometry/algorithms/detail/overlay/add_rings.hpp#L145) where the area of the result comes out negative and is therefore rejected. Maybe this [comment](https://github.com/boostorg/geometry/blob/b18db001b813f486b014744e9dd920e894bc3523/include/boost/geometry/algorithms/detail/overlay/overlay.hpp#L383) is relevant also.

https://godbolt.org/z/7h5Y9nxev

```
#include <array>
#include <boost/geometry.hpp>
#include <boost/geometry/geometries/point_xy.hpp>
#include <boost/geometry/geometries/polygon.hpp>
#include <cmath>
#include <iostream>
#include <vector>

namespace bg = boost::geometry;

using SphericalPoint =
    bg::model::d2::point_xy<double, bg::cs::spherical_equatorial<bg::degree>>;

using SphericalPolygon =
    boost::geometry::model::polygon<SphericalPoint,
                                    false,  // counter-clockwise
                                    true,   // closed
                                    std::vector, std::vector, std::allocator,
                                    std::allocator>;

auto rad(double deg) -> double { return M_PI * deg / 180.0; }

auto deg(double rad) -> double { return 180.0 * rad / M_PI; }

auto create_truncated_sphere_unit_vectors(double cone_half_angle_deg)
    -> std::vector<std::array<double, 3>> {
    auto const ring_radius = sin(rad(cone_half_angle_deg));
    auto const ring_height = cos(rad(cone_half_angle_deg));

    std::vector<std::array<double, 3>> result;

    auto const num_ring_points = 361;

    result.reserve(num_ring_points);

    for (size_t i = 0; i < num_ring_points - 1; ++i) {
        auto const angle_deg = (360.0 * i) / num_ring_points;
        result.push_back({ring_radius * cos(rad(angle_deg)),
                          ring_radius * sin(rad(angle_deg)), ring_height});
    }

[...]

PR #1074, opened 2022-10-21
Fix special case of area for segments that pass through a pole.

vissarion:
This PR proposes a fix for https://github.com/boostorg/geometry/issues/1063

### DIFF
--- a/include/boost/geometry/formulas/area_formulas.hpp
+++ b/include/boost/geometry/formulas/area_formulas.hpp
@@ -372,7 +372,7 @@ public:
         CT const lon2r = get_as_radian<0>(p2);
         CT const lat2r = get_as_radian<1>(p2);
 
-        CT lon12r = lon1r - lon2r;
+        CT lon12r = lon2r - lon1r;
         math::normalize_longitude<radian, CT>(lon12r);
 
         if (lon12r == pi || lon12r == -pi)
@@ -387,7 +387,7 @@ public:
             CT cbet2 = cos(lat2r);
             CT sbet2 = sin(lat2r);
 
-            CT omg12 = lon1r - lon2r;
+            CT omg12 = lon2r - lon1r;
             CT comg12 = cos(omg12);
             CT somg12 = sin(omg12);
 
@@ -400,7 +400,7 @@ public:
 
         } else {
 
-            excess = trapezoidal_formula(lat1r, lat2r, -lon12r);
+            excess = trapezoidal_formula(lat1r, lat2r, lon12r);
         }
 
         return excess;
@@ -487,7 +487,7 @@ public:
 
         CT excess;
 
-        CT lon12r = lon1r - lon2r;
+        CT lon12r = lon2r - lon1r;
         math::normalize_longitude<radian, CT>(lon12r);
 
         if (lon12r == pi || lon12r == -pi)
@@ -503,7 +503,7 @@ public:
             if (!meridian && (i_res.distance)
                 < mean_radius<CT>(spheroid_const.m_spheroid) / CT(638))  // short segment
             {
-                excess = trapezoidal_formula(lat1r, lat2r, -lon12r);
+                excess = trapezoidal_formula(lat1r, lat2r, lon12r);
             }
             else
             {
@@ -587,7 +587,7 @@ public:
         CT const lon1r = get_as_radian<0>(p1);
         CT const lon2r = get_as_radian<0>(p2);
 
-        CT lon12 = lon1r - lon2r;
+        CT lon12 = lon2r - lon1r;
         math::normalize_longitude<radian, CT>(lon12);
 
         if (lon12 == pi || lon12 == -pi)
@@ -595,8 +595,8 @@ public:
             return true;
         }
 
-        CT p1_lon = lon1r - ( floor( lon1r / two_pi ) * two_pi );
-        CT p2_lon = lon2r - ( floor( lon2r / two_pi ) * two_pi );
+        CT p1_lon = lon1r - ( std::floor( lon1r / two_pi ) * two_pi );
+        CT p2_lon = lon2r - ( std::floor( lon2r / two_pi ) * two_pi );
 
         CT max_lon = (std::max)(p1_lon, p2_lon);
         CT min_lon = (std::min)(p1_lon, p2_lon);
--- a/include/boost/geometry/formulas/area_formulas.hpp
+++ b/include/boost/geometry/formulas/area_formulas.hpp
@@ -342,6 +342,15 @@ public:
         }
     }
 
+    static inline CT trapezoidal_formula(CT lat1r, CT lat2r, CT lon21r)
+    {
+        CT const c1 = CT(1);
+        CT const c2 = CT(2);
+        CT tan_lat1 = tan(lat1r / c2);
+        CT tan_lat2 = tan(lat2r / c2);
+
+        return c2 * atan(((tan_lat1 + tan_lat2) / (c1 + tan_lat1 * tan_lat2))* tan(lon21r / c2));
+    }
 
     /*
         Compute the spherical excess of a geodesic (or shperical) segment
@@ -354,8 +363,6 @@ public:
     static inline CT spherical(PointOfSegment const& p1,
                                PointOfSegment const& p2)
     {
-        CT const c1 = CT(1);
-        CT const c2 = CT(2);
         CT const pi = math::pi<CT>();
 
         CT excess;
@@ -365,10 +372,10 @@ public:
         CT const lon2r = get_as_radian<0>(p2);
         CT const lat2r = get_as_radian<1>(p2);
 
-        CT lon12 = lon1r - lon2r;
-        math::normalize_longitude<radian, CT>(lon12);
+        CT lon12r = lon1r - lon2r;
+        math::normalize_longitude<radian, CT>(lon12r);
 
-        if (lon12 == pi || lon12 == -pi)
+        if (lon12r == pi || lon12r == -pi)
         {
             return pi;
         }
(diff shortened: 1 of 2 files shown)


include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -354,43 +363,44 @@ class area_formulas
...
+
         CT excess;
 
-        if (LongSegment) // not for segments parallel to equator
+        CT const lon1r = get_as_radian<0>(p1);
+        CT const lat1r = get_as_radian<1>(p1);
+        CT const lon2r = get_as_radian<0>(p2);
+        CT const lat2r = get_as_radian<1>(p2);
+
+        CT lon12r = lon1r - lon2r;
```
awulkiew, line 375: I'm aware that this naming was used before but I'm curious about the convention. Shouldn't this variable be called `lon21r` since it's an angle from `lon2r` to `lon1r`?

Azimuth between points of coordinates `lon1, lat1` and `lon2, lat2` is called `a12` in the code. Other variables referred to the azimuth e.g. in direct and inverse formulas are also called this way. Basically the order of numbers defines the direction.

Here the order of numbers defines the subtraction. This is also fine but I'm wondering whether or not it may be confusing.

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -563,17 +581,22 @@ class area_formulas
...
+        CT lon12 = lon1r - lon2r;
+        math::normalize_longitude<radian, CT>(lon12);
+
+        if (lon12 == pi || lon12 == -pi)
+        {
+            return true;
+        }
+
+        CT p1_lon = lon1r - ( floor( lon1r / two_pi ) * two_pi );
+        CT p2_lon = lon2r - ( floor( lon2r / two_pi ) * two_pi );
```
awulkiew, line 599: `math::floor` or `std::floor`?

awulkiew (approved):

include/boost/geometry/formulas/area_formulas.hpp
vissarion: aah that was there already, but you are right, I should be more specific. `std::floor` should be better.

include/boost/geometry/formulas/area_formulas.hpp
vissarion: Good point! The wikipedia https://en.wikipedia.org/wiki/Geodesics_on_an_ellipsoid is also using what you propose (i.e. $\lambda_{12}=\lambda_2-\lambda_1$)
In my case since only the absolute value matters I change the order of the subtraction (following the convention you propose). That also simplifies the call to `trapezoidal_formula()` below.

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -563,17 +581,22 @@ class area_formulas
...
-        CT p2_lon = get_as_radian<0>(p2)
-                                - ( floor( get_as_radian<0>(p2) / two_pi )
-                                  * two_pi );
+        CT const pi = geometry::math::pi<CT>();
+        CT const two_pi = geometry::math::two_pi<CT>();
+
+        CT const lon1r = get_as_radian<0>(p1);
+        CT const lon2r = get_as_radian<0>(p2);
+
+        CT lon12 = lon1r - lon2r;
```
awulkiew, line 590: Here is the last one. :)

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -342,6 +342,15 @@ class area_formulas
         }
     }
 
+    static inline CT trapezoidal_formula(CT lat1r, CT lat2r, CT lon21r)
+    {
+        CT const c1 = CT(1);
+        CT const c2 = CT(2);
+        CT tan_lat1 = tan(lat1r / c2);
+        CT tan_lat2 = tan(lat2r / c2);
```
barendgehrels, line 350: Can be `const` as well

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -472,43 +487,46 @@ class area_formulas
 
         CT excess;
 
-        auto const half_pi = math::pi<CT>() / 2;
-        bool meridian = lon2r - lon1r == CT(0)
-            || lat1r == half_pi || lat1r == -half_pi
-            || lat2r == half_pi || lat2r == -half_pi;
+        CT lon12r = lon2r - lon1r;
+        math::normalize_longitude<radian, CT>(lon12r);
```
barendgehrels, line 491: -> `const`

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -472,43 +487,46 @@ class area_formulas
...
 
-        if (!meridian && (i_res.distance)
-            < mean_radius<CT>(spheroid_const.m_spheroid) / CT(638))  // short segment
+        if (lon12r == pi || lon12r == -pi)
```
barendgehrels, line 493: Is a double comparison `==` OK in this case?
Probably, but maybe you can add a comment

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -472,43 +487,46 @@ class area_formulas
...
-
-            CT cos_omg12p1 = CT(1) + cos_omg12;
-            CT cos_bet1p1 = CT(1) + cos_bet1;
-            CT cos_bet2p1 = CT(1) + cos_bet2;
-            excess = CT(2) * atan2(sin_omg12 * (sin_bet1 * cos_bet2p1 + sin_bet2 * cos_bet1p1),
-                                   cos_omg12p1 * (sin_bet1 * sin_bet2 + cos_bet1p1 * cos_bet2p1));
-            */
-
-            excess = alp2 - alp1;
+            bool meridian = lon12r == c0
```
barendgehrels, line 499: -> `const`

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -563,17 +581,22 @@ class area_formulas
...
+        CT const pi = geometry::math::pi<CT>();
+        CT const two_pi = geometry::math::two_pi<CT>();
+
+        CT const lon1r = get_as_radian<0>(p1);
+        CT const lon2r = get_as_radian<0>(p2);
+
+        CT lon12 = lon2r - lon1r;
+        math::normalize_longitude<radian, CT>(lon12);
+
+        if (lon12 == pi || lon12 == -pi)
```
barendgehrels, line 593: Same here, `==` OK?

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -563,17 +581,22 @@ class area_formulas
...
-        CT p2_lon = get_as_radian<0>(p2)
-                                - ( floor( get_as_radian<0>(p2) / two_pi )
-                                  * two_pi );
+        CT const pi = geometry::math::pi<CT>();
+        CT const two_pi = geometry::math::two_pi<CT>();
+
+        CT const lon1r = get_as_radian<0>(p1);
+        CT const lon2r = get_as_radian<0>(p2);
+
+        CT lon12 = lon2r - lon1r;
```
barendgehrels, line 590: -> `const`

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -563,17 +581,22 @@ class area_formulas
...
+        math::normalize_longitude<radian, CT>(lon12);
+
+        if (lon12 == pi || lon12 == -pi)
+        {
+            return true;
+        }
+
+        CT p1_lon = lon1r - ( std::floor( lon1r / two_pi ) * two_pi );
+        CT p2_lon = lon2r - ( std::floor( lon2r / two_pi ) * two_pi );
```
barendgehrels, line 599: -> `const`

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -354,43 +363,44 @@ class area_formulas
...
-                            - sbet1 * cbet2 * comg12,
-                            cbet2 * somg12);
-
-            CT alp2 = atan2(cbet1 * sbet2 * comg12
-                            - sbet1 * cbet2,
-                            cbet1 * somg12);
+            CT const cbet1_sbet2 = cbet1 * sbet2;
+            CT const sbet1_cbet2 = sbet1 * cbet2;
+            CT alp1 = atan2(cbet1_sbet2 - sbet1_cbet2 * comg12, cbet2 * somg12);
+            CT alp2 = atan2(cbet1_sbet2 * comg12 - sbet1_cbet2, cbet1 * somg12);
```
barendgehrels, line 397: -> `const` most of the constants above

barendgehrels (approved):
Thanks! Skipping the math details in the review

include/boost/geometry/formulas/area_formulas.hpp
vissarion: `lon12r` cannot be `const` since `normalize_longitude` is changing it.

include/boost/geometry/formulas/area_formulas.hpp
vissarion: OK I added a comment

include/boost/geometry/formulas/area_formulas.hpp
vissarion: cannot declare `const` here (see comment above)

include/boost/geometry/formulas/area_formulas.hpp
vissarion: yes, I added a similar comment

vissarion:
@barendgehrels thanks for the review comments, I replied to all and pushed two new commits for addressing them.

include/boost/geometry/formulas/area_formulas.hpp
barendgehrels: All right, missed that, thanks

PR #1078, opened 2022-10-24
Fix a few warnings

vissarion:


include/boost/geometry/util/math.hpp
```diff
@@ -859,16 +859,16 @@ inline void sin_cos_degrees(T const& x,
 {
     // In order to minimize round-off errors, this function exactly reduces
     // the argument to the range [-45, 45] before converting it to radians.
-    T remainder; int quotient;
 
-    remainder = math::mod(x, T(360));
-    quotient = floor(remainder / 90 + T(0.5));
-    remainder -= 90 * quotient;
+    T remainder = math::mod(x, T(360));
+    T quotient = std::floor(remainder / T(90) + T(0.5));
```
awulkiew, line 864: Would it make sense to use `const`?

include/boost/geometry/util/math.hpp
```diff
@@ -859,16 +859,16 @@ inline void sin_cos_degrees(T const& x,
...
+    remainder -= T(90) * quotient;
 
     // Convert to radians.
     remainder *= d2r<T>();
 
-    T s = sin(remainder), c = cos(remainder);
+    T s = sin(remainder);
+    T c = cos(remainder);
```
awulkiew, line 871: Here as well.

awulkiew (approved):

barendgehrels (approved):
Thanks!

Issue #1088, opened 2022-11-10
Boost 1.80.0 causes regression for custom point types


sehe:
As discovered here https://stackoverflow.com/a/74392274/85371 simple usecases where custom point types are used no longer compile with Boost 1.80.0:

```c++
#include <boost/geometry.hpp>
#include <boost/geometry/geometries/register/point.hpp>
#include <iostream>
namespace bg = boost::geometry;

struct MyPoint {
    double x, y;
    MyPoint(double x = 0, double y = 0) : x(x), y(y) {}
};

BOOST_GEOMETRY_REGISTER_POINT_2D(MyPoint, double, bg::cs::cartesian, x, y);

using MyLine      = bg::model::linestring<MyPoint>;
using MyMultiline = bg::model::multi_linestring<MyLine>;
using MyPolygon   = bg::model::polygon<MyPoint>;

int main() {
    MyMultiline inMultiline;
    MyPolygon   inPoly;
    bg::within(inMultiline, inPoly); // error in boost 1.80.0
}
```

See it on Compiler Explorer [with boost 1.79.0](https://compiler-explorer.com/z/dTr5azMTn) and failing with [boost 1.80.0](https://compiler-explorer.com/z/dTr5azMTn).

I have analyzed it on stackoverflow and came up with an explanation and patch that would work for mutable input geometries. And, indeed, subsequently it turned out that the regression was introduced by 577de29211a97c which ... fixes boundary_checker for const point types.

I don't think I'm qualified to come up with a proper fix, but hoping that at least these pointers will speed up the analysis.

PR #1091, opened 2022-11-11
Fix mutable point issue and typos in boundary_checker.

vissarion:
Fixes https://github.com/boostorg/geometry/issues/1088


awulkiew (approved):

barendgehrels (approved):
Thanks! Apparently we don't have a unit test for this yet (though we have `relate_const.cpp`...)

PR #1093, opened 2022-11-16
[overlay] fix cluster by adapting tolerance

### LINKED ISSUE #1081: bg::union_(polygon,ring) returns wrong result. Inner loop disappears.
image
Output1 and output2 should have same number of inner loops but output1 has two (correct) and output2 has one (wrong).

boost1.80. Visual Studio 2017.
```
void Test()
{
    using point = bg::model::point<double, 2, bg::cs::cartesian>;
    bg::model::polygon<point> polygon;
    polygon.outer().assign({
        {40,0},
        {50,100},
        {75,60},
        {60,60},
        {75,40},
        {40,00}
        });

    polygon.inners().push_back({
        {46.317137830014417,52.146832892889620},
        {59.931267382602982,67.241906564861040},
        {59.931267382602982,73.678978034969646},
        {60,75},
        {46.317137830014417,52.146832892889620}
        });

    bg::model::ring<point> ring{
        {65,70},
        {67,40},
        {65,63},
        {56.765190742466075,63.715020193369355},
        {59.931267382602982,67.241906564861154},
        {65,70}
    };

    bg::model::multi_polygon<bg::model::polygon<point>> output1;
    bg::union_(polygon, ring, output1);

    auto MovePoint = [](auto& point) {
        point.set<1>(point.get<1>() + 100);
    };
    bg::for_each_point(polygon, MovePoint);
    bg::for_each_point(ring, MovePoint);

    bg::model::multi_polygon<bg::model::polygon<point>> output2;
    bg::union_(polygon, ring, output2);

    assert(output1.size() == 1 && output2.size() == 1 && output1[0].inners().size() == output2[0].inners().size());

    return 0;
}
```

barendgehrels:
Fixes #1081


doc/release_notes.qbk
```diff
@@ -26,6 +26,7 @@
 [*Solved issues]
 
 * [@https://github.com/boostorg/geometry/issues/1076 1076] Union: in rare cases it might miss one polygon
+* [@https://github.com/boostorg/geometry/issues/1081 1081] Union: due to precision it might miss interior rings
```
barendgehrels, line 29: @vissarion if this will not make it in `1.81` then I will update the PR (later)

vissarion (approved):
Thanks! 

So threshold `1000` was too large and now `3` works for all the old cases plus the issue case. Is it possible to make this threshold adaptive in the future? Does it depend on the input?

barendgehrels:
> Thanks!
> 
> So threshold `1000` was too large and now `3` works for all the old cases plus the issue case. Is it possible to make this threshold adaptive in the future? Does it depend on the input?

That's right.

Yes, it might be possible to make it adaptive. I will study that later. For that I should create a new issue. It is convenient to use the input of the reported #1081 issue.

barendgehrels:
> > Thanks!
> > So threshold `1000` was too large and now `3` works for all the old cases plus the issue case. Is it possible to make this threshold adaptive in the future? Does it depend on the input?
> 
> That's right.
> 
> Yes, it might be possible to make it adaptive. I will study that later. For that I should create a new issue. It is convenient to use the input of the reported #1081 issue.

Made it right away #1094

PR #1102, opened 2022-12-14
Fix github actions

vissarion:
Github actions is failing currently. This is due to a change in ubuntu version used by the CI. It used to be 20.04 (see https://github.com/boostorg/geometry/actions/runs/3591991978/jobs/6047189522#step:1:4) but now it is 22.04.1 (see https://github.com/boostorg/geometry/actions/runs/3594906729/jobs/6179253324#step:1:4). This is expected since `ubuntu-latest` can be either 20.04 or 22.04 depending on the runner (see https://github.com/actions/runner-images#ongoing-migrations). 

This PR changes the runner from `ubuntu-latest` to `ubuntu-20.04`. If we want to switch to `22.04` we either a) have to fix issues with some old versions of compilers e.g. `clang-3.9` that cannot be installed in 22.04 (`libc6-dev : Breaks: libgcc-7-dev (< 7.5.0-6~) but 7.3.0-16ubuntu3 is to be installed`) or b) decide to stop testing with those compilers or c) use Ubuntu 20.04 for old compilers and 22.04 for new ones.

This PR also fixes [a deprecation warning for set-output CI command](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands).

### DIFF
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -4,6 +4,9 @@
 # Copyright (c) 2020 Mateusz Loskot <mateusz@loskot.net>
 # Copyright (c) 2020-2021 Adam Wulkiewicz, Lodz, Poland
 #
+# Copyright (c) 2022 Oracle and/or its affiliates.
+# Contributed and/or modified by Vissarion Fysikopoulos, on behalf of Oracle
+#
 # Use, modification and distribution is subject to the Boost Software License,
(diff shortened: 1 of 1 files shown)


barendgehrels (approved):
I assume that this is related to the key errors I got. Great, thanks a lot.

mloskot (approved):

Issue #1104, opened 2022-12-17
Example in release needs boost/geometry/extensions which is not shipped


sgn:
Hello,

Can you either ship `boost/geometry/extensions` in the release or remove those examples from release?

As of it is, the example are all useless because it needs  `boost/geometry/extensions`

vissarion:
Thanks for the report. The release you mean is version `1.81.0`. Indeed this is an issue that will be fixed in `1.82.0` (we cannot change/update the current release). See https://github.com/boostorg/geometry/pull/1105.

PR #1105, opened 2022-12-19
Fix two issues in examples 

vissarion:
Fix https://github.com/boostorg/geometry/issues/1104 (also fix another issue in examples).

### DIFF
--- a/example/ml02_distance_strategy.cpp
+++ b/example/ml02_distance_strategy.cpp
@@ -1,6 +1,9 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
 // Copyright (c) 2009-2012 Mateusz Loskot, London, UK.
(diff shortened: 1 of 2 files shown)


barendgehrels (approved):
Thanks

example/ml02_distance_strategy.cpp
```diff
@@ -23,14 +26,14 @@ int main()
   point_xy p2(5.0, 0.0);
 
   // 1) This is direct call to Pythagoras algo
-  typedef strategy::distance::pythagoras<point_xy, point_xy, double> strategy1_type;
+  using strategy1_type = strategy::distance::pythagoras<double>;
```
awulkiew, line 29: I'm wondering, should we use umbrella strategies in examples and documentation now?

awulkiew (approved):

PR #1120, opened 2023-03-14
Fix buffer tests by using more accurate area computation

vissarion:
This PR changes the way area is computed in the unit tests of buffered geographic geometries. The problem is that for small geometries all strategies (andoyer, thomas, vincenty) but the series expansion one (karney) return inaccurate and/or unstable results. That is with a small change in the input or the number type the changes in the output are large.

For example consider [this](https://github.com/boostorg/geometry/commit/6a7224e369b958765daa719435371978863efe15#diff-4a770cf757384c3154247660df0dde42a4bc692d1b1e470e5c2b81a50c0640b9R71) unit test; We can get the following values for different strategies and number types

|   | andoyer  | thomas  | vincenty  | 
|---|---|---|---|
| float  |  31638 | 31633  | 31406  |   
| double  | 31450  | 31414  | 31343  |   
|  long double |  31469 | 31414  | 31403  |   

For floats the results of geographic buffer are not correct (e.g. creation of not valid geometries and assertions of zero denominators). 

*Note:* Buffer tests are now passing on [circleCI](https://app.circleci.com/pipelines/github/vissarion/geometry/258/workflows/74f7489b-2b6e-4eb2-8904-131f9335bd42/jobs/1817)


test/algorithms/buffer/test_buffer.hpp
```diff
@@ -357,7 +358,8 @@ void test_buffer(std::string const& caseid,
 
     if (settings.test_area)
     {
-        auto const area = bg::area(buffered, strategy);
+        bg::strategies::buffer::geographic<bg::strategy::karney, bg::srs::spheroid<double>> kstrategy;
```
awulkiew, line 361: Is `test_buffer` function only used for geographic CS?

Since different spheroids are used above, shouldn't they be passed here, i.e. extracted from `strategy`? This probably is not a problem since the issue is not correctness but consitency across compilers but still.

test/algorithms/buffer/buffer_geo_spheroid.cpp
```diff
@@ -57,24 +57,29 @@ void test_point(std::string const& label, Spheroid const& spheroid, double expec
...
-    using test_type = default_test_type;
-
     using point_t = bg::model::point<test_type, 2, bg::cs::geographic<bg::degree>>;
 
     // Use the default spheroid
     bg::srs::spheroid<test_type> def_spheroid;
-    test_linestring<bg::strategy::andoyer, true, point_t>("line_def", def_spheroid, 7996.0, 8093.0);
-    test_point<bg::strategy::andoyer, true, point_t>("point_def", def_spheroid, 31450.0);
+    test_linestring<bg::strategy::andoyer, true, point_t>("line_def", def_spheroid, 8046.0, 8143.0);
+    test_point<bg::strategy::andoyer, true, point_t>("point_def", def_spheroid, 31414.0);
```
barendgehrels, line 68: I don't see what really changed, apart from the expectations.

But you mentioned it's a personal test, so that's fine.

vissarion:
@barendgehrels @awulkiew now it is ready for review

vissarion:
If you agree with the changes of this PR I can also apply the change (always compute area with karney strategy) to other geographic buffer unit tests as well.

test/algorithms/buffer/buffer_geo_spheroid.cpp
```diff
@@ -16,6 +18,38 @@ namespace
...
+                std::enable_if_t<! bg::util::is_box<Geometry>::value> * = nullptr) const
+    {
+        return bg::strategy::area::geographic
+            <
+                bg::strategy::karney,
+                bg::strategy::default_order<bg::strategy::karney>::value,
+                Spheroid, CalculationType
+            >(base_t::m_spheroid);
+    }
+};
```
barendgehrels, line 51: Thanks. So it uses karney, whatever the other formula is. Fine for me.
Maybe this option should be documented, it's good to know for users.
But that can be done in another PR.

barendgehrels (approved):

test/algorithms/buffer/buffer_geo_spheroid.cpp
vissarion: Yes, it needs some further investigation. I will work on it. We know that those formulas (andoyer, thomas, vincenty) are unstable in same corner cases (antipodal points, points very close to each other, meridian/equatorial cases). Those results can be OK/expected for distance or azimuth but if they are used inside other algorithms like area or buffer the results can be unexpected. In general the Geographic part of the library is not documented.

awulkiew (approved):

PR #1122, opened 2023-03-15
Update CMakeLists.txt

vissarion:


mloskot:
These list actually states dependencies of tests, doesn't it?
Or is it dependencies of the core library, or both?

PR #1129, opened 2023-04-10
use auto instead of iterator types, and related

barendgehrels:
It's a big PR but quite mechanical

### DIFF
--- a/include/boost/geometry/algorithms/centroid.hpp
+++ b/include/boost/geometry/algorithms/centroid.hpp
@@ -43,7 +43,6 @@
 #include <boost/geometry/algorithms/assign.hpp>
 #include <boost/geometry/algorithms/convert.hpp>
 #include <boost/geometry/algorithms/detail/centroid/translating_transformer.hpp>
-#include <boost/geometry/algorithms/detail/interior_iterator.hpp>
 #include <boost/geometry/algorithms/detail/point_on_border.hpp>
 #include <boost/geometry/algorithms/detail/visit.hpp>
 #include <boost/geometry/algorithms/is_empty.hpp>
@@ -255,11 +254,9 @@ struct centroid_polygon_state
     {
         centroid_range_state::apply(exterior_ring(poly), transformer, strategy, state);
 
-        typename interior_return_type<Polygon const>::type
-            rings = interior_rings(poly);
-
-        for (typename detail::interior_iterator<Polygon const>::type
-                it = boost::begin(rings); it != boost::end(rings); ++it)
+        auto const& rings = interior_rings(poly);
+        auto const end = boost::end(rings);
+        for (auto it = boost::begin(rings); it != end; ++it)
         {
             centroid_range_state::apply(*it, transformer, strategy, state);
         }
@@ -352,15 +349,12 @@ struct centroid_multi
                 Point
             >::type state;
 
-        for (typename boost::range_iterator<Multi const>::type
-                it = boost::begin(multi);
-            it != boost::end(multi);
-            ++it)
+        for (auto it = boost::begin(multi); it != boost::end(multi); ++it)
         {
             Policy::apply(*it, transformer, strategy, state);
         }
 
-        if ( strategy.result(state, centroid) )
+        if (strategy.result(state, centroid))
         {
             // translate the result back
             transformer.apply_reverse(centroid);
--- a/include/boost/geometry/algorithms/densify.hpp
+++ b/include/boost/geometry/algorithms/densify.hpp
@@ -78,11 +78,10 @@ struct densify_range
     static inline void apply(FwdRng const& rng, MutRng & rng_out,
                              T const& len, Strategies const& strategies)
     {
-        typedef typename boost::range_iterator<FwdRng const>::type iterator_t;
         typedef typename boost::range_value<FwdRng>::type point_t;
 
-        iterator_t it = boost::begin(rng);
-        iterator_t end = boost::end(rng);
+        auto it = boost::begin(rng);
+        auto const end = boost::end(rng);
 
         if (it == end) // empty(rng)
         {
@@ -92,7 +91,7 @@ struct densify_range
         auto strategy = strategies.densify(rng);
         push_back_policy<MutRng> policy(rng_out);
 
-        iterator_t prev = it;
+        auto prev = it;
         for ( ++it ; it != end ; prev = it++)
         {
             point_t const& p0 = *prev;
@@ -123,9 +122,8 @@ struct densify_ring
         if (boost::size(ring) <= 1)
             return;
 
-        typedef typename point_type<Geometry>::type point_t;
-        point_t const& p0 = range::back(ring);
-        point_t const& p1 = range::front(ring);
+        auto const& p0 = range::back(ring);
+        auto const& p1 = range::front(ring);
 
         auto strategy = strategies.densify(ring);
         push_back_policy<GeometryOut> policy(ring_out);
--- a/include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
@@ -59,6 +59,7 @@
 #include <boost/geometry/algorithms/detail/sections/section_box_policies.hpp>
 
 #include <boost/geometry/views/detail/closed_clockwise_view.hpp>
+#include <boost/geometry/util/for_each_with_index.hpp>
 #include <boost/geometry/util/range.hpp>
 
 
@@ -295,12 +296,9 @@ struct buffered_piece_collection
         // and only applicable for linear features
         // (in a multi linestring with many short lines, the #endpoints can be
         // much higher)
-        for (typename boost::range_iterator<std::vector<point_type> const>::type it
-             = boost::begin(m_linear_end_points);
-             it != boost::end(m_linear_end_points);
-             ++it)
+        for (auto const& p : m_linear_end_points)
         {
-            if (detail::equals::equals_point_point(turn.point, *it, m_strategy))
+            if (detail::equals::equals_point_point(turn.point, p, m_strategy))
             {
                 turn.is_linear_end_point = true;
             }
@@ -324,12 +322,9 @@ struct buffered_piece_collection
                                     enriched_map_buffer_include_policy());
 
         // Sort turns over offsetted ring(s)
-        for (typename mapped_vector_type::iterator mit
-            = mapped_vector.begin();
-            mit != mapped_vector.end();
-            ++mit)
+        for (auto& pair : mapped_vector)
         {
-            std::sort(mit->second.begin(), mit->second.end(), buffer_less());
+            std::sort(pair.second.begin(), pair.second.end(), buffer_less());
         }
     }
 
@@ -342,17 +337,14 @@ struct buffered_piece_collection
 
         // Deflated rings may not travel to themselves, there should at least
         // be three turns (which cannot be checked here - TODO: add to traverse)
-        for (typename boost::range_iterator<turn_vector_type>::type it =
-            boost::begin(m_turns); it != boost::end(m_turns); ++it)
+        for (auto& turn : m_turns)
         {
-            buffer_turn_info_type& turn = *it;
             if (! turn.is_turn_traversable)
             {
                 continue;
             }
-            for (int i = 0; i < 2; i++)
+            for (auto& op : turn.operations)
             {
-                buffer_turn_operation_type& op = turn.operations[i];
                 if (op.enriched.get_next_turn_index() == static_cast<signed_size_type>(turn.turn_index)
                     && m_pieces[op.seg_id.piece_index].is_deflated)
                 {
@@ -385,10 +377,8 @@ struct buffered_piece_collection
 
         bool const deflate = m_distance_strategy.negative();
 
-        for (typename boost::range_iterator<turn_vector_type>::type it =
-            boost::begin(m_turns); it != boost::end(m_turns); ++it)
+        for (auto& turn : m_turns)
         {
-            buffer_turn_info_type& turn = *it;
             if (turn.is_turn_traversable)
             {
                 if (deflate && turn.count_in_original <= 0)
@@ -408,13 +398,8 @@ struct buffered_piece_collection
 
     inline void update_turn_administration()
     {
-        std::size_t index = 0;
-        for (typename boost::range_iterator<turn_vector_type>::type it =
-            boost::begin(m_turns); it != boost::end(m_turns); ++it, ++index)
+        for_each_with_index(m_turns, [this](std::size_t index, auto& turn)
         {
-            buffer_turn_info_type& turn = *it;
-
-            // Update member used
             turn.turn_index = index;
 
             // Verify if a turn is a linear endpoint
@@ -422,7 +407,7 @@ struct buffered_piece_collection
             {
                 check_linear_endpoints(turn);
             }
-        }
+        });
     }
 
     // Calculate properties of piece borders which are not influenced
@@ -434,11 +419,8 @@ struct buffered_piece_collection
     // - (if pieces are reversed)
     inline void update_piece_administration()
     {
-        for (typename piece_vector_type::iterator it = boost::begin(m_pieces);
-            it != boost::end(m_pieces);
-            ++it)
+        for (auto& pc : m_pieces)
         {
-            piece& pc = *it;
             piece_border_type& border = pc.m_piece_border;
             buffered_ring<Ring> const& ring = offsetted_rings[pc.first_seg_id.multi_index];
 
@@ -791,7 +773,7 @@ struct buffered_piece_collection
     {
         BOOST_GEOMETRY_ASSERT(boost::size(range) != 0u);
 
-        typename Range::const_iterator it = boost::begin(range);
+        auto it = boost::begin(range);
 
         // If it follows a non-join (so basically the same piece-type) point b1 should be added.
         // There should be two intersections later and it should be discarded.
@@ -928,18 +910,17 @@ struct buffered_piece_collection
     // Those can never be traversed and should not be part of the output.
     inline void discard_rings()
     {
-        for (typename boost::range_iterator<turn_vector_type const>::type it =
-            boost::begin(m_turns); it != boost::end(m_turns); ++it)
+        for (auto const& turn : m_turns)
         {
-            if (it->is_turn_traversable)
+            if (turn.is_turn_traversable)
             {
-                offsetted_rings[it->operations[0].seg_id.multi_index].has_accepted_intersections = true;
-                offsetted_rings[it->operations[1].seg_id.multi_index].has_accepted_intersections = true;
+                offsetted_rings[turn.operations[0].seg_id.multi_index].has_accepted_intersections = true;
+                offsetted_rings[turn.operations[1].seg_id.multi_index].has_accepted_intersections = true;
             }
             else
             {
-                offsetted_rings[it->operations[0].seg_id.multi_index].has_discarded_intersections = true;
-                offsetted_rings[it->operations[1].seg_id.multi_index].has_discarded_intersections = true;
+                offsetted_rings[turn.operations[0].seg_id.multi_index].has_discarded_intersections = true;
+                offsetted_rings[turn.operations[1].seg_id.multi_index].has_discarded_intersections = true;
             }
         }
     }
@@ -952,12 +933,8 @@ struct buffered_piece_collection
         // any of the robust original rings
         // This can go quadratic if the input has many rings, and there
         // are many untouched deflated rings around
-        for (typename std::vector<original_ring>::const_iterator it
-            = original_rings.begin();
-            it != original_rings.end();
-            ++it)
+        for (auto const& original : original_rings)
         {
-            original_ring const& original = *it;
             if (original.m_ring.empty())
             {
                 continue;
@@ -999,12 +976,8 @@ struct buffered_piece_collection
     // be discarded
     inline void discard_nonintersecting_deflated_rings()
     {
-        for(typename buffered_ring_collection<buffered_ring<Ring> >::iterator it
-            = boost::begin(offsetted_rings);
-            it != boost::end(offsetted_rings);
-            ++it)
+        for (auto& ring : offsetted_rings)
         {
-            buffered_ring<Ring>& ring = *it;
             if (! ring.has_intersections()
                 && boost::size(ring) > 0u
                 && geometry::area(ring, m_strategy) < 0)
@@ -1019,10 +992,8 @@ struct buffered_piece_collection
 
     inline void block_turns()
     {
-        for (typename boost::range_iterator<turn_vector_type>::type it =
-            boost::begin(m_turns); it != boost::end(m_turns); ++it)
+        for (auto& turn : m_turns)
         {
-            buffer_turn_info_type& turn = *it;
             if (! turn.is_turn_traversable)
             {
                 // Discard this turn (don't set it to blocked to avoid colocated
@@ -1055,21 +1026,16 @@ struct buffered_piece_collection
 
     inline void reverse()
     {
-        for(typename buffered_ring_collection<buffered_ring<Ring> >::iterator it = boost::begin(offsetted_rings);
-            it != boost::end(offsetted_rings);
-            ++it)
+        for (auto& ring : offsetted_rings)
         {
-            if (! it->has_intersections())
+            if (! ring.has_intersections())
             {
-                std::reverse(it->begin(), it->end());
+                std::reverse(ring.begin(), ring.end());
             }
         }
-        for (typename boost::range_iterator<buffered_ring_collection<Ring> >::type
-                it = boost::begin(traversed_rings);
-                it != boost::end(traversed_rings);
-                ++it)
+        for (auto& ring : traversed_rings)
         {
-            std::reverse(it->begin(), it->end());
+            std::reverse(ring.begin(), ring.end());
         }
     }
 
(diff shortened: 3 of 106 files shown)


include/boost/geometry/algorithms/centroid.hpp
```diff
@@ -352,15 +350,12 @@ struct centroid_multi
                 Point
             >::type state;
 
-        for (typename boost::range_iterator<Multi const>::type
-                it = boost::begin(multi);
-            it != boost::end(multi);
-            ++it)
+        for (auto it = boost::begin(multi); it != boost::end(multi); ++it)
```
barendgehrels, line 353: NOTE: we cannot do a range based for loop here
(as @awulkiew pointed out once).

In some places I changed it to a range based for loop, but these places are internal (for example turns).

include/boost/geometry/algorithms/densify.hpp
```diff
@@ -92,11 +89,11 @@ struct densify_range
...
         push_back_policy<MutRng> policy(rng_out);
 
-        iterator_t prev = it;
+        auto prev = it;
         for ( ++it ; it != end ; prev = it++)
         {
-            point_t const& p0 = *prev;
-            point_t const& p1 = *it;
+            auto const& p0 = *prev;
+            auto const& p1 = *it;
```
awulkiew, line 96: Note that here iterator could return non-true reference type. Previously this type was converted to `point_t` which was value_type vs reference_type returned by interator. If this non-true reference type was not adapted to Point concept the call to convert below could fail to compile. I don't know if that's a problem.

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -791,7 +775,7 @@ struct buffered_piece_collection
     {
         BOOST_GEOMETRY_ASSERT(boost::size(range) != 0u);
 
-        typename Range::const_iterator it = boost::begin(range);
+        auto it = std::cbegin(range);
```
awulkiew, line 778: To double-check, is `std` intended here? `boost::size` is called above.

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -409,13 +399,10 @@ struct buffered_piece_collection
...
         std::size_t index = 0;
-        for (typename boost::range_iterator<turn_vector_type>::type it =
-            boost::begin(m_turns); it != boost::end(m_turns); ++it, ++index)
+        for (auto& turn : m_turns)
         {
-            buffer_turn_info_type& turn = *it;
-
             // Update member used
-            turn.turn_index = index;
+            turn.turn_index = index++;
```
awulkiew, line 405: `for_each_with_index` is used below. Would it make sense to use it here as well?

include/boost/geometry/algorithms/detail/calculate_point_order.hpp
```diff
@@ -36,7 +36,7 @@ struct clean_point
         , m_is_azi_valid(false), m_is_azi_diff_valid(false)
     {}
 
-    typename boost::iterators::iterator_reference<Iter>::type ref() const
+    auto ref() const
```
awulkiew, line 39: This way the reference will be converted to value. You should probably return `decltype(auto)`.

include/boost/geometry/algorithms/detail/counting.hpp
```diff
@@ -84,15 +84,12 @@ template <typename SingleCount>
...
     template <typename MultiGeometry>
-    static inline std::size_t apply(MultiGeometry const& geometry)
+    static inline std::size_t apply(MultiGeometry const& multi)
     {
         std::size_t n = 0;
-        for (typename boost::range_iterator<MultiGeometry const>::type
-                 it = boost::begin(geometry);
-             it != boost::end(geometry);
-             ++it)
+        for (auto const& geometry : multi)
```
awulkiew, line 90: Range-based for loop is probably incorrect here.

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
barendgehrels: It's currently a const version only.
But I can add an overload.

include/boost/geometry/algorithms/detail/envelope/range_of_boxes.hpp
```diff
@@ -258,29 +249,27 @@ struct envelope_range_of_boxes
...
 
         coordinate_type const min_longitude = constants::min_longitude();
         coordinate_type const max_longitude = constants::max_longitude();
         coordinate_type const period = constants::period();
 
         interval_range_type intervals;
-        for (iterator_type it = boost::begin(range_of_boxes);
-             it != boost::end(range_of_boxes);
-             ++it)
+        for (auto const& box : range_of_boxes)
```
awulkiew, line 264: While it's true that currently this function is used only internally I wouldn't be so eager to replace this loop with range-based version. It's because technically we could introduce MultiBox concept in the future. In general I propose to leave normal loops in algorithms taking ranges as template parameters as if they were geometries because replacing them may have unintended consequences for users.

include/boost/geometry/algorithms/detail/num_distinct_consecutive_points.hpp
```diff
@@ -48,22 +48,20 @@ struct num_distinct_consecutive_points
...
-        iterator current = boost::begin(range);
-        iterator const end = boost::end(range);
+        auto current = boost::begin(range);
+        auto const end = boost::end(range);
         std::size_t counter(0);
         do
         {
             ++counter;
-            iterator next = std::find_if(current, end, [&](auto const& pt) {
+            auto next = std::find_if(current, end, [&](auto const& pt) {
```
awulkiew, line 64: Though here the names doesn't bother me because I can see what they are from the context.

include/boost/geometry/algorithms/detail/overlay/assign_parents.hpp
```diff
@@ -325,17 +319,15 @@ inline void assign_parents(Geometry1 const& geometry1,
...
                 ring_info_type& outer = ring_map[id_of_positive];
-                index = 0;
-                for (vector_iterator_type it = boost::begin(vector);
-                    it != boost::end(vector); ++it, ++index)
+                for_each_with_index(vector, [&](std::size_t index, auto const& item)
                 {
                     if (index != index_positive)
                     {
-                        ring_info_type& inner = ring_map[it->id];
+                        ring_info_type& inner = ring_map[item.id];
```
awulkiew, line 326: Out of curiosity, why have you left `ring_info_type`?

include/boost/geometry/algorithms/detail/overlay/intersection_insert.hpp
```diff
@@ -380,12 +379,10 @@ struct intersection_of_linestring_with_areal
...
-        int index = 0;
-        for(typename std::deque<turn_info>::const_iterator
-            it = turns.begin(); it != turns.end(); ++it)
-        {
-            debug_follow(*it, it->operations[0], index++);
-        }
+        for_each_with_index(turns, [](auto index, auto const& turn)
+            {
+                debug_follow(turn, turn.operations[0], index);
+            });
```
awulkiew, line 385: The indentation is inconsistent with other calls. Besides we should probably mimic loops as it is correctly done in other places.

include/boost/geometry/algorithms/detail/partition.hpp
```diff
@@ -629,8 +608,7 @@ class partition
                                        IteratorVector& iterator_vector,
                                        ExpandPolicy const& expand_policy)
     {
-        for(typename boost::range_iterator<ForwardRange const>::type
-                it = boost::begin(forward_range);
+        for (auto it = boost::begin(forward_range);
             it != boost::end(forward_range);
             ++it)
```
awulkiew, line 613: Would it make sense to put the loop in one line? It should be lesser than 100 characters.

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
```diff
@@ -305,7 +304,7 @@ inline bool calculate_from_inside(Geometry1 const& geometry1,
         std::size_t const q_seg_jk = (q_seg_ij + 1) % seg_count2;
         // TODO: the following function should return immediately, however the worst case complexity is O(N)
         // It would be good to replace it with some O(1) mechanism
-        range2_iterator qk_it = find_next_non_duplicated(boost::begin(range2),
+        auto qk_it = find_next_non_duplicated(boost::begin(range2),
                                                          range::pos(range2, q_seg_jk),
```
awulkiew, line 308: Indentation should probably be changed as well.

include/boost/geometry/algorithms/detail/relate/follow_helpers.hpp
```diff
@@ -256,9 +256,8 @@ class exit_watcher
         segment_identifier const& other_id = turn.operations[other_op_id].seg_id;
         overlay::operation_type exit_op = turn.operations[op_id].operation;
 
-        typedef typename std::vector<point_info>::iterator point_iterator;
         // search for the entry point in the same range of other geometry
-        point_iterator entry_it = std::find_if(m_other_entry_points.begin(),
+        auto entry_it = std::find_if(m_other_entry_points.begin(),
                                                m_other_entry_points.end(),
```
awulkiew, line 261: Indentation should probably be changed as well.

include/boost/geometry/algorithms/detail/relate/point_point.hpp
```diff
@@ -67,20 +67,25 @@ std::pair<bool, bool> point_multipoint_check(Point const& point,
     // point_in_geometry could be used here but why iterate over MultiPoint twice?
     // we must search for a point in the exterior because all points in MultiPoint can be equal
 
-    typedef typename boost::range_iterator<MultiPoint const>::type iterator;
-    iterator it = boost::begin(multi_point);
-    iterator last = boost::end(multi_point);
-    for ( ; it != last ; ++it )
+    auto it = boost::begin(multi_point);
+    auto last = boost::end(multi_point);
+    for ( ; it != last ; ++it)
```
awulkiew, line 72: These iterators could probably be put inside the loop. At least `it` because with `last`/`end` there could be a debate about performance.

include/boost/geometry/algorithms/detail/relate/topology_check.hpp
```diff
@@ -181,11 +181,9 @@ struct topology_check<MultiLinestring, Strategy, multi_linestring_tag>
 
         m_has_interior = false;
 
-        typedef typename boost::range_iterator<MultiLinestring const>::type ls_iterator;
-        for ( ls_iterator it = boost::begin(m_mls) ; it != boost::end(m_mls) ; ++it )
+        for (auto it = boost::begin(m_mls); it != boost::end(m_mls); ++it)
         {
-            typename boost::range_reference<MultiLinestring const>::type
-                ls = *it;
+            auto ls = *it;
```
awulkiew, line 186: This would cause a linestring to be copied. `auto const&` should probably be used here instead.

include/boost/geometry/algorithms/detail/within/within_no_turns.hpp
```diff
@@ -90,44 +95,42 @@ struct within_no_turns<Geometry1, Geometry2, polygon_tag, polygon_tag>
...
+        }
         // Now check if holes of G2 aren't inside G1
-        typedef typename boost::range_const_iterator
-            <
-                typename geometry::interior_type<Geometry2>::type
-            >::type iterator2;
-        for ( iterator2 it = boost::begin(geometry::interior_rings(geometry2)) ;
-              it != boost::end(geometry::interior_rings(geometry2)) ;
-              ++it )
+        auto&& rings2 = geometry::interior_rings(geometry2);
```
awulkiew, line 108: Same here.

include/boost/geometry/algorithms/for_each.hpp
```diff
@@ -326,9 +326,7 @@ struct for_each_polygon
             return false;
         }
 
-        typename interior_return_type<Polygon>::type
-            rings = interior_rings(poly);
-
+        auto&& rings = interior_rings(poly);
```
awulkiew, line 329: Forwarding reference, are we planning to forward rings?

include/boost/geometry/io/svg/write.hpp
```diff
@@ -164,15 +157,11 @@ struct svg_poly
 
         // Inner rings:
         {
-            typename interior_return_type<Polygon const>::type
-                rings = interior_rings(polygon);
-            for (typename detail::interior_iterator<Polygon const>::type
-                    rit = boost::begin(rings); rit != boost::end(rings); ++rit)
+            auto&& rings = interior_rings(polygon);
```
awulkiew, line 160: Forwarding reference.

awulkiew (changes requested):
That's a big PR. Thanks!
I saw some minor issues though.

include/boost/geometry/algorithms/line_interpolate.hpp
```diff
@@ -89,11 +89,10 @@ struct interpolate_range
                              PointLike & pointlike,
                              Strategies const& strategies)
     {
-        typedef typename boost::range_iterator<Range const>::type iterator_t;
         typedef typename boost::range_value<Range const>::type point_t;
 
-        iterator_t it = boost::begin(range);
-        iterator_t end = boost::end(range);
+        auto it = boost::begin(range);
+        auto end = boost::end(range);
```
vissarion, line 95: `auto const` ?

vissarion (approved):
Thanks, LGTM in general with some (minor) comments.

include/boost/geometry/algorithms/densify.hpp
barendgehrels: I'm not sure either. Reverted (it was not related to the iterator anyway)
:heavy_check_mark:

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
barendgehrels: Reverted. First I changed some `begin` but later decided it's better to limit this to iterators only. This was a left over.
:heavy_check_mark: 

Unless there are objections, a separate PR could change `boost::begin`  to `std::begin` everywhere (+`end`)

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
barendgehrels: :heavy_check_mark:

include/boost/geometry/algorithms/detail/calculate_point_order.hpp
barendgehrels: :heavy_check_mark: reverted

include/boost/geometry/algorithms/detail/counting.hpp
barendgehrels: :heavy_check_mark: right, reverted

include/boost/geometry/algorithms/detail/envelope/range_of_boxes.hpp
barendgehrels: :heavy_check_mark: good point, reverted (keeping `box`)

include/boost/geometry/algorithms/detail/overlay/assign_parents.hpp
```diff
@@ -260,39 +261,32 @@ inline void assign_parents(Geometry1 const& geometry1,
...
+            vector[index] = helper(pair.first, pair.second.get_area());
             helper& item = vector[index];
-            switch(it->first.source_index)
+            switch(pair.first.source_index)
             {
                 case 0 :
-                    geometry::envelope(get_ring<tag1>::apply(it->first, geometry1),
-                                       item.envelope, strategy);
+                    geometry::envelope(get_ring<tag1>::apply(pair.first, geometry1),
+                                    item.envelope, strategy);
```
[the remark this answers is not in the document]
barendgehrels: I see, now it was not. :heavy_check_mark:

include/boost/geometry/algorithms/detail/overlay/intersection_insert.hpp
barendgehrels: :heavy_check_mark:

include/boost/geometry/algorithms/detail/relate/point_point.hpp
barendgehrels: :heavy_check_mark: fixed and renamed `last` to `end` as done elsewhere

include/boost/geometry/algorithms/detail/relate/topology_check.hpp
barendgehrels: :+1: well spotted, thanks
:heavy_check_mark: changed to `auto const&` and a few lines later (for points) as well.

include/boost/geometry/algorithms/detail/within/within_no_turns.hpp
```diff
@@ -57,25 +59,28 @@ struct within_no_turns<Geometry1, Geometry2, ring_tag, polygon_tag>
...
+        }
         // Now check if holes of G2 aren't inside G1
-        typedef typename boost::range_const_iterator
-            <
-                typename geometry::interior_type<Geometry2>::type
-            >::type iterator;
-        for ( iterator it = boost::begin(geometry::interior_rings(geometry2)) ;
-              it != boost::end(geometry::interior_rings(geometry2)) ;
-              ++it )
+        auto&& rings2 = geometry::interior_rings(geometry2);
```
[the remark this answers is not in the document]
barendgehrels: I copied this from https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/correct.hpp#L135
but indeed, it's not const there.

Changed to `auto const&`, indeed we use that more often

include/boost/geometry/algorithms/detail/within/within_no_turns.hpp
barendgehrels: :heavy_check_mark: fixed

include/boost/geometry/algorithms/detail/within/within_no_turns.hpp
```diff
@@ -90,44 +95,42 @@ struct within_no_turns<Geometry1, Geometry2, polygon_tag, polygon_tag>
...
                 // if it's also inside one of the G1 holes, it's ok
                 bool ok = false;
-                typedef typename boost::range_const_iterator
-                    <
-                        typename geometry::interior_type<Geometry1>::type
-                    >::type iterator1;
-                for ( iterator1 it1 = boost::begin(geometry::interior_rings(geometry1)) ;
-                      it1 != boost::end(geometry::interior_rings(geometry1)) ;
-                      ++it1 )
+                auto&& rings1 = geometry::interior_rings(geometry1);
```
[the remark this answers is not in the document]
barendgehrels: :heavy_check_mark: fixed

include/boost/geometry/algorithms/centroid.hpp
```diff
@@ -255,11 +255,9 @@ struct centroid_polygon_state
     {
         centroid_range_state::apply(exterior_ring(poly), transformer, strategy, state);
 
-        typename interior_return_type<Polygon const>::type
-            rings = interior_rings(poly);
-
-        for (typename detail::interior_iterator<Polygon const>::type
-                it = boost::begin(rings); it != boost::end(rings); ++it)
+        auto&& rings = interior_rings(poly);
```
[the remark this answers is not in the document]
barendgehrels: :heavy_check_mark: fixed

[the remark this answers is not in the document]
barendgehrels: :+1: :heavy_check_mark: fixed

include/boost/geometry/algorithms/centroid.hpp
```diff
@@ -255,11 +255,9 @@ struct centroid_polygon_state
...
+        auto const end = boost::end(rings);
```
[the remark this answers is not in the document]
barendgehrels: It's done elsewhere as well, for rings.

include/boost/geometry/algorithms/densify.hpp
```diff
@@ -78,11 +78,8 @@ struct densify_range
     static inline void apply(FwdRng const& rng, MutRng & rng_out,
                              T const& len, Strategies const& strategies)
     {
-        typedef typename boost::range_iterator<FwdRng const>::type iterator_t;
-        typedef typename boost::range_value<FwdRng>::type point_t;
-
-        iterator_t it = boost::begin(rng);
-        iterator_t end = boost::end(rng);
+        auto it = boost::begin(rng);
+        auto end = boost::end(rng);
```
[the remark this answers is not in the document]
barendgehrels: You see it occasionally, but usually iterators are never declared const.
I would not make them `auto const&` because it's not a reference (though it's syntactically correct)

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -801,7 +785,7 @@ struct buffered_piece_collection
             add_point(*it);
         }
 
-        for (++it; it != boost::end(range); ++it)
+        for (++it; it != std::cend(range); ++it)
```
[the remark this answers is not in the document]
barendgehrels: :+1: :heavy_check_mark: reverted

include/boost/geometry/algorithms/detail/distance/geometry_to_segment_or_box.hpp
```diff
@@ -203,25 +197,21 @@ class geometry_to_segment_or_box
         // consider all distances of the points in the geometry to the
         // segment or box
         comparable_return_type cd_min1(0);
-        point_iterator_type pit_min;
-        seg_or_box_const_iterator it_min1 = boost::const_begin(seg_or_box_points);
-        seg_or_box_const_iterator it_min2 = it_min1;
+        auto pit_min = points_begin(geometry);
+        auto it_min1 = boost::const_begin(seg_or_box_points);
```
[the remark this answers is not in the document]
barendgehrels: See above, we don't have that habit for iterators

include/boost/geometry/algorithms/detail/distance/geometry_to_segment_or_box.hpp
```diff
@@ -318,7 +301,7 @@ class geometry_to_segment_or_box
     {
         distance::creturn_t<MultiPoint, SegmentOrBox, Strategies> cd_min;
 
-        iterator_type it_min
+        auto it_min
```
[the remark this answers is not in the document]
barendgehrels: Same here

include/boost/geometry/algorithms/detail/envelope/range_of_boxes.hpp
```diff
@@ -258,29 +249,27 @@ struct envelope_range_of_boxes
 
         BOOST_GEOMETRY_ASSERT(! boost::empty(range_of_boxes));
 
-        iterator_type it_min = std::min_element(boost::begin(range_of_boxes),
-                                                boost::end(range_of_boxes),
-                                                latitude_less<min_corner>());
-        iterator_type it_max = std::max_element(boost::begin(range_of_boxes),
-                                                boost::end(range_of_boxes),
-                                                latitude_less<max_corner>());
+        auto it_min = std::min_element(boost::begin(range_of_boxes),
```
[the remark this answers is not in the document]
barendgehrels: Same here

include/boost/geometry/algorithms/line_interpolate.hpp
barendgehrels: Same here

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
barendgehrels: I did some more checks, we can't change it everywhere. Because we support geometries adapted to Boost.Range, but not `std::begin` etc.
We might (ever) change that, but that might break behavior for users having custom geometries.

include/boost/geometry/algorithms/for_each.hpp
awulkiew: Right, if it's non-const then we have to use `auto&&`. If we use `auto&` the lifetime of a temporary which might be returned by `interior_rings()` will not be extended.

include/boost/geometry/algorithms/centroid.hpp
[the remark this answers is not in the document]
awulkiew: I'm not sure if this is the reason but the effect is that end will not be calculated at the end of each iteration.

include/boost/geometry/algorithms/detail/distance/geometry_to_segment_or_box.hpp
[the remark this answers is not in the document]
awulkiew: Defining it as `auto const` will prevent us from changing the iterator e.g incrementg it. If that's ok then it could be `const` like in case of end/last. Think about a pointer to const (`const_iterator`) vs const pointer (`auto const`).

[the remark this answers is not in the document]
barendgehrels: Yes, I changed it where @vissarion requested this. I completely agree now.

But specifically in this case, this iterator is meant to change. It is an iterator pointing to the minimum. So it cannot be made `const` here.

include/boost/geometry/algorithms/for_each.hpp
barendgehrels: thanks

include/boost/geometry/algorithms/detail/calculate_point_order.hpp
awulkiew: FYI, revert was not required. Returning `decltype(auto)` would be enough.

include/boost/geometry/algorithms/for_each.hpp
```diff
@@ -326,9 +325,7 @@ struct for_each_polygon
             return false;
         }
 
-        typename interior_return_type<Polygon>::type
-            rings = interior_rings(poly);
-
+        auto& rings = interior_rings(poly);
```
awulkiew, line 328:
```suggestion
        auto&& rings = interior_rings(poly);
```

include/boost/geometry/util/for_each_with_index.hpp
```diff
@@ -30,6 +30,17 @@ inline void for_each_with_index(Container const& container, Function func)
     }
 }
 
+template <typename Container, typename Function>
+inline void for_each_with_index(Container& container, Function func)
+{
+    std::size_t index = 0;
+    for (auto it = std::begin(container); it != std::end(container); ++it, ++index)
```
awulkiew, line 37: Technically the type of index should be taken from Container/Range, e.g.: `boost::size_type<Container>::type` but I guess in practice `std::size_t` will be ok.

Are we sure that this utility will not be used for geometries adapted to Boost.Range Concept?
Would it make sense to use `boost::begin` and `boost::end` to be safe?

barendgehrels: :heavy_check_mark: fixed. While fixing it, I now also use Boost Range's size type

include/boost/geometry/algorithms/detail/calculate_point_order.hpp
barendgehrels: :heavy_check_mark: intuitively it looked the same, but I get it now. Changed. Thank you.

barendgehrels:
Results are pushed again. The last push also changes some other interior_ring_type occurrences, fixes a bug (concept) and adds two missing include files.
This is caught by the new unit test (still to be pushed).

barendgehrels:
Second commit adds tests (and fixes related concept).
Finally I'll have to add similar tests for (multi)linestring.
It proves to be quite useful, several functions do not support these geometries

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -201,8 +200,8 @@ struct wkt_poly
     {
         using ring = typename ring_type<Polygon const>::type;
 
-        auto const exterior = exterior_ring(poly);
-        auto const rings = interior_rings(poly);
+        auto const& exterior = exterior_ring(poly);
+        auto const& rings = interior_rings(poly);
```
barendgehrels, line 204: My bad, I changed this last December. We didn't have unit tests catching this.

test/geometries/custom_non_copiable/cnc_container.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+#ifndef GEOMETRY_TEST_CNC_CONTAINER_HPP
+#define GEOMETRY_TEST_CNC_CONTAINER_HPP
+
+#include <cstddef>
+#include <deque>
+
+// Define a custom class which is only movable and not copiable, and cannot be indexed.
+// Its derivatives will be linestring, ring, multi*
+// For readability purposes we use the "cnc" abbreviation as "custom_non_copiable"
```
barendgehrels, line 18: This class originates from `all_custom_geometries` but takes it a bit further, making it non-copyable.
For test purposes it has to be indexable, but that's also customized (the library does not see that).

Also the adaptations are more modern (requiring less typedefs, more readable)

awulkiew (approved):

awulkiew:
Thanks!

barendgehrels:
@vissarion many tests are added, I'll assume you are still OK. I will merge one of the coming days.

PR #1140, opened 2023-04-23
Drop dependencies, replace boost with std, remove includes

awulkiew:
Drop Boost.SmartPtr dependency
Drop Boost.Utility (Call Traits) dependency
Replace `boost::array` with `std::array`
Replace `boost::minmax_element` with `std::minmax_element` Remove unneeded headers of Boost.Tuple and Boost.Algorithm Include specific headers from Boost.StringAlgo and Boost.Thread
Use auto for iterator types

### DIFF
--- a/include/boost/geometry/strategies/geographic/closest_points_pt_seg.hpp
+++ b/include/boost/geometry/strategies/geographic/closest_points_pt_seg.hpp
@@ -1,5 +1,7 @@
 // Boost.Geometry
 
+// Copyright (c) 2023 Adam Wulkiewicz, Lodz, Poland.
+
 // Copyright (c) 2021, Oracle and/or its affiliates.
 
 // Contributed and/or modified by Vissarion Fysikopoulos, on behalf of Oracle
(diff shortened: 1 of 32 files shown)


barendgehrels (approved):
Thanks! Looks good.

The only question I have: do we still separate commits between extensions and normal code, or is it equally convenient to mix it?

include/boost/geometry/strategies/geographic/closest_points_pt_seg.hpp
```diff
@@ -12,8 +14,6 @@
 
 #include <algorithm>
```
tinko92, line 15: Is this include possibly unused too?

awulkiew:
@barendgehrels I didn't remember that we do this. Certainly this would make cherry-picking easier into master. I'll separate the PR into two commits.

include/boost/geometry/strategies/geographic/closest_points_pt_seg.hpp
awulkiew: Yes, and several others in this file. They were probably copied from spherical version. I'll remove them.

awulkiew:
@vissarion are you ok with these changes?

PR #1146, opened 2023-05-08
[experimental] update test-demo with version with lambdas

barendgehrels:
Hi, this is an experimental version of partition using lambdas.

For demonstration and evaluation purposes.

I'm curious to your findings.


test/robustness/within/partition_within.cpp
```diff
@@ -311,27 +242,46 @@ void call_within(std::size_t size, std::size_t count)
...
+            [&quadtree_count](auto const& p, auto const& ri)
+            {
+                if (bg::within(p, ri.ring))
+                {
+                    quadtree_count++;
+                }
+                return true;
+            },
+            box_visitor
+            );
```
barendgehrels, line 282: This shows how the function could be used.
5 lambdas (2 expand, 2 overlaps, 1 main) plus an optional lambda (called per box),
instead of the earlier policies with `apply`

barendgehrels: The program itself creates SVG's like this:

image

awulkiew: A lot better! I prefer function objects over objects with `apply()` function.

One nitpick. This is not quadtree, it's kdtree, a node in quadtree is divided into 4 children, here a division is done into 2, once per dimension. Furthermore partition is k-dimensional, quadtree is 2-dimensional.

Side note, I was thinking about redesigning the API for a long time and use features from the rtree. Note e.g. that in your example you shouldn't be forced to pass function objects defining how to expand a box or perform intersection. These operations are always done the same way. We probably also never use box visitors, besides debugging or visualization so box visitor could be hidden as well. The only additional information in your example that's needed is the way of getting a box from item (taking point from point should work by default). In rtree this is done by IndexableGetter. There are also additional things like the type of operation (in this case intersects) and strategies (for defining CS) but they could be optional as well. I'm not saying to do this, only sharing what I was thinking about.

test/robustness/within/quadtree_partition.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+template
+<
+    typename BoxType,
+    typename ForwardRange1,
+    typename ForwardRange2
+>
+bool quadtree_partition(ForwardRange1 const& forward_range1,
+                        ForwardRange2 const& forward_range2,
+                        const std::function
```
awulkiew, line 97: I'd avoid using `std::function` because it's heavy. Why did you decide to use it over a function object?

test/robustness/within/quadtree_partition.hpp
```diff
@@ -0,0 +1,149 @@
...
+    explicit adapt_partition_visitor(const F& f)
+        : m_f(f)
+    {}
+
+    template<typename T1, typename T2>
+    bool apply(T1 const& i1, T2 const& i2) const
+    {
+        return m_f(i1, i2);
+    }
+};
```
awulkiew, line 37: All of the adaptors could probably be replaced with one simply forwarding parameters to function objects:
```
template <typename F>
struct adapt_partition_visitor
{
    F m_f;

    explicit adapt_partition_visitor(F&& f)
        : m_f(std::move(f))
    {}

    template <typename ...Ts>
    auto apply(Ts&& ...is) const
    {
        return m_f(std::forward<Ts>(is)...);
    }
};
```

test/robustness/within/partition_within.cpp
barendgehrels: Thanks!

I think it's not exactly a kdtree either. You are right, we divide into halves and not into 4. But always at the middle, in kdtree that it is not necessarily the case. Anyway, I can rename it of course. But it cannot be called `partition`... (unless we replace it).
Plus we have those additional boxes for cases where geometries exceed the halves. But sure, they are not part of the quadtree either.

I see your point. Yes, taking the box from an object is usually enough. Or point. But it could be done in a different way as well. For example: a box with a margin (for (max) distance calculations). It's flexible now. But that flexibility is probably used rarely, if at all. We use partition only internally, I think. It was introduced before our index so at that moment it was not a choice. So yes, it can be a bit simpler. The box visitor is indeed only used for debugging (at least at this moment).

Anyway, additionally, if we go this way, we don't need those visitor-adaptor(s), we can deprecate the old implementation and let the new one immediately use function objects.

test/robustness/within/quadtree_partition.hpp
barendgehrels: Indeed, works perfectly. Thanks!
:heavy_check_mark:

test/robustness/within/partition_within.cpp
awulkiew: Splitting planes can be placed at various locations in a kd-tree. Classic approach is to place them at the median object. Another strategy is to cut empty space out. There is also version using surface area heuristic. There are probably more. It depends on a use case. Splitting in half is probably a poor choice in general good only for objects distributed uniformly but such tree is still kd-tree.

Btw, in kd-tree splitting doesn't have to be done one dimension after another. The space can be split e.g. in X axis cutting out empty space and then again in X axis at the middle object, etc. But no matter what strategy is used this is still kd-tree.

Regarding quadtrees. There are versions of quadtrees have splitting planes at different locations than geometric middle of a node. AFAIR there are also quadtrees with child nodes partially overlapping each other which makes them similar to rtrees.

Obviously partition is neither quadtree nor kd-tree because it's not a data structure.

barendgehrels: It's used internally around 12 times.
Unfortunately it's not in namespace `detail` so officially it is part of the interface.

test/robustness/within/Jamfile
```diff
@@ -0,0 +1,17 @@
+# Boost.Geometry (aka GGL, Generic Geometry Library)
+# Robustness Test - convex_hull
```
vissarion, line 2: this is not convex hull, typo?

test/robustness/within/lambda_partition.hpp
```diff
@@ -0,0 +1,98 @@
...
+    // The all_policies (it is never run in a different way) could be
+    // a compiletime part of that options structure as well.
+    return partition
+        <
+            BoxType,
+            detail::partition::include_all_policy,
+            detail::partition::include_all_policy
+        >::apply(forward_range1, forward_range2, av,
+                 aev1, aov1,
+                 aev2, aov2, 16, apbv);
```
vissarion, line 93: maybe add a comment that 16 is the end of the iteration (the max number of points in a partition box)

vissarion (approved):
Thanks, this is cool! I am OK with merging now.
I will have a deeper look next week. 
Regarding the name I also agree it is neither quad nor kd-tree (it is not a tree at all of course and it actually combines the default splitting strategy "cut in middle" of quad trees with the kd-tree per-dimension partitioning). So I think it is better to just call it "partitioning".

test/robustness/within/lambda_partition.hpp
barendgehrels: I changed it to an options structure, as I planned earlier.
Is something like this fine?
:heavy_check_mark:

test/robustness/within/Jamfile
barendgehrels: Indeed, I copied another file and forgot to adapt the title.
:heavy_check_mark: fixed

barendgehrels: > Thanks, this is cool! I am OK with merging now. I will have a deeper look next week. Regarding the name I also agree it is neither quad nor kd-tree (it is not a tree at all of course and it actually combines the default splitting strategy "cut in middle" of quad trees with the kd-tree per-dimension partitioning). So I think it is better to just call it "partitioning".

Renamed now to `partitioning` (currently in namespace `experimental`)

barendgehrels:
I updated this PR:
* adding a lambda version over one range as well (like the original had)
* applying the lambda version in `buffer`, which nicely removes many helper structures.

It's still in namespace `experimental` (so don't merge yet), I can't put it in namespace `geometry` because it conflicts with the structure `partition`. Any suggestions?

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -358,22 +358,59 @@ struct buffered_piece_collection
...
+
+        struct specific_options : experimental::partition_options
+        {
+            struct include_turn_policy
+            {
+                static inline bool apply(buffer_turn_info_type const& turn)
+                {
+                    return turn.is_turn_traversable && ! turn.within_original;
+                }
+            };
```
barendgehrels, line 371: Doing this, I noticed that we sometimes use the `include_turn_policy` but sometimes have it in the get_box/overlaps functors. For that reason, if this functionality is always there, we could also choose to get rid of these options.
(the one left is the default partition size but we can specify it as an optional parameter).
What do you think?

barendgehrels: I omitted it, it was really just about expanding the bounding box.

barendgehrels:
> I updated this PR:
> 
> * adding a lambda version over one range as well (like the original had)
> * applying the lambda version in `buffer`, which nicely removes many helper structures.
> 
> It's still in namespace `experimental` (so don't merge yet), I can't put it in namespace `geometry` because it conflicts with the structure `partition`. Any suggestions?

I'm now using `detail::lambda_partition`. 
Still to do:
* use it in `disjoint`
* use it in `is_valid`
* use it in `relate`
So it's getting there.

Then we can deprecate `partition` itself because it is not used inside the library.

In a next Boost version, we can delete it and rename `partition_lambda` to just `partition`.

barendgehrels:
gcc 5 now reports:
```
/home/gehrels/git_public/boost/boost/geometry/algorithms/detail/overlay/get_turns.hpp:524:32: error: request for member ‘bounding_box’ in ‘sec’, which is of non-class type ‘const int’
                geometry::expand(box, sec.bounding_box, strategy);
```
This is inconvenient and wrong, it is there and correct. And it is only for the second lambda...

Any idea how to circumvent this?

barendgehrels:
> which is of non-class type ‘const int’

It confuses the two overloads of `partition_lambda`... If I use different names, it works. But that is less convenient....
To be continued.

PR #1147, opened 2023-05-17
Several fixes of errors and warnings

vissarion:
This PR fixes several warnings and compile errors. ~It also upgrades circleCI docker image.~

### DIFF
--- a/include/boost/geometry/index/detail/rtree/query_iterators.hpp
+++ b/include/boost/geometry/index/detail/rtree/query_iterators.hpp
@@ -125,7 +125,7 @@ public:
     {
         return r.m_impl.is_end();
     }
-    
+
 private:
     visitors::spatial_query_incremental<MembersHolder, Predicates> m_impl;
 };
@@ -217,7 +217,7 @@ public:
     virtual ~query_iterator_base() {}
 
     virtual query_iterator_base * clone() const = 0;
-    
+
     virtual bool is_end() const = 0;
     virtual reference dereference() const = 0;
     virtual void increment() = 0;
@@ -285,6 +285,8 @@ public:
         : m_ptr(o.m_ptr.get() ? o.m_ptr->clone() : 0)
     {}
 
+    query_iterator(query_iterator&&) = default;
+
     query_iterator & operator=(query_iterator const& o)
     {
         if ( this != boost::addressof(o) )
@@ -294,21 +296,7 @@ public:
         return *this;
     }
 
-    query_iterator(query_iterator && o)
-        : m_ptr(0)
-    {
-        m_ptr.swap(o.m_ptr);
-    }
-
-    query_iterator & operator=(query_iterator && o)
-    {
-        if ( this != boost::addressof(o) )
-        {
-            m_ptr.swap(o.m_ptr);
-            o.m_ptr.reset();
-        }
-        return *this;
-    }
+    query_iterator& operator=(query_iterator &&) = default;
 
     reference operator*() const
     {
--- a/test/views/reversible_closeable.cpp
+++ b/test/views/reversible_closeable.cpp
@@ -15,6 +15,8 @@
 
 #include <geometry_test_common.hpp>
 
+#include <boost/algorithm/string/trim.hpp>
+
 #include <boost/geometry/views/closeable_view.hpp>
 #include <boost/geometry/views/reversible_view.hpp>
 


include/boost/geometry/index/detail/rtree/query_iterators.hpp
```diff
@@ -294,12 +294,6 @@ class query_iterator
         return *this;
     }
 
-    query_iterator(query_iterator && o)
```
vissarion, line 297: @awulkiew this is just my quick fix to make tests passing.

.circleci/config.yml
```diff
@@ -19,8 +19,8 @@ environment: &environment
 # Docker image
 docker: &docker
   docker:
-      # gcc-6.0, ruby-2.3, sudo, curl
-    - image: circleci/ruby:2.3
+      # ubuntu 22.04.2 LTS, build-essential, gcc-11
```
vissarion, line 22: the main reason for this upgrade is to fix the error 

```
./boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp:408:39: error: ‘check_linear_endpoints’ was not declared in this scope
```

that is related to a gcc-6 bug. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71546

awulkiew:
Oops, it seems I fixed several of them without creating a PR.

include/boost/geometry/index/detail/rtree/query_iterators.hpp
awulkiew: @vissarion I don't think that this is necessary. Which test doesn't pass?

.circleci/config.yml
awulkiew: Gcc-6 does support C++14 so I don't think that bumping version of the compiler that is used for testing is the correct solution. Furthermore I'm not sure that this bug is the correct one.

A workaround would be to call the member function explicitly writing `this` keyword.

vissarion: I though it was a good opportunity to also upgrade the circleCI. Anyway, I am OK with the workaround and I remove the circleCI update from this PR. BTW do we want to do that upgrade in a different PR?

include/boost/geometry/index/detail/rtree/query_iterators.hpp
awulkiew: Right, I can reproduce it with gcc-6. It is caused by the replacement of `scoped_ptr` with `unique_ptr`. So direct solution would be to pass `nullptr` instead.

But the solution you prosed is even better (almost). AFAIU your intention was to force the compiler to generate move ctor automatically. But since copy ctor and assignements are defined AFAIR this will not happen. Instead copy ctor will be called which will harm the performance. Instead move ctor should be declared as `default`. And since we're at it I think move assignment operator could be `default` as well, hence:
```
query_iterator(query_iterator&&) = default;
query_iterator& operator=(query_iterator &&) = default;
```

.circleci/config.yml
awulkiew: Well it depends, do we want to catch such issues in the future?

awulkiew (approved):

vissarion:
I tried to clean up a bit with a rebase.

include/boost/geometry/index/detail/rtree/query_iterators.hpp
vissarion: Sure, why not. Do we still need the current PR. Either we close it and you fix the issues there or we merge it and you create a PR with additional fixes. I am OK with both. 

Also there are some tests related to envelop that are failing but I want to fix them in a separate PR.

awulkiew: I propose to merge this PR.

.circleci/config.yml
vissarion: Good point. On the other hand we may miss issues that caught by newer versions. BTW, does it make sense to allow github actions run the full test-suite instead of the minimal one?

awulkiew: Yes, and my guess is that more users compile with newer compiler and if that's true then testing it would be more important but I don't know really.

Regarding GH actions. I can't remember why have we decided to do it this way. Probably becasue running all of them would take a lot of time. But there may be some technical difficulties. I think the discussions should be in relevant PR. We should also check if there are some slots distributed across all of the Boost submodules because with long tests we'd take slots from other libraries.

vissarion:
@barendgehrels are you OK with merging this PR?

PR #1156, opened 2023-06-01
Update circleCI scripts and image

vissarion:
Following the discussion [here](https://github.com/boostorg/geometry/discussions/1155) this PR proposes:

- to update the circleCI image to a next-gen convenience image (read [this](https://circleci.com/docs/next-gen-migration-guide/) for the advantages of this change). Moreover, following the discussion [here](https://github.com/boostorg/geometry/pull/1147) the compiler is set to gcc-6 to possible catch library issues,
- add ~~a closing job to circleCI to cover~~ missing tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,10 @@
 # Copyright (c) 2018 Mateusz Loskot <mateusz@loskot.net>
 # Copyright (c) 2018-2021 Adam Wulkiewicz, Lodz, Poland
 #
+# This file was modified by Oracle on 2023.
+# Modifications copyright (c) 2023 Oracle and/or its affiliates.
+# Contributed and/or modified by Vissarion Fysikopoulos, on behalf of Oracle
+#
 # Use, modification and distribution is subject to the Boost Software License,
 # Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
 # http://www.boost.org/LICENSE_1_0.txt)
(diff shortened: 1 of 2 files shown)


barendgehrels (approved):
I'm not an expert here.
I welcome improvements so I approve this.

.circleci/config.yml
```diff
@@ -416,6 +428,15 @@ jobs:
           root: ~/project
           paths:
             - shared-coverage/index_detail
+  remaining_tests:
+    <<: *config
+    steps:
+      - *attach_workspace
+      - run: ./$BOOST_DIR/libs/geometry/.circleci/run_test.sh
```
awulkiew, line 435: What was the intention here? This script requires parameters, without it it will not run anything.

awulkiew:
Thanks for this update!

Btw, I'd be ok with using the default gcc version. It's probably more commonly used (I assume it's ubuntu 20.04) and requires to do less manual steps which have tendency to fail.

FYI, currently the coverage step is failing due to some issues with package repositories so this change could fix it, maybe.

Have you tested whether or not it works with your fork by pushing develop or master branch and checking at CircleCI?
Even if you havent I'd be ok with pushing this change and finding out this way. In the worst case we could always reverse it.

vissarion:
Hi @awulkiew thanks for the review. I revised my PR by adding a more straightforward solution, i.e. manually adding the missing tests. The previous approach by adding a closing job was end up too complicated (since I had to add also special cases to the `run_test.sh` and I think it doesn't worth the trouble. Now I am running circleci tests in my fork.

vissarion:
There is also this related PR https://github.com/boostorg/geometry/pull/1154 where I need your feedback.

awulkiew:
Thanks, I'm ok with merging it.

Regarding testing in your fork, AFAIR you have to push them to your develop branch because otherwise minimal test is tested. You can also modify the script temporarily to always run it for all branches. Unless I'm mistaken.

awulkiew (approved):

PR #1157, opened 2023-06-01
[algorithms] Check const Ring concept in calculate_point_order

awulkiew:
This allows to compile the algorithm with non-mutable rings.


vissarion (approved):
Thanks!

PR #1160, opened 2023-06-07
Fix Visual Studio warnings

awulkiew:
- C4100 Unreferenced formal parameter
- C4127 Conditional expression is constant
- C4456 Declaration hides previous local declaration
- C4701 Potentially uninitialized local variable used
- C4702 Unreachable code

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
@@ -353,7 +353,7 @@ private :
               TiRStrategy const& strategy,
               geometry::strategy::buffer::place_on_ring_type place_on_ring, State& state) const
     {
-        return strategy.apply(point, p1, p2, place_on_ring, m_is_convex, state, get_full_ring());
+        return strategy.apply(point, p1, p2, place_on_ring, m_is_convex, state);
     }
 
     template <typename It, typename Box, typename Strategy>
--- a/include/boost/geometry/algorithms/detail/overlay/pointlike_pointlike.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/pointlike_pointlike.hpp
@@ -1,7 +1,8 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
-// Copyright (c) 2014-2020, Oracle and/or its affiliates.
+// Copyright (c) 2023 Adam Wulkiewicz, Lodz, Poland.
 
+// Copyright (c) 2014-2020, Oracle and/or its affiliates.
 // Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
 // Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
 
@@ -32,6 +33,8 @@
 #include <boost/geometry/core/tags.hpp>
 
 #include <boost/geometry/policies/compare.hpp>
+
+#include <boost/geometry/util/condition.hpp>
 
 
 namespace boost { namespace geometry
@@ -256,8 +259,8 @@ struct multipoint_multipoint_point
     {
         typedef geometry::less<void, -1, typename Strategy::cs_tag> less_type;
 
-        if ( OverlayType != overlay_difference
-             && boost::size(multipoint1) > boost::size(multipoint2) )
+        if (BOOST_GEOMETRY_CONDITION(OverlayType != overlay_difference)
+            && boost::size(multipoint1) > boost::size(multipoint2))
         {
             return multipoint_multipoint_point
                 <
--- a/include/boost/geometry/algorithms/simplify.hpp
+++ b/include/boost/geometry/algorithms/simplify.hpp
@@ -3,6 +3,7 @@
 // Copyright (c) 2007-2015 Barend Gehrels, Amsterdam, the Netherlands.
 // Copyright (c) 2008-2015 Bruno Lalande, Paris, France.
 // Copyright (c) 2009-2015 Mateusz Loskot, London, UK.
+// Copyright (c) 2023 Adam Wulkiewicz, Lodz, Poland.
 
 // This file was modified by Oracle on 2018-2022.
 // Modifications copyright (c) 2018-2022 Oracle and/or its affiliates.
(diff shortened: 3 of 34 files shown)


include/boost/geometry/algorithms/detail/overlay/pointlike_pointlike.hpp
```diff
@@ -256,8 +259,9 @@ struct multipoint_multipoint_point
     {
         typedef geometry::less<void, -1, typename Strategy::cs_tag> less_type;
 
-        if ( OverlayType != overlay_difference
-             && boost::size(multipoint1) > boost::size(multipoint2) )
+        if (BOOST_GEOMETRY_CONDITION(
```
awulkiew, line 262: What do you think about adding a macro allowing to conditionally use C++17 `if constexpr`?

include/boost/geometry/strategies/cartesian/turn_in_ring_winding.hpp
```diff
@@ -107,13 +108,12 @@ class turn_in_ring_winding
...
 
-    template <typename Point, typename PointOfSegment, typename Ring>
+    template <typename Point, typename PointOfSegment>
     static inline void apply_on_boundary(Point const& point,
                              PointOfSegment const& s1,
                              PointOfSegment const& s2,
                              place_on_ring_type place_on_ring,
-                             counter& the_state,
-                             Ring const& full_ring)
+                             counter& the_state)
```
awulkiew, line 116: @barendgehrels Is that ok or should I rather leave the parameter and remove the name?

vissarion (approved):
Thanks!

include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
```diff
@@ -353,7 +353,7 @@ private :
               TiRStrategy const& strategy,
               geometry::strategy::buffer::place_on_ring_type place_on_ring, State& state) const
     {
-        return strategy.apply(point, p1, p2, place_on_ring, m_is_convex, state, get_full_ring());
+        return strategy.apply(point, p1, p2, place_on_ring, m_is_convex, state);
```
barendgehrels, line 356: Hmm, this was probably a left over, only used for debugging.
Deletion looks good to me, thanks.

include/boost/geometry/algorithms/detail/overlay/pointlike_pointlike.hpp
barendgehrels: It can probably be integrated into `BOOST_GEOMETRY_CONDITION` ? Such that we don't need another one?

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -482,7 +483,7 @@ public :
             // Do not duplicate the closing point
             auto rot_end = boost::end(ring);
             std::size_t rot_index = index;
-            if (is_closed_in && size > 1)
+            if (BOOST_GEOMETRY_CONDITION(is_closed_in && size > 1))
```
barendgehrels, line 486: Related to `if constexpr`, isn't it, in cases like this, better to only apply the const part on the part being const?
I mean: `if (BOOST_GEOMETRY_CONDITION(is_closed_in) && size > 1)` ?

include/boost/geometry/algorithms/simplify.hpp
```diff
@@ -497,7 +498,7 @@ public :
             simplify_range<0>::apply(rotated, out, max_distance, impl, strategies);
 
             // Open output if needed
-            if (! is_closed_out && boost::size(out) > 1)
+            if (BOOST_GEOMETRY_CONDITION(! is_closed_out && boost::size(out) > 1))
```
barendgehrels, line 501: Same here

include/boost/geometry/formulas/area_formulas.hpp
```diff
@@ -380,7 +381,7 @@ class area_formulas
             return pi;
         }
 
-        if (LongSegment && lat1r != lat2r) // not for segments parallel to equator
+        if (BOOST_GEOMETRY_CONDITION(LongSegment && lat1r != lat2r)) // not for segments parallel to equator
```
barendgehrels, line 384: And here, and more.

include/boost/geometry/strategies/cartesian/turn_in_ring_winding.hpp
barendgehrels: It's OK, thanks. I also answered elsewhere before seeing this comment.

barendgehrels (approved):
Looks good to me in general, the only thing is that I would like to separate the parts really being const from the parts that might vary.

include/boost/geometry/algorithms/detail/overlay/pointlike_pointlike.hpp
awulkiew: @barendgehrels Unfortunately we do need two separate macros. AFAIR there are rare places where a condition can be constant or not, e.g. some policy can define a variable as static or dynamic member. Of course we could agree to drop static members in such cases.

Another thing is that I think we should consider something which looks more like the language feature. I'd also prefer to keep `if` so to not include it in the macro in order to allow the editor to highlight the key word to make it more readable. So maybe something like this:
```
if constexpr (OverlayType != overlay_difference)
vs
if BOOST_GEOMETRY_CONSTEXPR (OverlayType != overlay_difference)
```

awulkiew: See the condition below in pj_transform.hpp line 255:
```
if( BOOST_GEOMETRY_CONDITION(srcdefn.vto_meter != 1.0 && dimension > 2) )
```
where `srcdefn.vto_meter` may or may not be static but `dimension` is always static. In this case `if constexpr` should be used for `dimension` and `BOOST_GEOMETRY_CONDITION` for `srcdefn.vto_meter`. Or `vto_meter` should always be non-static member, also in static projections .

include/boost/geometry/algorithms/simplify.hpp
awulkiew: done

awulkiew:
Thanks. I separated them.

include/boost/geometry/algorithms/detail/overlay/pointlike_pointlike.hpp
barendgehrels: OK, I see, and saw it in your new PR this morning. Cool.

Issue #1158, opened 2023-06-01
Convex hull function returning invalid polygon


andrea-cimatoribus-pix4d:
I discovered a behaviour that seems badly incorrect in `boost::geometry::convex_hull` function, in relation to collinear points. It's clearly an edge case, but I think it points to some underlying weakness of the algorithm, that I could not identify. The issue is summarised by the following image:
![ch](https://github.com/boostorg/geometry/assets/121491611/1984c9ea-c059-4ff3-b1ac-e7c6c532301b)
which is the minimal example I could extract from the original dataset that presented the issue.

The full code to reproduce is the following (without the plots, a quick hack in Python):
```cpp
#include <boost/geometry.hpp>
#include <boost/geometry/geometries/point_xy.hpp>

#include <iostream>
                                         
using Point = boost::geometry::model::d2::point_xy<double>;
using Points = boost::geometry::model::multi_point<Point>;                                                                                                            
using BoostPolygon = boost::geometry::model::polygon<Point, true, false>;

void check_hull(const Points& points)
{
    BoostPolygon hull;
    boost::geometry::convex_hull(points, hull);

    // just in case
    boost::geometry::correct(hull);

    std::cout << "\nTest case:\n";
    for (const auto& p : points)
        std::cout << "(" << p.x() << ", " << p.y() << ")\n";

    std::cout << "\n";
    if (boost::geometry::is_valid(hull))
    {
        std::cout << "Valid convex hull!\n";
    }
    else
    {
        std::cout << "Invalid convex hull:\n";
        for (const auto& p : hull.outer())
            std::cout << "(" << p.x() << ", " << p.y() << ")\n";
    }
}

int main(int, char**)
{
    Points points;
    boost::geometry::append(points, Point(1941.6475737576565734, 554.21197550796682663));
    boost::geometry::append(points, Point(2201.1796067026721175, 604.83253590728611471));
    boost::geometry::append(points, Point(2367.1836939680897558, 1898.3918136409306499));
    boost::geometry::append(points, Point(1856.9044662310534477, 2320.8057089752910542));
    // troublesome points
    boost::geometry::append(points, Point(2000.0000000000002274, 551.77450949119793222));
    boost::geometry::append(points, Point(1999.9999999999995453, 1721.4008337980080796));
    boost::geometry::append(points, Point(1999.9999999999993179, 1966.6530570371892281));

    check_hull(points);

    return 0;
}
```

[...]

Issue #1148, opened 2023-05-19
Incorrect results of boost::geometry::math::smaller


vissarion:
There are some failing tests in [circleCI](https://app.circleci.com/pipelines/github/boostorg/geometry/393/workflows/0f08835f-b25a-4112-b290-47348ca4a762/jobs/19548). 

The commit that introduced that fail was https://github.com/boostorg/geometry/commit/092ab9da349dd0dab8b610407b0275b192d0477c in particular the replacement of `boost::minmax_element` with `std::minmax_element`. 

However, it appears that the issue is related to `boost::geometry::math::smaller` and the inconsistent (and wrong) results it returns in the case of floating points numbers that are very close each other. e.g. (1, 1-eps), (1, 1-eps/2) etc (where `eps=std::numeric_limits<T>::epsilon()`.

I will dig more into it next week.

PR #1162, opened 2023-06-16
Fix convex hull issue by passing compare strategies

vissarion:
This PR propose a fix for https://github.com/boostorg/geometry/issues/1158

First, it enables passing (umbrella) strategies to less policy. Then different (umbrella) strategies can parameterize the less policy by different compare strategy. 

For, example floating point comparison in convex hull strategies does *not* need to be done approximately. Relate strategies still use the approximate comparison of fp numbers (`math::equals()` function).

Side note: I am not sure if approximate comparison of numbers is needed for relate strategy. I just left it as is in order for the tests to pass, but I think it needs some investigation. Related issue https://github.com/boostorg/geometry/issues/1148 

Next steps (to be implemented in new PRs)
- do the same for `greater` and `equal_to`
- move code from `strategies/compare.hpp` to `strategies/cartesian/compare.hpp` (if I understand the strategies code structure correctly, @awulkiew please correct me if I am wrong here)

### DIFF
--- a/include/boost/geometry/algorithms/detail/convex_hull/graham_andrew.hpp
+++ b/include/boost/geometry/algorithms/detail/convex_hull/graham_andrew.hpp
@@ -224,7 +224,6 @@ private:
             point_type const& left, point_type const& right,
             SideStrategy const& side)
     {
-        //std::cout << "add to hull" << std::endl;
         output.push_back(left);
         for (auto const& i : input)
         {
--- a/include/boost/geometry/algorithms/detail/convex_hull/graham_andrew.hpp
+++ b/include/boost/geometry/algorithms/detail/convex_hull/graham_andrew.hpp
@@ -190,7 +190,7 @@ private:
 
         point_type most_left, most_right;
 
-        geometry::less<point_type, -1, Strategy> less;
+        geometry::less_exact<point_type, -1, Strategy> less;
 
         detail::convex_hull::get_extremes(in_proxy, most_left, most_right, less);
 
--- a/include/boost/geometry/algorithms/detail/convex_hull/graham_andrew.hpp
+++ b/include/boost/geometry/algorithms/detail/convex_hull/graham_andrew.hpp
@@ -2,8 +2,8 @@
 
 // Copyright (c) 2007-2012 Barend Gehrels, Amsterdam, the Netherlands.
 
-// This file was modified by Oracle on 2014-2021.
-// Modifications copyright (c) 2014-2021 Oracle and/or its affiliates.
+// This file was modified by Oracle on 2014-2023.
+// Modifications copyright (c) 2014-2023 Oracle and/or its affiliates.
 
 // Contributed and/or modified by Vissarion Fysikopoulos, on behalf of Oracle
 // Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
@@ -190,8 +190,7 @@ private:
 
         point_type most_left, most_right;
 
-        // TODO: User-defined CS-specific less-compare
-        geometry::less<point_type> less;
+        geometry::less<point_type, -1, Strategy> less;
 
         detail::convex_hull::get_extremes(in_proxy, most_left, most_right, less);
 
@@ -225,6 +224,7 @@ private:
             point_type const& left, point_type const& right,
             SideStrategy const& side)
     {
+        //std::cout << "add to hull" << std::endl;
         output.push_back(left);
         for (auto const& i : input)
         {
--- a/include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
+++ b/include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
@@ -2,7 +2,7 @@
 
 // Copyright (c) 2017 Adam Wulkiewicz, Lodz, Poland.
 
-// Copyright (c) 2014-2021, Oracle and/or its affiliates.
+// Copyright (c) 2014-2023, Oracle and/or its affiliates.
 
 // Contributed and/or modified by Vissarion Fysikopoulos, on behalf of Oracle
 // Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
@@ -63,12 +63,13 @@ namespace detail { namespace disjoint
 class multipoint_multipoint
 {
 private:
-    template <typename Iterator, typename CSTag>
+    template <typename Iterator, typename Strategy>
     class unary_not_disjoint_predicate
-        : geometry::less<void, -1, CSTag>
+        : geometry::less<void, -1, Strategy>
     {
     private:
-        typedef geometry::less<void, -1, CSTag> base_type;
+        //TODO: pass strategy
+        typedef geometry::less<void, -1, Strategy> base_type;
 
     public:
         unary_not_disjoint_predicate(Iterator first, Iterator last)
@@ -96,8 +97,8 @@ public:
     {
         BOOST_GEOMETRY_ASSERT( boost::size(multipoint1) <= boost::size(multipoint2) );
 
-        using cs_tag = typename Strategy::cs_tag;
-        using less_type = geometry::less<void, -1, cs_tag>;
+        //TODO: pass strategy
+        using less_type = geometry::less<void, -1, Strategy>;
         using point1_type = typename boost::range_value<MultiPoint1>::type;
 
         std::vector<point1_type> points1(boost::begin(multipoint1),
@@ -108,7 +109,7 @@ public:
         using predicate_type = unary_not_disjoint_predicate
             <
                 typename std::vector<point1_type>::const_iterator,
-                cs_tag
+                Strategy
             >;
 
         return none_of(boost::begin(multipoint2),
--- a/include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
+++ b/include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
@@ -68,12 +68,11 @@ private:
         : geometry::less<void, -1, Strategy>
     {
     private:
-        //TODO: pass strategy
-        typedef geometry::less<void, -1, Strategy> base_type;
+        using less_type = geometry::less<void, -1, Strategy>;
 
     public:
         unary_not_disjoint_predicate(Iterator first, Iterator last)
-            : base_type(), m_first(first), m_last(last)
+            : less_type(), m_first(first), m_last(last)
         {}
 
         template <typename Point>
@@ -82,7 +81,7 @@ private:
             return std::binary_search(m_first,
                                       m_last,
                                       point,
-                                      static_cast<base_type const&>(*this));
+                                      static_cast<less_type const&>(*this));
         }
 
     private:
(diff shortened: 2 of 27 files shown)


include/boost/geometry/algorithms/detail/convex_hull/graham_andrew.hpp
```diff
@@ -225,6 +224,7 @@ class graham_andrew
             point_type const& left, point_type const& right,
             SideStrategy const& side)
     {
+        //std::cout << "add to hull" << std::endl;
```
barendgehrels, line 227: Is this a left over?

include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
```diff
@@ -63,12 +63,13 @@ namespace detail { namespace disjoint
 class multipoint_multipoint
 {
 private:
-    template <typename Iterator, typename CSTag>
+    template <typename Iterator, typename Strategy>
     class unary_not_disjoint_predicate
-        : geometry::less<void, -1, CSTag>
+        : geometry::less<void, -1, Strategy>
```
barendgehrels, line 68: Can we call it maybe: `CompareStrategy` ?

include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
```diff
@@ -63,12 +63,13 @@ namespace detail { namespace disjoint
...
     {
     private:
-        typedef geometry::less<void, -1, CSTag> base_type;
+        //TODO: pass strategy
```
barendgehrels, line 71: Is this still TODO or already done? Because it is passed now. Or do you want to pass an instance? If so, can you rephrase the comment?

include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
```diff
@@ -108,7 +109,7 @@ class multipoint_multipoint
         using predicate_type = unary_not_disjoint_predicate
             <
                 typename std::vector<point1_type>::const_iterator,
-                cs_tag
+                Strategy
```
barendgehrels, line 112: This (for example) looks like a breaking change. Can you point if there is something which could (potentially) cause non compilations for existing users?

include/boost/geometry/algorithms/detail/relate/point_point.hpp
```diff
@@ -223,7 +222,8 @@ struct multipoint_multipoint
     {
         // sort points from the 1 MPt
         typedef typename geometry::point_type<SortedMultiPoint>::type point_type;
-        typedef geometry::less<void, -1, CSTag> less_type;
+        //TODO: pass strategy
+        typedef geometry::less<void, -1, Strategy> less_type;
```
barendgehrels, line 226: I'm a bit worried that `geometry::less` is changed. This really looks like a breaking change to me...

So it fixes convex hull (good!) but it needs also changes in many other operations like `relate` and `is_valid` etc...

barendgehrels (changes requested):
As mentioned in my comments, this looks like breaking...

Can it be solved in a different way too? A way does not change `less` itself?

include/boost/geometry/policies/compare.hpp
```diff
@@ -41,7 +42,7 @@ template
 <
     typename Point = void,
     int Dimension = -1,
-    typename CSTag = void
+    typename Strategy = void
```
barendgehrels, line 45: In summary, this is the main thing I'm worried about.

barendgehrels:
Copying your description, maybe just:
* `less_with_compare_strategy` 
* `greater_with_compare_strategy`
* etc
Though these names are a bit long.

Alternatively the `less` could be meta-templated such that it accepts either `CSTag` or a `CompareStrategy`.

include/boost/geometry/algorithms/detail/convex_hull/graham_andrew.hpp
vissarion: Yes, will be removed (together with a few other comments).

include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
vissarion: I do not know, I kept it that way since this is the name used by apply function (below) that actually pass that type.

include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
vissarion: This is done, removed now.

include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
vissarion: All strategies (apart from convex hull strategy) calls the old code at the end so AFAIU there is no issue at this point. But there are breaking changes in this PR (see below).

include/boost/geometry/algorithms/detail/relate/point_point.hpp
vissarion: For everything apart from convex hull the same strategy is called. E.g. for cartesian it is `cartesian<ComparePolicy, EqualsPolicy, -1>` which is as the old one https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/strategies/compare.hpp#L165 (also called from default compare strategy) with the only difference that now it has the EqualsPolicy parameter which for all algorithms is as before (i.e. `strategy::compare::equals_epsilon` see `strategies/relate/cartesian.hpp` but for convex hull it is `strategy::compare::equals_exact` see `strategies/convex_hull/cartesian.hpp`). This change eventually change only convex hull strategies.

include/boost/geometry/policies/compare.hpp
vissarion: See above.

vissarion:
First, many thanks for the review!

> Copying your description, maybe just:
> 
>     * `less_with_compare_strategy`
> 
>     * `greater_with_compare_strategy`
> 
>     * etc
>       Though these names are a bit long.
> 
> 
> Alternatively the `less` could be meta-templated such that it accepts either `CSTag` or a `CompareStrategy`.

The breaking change I see is in the case some user define a less operator with a CSTag template parameter. I added specializations to support this case too. Please let me know what you think after these changes. 

Side comment, I think it is more correct (according to the library design) to pass a strategy to less than a cs_tag.

awulkiew:
Strategies are for CS-specific computation. AFAIU you want to compare coordinates strictly instead with epsilon in an algorithm. So the differentiation is made based on algorithm, not CS. This probably should not be a strategy but a policy (custom less policy) used only in convex_hull.

Consider this. The most general umbrella strategy can be passed to all algorithms, including `relate` and `convex_hull`. Which compare strategy should be returned by it?

awulkiew:
What do you think about defining less compare policy in convex_hull strategy?

vissarion:
> Strategies are for CS-specific computation. AFAIU you want to compare coordinates strictly instead with epsilon in an algorithm. So the differentiation is made based on algorithm, not CS. This probably should not be a strategy but a policy (custom less policy) used only in convex_hull.
> 
> Consider this. The most general umbrella strategy can be passed to all algorithms, including `relate` and `convex_hull`. Which compare strategy should be returned by it?

@awulkiew that's a good point!

In https://github.com/boostorg/geometry/pull/1162/commits/f719a26b442eb12c0f91fba5cbb980758164cd77 I defined a new less policy (to be used only by convex hull for now) that passes the right equals policy to the strategy. Thus the strategy is responsible for the CS specific computation and the less policy (in the algorithm level) is responsible for the comparison of coordinates (exact or with epsilon). Alternatively, I could have added another template to less (i.e. EqualsPolicy). 

@awulkiew what do you think? I think it is better this way that defining less inside convex hull strategy, this way strategy does not have any information on how we compare coordinates but this is passed by the algorithm. Now type alias `compare` is the same in both `relate` and `convex hull` strategies.

BTW a similar problem holds with `side` method on those strategies, where `relate` uses a default side strategy while `convex hull` a robust one, this is a non-CS specific differentiation.

include/boost/geometry/algorithms/detail/relate/point_point.hpp
barendgehrels: But `less` it is a public policy, meant for all library users. It’s not internal. So that we don’t use it like this does not mean that we can change it.

vissarion: This is updated and now the user can pass either a strategy or a cs tag as the third parameter to `less` (the second choice can be considered as deprecated and removed in the future). So in some sense now `less` is "enhanced". What do you think?

include/boost/geometry/algorithms/detail/disjoint/multipoint_geometry.hpp
barendgehrels: ok

include/boost/geometry/algorithms/detail/relate/point_point.hpp
barendgehrels: sound good!

include/boost/geometry/policies/compare.hpp
```diff
@@ -37,92 +38,244 @@ namespace boost { namespace geometry
...
+                strategy::compare::equals_epsilon
+            >::apply(left, right);
+    }
+};
+
+// for backward compatibility
+
+template <typename Point, int Dimension>
+struct less<Point, Dimension, boost::geometry::cartesian_tag>
+{
```
barendgehrels, line 128: Ok, I see. Indeed `less` is enhanced now, and it wouldn’t break, and these are the implementations. 
So I’m ok now

barendgehrels (approved):
thanks!

awulkiew:
Thanks, it looks better now. But I have one more concern. What if someone passes a different, non-robust or maybe non-cartesian side strategy (assuming it works) into the convex_hull? Would less_exact be the correct choice compatible with strategies other than robust side? Or am I missing something about the way how less compare is defined?

vissarion:
So here the less strategy does have to do with robust side strategy. It is related to how we sort inside convex hull. If you sort with approximate less then you will run into issues like the one this PR solves when e.g. you have input points with very similar x or y coordinates. And these issues will be independent from side strategy.

awulkiew:
Right, thanks for the explanation.
Since `less_exact` is explicitly only used in `convex_hull` does compare strategy has to be defined in umbrella strategies? AFAIU the only change that is needed is to use specific `less` in the algorithm. Or am I missing something?

vissarion:
You mean to pass just the (non umbrella) compare strategy to `less_exact` or even define `less_exact` inside `convex_hull`? I implemented that way because I need to change how numbers are compared inside  compare strategies (e.g. https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/strategies/compare.hpp#L93). Also, `less` and `less_exact` policies now take a strategy instead of a CStag which I think is more correct, right?

Moreover, even not tested, I think that there could be similar problems in the library where `std::sort` is used to sort points or turns. So, it is eventually not only convex hull to be affected.

awulkiew:
Yes, I meant having `less_exact` policy alongside `less` policy, both taking CSTag and using `less_exact` in `convex_hull`. The rationale behind this was to introduce least amount of changes.

> take a strategy instead of a CStag which I think is more correct

To be honest I don't know which is more correct. I feel that the whole less policy could itself be a strategy because it's almost wholly CS-specific. But taking the strategy seems to be more "correct" than taking CSTag.

awulkiew (approved):

PR #1168, opened 2023-07-01
Add if constexpr macro and replace condition macro.

awulkiew:
This is a proposal adding conditional support of C++17 `if constexpr`. I modified only several files for now so we can start a discussion. If we have an agreement I'll modify other places.

My intention was to have a solution as close as possible to the original language feature and to keep `if` keyword so editors can highlight it as usual, i.e.:
```
if constexpr (static_condition)
```
vs
```
if BOOST_GEOMETRY_CONSTEXPR (static_condition)
```

It's a followup of https://github.com/boostorg/geometry/pull/1160

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
@@ -1,7 +1,7 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
 // Copyright (c) 2012-2020 Barend Gehrels, Amsterdam, the Netherlands.
-// Copyright (c) 2022 Adam Wulkiewicz, Lodz, Poland.
+// Copyright (c) 2022-2023 Adam Wulkiewicz, Lodz, Poland.
 
 // This file was modified by Oracle on 2017-2022.
 // Modifications copyright (c) 2017-2022 Oracle and/or its affiliates.
@@ -44,7 +44,7 @@
 #include <boost/geometry/strategies/buffer.hpp>
 #include <boost/geometry/strategies/side.hpp>
 
-#include <boost/geometry/util/condition.hpp>
+#include <boost/geometry/util/constexpr.hpp>
 #include <boost/geometry/util/math.hpp>
 #include <boost/geometry/util/type_traits.hpp>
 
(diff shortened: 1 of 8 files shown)


include/boost/geometry/util/constexpr.hpp
```diff
@@ -0,0 +1,27 @@
...
+#ifndef BOOST_NO_CXX17_IF_CONSTEXPR
+
+#define BOOST_GEOMETRY_CONSTEXPR(CONDITION) constexpr (CONDITION)
+
+#else
+
+#define BOOST_GEOMETRY_CONSTEXPR(CONDITION) (BOOST_GEOMETRY_CONDITION(CONDITION))
+
+#endif
+
```
barendgehrels, line 25: :+1:

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
```diff
@@ -969,7 +969,7 @@ inline void buffer_inserter(GeometryInput const& geometry_input, OutputIterator
                  robust_policy, strategies);
 
     collection.get_turns();
-    if (BOOST_GEOMETRY_CONDITION(areal))
+    if BOOST_GEOMETRY_CONSTEXPR (areal)
```
barendgehrels, line 972: I like it

include/boost/geometry/algorithms/detail/buffer/buffer_inserter.hpp
```diff
@@ -1011,9 +1010,12 @@ inline void buffer_inserter(GeometryInput const& geometry_input, OutputIterator
...
 
-    if (BOOST_GEOMETRY_CONDITION(distance_strategy.negative() && areal))
+    if BOOST_GEOMETRY_CONSTEXPR (areal)
     {
-        collection.discard_nonintersecting_deflated_rings();
+        if (distance_strategy.negative())
+        {
+            collection.discard_nonintersecting_deflated_rings();
+        }
     }
```
barendgehrels, line 1019: Looks better than the original code

barendgehrels (approved):
I agree with this approach, thank you!

vissarion (approved):
Thanks! Looks great!

awulkiew:
@vissarion Thanks! This wasn't finished though. I'll create another PR with the rest. :)

vissarion:
> @vissarion Thanks! This wasn't finished though. I'll create another PR with the rest. :)

I know. I just merged it since there were the 1.83 and the enhancement labels and the beta is closing today. I guess there is no problem to have the rest of the support in 1.84.

Issue #1161, opened 2023-06-12
Incorrect results for geometries or inputs in geographic coordinate system at the edges longitude-latitude range.


rorviksam:
I have noticed that boost::geometry::covered_by, boost::geometry::within, boost::gemetry::intersection, and many more don't work or yield incorrect results in geographic coordinates if the input position (longitude, latitude) or involved geometries have positions at the edges of the longitude-latitude range. e.g.
```c++
    namespace bg = boost::geometry;
    namespace bgs = boost::geometry::strategy;
    using Point = bg::model::point<double, 2, bg::cs::geographic<bg::degree>>;
    using Polygon = bg::model::polygon<Point, true>;
    Point lon_lat{18.8569, 64.5507};
    Polygon sweden;
    bg::read_wkt(
        "POLYGON ((20.548611111111114 69.06, 20.590563888888887 69.05333333333333, "
        "20.621111111111112 69.0463861111111, 20.656116666666666 69.04277777777777, "
        "20.70583333333333 69.04110833333333, 20.75583333333333 69.03666666666666, "
        "20.780833333333334 69.02916666666667, 20.821394444444444 69.01194444444444, "
        "20.854166666666668 68.9975, 20.887777777777778 68.980275, 20.900277777777777 "
        "68.96055277777778, 20.895280555555555 68.94860833333334, 20.86388888888889 "
        "68.94249722222223, 20.832225 68.92916666666667, 20.86472777777778 68.91833055555556, "
        "20.881666666666668 68.90694444444445, 20.879719444444444 68.89777500000001, "
        "20.899166666666666 68.892775, 20.930836111111113 68.89583333333334, 20.969166666666666 "
        "68.89166666666668, 21.003333333333334 68.88555277777779, 21.04361111111111 "
        "68.87777777777777, 21.075836111111112 68.86638888888888, 21.11695 68.85860833333332, "
        "21.14944722222222 68.84333333333333, 21.174722222222222 68.83194444444445, "
        "21.206947222222222 68.82055277777778, 21.240000000000002 68.80860833333332, "
        "21.251669444444445 68.79333055555556, 21.267219444444443 68.77749722222222, "
        "21.294444444444448 68.76472222222222, 21.322499999999998 68.76361111111112, "
        "21.362222222222222 68.765275, 21.379722222222224 68.76555555555555, 21.384166666666665 "
        "68.75805555555556, 21.39388888888889 68.74472222222222, 21.40249722222222 "
[...]

Issue #480, opened 2018-05-08
Empty intersection on polys touching the north pole


remster:
Consider the test case:
```
namespace cartesian {
  using point = boost::geometry::model::d2::point_xy<double, boost::geometry::cs::cartesian>;
  using linestring = boost::geometry::model::linestring<point>;
  using polygon = boost::geometry::model::polygon<point>;
  using multipolygon = boost::geometry::model::multi_polygon<polygon>;
  using mpoint = boost::geometry::model::multi_point<point>;
}

namespace spherical {
  using point = boost::geometry::model::d2::point_xy<double, boost::geometry::cs::spherical_equatorial<boost::geometry::degree> >;
  using linestring = boost::geometry::model::linestring<point>;
  using polygon = boost::geometry::model::polygon<point>;
  using multipolygon = boost::geometry::model::multi_polygon<polygon>;
  using mpoint = boost::geometry::model::multi_point<point>;
  const double earthRadiusMts = 6371000;
}

BOOST_AUTO_TEST_CASE(usaIntersectBigTile90) {
  auto bigTileWKt     = "POLYGON((-182.25 -2.25,-182.25 85,-90 85,4.5 85,4.5 -2.25,-90 -2.25,-182.25 -2.25))";
  auto bigTileWKtOn90 = "POLYGON((-182.25 -2.25,-182.25 90,-90 90,4.5 90,4.5 -2.25,-90 -2.25,-182.25 -2.25))";
[...]

PR #1177, opened 2023-07-14
[within] Fix the case when a segment has as an endpoint a pole

vissarion:
This PR fixes the cases were a geometry has a pole as a point. 
Fixes https://github.com/boostorg/geometry/issues/1161 and (most probably) https://github.com/boostorg/geometry/issues/480.


awulkiew (approved):
LGTM, thanks.

vissarion:
@barendgehrels are you OK with merging this?

barendgehrels (approved):
Looks good to me

Issue #1184, opened 2023-08-15
boost::geometry::intersection algorithm creates spike


vschoech:
My `tc::geo::polygon<int>` type is actually a **multi-polygon**, using a polygon type that is based on **int**, oriented **counter-clockwise** and **open** (not closed). I am using **boost 1.79.0**.

Please consider the following example as it occurred in our code: 

```C++
tc::geo::polygon<int> polygonA;
boost::geometry::read_wkt("MULTIPOLYGON(((1169 177,2004 177,2004 1977,1262 1977,1169 177)))", polygonA);
tc::geo::polygon<int> polygonB;
boost::geometry::read_wkt("MULTIPOLYGON(((1179 371,1175 287,1179 287,1179 371)))", polygonB);
tc::geo::polygon<int> polygonC;
boost::geometry::intersection(polygonA, polygonB, polygonC);
tc::geo::polygon<int> polygonD = polygonC;
boost::geometry::remove_spikes(polygonD);
// polygonC=polygon(MULTIPOLYGON(((1179 370,1179 365,1175 287,1179 287,1179 370))))
// polygonD=polygon(MULTIPOLYGON(((1179 365,1175 287,1179 287,1179 365))))
```

It was my expectation that when the input does not contain spikes, neither does the output from boost::geometry::intersection. Please clarify whether my assumption is wrong or there is a bug in boost::geometry::intersection.

See also:
- https://github.com/boostorg/geometry/issues/603
- https://github.com/boostorg/geometry/issues/1022

vschoech:
Thank you for the quick turn-around! I've seen it in 1.79, but I assume you reproduced in 1.83?

barendgehrels:
Yes, it's reproduced in 1.83

barendgehrels:
I've a concept fix, will create a PR later today (if the concept is fine)

vschoech:
@barendgehrels Thank you! :)

vschoech:
We updated our code base to boost 1.87.0 and I'm happy to confirm that I cannot reproduce this issue any more. ✔️

PR #1185, opened 2023-08-16
[intersection] remove closing point correctly

barendgehrels:
* the function (in namespace detail) is split and two parts are renamed on purpose, because functionality is changed
* tests are added
* intersection.cpp split into new file intersection_integer.cpp
* testing point count now if specified

Fixes: #1184


test/algorithms/set_operations/intersection/test_intersection.hpp
```diff
@@ -179,13 +196,12 @@ typename bg::default_area_result<G1>::type test_intersection(std::string const&
...
     if (! settings.debug)
     {
-        // Check _inserter behaviour with stratey
-        typedef typename bg::strategy::intersection::services::default_strategy
-            <
-                typename bg::cs_tag<point_type>::type
-            >::type strategy_type;
+        // Check inserter behaviour with stratey
+        using strategy_type
+            = typename bg::strategies::relate::services::default_strategy<G1, G2>::type;
```
barendgehrels, line 201: This was a left-over related to the change to umbrella strategies.
Because the define `BOOST_GEOMETRY_TEST_ONLY_ONE_TYPE`, this code was effectively not tested for some years... but the new unit test does not use that define.

test/algorithms/set_operations/intersection/intersection.cpp
```diff
@@ -869,21 +842,6 @@ int test_main(int, char* [])
...
-    test_ticket_10868<int64_t>("MULTIPOLYGON(((33520458 6878575,33480192 14931538,31446819 18947953,30772384 19615678,30101303 19612322,30114725 16928001,33520458 6878575)))");
-#endif
-
-    if (BOOST_GEOMETRY_CONDITION(sizeof(long) * CHAR_BIT >= 64))
-    {
-        test_ticket_10868<long>("MULTIPOLYGON(((33520458 6878575,33480192 14931538,31446819 18947953,30772384 19615678,30101303 19612322,30114725 16928001,33520458 6878575)))");
-    }
-
-    test_ticket_10868<long long>("MULTIPOLYGON(((33520458 6878575,33480192 14931538,31446819 18947953,30772384 19615678,30101303 19612322,30114725 16928001,33520458 6878575)))");
-#endif
```
barendgehrels, line 886: Moved to `intersection_integer.cpp` - but it was excluded anyway, and still fails, for reasons not investigated.
Apart from that, verifying WKT expectations is cumbersome and normally not done in these tests.

test/algorithms/set_operations/intersection/intersection_integer.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+
+template <typename Polygon>
+void test_areal()
+{
+    static bool is_open = bg::closure<Polygon>::value == bg::open;
+
+    ut_settings settings;
+    settings.test_point_count = true;
```
barendgehrels, line 46: This is the way, in these tests, to verify if the closing point is removed.
Some test code (for example `test_point_count`) in this PR relates to that.

vissarion (approved):
Thanks, OK for merging!

PR #1187, opened 2023-08-18
Fix -Wunused-parameter warning

georgthegreat:


barendgehrels (approved):
Thanks!

georgthegreat:
@barendgehrels, can we merge this?
I have a couple more patches to upstream, but I would like to avoid maintaining multiple open PRs

Issue #1186, opened 2023-08-18
union_ discards result because of improper inner ring assignment


AndreMeyerRMC:
In the following union_ - operation, result is empty even though both input shapes are valid. This seems to stem from an improper assignment of an inner ring. In the add_rings<GeometryOut> function, the first inner ring of the SECOND poly in polyB gets assigned to the first output poly, which ist the FIRST ring of polyB and as such completely outside of the inner ring.

Tested in with msvc 14.1 boost 1.82.0 and 1.83.0 and godbolt clang 16 boost 1.82.0
![boost_union2](https://github.com/boostorg/geometry/assets/86659038/571fdbff-cb05-4b02-b6f1-6b8b315e1bfa)

```
#include <iostream>

#include <boost/geometry.hpp>

using boost_coordinate_t = double;
using boost_point_2d = boost::geometry::model::d2::point_xy<boost_coordinate_t>;
using boost_polygon_2d = boost::geometry::model::polygon<boost_point_2d>;
using boost_multipolygon_2d = boost::geometry::model::multi_polygon<boost_polygon_2d>;

int main()
{
    auto polyA = boost::geometry::from_wkt<boost_polygon_2d>("POLYGON((\
-13848.144652755556 6710443.1496919869,\
-13847.699374792446 6710443.1496919869,\
-13847.810694283222 6710440.1096301023,\
-13848.255972246334 6710440.2884572418,\
-13848.144652755556 6710443.1496919869))");

    auto polyB = boost::geometry::from_wkt<boost_multipolygon_2d>("MULTIPOLYGON(((\
-13848.144652755556 6710443.1496919869,\
-13848.144652755556 6710443.3285191935,\
-13847.699374792446 6710443.3285191935,\
-13848.144652755556 6710443.1496919869\
)),((\
-13848.144652755556 6710443.1496919869,\
-13848.255972246334 6710440.2884572418,\
-13847.810694283222 6710440.1096301023,\
-13847.699374792446 6710443.1496919869,\
-13847.365416320112 6710442.9708647905,\
-13846.029582430780 6710442.9708647905,\
-13846.029582430780 6710443.3285191935,\
-13846.029582430780 6710443.6861736169,\
-13845.918262940002 6710444.0438280506,\
-13845.918262940002 6710444.5803097384,\
-13845.918262940002 6710444.7591369776,\
-13845.918262940002 6710445.1167914625,\
-13845.472984976890 6710444.5803097384,\
[...]

Issue #1183, opened 2023-08-14
union_ returns incomplete result polygon


AndreMeyerRMC:
union_ returns an incomplete result polygon in a special case where large portions of the two input polygons overlap

Tested with VS2017 (msvc-14.1), boost Versions 1.79.0, 1.82.0, 1.83.0

![boost_union](https://github.com/boostorg/geometry/assets/86659038/481aebb6-002c-4f84-ba2e-ee09f98ea6c1)

The upper small branch of the polygon is comletely lost in the result. The two shapes have a few points in common but do not overlap exactly in the vicinity of the lost branch.

```
using boost_coordinate_t = double;
using boost_point_2d = boost::geometry::model::d2::point_xy<boost_coordinate_t>;
using boost_polygon_2d = boost::geometry::model::polygon<boost_point_2d>;
using boost_multipolygon_2d = boost::geometry::model::multi_polygon<boost_polygon_2d>;

auto polyA = boost::geometry::from_wkt<boost_multipolygon_2d>("MULTIPOLYGON(((\
-38841.000561763794 6721272.4648206839,\
-38869.843144000479 6721327.7877584957,\
-38880.685990792437 6721344.0451435195,\
-38902.349206128216 6721371.1493254723,\
-38937.993505971914 6721407.9151819283,\
-38925.264019448201 6721389.3887558663,\
-38925.240186032526 6721389.3524003429,\
-38903.891993208345 6721355.1907963008,\
-38909.181455691403 6721352.2237809654,\
-38926.953387540903 6721388.3157829558,\
-38940.906008282145 6721408.6223519640,\
-38961.796165266409 6721432.5740491701,\
-38961.819650679929 6721432.6019898951,\
-38977.627018370375 6721452.1204930553,\
-38977.724235665897 6721452.2645238554,\
-38997.157657657241 6721487.2743892670,\
-38997.156729985494 6721487.2749042027,\
-38997.205159411656 6721487.3621566473,\
-38995.456469315024 6721488.3327662209,\
-38976.018586129794 6721453.3126770724,\
-38959.181300235970 6721432.6429256191,\
-38900.879954713615 6721372.5071140006,\
-38900.816775977277 6721372.4353842726,\
-38879.096795985606 6721345.2601805655,\
-38879.046005852288 6721345.1907021403,\
-38868.149375046065 6721328.8526752470,\
-38868.094590395332 6721328.7601068122,\
-38839.231040705585 6721273.3969513429,\
[...]

Issue #1108, opened 2023-01-23
Union of ring and ring returns empty result.


Mitsuhiko-Matsukawa:
boost1.81.0
Visual Studio 2019. x64 debug

This is a regression from 1.80.0.
If I revert a commit a908a21efeb773a88ec2c8999ba592dc161649e7, it works well.
However, this commit is useful and it's not good to revert.
We need to fix overlay, but I'm sorry I have no idea so far.

If I change #if 1 to 0, it works well.

![image](https://user-images.githubusercontent.com/17040347/213965367-4e5c2441-eb18-44f6-84c8-630ac683bf70.png)

```
void Test()
{
    namespace bg = boost::geometry;
    using point = bg::model::point<double, 2, bg::cs::cartesian>;
    using ring = bg::model::ring<point>;
    using multi_polygon = bg::model::multi_polygon<bg::model::polygon<point>>;
    ring input0, input1;

    input0.assign({
        {17,-2},
        {18,-1.269679093926235014},
        {12,0},
        {22,0},
        {17,-2},
    });

    input1.assign({
        {22,1},
        {22,0},
        {14,0},
#if 1
        {18,-1.2696790939262529996},
#else
        {18,-1.269679093926235014},
#endif
        {12,0},
        {22,1},
    });

    assert(bg::is_valid(input0));
    assert(bg::is_valid(input1));
    multi_polygon output;
    bg::union_(input0, input1, output);
    assert(!output.empty());
}
```

barendgehrels:
I can reproduce this as well, and indeed the get_cluster threshold is important here.
Thanks for this information.

barendgehrels:
I think we have to make the threshold in `get_cluster` higher. Then there is a cluster (obviously). This case is fixed then, and it breaks the cases you reported in #1081 

These case(s) should then be fixed in another way. As far as I can see, there is something wrong or missing in the turn information, from that one cluster (two turns) it should be able to find the next intersection point (which is: itself) and create the inner ring.

Debug info I've now there is:
![image](https://user-images.githubusercontent.com/334849/214595198-989332b2-4f2b-4082-a321-59faf7181a99.png)

It's a bit messy, but the important info is in red: No next IP. That's wrong, it should find one. To be continued.

Mitsuhiko-Matsukawa:
Thank you for chasing this issue.
I agree that the turn information is something wrong and we need to investigate it.

[...]

PR #1189, opened 2023-08-23
[fix] overlay threshold for sort_by_side

barendgehrels:
Fixes: issue #1186


include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
```diff
@@ -325,7 +325,7 @@ public :
                 double
             >::type;
 
-        ct_type const tolerance = 1000000000;
+        static auto const tolerance = common_approximately_equals_epsilon<ct_type>::value();
```
vissarion, line 328: this is the actual change right?

vissarion (approved):
Thanks! 

It would help a lot if you comment on the geometric intuition of the value 100 that is used as threshold.

Regarding https://github.com/boostorg/geometry/issues/1183 is there some other combination of values that fixes that or you believe it is of different nature?

include/boost/geometry/algorithms/detail/overlay/sort_by_side.hpp
barendgehrels: Indeed!

barendgehrels: I'll document the `100` later. It's not new, it's moved. 
It is recently (earlier this year) adapted from a larger value to this value, considering another issue and all current tests.
It is used for clustering turns, and now also for sort by side. In both cases the exact value is not that important, but it should not be 1 (for floating point comparisons) and not be too large (obviously). The tolerance which was now replaced was very large, I don't remember exactly why.

barendgehrels: See also #1108

PR #1191, opened 2023-08-28
Modernize noexcept specifications

georgthegreat:
As boost::geometry mandates C++14 since 1.75, I assume this can be merged in C++03-incompatible form.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
@@ -46,10 +46,7 @@ public:
         message += method;
     }
 
-    virtual ~turn_info_exception() throw()
-    {}
-
-    virtual char const* what() const throw()
+    virtual char const* What() const noexcept
     {
         return message.c_str();
     }
--- a/include/boost/geometry/algorithms/centroid.hpp
+++ b/include/boost/geometry/algorithms/centroid.hpp
@@ -91,7 +91,7 @@ public:
     \brief Returns the explanatory string.
     \return Pointer to a null-terminated string with explanatory information.
     */
-    virtual char const* what() const throw()
+    virtual char const* what() const noexcept
     {
         return "Boost.Geometry Centroid calculation exception";
     }
--- a/include/boost/geometry/algorithms/detail/has_self_intersections.hpp
+++ b/include/boost/geometry/algorithms/detail/has_self_intersections.hpp
@@ -53,7 +53,7 @@ public:
 
     inline overlay_invalid_input_exception() {}
 
-    virtual char const* what() const throw()
+    virtual char const* what() const noexcept
     {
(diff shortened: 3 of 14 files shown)


barendgehrels (approved):
Looks good to me, indeed C++14 syntax is fine now

barendgehrels:
One thing, @vissarion , is it better to split it into two commits?
One for library code, one for extensions?

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -46,10 +46,10 @@ class turn_info_exception : public geometry::exception
         message += method;
     }
 
-    virtual ~turn_info_exception() throw()
+    virtual ~turn_info_exception()
     {}
```
barendgehrels, line 50: We can better remove these constructors and destructors, in most cases we don't define them.

georgthegreat: Maybe we can merge this as is?
I will do the cleanup later on.

barendgehrels: You probably have to change it anyway, because the commit mixes extensions and normal code...
But of course, indeed, we could do the cleanup in a separate PR tool.

georgthegreat: I do not understand the problem with the extensions directory.

Could you, please, explain it to me?

georgthegreat: Btw, it is unclear if these can be removed or defaulted.

This is more like a codestyle consideration.

barendgehrels: Indeed, it is. I think they can all be removed, don’t see what it adds.

About `extensions`: we often merge from develop to master by cherry-picking commits. Master does not contain anything from the `extensions` folder. So then cherry-picking does not work or needs manual editing.

georgthegreat:
I have removed changes in extensions/

awulkiew:
@georgthegreat Since you removed the unneded destructors would you be willing to make some other related improvements as well or would you prefer to merge this as it is?

The improvements I'm thinking about are:
- the removal of unneeded default constructors or making them  `= default` in case their existance is documentation-related
- replacement of `virtual` in `what()` declaration with `override`

georgthegreat:
@awulkiew, I am ok with doing this, but I want this PR merged.
Other improvements are definitely out of scope of this PR.

barendgehrels:
Thanks for splitting it.

georgthegreat:
Let's merge it.

vissarion (approved):
Thanks!

georgthegreat:
@awulkiew, virtual -> override replacement is here:
https://github.com/boostorg/geometry/pull/1198/files

PR #1194, opened 2023-09-01
[strategies] Fix msvc conversion warning in centroid::average

awulkiew:
Msvc complains about the conversion from integer to floating point at some higher warning level.


barendgehrels (approved):
Looks good to me

vissarion (approved):
LGTM, thanks!

PR #1195, opened 2023-09-03
Switch to `boost::core::invoke_swap`. Add missing includes.

Lastique:
`boost::swap` is deprecated and will be removed. Use `boost::core::invoke_swap` as a replacement.

Also add missing includes.

### DIFF
--- a/include/boost/geometry/formulas/karney_inverse.hpp
+++ b/include/boost/geometry/formulas/karney_inverse.hpp
@@ -32,6 +32,7 @@
 #define BOOST_GEOMETRY_FORMULAS_KARNEY_INVERSE_HPP
 
 
+#include <boost/core/invoke_swap.hpp>
 #include <boost/math/constants/constants.hpp>
 #include <boost/math/special_functions/hypot.hpp>
 
@@ -191,7 +192,7 @@ public:
         if (swap_point < 0)
         {
             lon12_sign *= -1;
-            swap(lat1, lat2);
+            boost::core::invoke_swap(lat1, lat2);
         }
 
         // Enforce lat1 to be <= 0.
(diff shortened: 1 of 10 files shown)


include/boost/geometry/formulas/karney_inverse.hpp
```diff
@@ -448,9 +449,9 @@ class karney_inverse
 
         if (swap_point < 0)
         {
-            swap(sin_alpha1, sin_alpha2);
-            swap(cos_alpha1, cos_alpha2);
-            swap(result.geodesic_scale, M21);
+            boost::core::invoke_swap(sin_alpha1, sin_alpha2);
+            boost::core::invoke_swap(cos_alpha1, cos_alpha2);
+            boost::core::invoke_swap(result.geodesic_scale, M21);
```
barendgehrels, line 454: We can, if I see it correctly, replace these calls (here) with just `std::swap`
And maybe on more places.
But for now this PR looks okay to me.

barendgehrels (approved):

vissarion (approved):
Thanks, I am OK with merging. 

ps. there are some CI tests failing but probably not related to this PR

Lastique:
The test failures show that some Boost.Geometry headers are incomplete and probably broken. I didn't know how to fix it, and it wasn't the point of this PR anyway.

For example, [this](https://github.com/boostorg/geometry/actions/runs/6061347079/job/16446454333#step:7:281) seems to rely on `get` being found by ADL, but ADL is not performed when template parameters are explicitly specified. At least, not in C++17 that is used by default in gcc 11. Or, if that `get` should be available without ADL, it is not available, so an include must be missing (I don't know which).

PR #1197, opened 2023-09-08
Use boost::size to allow accessing ring size via range_size method

georgthegreat:


georgthegreat:
@vissarion, could you, please, take a look?

vissarion (approved):
Thanks!

barendgehrels (approved):
Thanks!

Issue #1169, opened 2023-07-03
[regression] MSVC Warning C4459: declaration of 'indices' hides global declaration


wanghan02:
There is a fixed [ticket#835](https://github.com/boostorg/geometry/issues/835). Now the warning pops out again with boost 1.82.0 at a different location. MSVC version is VS2022 17.1.0.

```
boost_1_82_0\boost\geometry\algorithms\detail\overlay\colocate_clusters.hpp(89,1): warning C4459: declaration of 'indices' hides global declaration
boost_1_82_0\boost\multi_array\base.hpp(70,32): note: see declaration of 'boost::`anonymous-namespace'::indices'
boost_1_82_0\boost\geometry\algorithms\detail\overlay\handle_colocations.hpp(100): note: see reference to function template instantiation 'void boost::geometry::detail::overlay::colocate_clusters<Clusters,Turns>(const Clusters &,Turns &)' being compiled
        with
        [
            Clusters=cluster_type,
            Turns=turn_container_type
        ]
```

wanghan02:
This is still a problem in boost 1.83.0.

barendgehrels:
Fixed by #1202

Issue #1138, opened 2023-04-21
Regression 1.78.0 - 1.79.0 (still in 1.82.0): difference of two polygons gives unexpected empty result


gastald88:
I have two polygons p1 and p2, where the difference p1 - p2 is clearly non-empty (see image, p1 blue, p2 orange), but the difference operation returns an empty polygon.

![image](https://user-images.githubusercontent.com/37936526/233601244-b9849164-f69c-4fe0-b408-35766b321960.png)

Here below a godbolt link to replicate this problem:
https://godbolt.org/z/YTjM3GG1G

The result is correct in boost version 1.78.0 (you can set at the godbolt platform in the compiler arguments `-I/opt/compiler-explorer/libs/boost_1_78_0)`

From version 1.79.0 on (until the 1.82.0 that I recentely tested on my local environment) the result is incorrect.

I also tried to define the macro BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE, but the bug is still there.

vissarion:
Thanks for filing this issue.

I can reproduce your issue locally.  

However, defining BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE (before all includes) is working for me.
This seems to be one regression of rescaling removal in 1.79.

barendgehrels:
Fixed by commit eb879fe62
It now also works without rescaling.
I will create a unit test and close the issue.

barendgehrels:
Unit test in #1202

Issue #1103, opened 2022-12-16
Regression 1.79.0: union operation of two polygons outputs zero polygons


paulk125:
I have two polygons. Both are nearly unit squares. One is from (0,epsilon)-(1,1) and the other is from (0,-1)-(1,0). So these two squares are not quite touching. I would expect the geometry union operation to either return one merged polygon or two separate polygons. Instead it returns zero polygons.

Here is a godbolt link to duplicate the problem:
https://godbolt.org/z/P16oKM6Pn

Note that this bug affects 1.79 and later. If you change the compiler option in the godbolt link to include the 1.78 boost headers, then the union operation works fine:

`-I/opt/compiler-explorer/libs/boost_1_78_0`

barendgehrels:
Unit test is added.

PR #1202, opened 2023-09-14
Various minor improvements

barendgehrels:
Fixes issue #1169 
Adds unit tests for issue #1138 which was already fixed
Adds unit test for issue #1103 which was already fixed
Adds comments for epsilon as suggested by @vissarion 
And some stylistic issues (const placements, includes)


include/boost/geometry/io/dsv/write.hpp
```diff
@@ -27,6 +27,7 @@
 #include <boost/range/begin.hpp>
 #include <boost/range/end.hpp>
 #include <boost/range/value_type.hpp>
+#include <boost/range/size.hpp>
```
barendgehrels, line 30: After updating Boost to the most recent develop on my notebook (because of swap), this missing include was problematic for some unit tests, which is the reason why I checked all usages.

vissarion: indeed, this is caught by github actions too.

vissarion (commented):
Thanks! I am OK in general but I think it's worth it to fix the missing header issue.

barendgehrels:
> Thanks! I am OK in general but I think it's worth it to fix the missing header issue.

For me it did not fail. But I added the include there too and it should now be fine. Thank you.

barendgehrels:
Now all tests pass. @vissarion could you approve?

vissarion (approved):
Sure, thanks!

PR #1204, opened 2023-09-30
Fixes for miscellaneous gcc/clang warnings in the test suite.

tinko92:
This fixes a couple of warnings to declutter the output of running `b2 test` with GCC 13 or Clang 16 (and possibly other versions which I do not have installed):
* a missing template keyword warning with GCC (which looks like a compiler bug to me but the workaround only costs two brackets),
* some warnings for comparing signed and unsigned integers,
* missing parentheses warnings,
* an unused parameter warning (just commented out because the lining that would have used it is also just commented out).

The remaining warnings (with the compilers that I have) come from Boost.Polygon and Boost.Test.

### DIFF
--- a/include/boost/geometry/formulas/meridian_direct.hpp
+++ b/include/boost/geometry/formulas/meridian_direct.hpp
@@ -84,8 +84,8 @@ public:
             {
                 result.reverse_azimuth =  pi;
             }
-            else if (result.lat2 < -half_pi &&
-                     result.lat2 >  -one_and_a_half_pi)
+            else if (result.lat2 > -one_and_a_half_pi &&
+                     result.lat2 < -half_pi)
             {
                 result.reverse_azimuth =  c0;
             }
--- a/test/algorithms/buffer/buffer_linestring_geo.cpp
+++ b/test/algorithms/buffer/buffer_linestring_geo.cpp
@@ -71,7 +71,7 @@ void test_linestring()
     }
 
     settings.test_area = false;
-    auto const n = sizeof(testcases_aimes) / sizeof(testcases_aimes[0]);
+    int const n = sizeof(testcases_aimes) / sizeof(testcases_aimes[0]);
 
     // Cases (ouf of 197) where the guessed area estimations are not met.
     // If this needs to be changed, be sure to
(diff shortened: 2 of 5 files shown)


vissarion (approved):
Thanks!

include/boost/geometry/formulas/meridian_direct.hpp
```diff
@@ -84,7 +84,7 @@ class meridian_direct
             {
                 result.reverse_azimuth =  pi;
             }
-            else if (result.lat2 < -half_pi &&
+            else if ((result.lat2 < -half_pi) &&
```
barendgehrels, line 87: Weird that it warns for this, and not for the rest. We never have brackets here, unless and/or are combined.
But anyway - if it fixes a warning, it's fine.

barendgehrels (approved):
Thanks!

include/boost/geometry/formulas/meridian_direct.hpp
tinko92: I couldn't really make sense of it and the warning is only shown with GCC IIRC. `lat2`, assuming I didn't miss an alternative definition of the result type is just not a templated field. My only guess (which I don't find convincing) is that it maybe misreads part of
```
else if (result.lat2 < -half_pi &&
         result.lat2 >  -one_and_a_half_pi)

```
As `result.lat2</*boolean template argument*/> - one_and_a_half_pi`. But then the logic that decides whether to emit a warning here would have to be different from the logic that correctly interprets and compiles that code and I would not expect a bug like that in GCC and I have no experience debugging GCC.

vissarion: Are you getting the same warning with gcc 12?

tinko92: Yes, I get the warning with both gcc12 and gcc13.

tinko92: Changing the second line of the condition to `-one_and_a_half_pi < result.lat2` also makes the warning disappear.

barendgehrels: Interesting...

barendgehrels: Could it be that it somehow interprets `< -half_pi && result.lat2 >` as a template?
Maybe you can try to reverse the conditions
```
(result.lat2 >  -one_and_a_half_pi 
 && result.lat2 < -half_pi)
         
```

barendgehrels: or probably not, it was a warning. The reversal you suggested is also fine.

PR #1205, opened 2023-10-02
Remove some access.hpp includes in overlay

tinko92:
This removes the include of access.hpp from some files in the overlay algorithm that don't use it (AFAICS they don't use get, set or the {max,min}_corner constants) and moves the box_box_intersection implementation out of the overlay algorithm where it is not used directly.

(This is a by-product of my larger attempt to locate and separate direct computations involving coordinates from the overlay-code, which I think is a prerequisite for implementing and testing numerically robust strategies. Because I don't know how long that might take, whether it will succeed and how much the develop-branch might change in the mean time, I thought it made sense to push this self-contained change now.)

### DIFF
--- a/include/boost/geometry/algorithms/detail/intersection/box_box.hpp
+++ b/include/boost/geometry/algorithms/detail/intersection/box_box.hpp
@@ -16,7 +16,7 @@
 
 
 #include <boost/geometry/algorithms/detail/intersection/interface.hpp>
-#include <boost/geometry/algorithms/detail/overlay/intersection_box_box.hpp>
+#include <boost/geometry/algorithms/detail/intersection/box_box_implementation.hpp>
 
 
 namespace boost { namespace geometry
--- a/include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
@@ -39,7 +39,6 @@
 #include <boost/geometry/algorithms/detail/sections/section_functions.hpp>
 #include <boost/geometry/algorithms/detail/sections/sectionalize.hpp>
 
-#include <boost/geometry/core/access.hpp>
 #include <boost/geometry/core/assert.hpp>
 #include <boost/geometry/core/coordinate_dimension.hpp>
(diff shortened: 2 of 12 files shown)


include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
```diff
@@ -741,7 +740,7 @@ struct get_turns_cs
         else if (c < left) return -1;
         else if (c > right) return 1;
         else return 0;
-    }
+    }*/
```
barendgehrels, line 743: You can remove this one. I don't remember exactly but I probably should have cleaned it up earlier.

include/boost/geometry/algorithms/detail/intersection/box_box.hpp
```diff
@@ -16,7 +16,7 @@
 
 
 #include <boost/geometry/algorithms/detail/intersection/interface.hpp>
-#include <boost/geometry/algorithms/detail/overlay/intersection_box_box.hpp>
+#include <boost/geometry/algorithms/detail/intersection/box_box_implementation.hpp>
```
barendgehrels, line 19: Why is this? I'm not against, but we don't have this convention.

barendgehrels (approved):
Thanks!

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
tinko92: I removed the get_side function that was commented out and the commented out that was calling it. If the latter was meant to be kept, I can change the commit.

include/boost/geometry/algorithms/detail/intersection/box_box.hpp
barendgehrels: All right, clear, thanks!

include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
barendgehrels: This is completely fine, thanks!

vissarion (approved):
Thanks LGTM.

Issue #1208, opened 2023-10-16
Convex hull: wrong results (counter-example on 4 points)


atalon-lip6:
Hello,
In our project (topology-toolkit) we wanted to use Boost instead of Qhull to have fewer third party libraries. We had some bug appearing when using Boost implemantation for convex hull, and after some investigation I found a minimal counter-example.

Three points are aligned, and a fourth one (a2596) make a 90° angle with this line. Whether or not the middle point of the vertical segment is included may depend on precision errors.
However, the point with largest x-coordinates should obviously be taken. Boost does not include it and includes the middle point in the convex hull, hence it is wrong.

I static cast doubles into doubles because it could happen in our program. If we remove the static_casts, then the output is still wrong but boost returns different points (not taking the point with larger y-coordinate):
0.0285095558699018,0
0.0469472760530917,0
0.0285095558699018,0
0.0285095558699018,0

Here is the cpp code of the example I am talking about:
```

#include <boost/geometry.hpp>
#include <boost/geometry/geometries/polygon.hpp>
#include <boost/geometry/geometries/adapted/boost_tuple.hpp>

#include <iostream>


namespace bg = boost::geometry;
BOOST_GEOMETRY_REGISTER_BOOST_TUPLE_CS(cs::cartesian)

using Point = boost::tuple<double, double>;
using Polygon =  boost::geometry::model::polygon<Point>;
using Mpoints =  boost::geometry::model::multi_point<Point>;

int main(void)
{
  std::cout << std::setprecision(15);

  double a2738[2] = {0.028509555869901827,0.028509555869901813};
  double a2899[2]= {0.02850955586990183, 0.017233466136418303};
  double a2537[2] = {0.02850955586990183, 0};
  double a2596[2] = {0.046947276053091694, 0};

  Mpoints multiPoints;
  Polygon hull;

  boost::geometry::append(multiPoints, Point(static_cast<double> (a2738[0]), static_cast<double>(a2738[1])));

  boost::geometry::append(multiPoints, Point(static_cast<double> (a2899[0]), static_cast<double>(a2899[1])));
  boost::geometry::append(multiPoints, Point(static_cast<double> (a2537[0]), static_cast<double>(a2537[1])));
  boost::geometry::append(multiPoints, Point(static_cast<double> (a2596[0]), static_cast<double>(a2596[1])));

  for (const auto &pt : multiPoints)
    std::cout << "Point:  " << pt.get<0>() << "," << pt.get<1>() << std::endl;
[...]

PR #1209, opened 2023-10-16
Use fp_equals comparison in side_strategy for convex_hull.

tinko92:
This changes the default side strategy for convex_hull for cartesian coordinates so that it uses an exact comparison to zero for the side value. This is IMHO the only way to use that strategy to guarantee exact (and thereby consistent) results, which was the intention behind the strategy. The motivation is issue #1208 , which is fixed by this change. I see no testcase breaking.

From git blame, I don't see the motivation for the current non-exact comparison. If it was to be more lenient in is_convex, which uses the same umbrella strategy, then maybe those should be separate umbrella strategies. I think in the construction of the hull this non-robustness is more problematic. If it was done like this to avoid near-collinear points on edges of the output, maybe there should be a separate iteration over the hull after construction with another side-strategy.

### DIFF
--- a/include/boost/geometry/strategies/convex_hull/cartesian.hpp
+++ b/include/boost/geometry/strategies/convex_hull/cartesian.hpp
@@ -46,7 +46,7 @@ public:
     static auto side()
     {
         using side_strategy_type
-            = strategy::side::side_robust<CalculationType>;
+            = strategy::side::side_robust<CalculationType, strategy::side::fp_equals_policy>;
         return side_strategy_type();
     }
 


atalon-lip6:
I can confirm it solves a case where Boost was not including one point with maximum x-coordinates in the convex hull.

vissarion:
Let's wait for @barendgehrels to approve, who has been involved in those changes of robust and non robust strategies.

barendgehrels (approved):
Sorry for the delay. I'm OK

PR #1214, opened 2023-11-11
Fix long compilation times for epsg

### LINKED ISSUE #1006: bg::projections::detail::epsg_to_parameters causes excessive compile times
Hi,

First of all, thanks for this excellent library! 

# Intro

Just to notify about an issue that might be of interest for those using the  projections part of boost geometry,
limited to the epsg projection engine.

To be clear: This is not an issue about the correctness of the code, rather practical implications regarding excessive compile times.

The function in question is the  boost::geometry::projections::detail::epsg_to_parameters(), that does the job of converting an epsg code to parameters that describes the transformation.

E.g. given epsg code (an int) return back boost::geometry::srs::parameters<>.

This is realized using a static array inside the mentioned function, declarative and clear to read.

## the issue

The only issue is the time that the compiler need in order to resolve the expressions in the static table,
 chained call, 
starting with parameters<>(projection)(param1)....(param1) // the call operator is overloaded

Compile times for pre gcc 12.1 was already high (several minutes),
 but with the new gcc 12.1 it becomes > 30 minutes (33 minutes on ryzen 9 5950X series, only one core used).

On godbolt, the mininal sample below works for pre 12.1 compilers, but time-out on 12.1.

```c++
#include <boost/geometry/srs/projections/dpar.hpp>
#include <boost/geometry/srs/projections/epsg.hpp>

namespace test {
    size_t spend_massive_compile_time(int epsg_code) {
        auto dp = boost::geometry::projections::detail::epsg_to_parameters(epsg_code);
        return dp.size();
    }
}

int main(int argc, char **argv) {
    return test::spend_massive_compile_time(32631); // should return 6
}
``` 

### Simplify the expressions in the table

I tried to rework the static table, trying to avoid using chained expressions, 
rather use use plain vector<parameter<T>> {}.

It was kind of promising, 10 times faster!
 - so 3 minutes instead of 30 minutes. 

Looking at the data that goes into the static table, it's basically code,'numbers', strongly typed,
so moving along in a direction where the data that drives the resulting parameters are simple/fast compile-time types, 
it might be possible to get down to 'zero' compile-time, 
- at the cost of some run-time. (because the construction of the wanted object must be done at runtime)

E.g. creating the parameters only when really asked for, 
by means of typed constants/tuples, that are cheap during compile time.

A local test with initial approach, using the same table definitions, 
 reduces the compile time from 33 minutes to 3 minutes. (promising) 
So it illustrates that it works, even though 3 minutes to compile a table of approx 5k rows is still excessive.

Further progress on this shows that compiling a simple POD approach,
takes only 0.8 seconds. (compared to 33 minutes gcc 12.1, or 3 minutes pre gcc 12). 

**So approx 2000 times faster than the current situation for gcc 12.1** (and any boost geometry version, is my guess).

I think with a little work on this approach it might be worth considering, 
unless there are other more obvious solutions around that I have overlooked.

Sketches for this approach looks like:

```c++
#include <boost/geometry/srs/projections/dpar.hpp>


namespace boost::geometry::projections::detail {
    namespace epsg_data {
        /**
         * @brief simplest possible pod to capture a parameter definition
         */
        struct epsg_p_def {
            int e_type;///< unique enum type, one of name_xxx enums.
            int e_val{0}; ///< sub-enum type, one of value_xxx enums
            int n_args{0};///< number of floating point args, dependent on e_type,e_val
            double args[7];//most parameters are 1, towgs84 3 or 7
        };
        /**
         * @brief  pod to keep one epsg def, a set of epsg_p_defs.
         */
        struct epsg_def {
            epsg_p_def p_defs[13];// currently number of p_defs are 13(!)
        };
        struct epsg_entry {
            int code;
            epsg_def parameters;
        };
        using namespace srs::dpar;

        static const epsg_entry arr[] = {
            {2000, {{{proj,proj_tmerc},{lat_0,0,1,{0}},{lon_0,0,1,{-62}},{k,0,1,{0.9995000000000001}},{x_0,0,1,{400000}},{y_0,0,1,{0}},{ellps,ellps_clrk80},{units,units_m},{no_defs}}}},
            {2001, {{{proj,proj_tmerc},{lat_0,0,1,{0}},{lon_0,0,1,{-62}},{k,0,1,{0.9995000000000001}},{x_0,0,1,{400000}},{y_0,0,1,{0}},{ellps,ellps_clrk80},{units,units_m},{no_defs}}}},
            {2002, {{{proj,proj_tmerc},{lat_0,0,1,{0}},{lon_0,0,1,{-62}},{k,0,1,{0.9995000000000001}},{x_0,0,1,{400000}},{y_0,0,1,{0}},{ellps,ellps_clrk80},{towgs84,0,7,{725,685,536,0,0,0,0}},{units,units_m},{no_defs}}}},
            {2003, {{{proj,proj_tmerc},{lat_0,0,1,{0}},{lon_0,0,1,{-62}},{k,0,1,{0.9995000000000001}},{x_0,0,1,{400000}},{y_0,0,1,{0}},{ellps,ellps_clrk80},{towgs84,0,7,{72,213.7,93,0,0,0,0}},{units,units_m},{no_defs}}}},
            {2004, {{{proj,proj_tmerc},{lat_0,0,1,{0}},{lon_0,0,1,{-62}},{k,0,1,{0.9995000000000001}},{x_0,0,1,{400000}},{y_0,0,1,{0}},{ellps,ellps_clrk80},{towgs84,0,7,{174,359,365,0,0,0,0}},{units,units_m},{no_defs}}}},
            {2005, {{{proj,proj_tmerc},{lat_0,0,1,{0}},{lon_0,0,1,{-62}},{k,0,1,{0.9995000000000001}},{x_0,0,1,{400000}},{y_0,0,1,{0}},{ellps,ellps_clrk80},{units,units_m},{no_defs}}}},
          // some more thousands rows follows here
        };
    }
}

```

### Make the table/instantiation part of a .cpp  file 

If the table was created as part of the compilation of the boost library itself, instead of header-only for this part, that  might work.
Maybe in combination with the speedup approach, 30 minutes is still some time, for a single file.

A variant is to control the inline/static part with a compile-time define, so that affected users can choose to compile the table-part in a separate unit (avoiding repeated compilations of the stuff that takes time)
Drawbacks with this approach is clearly that it breaks header-only, so the 'hybrid' user def constant specified approach could work, helping out the users that are affected.

tinko92:
This PR is intended to fix #1006 . @vissarion , I saw your comment about GCC 12.1 being an outlier at the time but unfortunately it now occurs with GCC 13.2.1 and GCC 12.3.0 too. I also get long (40s-60s on an Intel i7-12800H) compilation times with Clang 16 (haven't tried other versions) if I use any kind of optimization (-O{1,2,3}), which seems unreasonable for a <10k line header that contains only a list of static data. I didn't record the exact numbers but I also saw excessive compilation times for this header with MSVC.

This PR changes the code_element type to a statically sized object without heap allocations (similar to the one proposed by @sigbjorn in #1006 ). Since the keys are all known at compile-time, the binary search can also be replaced with perfect hashing.

A test for the equality of the parameters-arrays in this PR and the one currently in boost is implemented here: https://gist.github.com/tinko92/3b55c5288b8ff229e541cfa76257f892

The code that was used to generate the code in this PR from the old one is found here (uses std::format and magic_enum):
https://gist.github.com/tinko92/869d610859e53a2a1906b7ae19b67e8c

The perfect hashing functions are generated using https://github.com/rurban/nbperf and then slightly edited (formatting, constexpr).

# Advantages
- 600x faster compilation with affected GCC versions, 50x faster compilation with Clang if optimizations are enabled, generally <1s compilation times across all tested configurations.
- Worst-case O(1) lookup rather than O(log(n)).
- No more ~4k heap allocations when epsg_to_parameters is first called.
- Functions in the binary are a lot smaller, all the data is in .rodata.
- Overall compiled binary size for the example in issue #1006 is 560kB (~40%) smaller with Clang 16 and -O2. For equivalent examples with esri.hpp and iau2000.hpp, the size reductions are 100kB (~53%) and 332kB (~64%) respectively. Greater savings with GCC.

# Disadvantages
- Adding/removing transformations requires rehashing. This could be avoided by reverting to binary search at the cost of O(1) lookup.
- Code is less readable and longer than before (but imho still reasonably readable).

I was hoping, it would also lead to smaller binaries when the code is known at compile-time (At least Clang seems to be able in principle to optimize away this kind of hashing with static keys: https://godbolt.org/z/Kq66anToo) but that did not work out unfortunately, maybe I am missing something to make that work.


barendgehrels:
@tinko92 I didn't react at that time - because I wanted to think about it (and later forgot).
I'm not in favour of this. It adds too much complexity. Can we close it?

tinko92:
> @tinko92 I didn't react at that time - because I wanted to think about it (and later forgot). I'm not in favour of this. It adds too much complexity. Can we close it?

I understand, yes it can be closed.

If the main complexity objection is with the perfect hashing, maybe a different PR that just replaces the current chaining (which I think leads to the long compile times) with a simpler initialisation approach can be considered to address #1006 ? I am currently very time-constrained, but I will try to look into this again in summer.

barendgehrels:
Thank you! And thanks for the ref to #1006 again. It's listed as close because they found another solution - so we can verify the need and the workflow.

Issue #1223, opened 2023-12-10
MSVC CI may not take effect


FantasqueX:
As you can see in [the latest CI result](https://github.com/boostorg/geometry/actions/runs/7101383204/job/19329377831), `warning: Did not find command for MSVC toolset. If you have Visual Studio 2017 installed you will need to specify the full path to the command, set VS150COMNTOOLS for your installation, or build from the 'Visual Studio Command Prompt for VS 2017'.` and no test is executed.

The problem is that there is no msvc-15 or msvc-16. MSVC supports c++14 from vs2015 a.k.a. msvc-14.0. b2_toolset should be
```
b2_toolset: [
  msvc-14.0,
  msvc-14.1,
  msvc-14.2,
  msvc-14.3,
]
```

vissarion:
Thanks for opening this. It seems that msvc-14.0 -- msvc-14.2 return the same error (see https://github.com/boostorg/geometry/pull/1224).  Probably there is a need for extra setting to make them work on CI.

FantasqueX:
On my machine (Windows 11 with vs2019 installed), `./b2 toolset=msvc-14.2` works well.
According to [Github action documentation](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md), both vc142 and vc143 of vs2022 are installed. I don't know why b2 cannot find vc142. Maybe some more investigation needs to be done.

barendgehrels:
@vissarion , this looks outdated, can we close it?

barendgehrels:
OK thanks for checking. Then we can keep it open.
Personally, I don't use `b2` anymore.

PR #1224, opened 2023-12-12
[ci] Fix mscv versions

vissarion:
Closes https://github.com/boostorg/geometry/issues/1223

### DIFF
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -282,17 +282,77 @@ jobs:
       fail-fast: false
       matrix:
         b2_toolset: [
-          msvc-14.3,
-          msvc-15,
-          msvc-16,
+          msvc-14.3
         ]
 
         include:
           - b2_toolset: msvc-14.3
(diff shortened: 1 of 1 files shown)


barendgehrels (approved):
Looks OK to me

barendgehrels:
> @barendgehrels even if it looks OK it isn't. The tests look green but there are the same warnings as before. See [#1223 (comment)](https://github.com/boostorg/geometry/issues/1223#issuecomment-1853623770)

Okay... I didn't work in this CI yet, so you can ignore my observations.

vissarion:
OK, it seems there is a way to use `msvc-14.2` (Visual Studio 19) in `windows-2019` runner. The other two available runners `windows-2022` and `windows-2025` or `windows-latest` both support `msvc-14.3` (Visual Studio 22). Finally, there is not choice to run tests on older versions i.e. `msvc-14.1` (Visual Studio 17) because there is no runner in github actions that supports that version.

Issue #1230, opened 2024-01-26
union_ perturbs a point and returns wrong result


vissarion:
Boost geometry returns a wrong result by actually perturb slightly a point of the input that should appear in the output as is. 

In particular, the following program 

```cpp
#include <boost/geometry.hpp>

int main(){

    using Point = boost::geometry::model::d2::point_xy<double>;
    using Polygon = boost::geometry::model::polygon<Point>;

    Polygon p1;
    boost::geometry::read_wkt("POLYGON((5 0,15 25,25 0,15 5,5 0))", p1);
    Polygon p2;
    boost::geometry::read_wkt("POLYGON((25 0,0 15,30 15,22 10,25 0))", p2);
    Polygon p3;
    boost::geometry::read_wkt("POLYGON((5 0,7 10,0 15,10 15,15 25,20 15,30 15,22 10,25 0,15 5,5 0))", p3);

    boost::geometry::model::multi_polygon<Polygon> result;
    boost::geometry::union_(p1, p2, result);

    boost::geometry::model::multi_polygon<Polygon> result2;
    boost::geometry::union_(result, p3, result2);

    std::cout << boost::geometry::wkt(result2) << std::endl;

    return 0;
}
```

returns `MULTIPOLYGON(((8.88178e-16 15,10 15,15 25,20 15,30 15,22 10,25 0,15 5,5 0,7 10,8.88178e-16 15)))` in boost 1.84 while returns the correct result `MULTIPOLYGON(((0 15,10 15,15 25,20 15,30 15,22 10,25 0,15 5,5 0,7 10,0 15)))` in 1.78.  
The perturbed point here is POINT(0 15).

This is probably a regression of removing rescaling as a default in 1.79.

PR #1237, opened 2024-01-31
Fix numerical issue for segments sharing a common point

vissarion:
Fixes https://github.com/boostorg/geometry/issues/1230


include/boost/geometry/strategies/cartesian/intersection.hpp
```diff
@@ -459,6 +459,7 @@ struct cartesian_segments
 
         bool collinear = sides.collinear();
 
+        //TODO: remove this
```
barendgehrels, line 462: In this PR?

test/algorithms/Jamfile
```diff
@@ -39,9 +39,6 @@ test-suite boost-geometry-algorithms
     [ run is_valid.cpp                 : : : : algorithms_is_valid ]
     [ run is_valid_failure.cpp         : : : : algorithms_is_valid_failure ]
     [ run is_valid_geo.cpp             : : : : algorithms_is_valid_geo ]
-    [ run is_valid.cpp                 : : : <define>BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE : algorithms_is_valid_alternative ]
-    [ run is_valid_failure.cpp         : : : <define>BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE : algorithms_is_valid_failure_alternative ]
-    [ run is_valid_geo.cpp             : : : <define>BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE : algorithms_is_valid_geo_alternative ]
     [ run line_interpolate.cpp         : : : : algorithms_line_interpolate ]
```
barendgehrels, line 42: In general, I think we should remove these in a separate PR.
Now it looks just like it's done because the change still gives errors in rescaling code.
By chance, I added all these in the cmake PR - but if it is already removed, that part is neither necessary there anymore (or it will break after merging this PR).

Can you please split it?

test/algorithms/is_valid.cpp
```diff
@@ -1433,14 +1433,8 @@ BOOST_AUTO_TEST_CASE( test_with_NaN_coordinates )
...
 
-    linestring_type ls1, ls2;
-    bg::read_wkt("LINESTRING(1 1,1.115235e+308 1.738137e+308)", ls1);
-    bg::read_wkt("LINESTRING(-1 1,1.115235e+308 1.738137e+308)", ls2);
-
-    // the intersection of the two linestrings is a new linestring
-    // (multilinestring with a single element) that has NaN coordinates
     multi_linestring_type mls;
-    bg::intersection(ls1, ls2, mls);
+    bg::read_wkt("MULTILINESTRING((nan nan))", mls);
```
barendgehrels, line 1437: Can these changes be explained?

test/algorithms/set_operations/intersection/intersection_linear_linear.cpp
```diff
@@ -1368,7 +1368,7 @@ BOOST_AUTO_TEST_CASE( test_intersection_ml_ml_degenerate )
                       (9.98265 0.00543606),(9.09826 -100.515944),\
                       (7.08745 -329.0674155),(5.06428 -559.024344),\
                       (3.23365 -767.0972558),(3.16036 -775.427199),\
-                      (-0.7654 8.88178e-16,-0.7654 0,5 3))"),
```
barendgehrels, line 1371: So we don't test these degenerate cases anymore?
I'm not against it, but I wonder why - it breaks the new code? Why?

include/boost/geometry/policies/relate/intersection_points.hpp
```diff
@@ -61,6 +62,21 @@ struct segments_intersection_points
...
+    template<typename SegmentIntersectionInfo, typename Point>
+    static inline return_type
+    segments_share_common_point(side_info const&, SegmentIntersectionInfo const& sinfo,
+                                Point const& p)
+    {
+        return_type result;
+        result.count = 1;
+        boost::geometry::assign(result.intersections[0], p);
+
+        // Temporary - this should go later
```
barendgehrels, line 74: When will this go? In this PR?

include/boost/geometry/strategies/cartesian/intersection.hpp
vissarion: ideally in the PR that removes rescaling (I added a note)

test/algorithms/Jamfile
vissarion: I need this is this PR because there tests are failing with the new fix. I will place them in a different commit.

test/algorithms/is_simple.cpp
```diff
@@ -263,14 +263,8 @@ BOOST_AUTO_TEST_CASE( test_geometry_with_NaN_coordinates )
...
 
-    linestring_type ls1, ls2;
-    bg::read_wkt("LINESTRING(1 1,1.115235e+308 1.738137e+308)", ls1);
-    bg::read_wkt("LINESTRING(-1 1,1.115235e+308 1.738137e+308)", ls2);
-
-    // the intersection of the two linestrings is a new linestring
-    // (multilinestring with a single element) that has NaN coordinates
     multi_linestring_type mls;
-    bg::intersection(ls1, ls2, mls);
+    bg::read_wkt("MULTILINESTRING((nan nan))", mls);
```
vissarion, line 267: This is the same as the `is_valid` case.

test/algorithms/set_operations/intersection/intersection_linear_linear.cpp
vissarion: This is the result of the intersection (the inputs are a few lines above and they are not changed). So the results here are a bit different with the new code. Actually, they look better since they create multilinestring without the almost duplicate points (-0.7654 8.88178e-16) and (-0.7654 0) that is created since (-0.7654 0) is contained in both input multilinstrings.

include/boost/geometry/policies/relate/intersection_points.hpp
vissarion: This is copy paste from the segments_crosses() function above. I do not understand it either, but without that line the results are not correct, thus the information assigned to fractions is useful and cannot go now.

vissarion: Do you have any idea of the reasoning behind the original comment?

test/algorithms/is_valid.cpp
barendgehrels: I think the last.
I didn't add these cases here, I think @awulkiew  did.
Probably for the reason that the intersection indeed returned something like that, and then the `is_valid` functionality should check these results.

test/algorithms/set_operations/intersection/intersection_linear_linear.cpp
barendgehrels: Okay, that looks fine to me. Thanks!

include/boost/geometry/policies/relate/intersection_points.hpp
barendgehrels: Actually, the whole fractions could be revised / simplified if rescaling goes.
It might be that the comment is related to that.

barendgehrels (approved):
Thanks for handling my comments.

I still don't understand all the details of the change itself (though it's not that much), but I think it's fine, and if all tests pass (we have so many, fortunately) - it should be completely fine.

vissarion:
I should switch back to writing a short intro to my PRs :)

Let me try to briefly explain the issue and the solution now. 

The intersection of two segments is computed here 
https://github.com/boostorg/geometry/pull/1237/files#diff-bad1f35b9d58f7f24c666340327c6ac0fa1ac0ffb5b2ae59fa905e15699c0550R547

For the case of two segments with a common edge point (vertex) we apriori know the intersection and we do not have to compute it. Thus we baypass the segment intersection here:
https://github.com/boostorg/geometry/pull/1237/files#diff-bad1f35b9d58f7f24c666340327c6ac0fa1ac0ffb5b2ae59fa905e15699c0550R537

This is faster but more importantly more accurate we explicitly return the common point instead of computing the intersection which is sensitive to numerical inaccuracies.

test/algorithms/is_valid.cpp
awulkiew: It was added by @mkaravel:
https://github.com/boostorg/geometry/commit/73960f31ca3e6483a5846bbf0b6ad601a4640fbe
It seems such geometry was causing infinite loops. I think we should leave this test.

PR #1242, opened 2024-02-10
pass strategy to sort by side

barendgehrels:
Small PR, passing the strategy i/o the side strategy in several places, which was a TODO already.
Also documents sort_by_side better.
It is a preparation step for a next PR.

This should not change anything (apart from users, by chance using `sort_by_side` - but it's all in namespace `detail`, so not considered as a breaking change)


barendgehrels:
@vissarion is this OK?

vissarion (approved):
LGTM, thanks!

awulkiew:
A little late but yes, it looks good. Thanks.

PR #1243, opened 2024-02-13
Add missing headers and remove unused typedefs

vissarion:
fixes failing CI header tests (and removes some unused typedefs appeared as warnings)

### DIFF
--- a/include/boost/geometry/algorithms/detail/within/within_no_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/within/within_no_turns.hpp
@@ -18,6 +18,8 @@
 #ifndef BOOST_GEOMETRY_ALGORITHMS_DETAIL_WITHIN_WITHIN_NO_TURNS_HPP
 #define BOOST_GEOMETRY_ALGORITHMS_DETAIL_WITHIN_WITHIN_NO_TURNS_HPP
 
+#include <boost/type_traits/is_base_of.hpp>
+
 #include <boost/geometry/algorithms/detail/point_on_border.hpp>
 #include <boost/geometry/algorithms/detail/within/point_in_geometry.hpp>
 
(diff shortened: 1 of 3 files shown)


awulkiew (approved):

include/boost/geometry/algorithms/detail/within/within_no_turns.hpp
```diff
@@ -18,6 +18,8 @@
 #ifndef BOOST_GEOMETRY_ALGORITHMS_DETAIL_WITHIN_WITHIN_NO_TURNS_HPP
 #define BOOST_GEOMETRY_ALGORITHMS_DETAIL_WITHIN_WITHIN_NO_TURNS_HPP
 
+#include <boost/type_traits/is_base_of.hpp>
```
barendgehrels, line 21: Sorry, I had not seen this PR before.
And I tackled the same problem because I got a mail that my merge was red.

I think this  line is not the right fix - we use it from `std` now (in several other places)

include/boost/geometry/util/rational.hpp
```diff
@@ -16,6 +16,7 @@
 
 #include <boost/rational.hpp>
 #include <boost/numeric/conversion/bounds.hpp>
+#include <boost/numeric/conversion/converter.hpp>
```
barendgehrels, line 19: So indeed this fixed it.
I fixed it in another way (again, because I didn't notice this earlier, sorry) and I think it is a better fix. See #1246

barendgehrels:
I see this PR is already 2 weeks old. My bad for not reviewing of course.
But I think a CI fix should get some priority - feel free to poke such a thing, or even merge unreviewed

vissarion:
Closing as duplicate of https://github.com/boostorg/geometry/pull/1246

PR #1246, opened 2024-02-24
fix compilation of converter and is_base_of

barendgehrels:
This fixes two new errors:
* `is_base_of` was called from `Boost`, while in the rest of the library it was called from `std`. Fixed ✅ 
* `converter` reported an error, apparently it is changed in the underlying Boost library. Fixed ✅ 

The first one is trivial. For the second one (related to Boost.Rational) there were three options
* fix `converter` call - with the risk that it will happen again
* add `numeric_cast_traits` - I got it working, but it is quite a lot of code, and it will give problems if a library user does it also for Boost.Rational 
* add another level of indirection, as we often do, so add a `geometry::detail::numeric_cast`

I selected the third option. If affected quite some files, but the changes were mechanic.

Add error 1: this was the message
```
In file included from /home/runner/work/geometry/geometry/boost-root/libs/geometry/test/headers/main.cpp:14:
./boost/geometry/algorithms/detail/within/within_no_turns.hpp:145:34: error: ‘is_base_of’ is not a member of ‘boost’; did you mean ‘std::is_base_of’?
  145 |           bool IsMulti1 = boost::is_base_of<geometry::multi_tag, Tag1>::value,
      |                                  ^~~~~~~~~~

```

Add error 2: this was the message
```
gcc.compile.c++ bin.v2/libs/geometry/test/headers/h-util-rational.test/gcc-11/debug/threading-multi/visibility-hidden/h-util-rational.o
In file included from /home/runner/work/geometry/geometry/boost-root/libs/geometry/test/headers/main.cpp:14:
./boost/geometry/util/rational.hpp:150:8: error: ‘converter’ is not a class template
  150 | struct converter<int, rational<T>, Traits, OverflowHandler, Float2IntRounder, RawConverter, UserRangeChecker>
      |        ^~~~~~~~~
./boost/geometry/util/rational.hpp:167:112: error: wrong number of template arguments (7, should be 6)
  167 | struct converter<double, rational<T>, Traits, OverflowHandler, Float2IntRounder, RawConverter, UserRangeChecker>
      |                                                                                                                ^
./boost/geometry/util/rational.hpp:150:109: note: provided for ‘template<class T, class Traits, class OverflowHandler, class Float2IntRounder, class RawConverter, class UserRangeChecker> struct boost::numeric::converter’
  150 | struct converter<int, rational<T>, Traits, OverflowHandler, Float2IntRounder, RawConverter, UserRangeChecker>
      |                                                                                                             ^
```

### DIFF
--- a/include/boost/geometry/algorithms/convert.hpp
+++ b/include/boost/geometry/algorithms/convert.hpp
@@ -24,7 +24,6 @@
 #include <cstddef>
 #include <type_traits>
 
-#include <boost/numeric/conversion/cast.hpp>
 #include <boost/range/begin.hpp>
 #include <boost/range/end.hpp>
 #include <boost/range/size.hpp>
@@ -45,6 +44,7 @@
 
 #include <boost/geometry/geometries/concepts/check.hpp>
 
+#include <boost/geometry/util/numeric_cast.hpp>
 #include <boost/geometry/util/range.hpp>
 
 #include <boost/geometry/views/detail/closed_clockwise_view.hpp>
@@ -80,7 +80,7 @@ struct point_to_box
         typedef typename coordinate_type<Box>::type coordinate_type;
 
         set<Index, Dimension>(box,
-                boost::numeric_cast<coordinate_type>(get<Dimension>(point)));
+                util::numeric_cast<coordinate_type>(get<Dimension>(point)));
         point_to_box
             <
                 Point, Box,
--- a/include/boost/geometry/algorithms/detail/within/within_no_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/within/within_no_turns.hpp
@@ -19,10 +19,9 @@
 #define BOOST_GEOMETRY_ALGORITHMS_DETAIL_WITHIN_WITHIN_NO_TURNS_HPP
 
 
-#include <type_traits>
-
 #include <boost/geometry/algorithms/detail/point_on_border.hpp>
 #include <boost/geometry/algorithms/detail/within/point_in_geometry.hpp>
+#include <boost/geometry/util/type_traits.hpp>
 
 namespace boost { namespace geometry {
 
@@ -145,8 +144,8 @@ template <typename Geometry1,
           typename Geometry2,
           typename Tag1 = typename geometry::tag<Geometry1>::type,
           typename Tag2 = typename geometry::tag<Geometry2>::type,
-          bool IsMulti1 = std::is_base_of<geometry::multi_tag, Tag1>::value,
-          bool IsMulti2 = std::is_base_of<geometry::multi_tag, Tag2>::value>
+          bool IsMulti1 = util::is_multi<Geometry1>::value,
+          bool IsMulti2 = util::is_multi<Geometry2>::value>
 struct within_no_turns_multi
 {
     template <typename Strategy> static inline
--- a/include/boost/geometry/algorithms/detail/within/within_no_turns.hpp
+++ b/include/boost/geometry/algorithms/detail/within/within_no_turns.hpp
@@ -18,6 +18,9 @@
 #ifndef BOOST_GEOMETRY_ALGORITHMS_DETAIL_WITHIN_WITHIN_NO_TURNS_HPP
 #define BOOST_GEOMETRY_ALGORITHMS_DETAIL_WITHIN_WITHIN_NO_TURNS_HPP
 
+
+#include <type_traits>
+
 #include <boost/geometry/algorithms/detail/point_on_border.hpp>
 #include <boost/geometry/algorithms/detail/within/point_in_geometry.hpp>
 
@@ -142,8 +145,8 @@ template <typename Geometry1,
           typename Geometry2,
           typename Tag1 = typename geometry::tag<Geometry1>::type,
           typename Tag2 = typename geometry::tag<Geometry2>::type,
-          bool IsMulti1 = boost::is_base_of<geometry::multi_tag, Tag1>::value,
-          bool IsMulti2 = boost::is_base_of<geometry::multi_tag, Tag2>::value>
+          bool IsMulti1 = std::is_base_of<geometry::multi_tag, Tag1>::value,
+          bool IsMulti2 = std::is_base_of<geometry::multi_tag, Tag2>::value>
 struct within_no_turns_multi
 {
     template <typename Strategy> static inline
--- a/include/boost/geometry/util/numeric_cast.hpp
+++ b/include/boost/geometry/util/numeric_cast.hpp
@@ -12,7 +12,8 @@
 #include <boost/numeric/conversion/cast.hpp>
 #include <boost/rational.hpp>
 
-namespace boost { namespace geometry
+namespace boost { namespace geometry { namespace util
+
 {
 
 #ifndef DOXYGEN_NO_DETAIL
@@ -42,9 +43,6 @@ struct numeric_caster<Target, rational<T>>
 } // namespace detail
 #endif
 
-namespace util
-{
-
 // Calls either boost::numeric_cast, or functionality specific for Boost.Geometry
 // such as rational_cast for Boost.Rational
 template <typename Target, typename Source>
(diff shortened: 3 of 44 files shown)


awulkiew:
I noticed that numeric converter is failing and wanted to propose adding missing header
```
#include <boost/numeric/conversion/converter.hpp>
```
because it seems to fix the compilation. Is this one of the solutions or is it the fourth one?

Anyway, I think the solution you've choosen is better.

Btw, in util/rational.hpp there is also a specialization of `boost::numeric::bounds<>` for `rational<T>`. Is it still needed?

include/boost/geometry/util/numeric_cast.hpp
```diff
@@ -0,0 +1,54 @@
...
+struct numeric_caster<Target, rational<T>>
+{
+   static inline Target apply(rational<T> const& source)
+    {
+        return boost::rational_cast<Target>(source);
+    }
+};
+
+template <typename Target, typename Source>
+inline Target numeric_cast(Source const& source)
```
awulkiew, line 43: Why not put it in `bg::util` namespace with the rest of utilities?

include/boost/geometry/algorithms/convert.hpp
```diff
@@ -80,7 +80,7 @@ struct point_to_box
         typedef typename coordinate_type<Box>::type coordinate_type;
 
         set<Index, Dimension>(box,
-                boost::numeric_cast<coordinate_type>(get<Dimension>(point)));
+                geometry::detail::numeric_cast<coordinate_type>(get<Dimension>(point)));
```
awulkiew, line 83: This could be `detail::numeric_cast` or if this was implemented in `util` namspace `util::numeric_cast`.

barendgehrels: `detail` didn't always work, because we use it a lot on lower levels too.

include/boost/geometry/util/numeric_cast.hpp
barendgehrels: I noticed that `coordinate_cast` was also in `detail` so I put it there.
Also it is not really meant for end users (but `util` neither).
I didn't think of `util` - yes, maybe it is better there. I'll change it tomorrow.

awulkiew:
It seems PR https://github.com/boostorg/geometry/pull/1243 created by @vissarion also fixes the conversion issue.

barendgehrels:
> I noticed that numeric converter is failing and wanted to propose adding missing header

To be honest, I didn't try this. The error message indicated something else. But of course, if you tried it, and it worked, it might have solved it.

> 
> Anyway, I think the solution you've choosen is better.

OK let's keep it then.
 
> Btw, in util/rational.hpp there is also a specialization of `boost::numeric::bounds<>` for `rational<T>`. Is it still needed?

Yes, I saw it too. I didn't want to mix in other things. But it might be that it is not needed anymore.

include/boost/geometry/util/numeric_cast.hpp
barendgehrels: ✅ changed it to `util::`

include/boost/geometry/algorithms/convert.hpp
barendgehrels: for `util` the `geometry::` prefix was not necessary, it's convenient

barendgehrels:
> > Btw, in util/rational.hpp there is also a specialization of `boost::numeric::bounds<>` for `rational<T>`. Is it still needed?
> 
> Yes, I saw it too. I didn't want to mix in other things. But it might be that it is not needed anymore.

I think it is still needed, because of this code in `assign_values.hpp`:
```
        coordinate_type const highest = geometry::bounds<coordinate_type>::highest();
        coordinate_type const lowest = geometry::bounds<coordinate_type>::lowest();
```
Now this is actually not yet optimal, because that is defined like this, in `is_inverse_spheroidal_coordinates.hpp`:
```
namespace boost { namespace geometry
{

template<class CT>
struct bounds
{
  static CT lowest  () { return boost::numeric::bounds<CT>::lowest(); }
  static CT highest () { return boost::numeric::bounds<CT>::highest(); }
};
```

So it looks like a generic structure. But most of the code uses it from `boost::numeric::` directly.
So I think that should go to a specific  headerfile in `util` as well. And then we might make it for Boost.Rational specifically, like I did for `numeric_cast`.

Anyway, for Boost.Rational it is still needed now, and it is better to handle it in a separate PR.

include/boost/geometry/algorithms/detail/within/within_no_turns.hpp
```diff
@@ -142,8 +145,8 @@ template <typename Geometry1,
           typename Geometry2,
           typename Tag1 = typename geometry::tag<Geometry1>::type,
           typename Tag2 = typename geometry::tag<Geometry2>::type,
-          bool IsMulti1 = boost::is_base_of<geometry::multi_tag, Tag1>::value,
-          bool IsMulti2 = boost::is_base_of<geometry::multi_tag, Tag2>::value>
+          bool IsMulti1 = std::is_base_of<geometry::multi_tag, Tag1>::value,
+          bool IsMulti2 = std::is_base_of<geometry::multi_tag, Tag2>::value>
```
awulkiew, line 149: This could also be replaced with `util::is_multi<Geometry2>::value` from `<boost/geometry/util/type_traits.hpp>`

awulkiew:
> it is better to handle it in a separate PR

I agree. Thanks for the investigation and explanation.

awulkiew (approved):

include/boost/geometry/util/numeric_cast.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace util
+{
+
+// Calls either boost::numeric_cast, or functionality specific for Boost.Geometry
+// such as rational_cast for Boost.Rational
+template <typename Target, typename Source>
+inline Target numeric_cast(Source const& source)
+{
+    return detail::numeric_caster<Target, Source>::apply(source);
```
awulkiew, line 53: It seems that there is namespace `util::detail` so:
```suggestion
    return geometry::detail::numeric_caster<Target, Source>::apply(source);
```
Or alternatively `numeric_caster` could be moved from `geometry::detail`.

include/boost/geometry/algorithms/detail/within/within_no_turns.hpp
barendgehrels: That's right. So this was not the right fix either ;-)
✅ fixed, and on 3 other places as well.

include/boost/geometry/util/numeric_cast.hpp
barendgehrels: ✅ fixed (will update PR)

barendgehrels:
The file `within_no_turns.hpp` is now updated.
But the whole file is never used anywhere, as far as I can see. And all is in `detail` so it is not public.
So we can deprecate it (I remember this was already discussed earlier a while ago).

Also there are no unit tests.

Anyway, it's now fixed, but can be removed later.

vissarion (approved):
Thanks. I agree with the changes. I will close https://github.com/boostorg/geometry/pull/1243 as duplicate.

vissarion:
> Question then: how did CI detect it didn't work?

We have a "headers" script in CI that tests all the header files of the library if they compile independently (related issue https://github.com/boostorg/geometry/issues/523)

See
https://github.com/boostorg/geometry/actions/runs/8046744097/job/21974572517 
https://github.com/boostorg/geometry/blob/develop/.github/workflows/headers.yml

barendgehrels:
> > Question then: how did CI detect it didn't work?
> 
> We have a "headers" script in CI that tests all the header files of the library if they compile independently (related issue #523)
> 
> See https://github.com/boostorg/geometry/actions/runs/8046744097/job/21974572517 https://github.com/boostorg/geometry/blob/develop/.github/workflows/headers.yml

👍 thanks, good to know! And thanks for approval and merging!

PR #1247, opened 2024-02-24
Use if constexpr macro instead of condition macro.

awulkiew:
This PR replaces uses of `BOOST_GEOMETRY_CONDITION` with `BOOST_GEOMETRY_CONSTEXPR` if possible and it is not destructive to readability.

Unfortunately it as less straightforward than I thought it would be because with MSVC and level 4 warnings enabled "unreachable code" warning is generated of some code exists outside `if constexpr` block. In such case an additional `else` block has to be added to prevent the warning. This means there are more blocks and indentations.

Constexpr conditions has to be checked separately from regular conditions. This means that some of the code has to be repeated e.g. if it has to be executed for both compile-time or run-time condition. This also can add more blocks and indentations.

For the reasons above I left `BOOST_GEOMETRY_CONDITION` in formulas because code like this:
```
if (Foo == 0)
{
   return bar;
}
// update bar
if (Foo == 1)
{
   return bar;
}
// update bar
etc.
```
would force me to add `else` brackets. Unless we agreed to write `if constexpr` differently than "regular" `if`s, with different brackets placement, indentations, etc. I can't think of any way that would look good.

So if you think that some of the changes look bad feel free to point it out so I'll revert it to `BOOST_GEOMETRY_CONDITION`.

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/append_no_dups_or_spikes.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/append_no_dups_or_spikes.hpp
@@ -1,10 +1,10 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
 // Copyright (c) 2007-2012 Barend Gehrels, Amsterdam, the Netherlands.
+// Copyright (c) 2023 Adam Wulkiewicz, Lodz, Poland.
 
 // This file was modified by Oracle on 2014-2020.
 // Modifications copyright (c) 2014-2020 Oracle and/or its affiliates.
-
 // Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
 
 // Use, modification and distribution is subject to the Boost Software License,
@@ -29,7 +29,7 @@
 
 #include <boost/geometry/core/closure.hpp>
 
-#include <boost/geometry/util/condition.hpp>
+#include <boost/geometry/util/constexpr.hpp>
 #include <boost/geometry/util/range.hpp>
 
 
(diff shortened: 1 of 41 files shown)


include/boost/geometry/algorithms/detail/overlay/append_no_dups_or_spikes.hpp
```diff
@@ -1,10 +1,10 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
 // Copyright (c) 2007-2012 Barend Gehrels, Amsterdam, the Netherlands.
```
barendgehrels, line 3: -> `2024`

include/boost/geometry/util/precise_math.hpp
```diff
@@ -457,7 +457,10 @@ RealNumber incircle(std::array<RealNumber, 2> const& p1,
     RealNumber det = A_13 * (A_21_x_A_32[0] - A_31_x_A_22[0])
       + A_23 * (A_31_x_A_12[0] - A_11_x_A_32[0])
       + A_33 * (A_11_x_A_22[0] - A_21_x_A_12[0]);
-    if(Robustness == 0) return det;
+    if (BOOST_GEOMETRY_CONDITION(Robustness == 0))
```
barendgehrels, line 460: Can this (and similar) made `CONSTEXPR` as well? Apart from needing the `else` branch, but you did that elsewhere too

barendgehrels (approved):
I'm OK with this change.

I made a remark about the copyright (sometimes 2023, sometimes 2024) but it's not important, and maybe you started it already last year.

barendgehrels:
Thanks for the change

include/boost/geometry/algorithms/detail/overlay/get_turn_info_ll.hpp
```diff
@@ -553,65 +559,75 @@ struct get_turn_info_linear_linear
...
+            }
         }
 
-        if ( is_q_spike
-          && ( BOOST_GEOMETRY_CONDITION(is_version_touches)
-            || inters.d_info().arrival[1] == 1 ) )
+        if (is_q_spike)
         {
-            if ( BOOST_GEOMETRY_CONDITION(is_version_touches) )
+            bool output_spike = false;
```
vissarion, line 600: this is also change the behavior e.g. the block 
```
            tp.operations[0].operation = operation_intersection;
            tp.operations[1].operation = operation_blocked;
            *out++ = tp;
            //tp.operations[0].operation = operation_intersection;
            tp.operations[1].operation = operation_intersection;
            *out++ = tp;
```
was executed in all cases but now only in touch or append_colinnear_opposite cases. 

Is this intentional?

awulkiew: If there is any change in behavior it was not intentional. But AFAIU there is no change here. We're talking about the block below at line 607(old) or 622(new) correct? In the old version it was executed:
```
if ( is_q_spike
  && ( BOOST_GEOMETRY_CONDITION(is_version_touches)
    || inters.d_info().arrival[1] == 1 ) )
```
now it's executed if `output_spike` is `true` and this is set in block:
```
if (is_q_spike)
```
and only
```
if BOOST_GEOMETRY_CONSTEXPR (is_version_touches)
```
or
```
else if (inters.d_info().arrival[1] == 1) // Version == append_collinear_opposite
```

For convenience here is [the old version](https://github.com/boostorg/geometry/blob/eb38231d367fe0b74fe8c2b4083d94870a86ac9d/include/boost/geometry/algorithms/detail/overlay/get_turn_info_ll.hpp) and [the new version](https://github.com/boostorg/geometry/blob/f25545b58e54857f56cfe90769d9bdafe4bca731/include/boost/geometry/algorithms/detail/overlay/get_turn_info_ll.hpp).

Or am I missing something?

include/boost/geometry/algorithms/detail/relate/point_point.hpp
```diff
@@ -69,7 +69,7 @@ std::pair<bool, bool> point_multipoint_check(Point const& point,
     // point_in_geometry could be used here but why iterate over MultiPoint twice?
     // we must search for a point in the exterior because all points in MultiPoint can be equal
 
-
+
```
vissarion, line 72: this is probably here by mistake.

vissarion (approved):
LGTM, thanks!

include/boost/geometry/algorithms/detail/relate/point_point.hpp
awulkiew: yes, fixed

PR #1248, opened 2024-02-25
fix: set turns in closed clusters as non traversable

barendgehrels:


include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -898,6 +877,61 @@ struct buffered_piece_collection
...
+            offsetted_rings, offsetted_rings, m_strategy);
+
+        for (const auto& cluster : m_clusters)
+        {
+            if (cluster.second.open_count == 0 && cluster.second.spike_count == 0)
+            {
+                // If the cluster is completely closed, mark it as not traversable.
+                for (auto index : cluster.second.turn_indices)
+                {
+                    m_turns[index].is_turn_traversable = false;
```
barendgehrels, line 904: This is the main fix. Especially for linestrings, with flat end, there were many turns, completely closed, but still marked as traversable. This could cause the buffer fail (no or partial output)

include/boost/geometry/algorithms/detail/overlay/cluster_info.hpp
```diff
@@ -27,11 +27,11 @@ struct cluster_info
...
     //! Number of open spaces (e.g. 2 for touch)
-    std::size_t open_count;
+    std::size_t open_count{0};
 
-    inline cluster_info()
-        : open_count(0)
-    {}
+    //! Number of spikes, where a segment goes to the cluster point
+    //! and leaves immediately in the opposite direction.
+    std::size_t spike_count{0};
```
barendgehrels, line 34: This (and related code) is also necessary, to have a way out from a cluster which is closed, but where still is a way in and out. This is also related to floating point issues - the part between the spike legs might be indetectable

include/boost/geometry/algorithms/detail/overlay/discard_duplicate_turns.hpp
```diff
@@ -137,6 +137,13 @@ inline void discard_duplicate_start_turns(Turns& turns,
                 {
                     for (std::size_t const& i : it->second)
                     {
+                        if (turns[i].cluster_id != turn.cluster_id)
+                        {
+                            // The turns are not part of the same cluster,
+                            // or one is clustered and the other is not.
+                            // This is not corresponding.
+                            continue;
```
barendgehrels, line 145: This was earlier not possible. But now that clusters are detected before, the start turn and its corresponding turn should be located on the same cluster.

barendgehrels: We might be able to delete the `corresponding_turn` functions because of this.
Will check that, but if possible, it will be a different PR

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
```diff
@@ -589,19 +589,22 @@ public :
...
+                = cinfo.turn_indices.count(op.enriched.get_next_turn_index()) > 0;
 
-        return to_start && to_start_index ? 4
-            : to_start ? 3
-            : next_in_same_cluster ? 1
-            : 2
+        return to_start && to_start_index ? 5
+            : to_start ? 4
+            : next_in_same_cluster ? 2
+            : 3
```
barendgehrels, line 607: The values just define the ordering. There is now 1 extra, lowest ordering: spike out

test/algorithms/buffer/buffer_variable_width.cpp
```diff
@@ -0,0 +1,184 @@
...
+             {6.0, 4.5, 1.5}, 429.831);
+    test_buffer("case4", "MULTILINESTRING((-18.00000000002045652536253328435122966766357421875 10.4999999997948929575386500800959765911102294921875,-2.99999999998012700785920969792641699314117431640625 10.4999999997948929575386500800959765911102294921875),(-2.99999999998012700785920969792641699314117431640625 10.4999999997948929575386500800959765911102294921875,18.00000000002045652536253328435122966766357421875 10.4999999997948947338954894803464412689208984375),(-2.99999999998012700785920969792641699314117431640625 10.4999999997948929575386500800959765911102294921875,17.99999999995061017443731543608009815216064453125 -4.5000000001943778471513724070973694324493408203125))",
+             {6.0, 4.5, 1.5}, 423.195);
```
barendgehrels, line 182: Cases 3 and 4 were failing before the fixes

include/boost/geometry/algorithms/detail/overlay/enrich_intersection_points.hpp
```diff
@@ -396,16 +397,36 @@ inline void enrich_intersection_points(Turns& turns,
...
-    discard_duplicate_start_turns(turns, geometry1, geometry2);
+    // Turns are often used by index (in clusters, next_index, etc)
+    // and turns may therefore NOT be DELETED - they may only be flagged as discarded
 
     bool has_cc = false;
-    bool const has_colocations
-        = detail::overlay::handle_colocations
+    bool has_colocations = false;
+
+    if (BOOST_GEOMETRY_CONDITION (! is_buffer))
```
awulkiew, line 406: `BOOST_GEOMETRY_CONSTEXPR` could be used instead.

include/boost/geometry/algorithms/detail/overlay/handle_colocations.hpp
```diff
@@ -357,7 +356,7 @@ inline bool handle_colocations(Turns& turns, Clusters& clusters,
     {
         std::cout << kv.first << std::endl;
         for (auto const& toi : kv.second)
-        {
+    {
```
awulkiew, line 359: probably wrong indent

include/boost/geometry/algorithms/detail/overlay/enrich_intersection_points.hpp
```diff
@@ -396,16 +397,36 @@ inline void enrich_intersection_points(Turns& turns,
...
+        // Gather cluster properties (using even clusters with
+        // discarded turns - for open turns)
+        detail::overlay::gather_cluster_properties
             <
-                Reverse1, Reverse2, OverlayType, Geometry1, Geometry2
-            >(turns, clusters, robust_policy);
+                Reverse1,
+                Reverse2,
+                OverlayType
+            >(clusters, turns, target_operation,
```
awulkiew, line 420: `target_operation` is known at compile-time. It could be passed as template parameter and then checked with `if BOOST_GEOMETRY_CONSTEXPR` inside `gather_cluster_properties`. But a different PR would probably be better for this.

include/boost/geometry/algorithms/detail/overlay/handle_colocations.hpp
```diff
@@ -475,7 +489,8 @@ inline void gather_cluster_properties(Clusters& clusters, Turns& turns,
             turn_operation_type& op = turn.operations[ranked.operation_index];
 
             if (set_startable
-                    && for_operation == operation_union && cinfo.open_count == 0)
+                && for_operation == operation_union
```
awulkiew, line 492: Here is the condition for `if constexpr`. Or is it not always the case that `for_operation` is known at compile-time?

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
```diff
@@ -791,6 +794,24 @@ public :
             return false;
         }
 
+        if (is_union && cinfo.open_count == 0 && cinfo.spike_count > 0)
```
awulkiew, line 797: `is_union` is `static const` so I guess it can cause warnings.

awulkiew (approved):

include/boost/geometry/algorithms/detail/overlay/handle_colocations.hpp
barendgehrels: Yes, it is known. I'll update.

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
barendgehrels: indeed, I'll update.
Thanks for your quick review.

include/boost/geometry/algorithms/detail/overlay/enrich_intersection_points.hpp
barendgehrels: As soon as your PR is merged ;-)

awulkiew: Ok, but in case you wasn't aware, this macro is already in develop:
https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/util/constexpr.hpp

barendgehrels: 👍 right! Will update on Wednesday

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
vissarion: Is there a comment that explain this orderings?

test/algorithms/buffer/buffer_variable_width.cpp
```diff
@@ -0,0 +1,184 @@
...
+    specific_buffer_distance_strategy distance_strategy;
+    boost::geometry::strategy::buffer::join_miter join_strategy;
+    boost::geometry::strategy::buffer::end_flat end_strategy;
+    boost::geometry::strategy::buffer::point_circle point_strategy;
+    boost::geometry::model::multi_polygon<polygon_t> result;
+    boost::geometry::detail::buffer::buffer_inserter
+        <
+            polygon_t
+        >(mls,
+            std::back_inserter(result),
```
vissarion, line 158: indentation issue here.

test/algorithms/buffer/buffer_variable_width.cpp
```diff
@@ -0,0 +1,184 @@
...
+        return true;
+    }
+
+    template <typename Point, typename OutputRange, typename DistanceStrategy>
+    bg::strategy::buffer::result_code apply(Point const& input_p1, Point const& input_p2,
+                                            bg::strategy::buffer::buffer_side_selector side,
+                                            DistanceStrategy const& distance,
+                                            OutputRange &output_range) const
+    {
+        const auto result =
```
vissarion, line 49: Maybe `auto const` which seems that is more common as a style in BG? 
Also why not `auto const&`?

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -898,6 +877,61 @@ struct buffered_piece_collection
...
+            return;
+        }
+
+        detail::overlay::gather_cluster_properties
+            <
+                false, false, overlay_buffer
+            >(m_clusters, m_turns, detail::overlay::operation_union,
+            offsetted_rings, offsetted_rings, m_strategy);
+
+        for (const auto& cluster : m_clusters)
```
vissarion, line 897: `auto const&`?

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -898,6 +877,61 @@ struct buffered_piece_collection
...
+        {
+            if (cluster.second.open_count == 0 && cluster.second.spike_count == 0)
+            {
+                // If the cluster is completely closed, mark it as not traversable.
+                for (auto index : cluster.second.turn_indices)
```
vissarion, line 902: what about `auto const&` here?

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -898,6 +877,61 @@ struct buffered_piece_collection
...
+            }
+        }
+    }
+
+    inline void make_traversable_consistent_per_cluster()
+    {
+        for (auto const& cluster : m_clusters)
+        {
+            bool is_traversable = false;
+            for (auto index : cluster.second.turn_indices)
```
vissarion, line 915: similarly, what about using `auto const&` here?

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
```diff
@@ -898,6 +877,61 @@ struct buffered_piece_collection
...
+                {
+                    // If there is one turn traversable in the cluster,
+                    // then all turns should be traversable.
+                    is_traversable = true;
+                    break;
+                }
+            }
+            if (is_traversable)
+            {
+                for (auto index : cluster.second.turn_indices)
```
vissarion, line 927: same here

vissarion (approved):
Thanks, LGTM in general.

test/algorithms/buffer/buffer_variable_width.cpp
barendgehrels: Yes it is. But on my office it's not, so I sometimes don't switch at the right moment...
Thanks.
BTW, in this case it should not be `&` because it's an assignment from a function result. (Though, technically, it's in line with C++ specs)

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
barendgehrels: Indeed. Will fix later today. Same for your similar comments, thanks.

test/algorithms/buffer/buffer_variable_width.cpp
barendgehrels: :whi

test/algorithms/buffer/buffer_variable_width.cpp
barendgehrels: :white

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
barendgehrels: ✅

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
barendgehrels: ✅

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
barendgehrels: ✅

include/boost/geometry/algorithms/detail/buffer/buffered_piece_collection.hpp
barendgehrels: ✅

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
barendgehrels: ✅ the comment was a bit above, but I didn't update it. Thanks. 
Comment is moved to function description, and a short comment here.

include/boost/geometry/algorithms/detail/overlay/handle_colocations.hpp
barendgehrels: ✅

include/boost/geometry/algorithms/detail/overlay/enrich_intersection_points.hpp
barendgehrels: Will do it in another PR indeed. Not today.

include/boost/geometry/algorithms/detail/overlay/handle_colocations.hpp
barendgehrels: I didn't update this one, and the ones below.
Because `for_operation` is passed as a normal parameter.
This needs more refactoring and it is better to separate that.

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
barendgehrels: Similarly here: not updated. Here it was not passed, but it is used in many places like this in this source file. And they should all be updated. 
I guess, because of the combination of conditions, there are no warnings (not sure though). In other places it was combined as well, in this source.
So also this should go to another PR.

PR #1251, opened 2024-02-28
Fixes compile error in tests and add removed tests for is_simple and is_valid

vissarion:
Following comment https://github.com/boostorg/geometry/pull/1237#discussion_r1501810625 this PR adds removed tests for is_simple and is_valid with geometries whose coordinates are close to the maximum double. 
The tests are passing with the fix of https://github.com/boostorg/geometry/pull/1237 but the results are different i.e. the intersection does not create NaN coordinates, it is more accurate but still can create invalid geometries.

This PR also fixes a compile error in tests (wrong namespace for numeric_cast).

### DIFF
--- a/test/util/rational.cpp
+++ b/test/util/rational.cpp
@@ -30,11 +30,11 @@ void test_coordinate_cast(std::string const& s, int expected_nom, int expected_d
 void test_numeric_cast()
 {
     const boost::rational<int> r1(3, 4);
-    BOOST_CHECK_CLOSE(bg::detail::numeric_cast<double>(r1), 0.75, 0.00001);
+    BOOST_CHECK_CLOSE(bg::util::numeric_cast<double>(r1), 0.75, 0.00001);
 
     const boost::rational<int> r2(10, 4);
-    BOOST_CHECK_CLOSE(bg::detail::numeric_cast<double>(r2), 2.5, 0.00001);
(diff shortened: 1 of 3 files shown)


test/util/rational.cpp
```diff
@@ -30,11 +30,11 @@ void test_coordinate_cast(std::string const& s, int expected_nom, int expected_d
...
 {
     const boost::rational<int> r1(3, 4);
-    BOOST_CHECK_CLOSE(bg::detail::numeric_cast<double>(r1), 0.75, 0.00001);
+    BOOST_CHECK_CLOSE(bg::util::numeric_cast<double>(r1), 0.75, 0.00001);
 
     const boost::rational<int> r2(10, 4);
-    BOOST_CHECK_CLOSE(bg::detail::numeric_cast<double>(r2), 2.5, 0.00001);
-    BOOST_CHECK_EQUAL(bg::detail::numeric_cast<int>(r2), 2);
+    BOOST_CHECK_CLOSE(bg::util::numeric_cast<double>(r2), 2.5, 0.00001);
+    BOOST_CHECK_EQUAL(bg::util::numeric_cast<int>(r2), 2);
```
barendgehrels, line 37: Oops. As soon as the Cmake PR is merged ,this won't happen anymore ;-)
Thanks for fixing.

barendgehrels (approved):
looks good to me

Issue #1250, opened 2024-02-27
`boost::geometry::buffer()` of linestring returns inner polygon


Raclamusi:
I tried to compute the buffer of `LINESTRING(3 2, 2 2, 2 4, 4 4, 4 2, 3 1)` with a distance of 0.5, but did not get the expected result and instead got the inner polygon.
This problem does not occur with the reverse order linestring, `LINESTRING(3 1, 4 2, 4 4, 2 4, 2 2, 3 2)`.

https://godbolt.org/z/Gfd85v5bG

**Expected Result**: `MULTIPOLYGON(((2.79289 1.5,2 1.5,1.5 1.5,1.5 2,1.5 4,1.5 4.5,2 4.5,4 4.5,4.5 4.5,4.5 4,4.5 2,4.5 1.79289,4.35355 1.64645,3.35355 0.646447,2.64645 1.35355,2.79289 1.5),(3 1.70711,3.5 2.20711,3.5 3.5,2.5 3.5,2.5 2.5,3 2.5,3 1.70711)))`
**Result with boost-1.84.0**: `MULTIPOLYGON(((3 1.70711,3 2.5,2.5 2.5,2.5 3.5,3.5 3.5,3.5 2.20711,3 1.70711)))`

![Screenshot](https://github.com/boostorg/geometry/assets/61970673/6123ac42-b498-4a86-b0ef-cf096fdead0d)

barendgehrels:
Thanks for the report, I can reproduce it. Not yet fixed by my last changes.

barendgehrels:
It is fixed now. I jumped on it because the case is, geometrically, so simple, it should really not fail.
But apparently it was still a rare case, needing a condition which should always have been there.
Great you found it and reported it.

barendgehrels:
The fix is merged, closing.

PR #1252, opened 2024-02-28
[fix] assign next turn

barendgehrels:
Fixes issue #1250

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/enrich_intersection_points.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/enrich_intersection_points.hpp
@@ -107,69 +107,86 @@ inline void enrich_sort(Operations& operations,
 }
 
 
+// Assign travel-to-vertex/ip index for each turn.
 template <typename Operations, typename Turns>
 inline void enrich_assign(Operations& operations, Turns& turns,
-                          bool check_turns)
+                          bool check_consecutive_turns)
 {
-    if (operations.empty())
-    {
-        return;
-    }
-
-    // Assign travel-to-vertex/ip index for each turning point.
-    // Iterator "next" is circular
-
-    geometry::ever_circling_range_iterator<Operations const> next(operations);
-    ++next;
-
-    for (auto const& indexed : operations)
+    for_each_with_index(operations, [&](std::size_t index, auto const& indexed)
     {
         auto& turn = turns[indexed.turn_index];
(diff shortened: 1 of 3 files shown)


test/algorithms/buffer/buffer_linestring.cpp
```diff
@@ -326,6 +345,12 @@ void test_all()
         using bg::strategy::buffer::end_round;
         test_one<linestring, polygon>("issue_1084", issue_1084, join_round(10), end_round(10), 13200.83, 10.0);
     }
+    {
+        using bg::strategy::buffer::join_miter;
+        using bg::strategy::buffer::end_flat;
+        test_one<linestring, polygon>("issue_1250", issue_1250, join_miter(5), end_flat(), 8.39277, 0.5);
+        test_one<linestring, polygon>("issue_1250_rev", issue_1250_rev, join_miter(5), end_flat(), 8.39277, 0.5);
```
barendgehrels, line 352: This cases is fixed. See the issue for a picture.
It was surprisingly simple case - and therefore I was surprised that it was failing, and similar other cases were not failing. The cause was that it increased the next-index, because there was only one turn in the outer ring, and therefore it pointed to the wrong next turn.

include/boost/geometry/algorithms/detail/overlay/enrich_intersection_points.hpp
```diff
@@ -490,13 +505,6 @@ inline void enrich_intersection_points(Turns& turns,
                     pair.second, turns,
                     geometry1, geometry2,
                     robust_policy, strategy);
-#ifdef BOOST_GEOMETRY_DEBUG_ENRICH
-        std::cout << "ENRICH-sort Ring " << pair.first << std::endl;
-        for (auto const& op : pair.second)
-        {
-            std::cout << op.turn_index << " " << op.operation_index << std::endl;
```
barendgehrels, line 497: this was without any context and not useful

test/algorithms/buffer/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@
 foreach(item IN ITEMS
     buffer
     buffer_gc
+    buffer_variable_width
```
barendgehrels, line 10: this is the case of last PR, but was not in the cmake yet because that PR was made before

vissarion (approved):
Thanks LGTM!

PR #1254, opened 2024-03-02
[util] move bounds to geometry::util

barendgehrels:
This is a follow up on `numeric_cast` - now for `bounds` - which is different and simpler because it was already a structure.
It also adds unit tests for it.
So `bounds` for Boost.Rational was stil necessary.

It also adds `CMakeLists.txt` for the folders which tests which were affected


include/boost/geometry/util/numeric_cast.hpp
```diff
@@ -30,21 +29,11 @@ struct numeric_caster
...
 
-// Specialization for Boost.Rational
-template <typename Target, typename T>
-struct numeric_caster<Target, rational<T>>
-{
-   static inline Target apply(rational<T> const& source)
-    {
-        return boost::rational_cast<Target>(source);
-    }
-};
```
barendgehrels, line 41: This was placed here in last commit, but it is more consistent with other approach if it is all in `rational.hpp` (it was there before as well)

awulkiew:
Thanks, I think `util` is a better place for this. We could also consider `math`.

With that said. What do you think about removing it after rewriting or deprecating algorithms using `bounds`? It's because the library supports any coordinate type even big numerical types with dynamic size that have no bounds, e.g. Boost.Mutiprecision. AFAIR in the past we made a decision to not use `assign_inverse` in the library. This utliity was left for backward compatibility and if it was used I'd consider this as a bug. What do you think?

Btw, C++14 has this: https://en.cppreference.com/w/cpp/language/attributes/deprecated

barendgehrels:
> Thanks, I think `util` is a better place for this. We could also consider `math`.
> 
> With that said. What do you think about removing it after rewriting or deprecating algorithms using `bounds`? It's because the library supports any coordinate type even big numerical types with dynamic size that have no bounds, e.g. Boost.Mutiprecision. AFAIR in the past we made a decision to not use `assign_inverse` in the library. This utliity was left for backward compatibility and if it was used I'd consider this as a bug. What do you think?
> 
> Btw, C++14 has this: https://en.cppreference.com/w/cpp/language/attributes/deprecated

Didn't answer this yet. But yes, it makes sense. We use it in some places, but not often. And yes, assign_inverse we can deprecate, I forgot it, but I guess it was mainly for spherical/geographic

barendgehrels:
There was one comment but I don't think it will change the PR.
Can it be approved? @awulkiew @vissarion

awulkiew:
lgtm thanks.

awulkiew (approved):

vissarion (approved):
Looks good to me, thanks!

PR #1255, opened 2024-03-03
Add alias templates for core and util traits

awulkiew:
Plus:
Add C++17 variable templates for traits behind #ifdef
Use aliases in several algorithms

### DIFF
--- a/include/boost/geometry/algorithms/append.hpp
+++ b/include/boost/geometry/algorithms/append.hpp
@@ -3,6 +3,7 @@
 // Copyright (c) 2007-2014 Barend Gehrels, Amsterdam, the Netherlands.
 // Copyright (c) 2008-2014 Bruno Lalande, Paris, France.
 // Copyright (c) 2009-2014 Mateusz Loskot, London, UK.
+// Copyright (c) 2024 Adam Wulkiewicz, Lodz, Poland.
 
 // This file was modified by Oracle on 2014-2023.
 // Modifications copyright (c) 2014-2023, Oracle and/or its affiliates.
@@ -60,7 +61,7 @@ struct to_range_point
     static inline void apply(Geometry& geometry, Point const& point,
                              signed_size_type = -1, signed_size_type = 0)
     {
-        typename geometry::point_type<Geometry>::type copy;
+        geometry::point_type_t<Geometry> copy;
         geometry::detail::conversion::convert_point_to_point(point, copy);
         traits::push_back<Geometry>::apply(geometry, copy);
     }
@@ -90,7 +91,7 @@ struct to_polygon_point
     static inline void apply(Polygon& polygon, Point const& point,
                              signed_size_type ring_index, signed_size_type = 0)
     {
-        using ring_type = typename ring_type<Polygon>::type;
+        using ring_type = ring_type_t<Polygon>;
 
         if (ring_index == -1)
         {
@@ -112,18 +113,16 @@ struct to_polygon_range
     static inline void apply(Polygon& polygon, Range const& range,
                              signed_size_type ring_index, signed_size_type = 0)
     {
-        using ring_type = typename ring_type<Polygon>::type;
-        using exterior_ring_type = typename ring_return_type<Polygon>::type;
-        using interior_ring_range_type = typename interior_return_type<Polygon>::type;
-
+        using ring_type = ring_type_t<Polygon>;
+        
         if (ring_index == -1)
         {
-            exterior_ring_type ext_ring = exterior_ring(polygon);
+            auto&& ext_ring = exterior_ring(polygon);
             to_range_range::apply<ring_type, Range>(ext_ring, range);
         }
         else if (ring_index < signed_size_type(num_interior_rings(polygon)))
         {
-            interior_ring_range_type int_rings = interior_rings(polygon);
+            auto&& int_rings = interior_rings(polygon);
             to_range_range::apply<ring_type, Range>(range::at(int_rings, ring_index), range);
         }
     }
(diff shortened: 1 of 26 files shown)


include/boost/geometry/algorithms/append.hpp
```diff
@@ -60,7 +61,7 @@ struct to_range_point
     static inline void apply(Geometry& geometry, Point const& point,
                              signed_size_type = -1, signed_size_type = 0)
     {
-        typename geometry::point_type<Geometry>::type copy;
+        geometry::point_type_t<Geometry> copy;
```
barendgehrels, line 64: This first confused me.
Because, what about `point_t`  ...

But it is correct.

From the standard: 
`using add_pointer_t = typename add_pointer<T>::type`
`using common_type_t = typename common_type<T...>::type`

We have to confirm to the second form.

include/boost/geometry/algorithms/clear.hpp
```diff
@@ -90,7 +91,7 @@ namespace dispatch
 template
 <
     typename Geometry,
-    typename Tag = typename tag_cast<typename tag<Geometry>::type, multi_tag>::type
+    typename Tag = tag_cast_t<tag_t<Geometry>, multi_tag>
```
barendgehrels, line 94: I'll finish my review but this all looks very good, cool improvement. Way more readable.

Thanks for all this work.

include/boost/geometry/util/type_traits_std.hpp
```diff
@@ -195,22 +198,21 @@ using transcribe_cref_t = typename transcribe_cref<From, To>::type;
...
+        >
+{};
 
 
 namespace util
 {
 
 template <typename T>
-using bare_type = remove_cptrref<T>;
+using bare_type [[deprecated]] = remove_cptrref<T>;
```
barendgehrels, line 215: I think this needs a comment saying why this is deprecated, and what is the replacement.

include/boost/geometry/util/type_traits_std.hpp
```diff
@@ -195,22 +198,21 @@ using transcribe_cref_t = typename transcribe_cref<From, To>::type;
     \endcode
 */
 template <bool IsConst, typename Type>
-struct add_const_if_c
-{
-    typedef std::conditional_t
+struct [[deprecated]] add_const_if_c
```
barendgehrels, line 201: Also here, can you add a comment?

include/boost/geometry/algorithms/append.hpp
awulkiew: Yes, the standard pattern unfortunately is different and `point_t` would probably be confusing.

On similar note. I was thinking about adding aliases for Boost.Range traits like `typename boost::range_value<Range>::type` in `util/range.hpp`. C++20 adds this as `std::ranges::range_value_t<>`. In Boost.Geometry there is `namespace bg::range` already. This seems to be the best place for these aliases even though the word `range` is repeated. However I don't have a strong opinion about it, I see several possibilities:
- `boost::geometry::range::range_value_t<>`
- `boost::geometry::range_value_t<>`
- `boost::geometry::range::value_t<>`
- `boost::geometry::ranges::range_value_t<>` (probably the closest to the standard)

Do you have any opinions?

include/boost/geometry/util/type_traits.hpp
```diff
@@ -307,7 +364,8 @@ using enable_if_geometry_collection_t = typename enable_if_geometry_collection<G
     \note Also a "ring" has areal properties within Boost.Geometry
     \ingroup core
 */
-using util::is_areal;
+template <typename T>
+using is_areal [[deprecated]] = util::is_areal<T>;
```
awulkiew, line 368: Before you ask, I'll add an explanation here as well. :)

include/boost/geometry/algorithms/append.hpp
barendgehrels: I would adhere to the standard. So that's the last one (incl. namespace) or the first one (with our namespace).

barendgehrels (approved):
As said, all cool, thanks.
We should add it to the docs too.

include/boost/geometry/util/type_traits_std.hpp
barendgehrels: Actually I was not aware of `cptrref` and find it hardly readable.
`bare_type` is clearer.

But anyway, there is `std::decay` tool, is it equivalent? Can't we deprecate both and use that instead (if it works)?

include/boost/geometry/algorithms/append.hpp
vissarion: I also agree with the 1st or last.

vissarion (approved):
Thanks, looks nice.

include/boost/geometry/util/type_traits_std.hpp
awulkiew: I've used this name because in the standard there are:
```
remove_const<>
remove_volatile<>
remove_reference<>
remove_pointer<>
```
and then there are:
```
remove_cv<>
remove_cvref<>
```
The latter first removes reference and then const volatile. Because originally `bare_type` was first removing reference, then pointer, then cv (because there can be a reference to pointer) I named it:
```
remove_cptrref<>
```
If we wanted to remove volatile too which we probably should do I'd call it:
```
remove_cvptrref
```
but originally volatile was not removed so I didn't do it.

AFAIU `std::decay` wouldn't remove pointer.

awulkiew:
Thanks for the reviews. Should I merge it now or wait until the release to make bugfixing easiter?

vissarion:
> Thanks for the reviews. Should I merge it now or wait until the release to make bugfixing easiter?

I think it is simpler to merge it after the release.

Issue #1244, opened 2024-02-22
Difference returns wrong result if hole touch the boundary of subtrahend


evgeniyfeder:
Hello
Actual result: Purple - orange = green.
<img src="https://github.com/boostorg/geometry/assets/7823804/a08ad779-e0d1-47f6-b0b7-f68e9af0622c" width="300" height="300">
<img src="https://github.com/boostorg/geometry/assets/7823804/31e6fc4f-e0af-48c6-a48f-6d3c1602916f" width="300" height="300">
Expected result: blue
<img src="https://github.com/boostorg/geometry/assets/7823804/6e233a1f-26f3-41e5-bbe0-687ac948b3a9" width="300" height="300">

I noticed that the problem was added in 1.77->1.78. Related PR could be: https://github.com/boostorg/geometry/pull/887

https://godbolt.org/z/s7zKKbG61

barendgehrels:
Thanks for the report and the potential related PR, I can reproduce it. Clockwise is fine, but counter clockwise is wrong.

vissarion:
https://github.com/boostorg/geometry/pull/1263

Issue #1231, opened 2024-01-26
invalid intersection for CCW polygons


vissarion:
The following program

```cpp
#include <boost/geometry.hpp>

int main() {

    using Point = boost::geometry::model::d2::point_xy<double>;
    using MultiPoint = boost::geometry::model::multi_point<Point>;
    using Polygon = boost::geometry::model::polygon<Point, false>;
    using Linestring = boost::geometry::model::linestring<Point>;
    using MultiPolygon = boost::geometry::model::multi_polygon<Polygon>;
    using MultiLinestring = boost::geometry::model::multi_linestring<Linestring>;

    Polygon p1, p2, p3;
    boost::geometry::read_wkt("POLYGON((0 5,-6 -17,12 17,0 5),(4 6,5 5,0 1,4 6))", p1);
    boost::geometry::read_wkt("POLYGON((3 9,-15 -5,13 -11,3 9))", p2);

    MultiPolygon mp;

    std::cout << boost::geometry::is_valid(p1) << std::endl;
    std::cout << boost::geometry::is_valid(p2) << std::endl;


    std::tuple<MultiPoint, MultiLinestring, MultiPolygon> result;
    boost::geometry::intersection(p1, p2, result);

    std::cout << boost::geometry::wkt(std::get<0>(result)) << std::endl;
    std::cout << boost::geometry::wkt(std::get<1>(result)) << std::endl;
    std::cout << boost::geometry::wkt(std::get<2>(result)) << std::endl;
    std::cout << boost::geometry::area(std::get<2>(result)) << std::endl;

    std::cout << boost::geometry::is_valid(std::get<2>(result)) << std::endl;

    return 0;
}
```

returns the invalid geometry 
`MULTIPOLYGON(((3.33333 8.33333,0 5,-3.40491 -7.48466,-1.21132 -7.95472,5.31429 4.37143,5 5,0 1,4 6,5 5,3.33333 8.33333)))` 
while for CW polygons ie. if we change line
` using Polygon = boost::geometry::model::polygon<Point, false>;`
above with 
` using Polygon = boost::geometry::model::polygon<Point, true>;`
it returns 
`MULTIPOLYGON(((3.33333 8.33333,5 5,5.31429 4.37143,-1.21132 -7.95472,-3.40491 -7.48466,0 5,3.33333 8.33333),(5 5,4 6,0 1,5 5)))`
which is valid and correct.

This behavior is tested in 1.84 and also 1.78. 
[...]

Issue #1229, opened 2024-01-16
Union producing self intersections after 1.78


jpan127:
`Minimal Reproducible Example`: https://godbolt.org/z/Tdb51rcY8

Hi, I have been `union_`ing a set of triangular polygons. Our repo recently updated from `1.71` to `1.82` which broke one of our union functions. The above example takes 2 triangles, that share an edge, and unions them. The result contains duplicate points and fails `boost::geometry::is_valid`.

All versions from `1.79 onwards` fails this example.

I noticed that the release notes (https://beta.boost.org/doc/libs/1_79_0/libs/geometry/doc/html/geometry/release_notes.html) show issue (https://github.com/boostorg/geometry/issues/838) which mentions `BOOST_GEOMETRY_NO_ROBUSTNESS`. Defining that for `1.71` also leads to the same failure if that gives any hints.

barendgehrels:
Thanks for the report. I cannot reproduce it anymore. It is fixed by 68a225daf

barendgehrels:
I'll close the issue because it is fixed and a testcase has been added.

cjamin:
Will this fix be released with Boost 1.85?

PR #1256, opened 2024-03-03
[test] add cases for issues #1229 #1231 #1244 and some test refactoring

barendgehrels:


test/algorithms/set_operations/difference/difference.cpp
```diff
@@ -22,15 +22,42 @@
...
 #include <algorithms/test_overlay.hpp>
 #include <algorithms/overlay/overlay_cases.hpp>
 #include <algorithms/overlay/multi_overlay_cases.hpp>
 
+namespace
+{
+
+// Change compiler defines to constexpr bools
+// to make conditions more readable
+// and to always compile all code.
```
barendgehrels, line 39: This is done in `difference.cpp` only, until now.
If you agree, I'll apply it in the others too. And we can apply it everywhere.
Advantages:
* `if` i/o `ifdef`
* code is always compiled
* easier to read

Also, we can harmonize it with the removing of the rescaling, as done in this source file.

test/algorithms/set_operations/difference/difference.cpp
```diff
@@ -41,10 +68,12 @@
     ( #caseid, caseid[0], caseid[1], clips1, -1, area1, clips2, -1, area2, \
                 clips3, -1, area1 + area2, settings)
 
-template <typename P>
+template <typename P, bool ClockWise>
```
barendgehrels, line 71: I added `ccw` tests for difference, it was not earlier there. Most cases work, but a few of them don't - they are now conditionally

test/algorithms/set_operations/difference/difference.cpp
```diff
@@ -498,66 +522,66 @@ void test_all()
...
-
-        test_one<polygon, box, ring>("star_box",
-            clip, example_star,
-            4, 20, 2.833333, 4, 16, 0.833333);
-
-        test_one<polygon, ring, box>("box_star",
-            example_star, clip,
-            4, 16, 0.833333, 4, 20, 2.833333);
+        using box = bg::model::box<P>;
+        using ring = bg::model::ring<P>;
```
barendgehrels, line 528: changes in the block below are mainly whitespace

test/algorithms/set_operations/difference/difference.cpp
```diff
@@ -577,24 +601,42 @@ void test_all()
...
     TEST_DIFFERENCE(issue_1138, 1, 203161.751, 2, 1237551.0171, 1);
 
+    if BOOST_GEOMETRY_CONSTEXPR(test_failures)
+    {
+        // Fails for both clockwise and counter clockwise in detecting the difference
+        // of the interior ring, touching the outer ring.
+        TEST_DIFFERENCE(issue_1231, 1, 203161.751, 2, 1237551.0171, 1);
+    }
+
+    if BOOST_GEOMETRY_CONSTEXPR(! is_ccw || test_failures)
```
barendgehrels, line 623: @awulkiew I assume this is fine, and an else branch is not needed in such cases

vissarion (approved):
Thanks, this will help tracking the bugs in setops.

PR #1260, opened 2024-03-13
remove unused source and deprecate many

barendgehrels:


barendgehrels:
If still possible, it should make it to 1.85, otherwise we should deprecate it one version later.

vissarion (approved):
LGTM, thanks!

PR #1263, opened 2024-03-26
Fix CCW and isolated turns

vissarion:
Fixes issues #1231, #1244 (and a few more test cases). I want to target 1.85 if possible.


include/boost/geometry/algorithms/detail/overlay/traversal_switch_detector.hpp
```diff
@@ -391,11 +391,19 @@ struct traversal_switch_detector
     {
         for (turn_type& turn : m_turns)
         {
+            constexpr auto order1 = geometry::point_order<Geometry1>::value;
+            constexpr bool reverse1 = (order1 == boost::geometry::counterclockwise)
+                ? ! Reverse1 : Reverse1;
+
+            constexpr auto order2 = geometry::point_order<Geometry2>::value;
+            constexpr bool reverse2 = (order2 == boost::geometry::counterclockwise)
+                ? ! Reverse2 : Reverse2;
```
barendgehrels, line 400: 👍 great fix, thanks a lot

test/algorithms/set_operations/intersection/intersection.cpp
```diff
@@ -314,11 +312,10 @@ void test_areal()
 #endif
 
     TEST_INTERSECTION(issue_1229, 0, -1, 0);
-    
+
     {
-        // For CCW this test case currently reports as invalid
         ut_settings settings;
-        settings.set_test_validity(! is_ccw);
+        settings.set_test_validity(true);
```
barendgehrels, line 318: This works - but you can also remove the settings and just call `TEST_INTERSECTION` (here and below)

barendgehrels (approved):

Issue #1262, opened 2024-03-26
Buffer gives incorrect result for negative distance


jdmarsh:
I have a polygon where I want to buffer inwards using a negative distance. The result I get is incorrect.

Polygon(red) and then the result of the buffer command "output" (green).
![image](https://github.com/boostorg/geometry/assets/972442/6aa5c629-8ff0-4fe9-bc5c-4df73345af9c)

The expected result of this buffer is for the output to be empty, instead the green triangle is huge.
- At the top left corner, it's buffered inwards but not by the -1.8 units specified.
- At the top right corner, the output goes outside the outer boundary of the polygon
- The center should certainly not be filled in

The code to generate this can be found here:
https://godbolt.org/z/18sq55oTG

barendgehrels:
Fixed recently in acaed28f1
I will create a unit test

barendgehrels:
Thanks for the report.
For the record, the PR was #1248

PR #1264, opened 2024-03-27
test: add test case for issue #1262

barendgehrels:


test/algorithms/buffer/test_buffer_svg_per_turn.hpp
```diff
@@ -15,6 +15,8 @@
 #ifndef BOOST_GEOMETRY_TEST_BUFFER_SVG_PER_TURN_HPP
 #define BOOST_GEOMETRY_TEST_BUFFER_SVG_PER_TURN_HPP
 
+#if defined(TEST_WITH_SVG_PER_TURN)
```
barendgehrels, line 18: This only needs to be compiled if this define is set.
Otherwise, `#include <boost/ptr_container/ptr_vector.hpp>` fails because `ptr_vector` is currently not in the CMakeLists and I don't want to add another dependency for this.

It's years ago I used this svg per turn, and I think it can be removed.

vissarion (approved):
LGTM, thanks!

PR #1265, opened 2024-03-27
Fix BOOST_GEOMETRY_CONSTEXPR call in index

vissarion:
This PR fixes the compilation error 
`error: macro "BOOST_GEOMETRY_CONSTEXPR" passed 2 arguments, but takes just 1`
introduced by cb2f53d9d90f9d602833e87ebf9699890b4f74d7

### DIFF
--- a/include/boost/geometry/index/detail/rtree/visitors/insert.hpp
+++ b/include/boost/geometry/index/detail/rtree/visitors/insert.hpp
@@ -328,11 +328,11 @@ protected:
         // Enlarge it in case if it's not bounding geometry type.
         // It's because Points and Segments are compared WRT machine epsilon
         // This ensures that leafs bounds correspond to the stored elements
-        if BOOST_GEOMETRY_CONSTEXPR (std::is_same<Element, value_type>::value
+        if (BOOST_GEOMETRY_CONSTEXPR ((std::is_same<Element, value_type>::value
                                     && ! index::detail::is_bounding_geometry
                                             <
                                                 typename indexable_type<translator_type>::type
-                                            >::value)
+                                            >::value)))
         {
             geometry::detail::expand_by_epsilon(m_element_bounds);
         }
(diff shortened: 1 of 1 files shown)


barendgehrels (approved):
lgtm

include/boost/geometry/index/detail/rtree/visitors/insert.hpp
```diff
@@ -328,11 +328,11 @@ class insert
         // Enlarge it in case if it's not bounding geometry type.
         // It's because Points and Segments are compared WRT machine epsilon
         // This ensures that leafs bounds correspond to the stored elements
-        if BOOST_GEOMETRY_CONSTEXPR (std::is_same<Element, value_type>::value
+        if (BOOST_GEOMETRY_CONSTEXPR ((std::is_same<Element, value_type>::value
```
awulkiew, line 331: Have you tested it with C++17 compiler?
My guess is that this would not compile because it'd expand to
```
if (constepxr ((...)))
```
The problem here is the comma. I think you only have to put `std::is_same<Element, value_type>::value` in parentheses.

vissarion: You are right, the problem is the first parenthesis, either putting `std::is_same<Element, value_type>::value` or the whole expression in parantheses is working, but I will follow your suggestion because it is simpler and focuses on the problem which is the comma. I will open a new PR.

vissarion: Do we want to add testing with C++17 on CI?

awulkiew: Yes, I think it's a good idea. We have optional C++17-dependent code: inline variables, `std::any` adaptation, `std::variant` adaptation, `if constexpr`.

PR #1271, opened 2024-04-05
[strategies] Fix spherical closest_points strategy for PointOfSegment different than model::point

awulkiew:
Currently it fails to compile due to multiple return types.


vissarion (approved):
thanks!

barendgehrels (approved):
Thanks
(we could add a test for this situation)

awulkiew:
> Thanks (we could add a test for this situation)

@barendgehrels would you prefer me to create a separate test case for user-defined types or maybe to replace BG models in all of them? If it's the latter then maybe we should use types different than BG models in all of the tests?

PR #1273, opened 2024-04-11
Enable C++17 testing in circleCI

vissarion:
This PR enables testing in circleCI the C++17 dependent code, following the discussion here: https://github.com/boostorg/geometry/pull/1265#discussion_r1544804957


awulkiew:
Would it be possible to force b2 to only compile the sources instead of running the tests?
I'm thinking about some mechanism automatically replacing `run` and `link` directive(?) with `compile` for targets in Jamfiles.

barendgehrels (approved):
Thanks

vissarion:
I have made a try, actually editing Jamfiles. I couldn't find a simpler way; it seems that b2 does not have an only build command. 

An alternative could be to use cmake.

vissarion:
cmake provides a much simpler way to do this (see https://github.com/boostorg/geometry/pull/1302) so I am closing this PR

PR #1274, opened 2024-04-21
Add enumerate to prepare for 17 and 23

barendgehrels:
I was recently pointed at `enumerate` which has similar behavior as our `...with_index`, but better.
So I propose to change that function with lambdas (I wrote it recently) into this view/range based approach, to be better prepared for C++17 (structured bindings) and C++23 (views::enumerate), avoid the lambda and the closures, and be more in line with the C++ direction.

If you like it, I'll change the other occurences of `for_each_with_index` too and we can deprecate it or remove it (it's in namespace `detail`)


jkuebart (approved):

awulkiew (approved):
That's interesting. Thanks!

barendgehrels:
Because of the approvals already, I changed the other occurences as well.
If @vissarion is also fine it can be merged.

include/boost/geometry/views/enumerate_view.hpp
```diff
@@ -0,0 +1,158 @@
...
+            >;
+
+        // Member variable index contains the zero-based index of the value in the range
+        std::size_t const index;
+
+        // Member variable value contains a const or non-const reference to the value itself
+        type& value;  
+    };
+
+ private:
```
vissarion, line 51: Please use the indentation of "public:" below. Similarly adjust the comment below.

vissarion (approved):
Thanks!

include/boost/geometry/algorithms/detail/overlay/check_enrich.hpp
```diff
@@ -26,6 +26,7 @@
 #include <boost/range/value_type.hpp>
 
 #include <boost/geometry/algorithms/detail/overlay/overlay_type.hpp>
+#include <boost/geometry/views/enumerate.hpp>
```
vissarion, line 29: this should be `#include <boost/geometry/views/enumerate_view.hpp>`

vissarion (changes requested):
There is a error in a header file include that should be changed before merging.

include/boost/geometry/algorithms/detail/overlay/check_enrich.hpp
barendgehrels: Oh - that slipped through my cmake. Thanks!
I got mails about this too, so that works great.

include/boost/geometry/views/enumerate_view.hpp
barendgehrels: ✅

include/boost/geometry/algorithms/detail/overlay/enrich_intersection_points.hpp
```diff
@@ -534,9 +540,15 @@ inline void enrich_intersection_points(Turns& turns,
...
-    //detail::overlay::check_graph(turns, for_operation);
+    constexpr bool do_check_graph = true;
+#else
+    constexpr bool do_check_graph = false;
 #endif
 
+    if BOOST_GEOMETRY_CONSTEXPR (do_check_graph)
+    {
+        detail::overlay::check_graph(turns, target_operation);
+    }
```
barendgehrels, line 551: (Slightly) related change which enforces compilation of `check_graph` (it was broken in 2 ways,
one by my PR, the other because `for_operation` is renamed somewhen).

@vissarion is this OK?

vissarion: yes, thanks.

vissarion (approved):

barendgehrels:
Github has currently performance problems.
Waiting until it can check again.

PR #1275, opened 2024-04-25
Initialize variables in ob_tran projection

vissarion:
This PR intents to prevent a maybe uninitialized warning that appear with recent versions of gcc (11 and 12).


barendgehrels (approved):
This should work indeed.
But why don't we initialize it where the variable is declared (the C++11 way)?

vissarion:
> This should work indeed. But why don't we initialize it where the variable is declared (the C++11 way)?

You are right, I was following the old style of the file, now they are initialized where they declared.

PR #1296, opened 2024-07-26
Some simplifications 

vissarion:


include/boost/geometry/algorithms/detail/is_simple/linear.hpp
```diff
@@ -68,12 +68,8 @@ template <typename Turn>
...
                                   signed_size_type last_index)
 {
-    return
-        (turn.operations[0].seg_id.segment_index == 0
-         && turn.operations[1].seg_id.segment_index == last_index)
-        ||
-        (turn.operations[0].seg_id.segment_index == 0
-         && turn.operations[1].seg_id.segment_index == last_index);
+    return turn.operations[0].seg_id.segment_index == 0
+        && turn.operations[1].seg_id.segment_index == last_index;
```
barendgehrels, line 72: I think what was meant here was that the second condition had `1` vs `0` instead.
But it's hard to guess. I didn't write it (I think). @awulkiew ?

barendgehrels (approved):
Thanks! One question but good looks

include/boost/geometry/algorithms/detail/is_simple/linear.hpp
vissarion: OK, I kept the second condition with reversed `0` and `1`. 

Out of curiosity, is it possible for the second condition to be true in the way turns are computed in boost geometry?

barendgehrels: Yes, I think so. But maybe better to check that before merging it.
I'll try to check one of the coming days.

barendgehrels: @vissarion I forgot, and don't have the time to do it now in detail (it would require a quite specific testcase).
The operations are symmetric (have the same role) and unordered, so I'm quite sure it's right now.
Feel free to merge it.

PR #1297, opened 2024-07-27
fix: remove remaining policies

barendgehrels:


barendgehrels:
Merging this already because of holiday season and it's hindering building with cmake in any branch.

PR #1298, opened 2024-07-27
[examples] update wxwidgets example

barendgehrels:
It's an old example but modernized, mainly the way how to make it, but also it accepts a command line now.

Building it is now trivial using cmake.

Running it like this:
`./wx_widgets_world_mapper ../../../data/world.wkt`

Giving

image


vissarion (approved):
Cool, thanks!

Issue #1293, opened 2024-07-25
Difference of polygons is incorrectly empty


janmoltz:
The `difference` method for polygons returns an incorrect empty result in some cases. This is an example:
```
#include <boost/geometry.hpp>
#include <boost/geometry/geometries/point_xy.hpp>
#include <boost/geometry/geometries/polygon.hpp>
#include <boost/geometry/geometries/multi_polygon.hpp>
typedef boost::geometry::model::d2::point_xy<double> BoostPoint;
typedef boost::geometry::model::polygon<BoostPoint> BoostPolygon;
typedef boost::geometry::model::multi_polygon<BoostPolygon> BoostMultiPolygon;

BoostPolygon polygon1;
boost::geometry::append(polygon1, BoostPoint(13.90250403054817, -108.7493905903629)); // when adding this point to polygon1 as the last point, it works
boost::geometry::append(polygon1, BoostPoint(18.1717071533203, -109.6161492466927)); // very similar to first point of polygon2; when changing coordinate to, for example, 18.1717071534203, it works
boost::geometry::append(polygon1, BoostPoint(20.38385009765624, -109.6161492466927));
boost::geometry::append(polygon1, BoostPoint(13.74227594050434, -109.9162293522026));
boost::geometry::append(polygon1, BoostPoint(13.80791011645497, -109.3242516085326));
boost::geometry::correct(polygon1);
BoostPolygon polygon2;
boost::geometry::append(polygon2, BoostPoint(18.17170715332031, -109.6161492466927));
boost::geometry::append(polygon2, BoostPoint(19.60749526774088, -109.5220427274138));
boost::geometry::append(polygon2, BoostPoint(11.19696044921875, -108.1424287761621));
boost::geometry::append(polygon2, BoostPoint(12.48744269320276, -108.7788241542253));
boost::geometry::append(polygon2, BoostPoint(13.84995095309101, -109.2413333582132));
boost::geometry::append(polygon2, BoostPoint(15.26117233479817, -109.5220427274138));
boost::geometry::append(polygon2, BoostPoint(16.69696044921874, -109.6161492466927));
boost::geometry::correct(polygon2);
BoostMultiPolygon result;
boost::geometry::difference(polygon1, polygon2, result);
std::cout << result.size() << std::endl; // 0, should be 1
```
The polygons look like this:
<img width="1036" alt="image" src="https://github.com/user-attachments/assets/c65b1904-9349-47ee-b83c-0bb06fe2e75a">
[...]

Issue #1295, opened 2024-07-26
boost::geometry::intersection, wrong result (result polygon is equal to the biggest of the input polygons)


yaroslav-zlot:
For 1.85.0 with MSVC2022 the result of boost::geometry::intersection for certain input is wrong.
Coordinate type: double

when some of the points of different input polygons are close to each other and
the values for the coordinates are very small (like 4.6030847803960675e-16)
boost::geometry::intersection produces completely wrong result:

In the test example one polygon is roughly contained in the other so the result of the intersection has to be 
about the same as the smaller polygon, but instead of it the result is the containing polygon.

Setting the values to zero, for those small values, workarounds it.

For 1.74.0 the result is expected.

In the pictures, the first polygon is in red, the second is in green and the intersection is in black, thinner than for the input polygons.

Wrong result for 1.85.0
![1 85_wrong_result](https://github.com/user-attachments/assets/ce0ee4cf-4f8b-4334-863a-943d5b68fbc2)

Expected result for 1.74.0 or 1.85.0 when small values are set to zero
![1 74_or_1 85_with_nearly_zero_coordinates_set_to_zero](https://github.com/user-attachments/assets/3f4bdc34-3b9b-48d7-a3e7-6d9dd5a02346)

```c++
#include "boost/geometry/algorithms/intersection.hpp"
#include "boost/geometry/algorithms/is_valid.hpp"
#include "boost/geometry/geometries/polygon.hpp"
#include "boost/geometry/geometries/multi_polygon.hpp"
#include "boost/geometry/geometries/point_xy.hpp"

#include <cassert>

int main(int, char *[])
{
    using point_2d = boost::geometry::model::d2::point_xy<double>;
    using BoostPolygon = boost::geometry::model::polygon<point_2d>;
    using BoostMultiPolygon = boost::geometry::model::multi_polygon<BoostPolygon>;
    using Ring = BoostPolygon::ring_type;

    auto verifyIntersectionAreaIsLessOrEqualThanAreasOfInputPolygons = [](const auto& polygon1, const auto& polygon2)
    {
        assert(boost::geometry::is_valid(polygon1) && boost::geometry::is_valid(polygon2));
        const auto areaOfPolygon1 = boost::geometry::area(polygon1);
        const auto areaOfPolygon2 = boost::geometry::area(polygon2);
        const double eps = 1e-7;
        if (areaOfPolygon1 >= eps && areaOfPolygon2 >= eps)
        {
[...]

Issue #1294, opened 2024-07-25
bug: geometry::buffer returns an empty multi-polygon in a very specific case


DJ-caddev:
BOOST v1.84 / v1.85

Using the following code, lOuts is empty in the end:
```c++
boost::geometry::model::polygon<boost::geometry::model::d2::point_xy<double>> lPolygon;
lPolygon.outer().push_back({730.35, 750.00});
lPolygon.outer().push_back({730.35, 740.00});
lPolygon.outer().push_back({726.02, 740.00});
lPolygon.outer().push_back({726.02, 735.00});
lPolygon.outer().push_back({730.35, 735.00});
lPolygon.outer().push_back({730.35, 0.00});
lPolygon.outer().push_back({0.00, 0.00});
lPolygon.outer().push_back({0.00, 750.00});
lPolygon.outer().push_back({730.35, 750.00});

static const boost::geometry::strategy::buffer::side_straight side_strategy;
const boost::geometry::strategy::buffer::distance_symmetric<double> distance_strategy(5);
boost::geometry::model::multi_polygon<boost::geometry::model::polygon<boost::geometry::model::d2::point_xy<double>>> lOuts;

static const boost::geometry::strategy::buffer::join_miter join_strategy;
static const boost::geometry::strategy::buffer::end_flat end_strategy;
static const boost::geometry::strategy::buffer::point_circle point_strategy;
boost::geometry::buffer(lPolygon, lOuts, distance_strategy, side_strategy, join_strategy, end_strategy, point_strategy);
```

The shape looks like this:
![image](https://github.com/user-attachments/assets/3bd5ed72-1187-4742-bb79-c5907a54cc6c)

If the small hole is moved a bit to the bottom, the boost function seems to work well.
This is a very specific case that crashes in our code.
Moreover, we need to keep "rectangular/cartesian" shapes (we cannot have arcs in the final shapes).

Do you guys have any clue of how to bypass this particular problem? Is this a wrong usage from on our side or something that Boost can fix directly?

If you need any detail, feel free to ask :)

barendgehrels:
Indeed, I can reproduce it. The code is right.

barendgehrels:
I've a working fix and will make a PR on Wednesday

barendgehrels:
This was fixed in July 30 (but the issue was not closed automatically, somehow - and not manually). PR #1300
Closing.

PR #1300, opened 2024-07-30
traverse first through non clustered turns, remove cluster exits, make priority consistent

barendgehrels:
Fixes issue #1293 #1294 and #1295

This PR fixes several issues. There are no regressions.

More detailed:
* it changes traversal such that it first traverses non clustered turns (simpler) and only then the remaining clustered ones
* that caused a regression in `issue_630_b` (a relatively simple case)
* that was fixed by removing `cluster_exits` (a welcome removal, see below)
* that caused three other regressions
* they could be fixed by using the priority approach, earlier used only for union/buffer. It is now also used for intersection/difference and in exactly the same way

Removing cluster_exits is welcome because it was designed as a simple approach, not needing many conditions. But in the end there are conditions there as well, and (apparently) bugs. So it is welcome to remove it again and keep the code having one approach, used for both union and intersection in the same way.


include/boost/geometry/algorithms/detail/overlay/traversal.hpp
```diff
@@ -231,26 +230,12 @@ public :
...
-
-        if BOOST_GEOMETRY_CONSTEXPR (OverlayType == overlay_buffer)
-        {
-            // Buffer does not use source_index (always 0).
-            return select_source_generic<&segment_identifier::multi_index>(
+        // Buffer and self-turns do not use source_index (always 0).
+        return OverlayType == overlay_buffer || is_self_turn<OverlayType>(turn)
+            ? select_source_generic<&segment_identifier::multi_index>(
+                        turn, candidate_seg_id, previous_seg_id)
+            : select_source_generic<&segment_identifier::source_index>(
```
barendgehrels, line 237: This is not part of the fix but makes code more concise and readable

include/boost/geometry/algorithms/detail/overlay/traversal.hpp
```diff
@@ -591,14 +576,14 @@ public :
...
-            if (op.enriched.count_left != 0 || op.enriched.count_right == 0)
+            if ((op.enriched.count_left != 0 || op.enriched.count_right == 0) && cinfo.spike_count > 0)
             {
                 // Check counts: in some cases interior rings might be generated with
                 // polygons on both sides. For dissolve it can be anything.
 
                 // If this forms a spike, going to/from the cluster point in the same
                 // (opposite) direction, it can still be used.
-                return cinfo.spike_count > 0 ? 1 : 0;
+                return 1;
```
barendgehrels, line 586: Returning `0` was a bug, it should only have returned 1.
So that is fixed and is necessary to avoid regressions with this PR

vissarion (approved):
Thanks!

PR #1301, opened 2024-08-06
feat: deprecate duplicate unused strategy distance_projected_point_ax

barendgehrels:
This deprecates `distance_projected_point_ax`

* it is not used inside the library
* it is a remainder of moving the strategy to `projected_point`
* it therefore contained duplicate point
* it is not documented and most probably not used by anyone
* this happened in 2014, most probably


include/boost/geometry/strategies/cartesian/distance_projected_point_ax.hpp
```diff
@@ -20,6 +20,8 @@
 #ifndef BOOST_GEOMETRY_STRATEGIES_CARTESIAN_DISTANCE_PROJECTED_POINT_AX_HPP
 #define BOOST_GEOMETRY_STRATEGIES_CARTESIAN_DISTANCE_PROJECTED_POINT_AX_HPP
 
+#include <boost/config/pragma_message.hpp>
+BOOST_PRAGMA_MESSAGE("This include file is deprecated and will be removed in Boost 1.87")
```
vissarion, line 24: Since this commit will be part of 1.87, should we change the message to 1.88?

vissarion (approved):
Thanks, this looks redundant indeed.

include/boost/geometry/strategies/cartesian/distance_projected_point_ax.hpp
barendgehrels: ✅

Issue #1282, opened 2024-05-14
Test libraries and executables should be declared EXCLUDE_FROM_ALL


pdimov:
The Boost convention is that `cmake --build .` only builds the libraries. Test libraries and executables are built with `cmake --build . --target tests`.

barendgehrels:
@vissarion I think this is handled now by one of your last PR's?

vfisikop:
> @vissarion I think this is handled now by one of your last PR's?

Yes, https://github.com/boostorg/geometry/pull/1302 which is waiting a review feedback. Upon merging this issue will be closed automatically.

vissarion:
> @vissarion I think this is handled now by one of your last PR's?

Yes, https://github.com/boostorg/geometry/pull/1302 which is waiting a review feedback. Upon merging this issue will be closed automatically.

PR #1280, opened 2024-05-09
[test] add cmake lists for 143 algorithms

barendgehrels:
This PR
* adds 143 algorithms
* because this caused a duplicate in `perimeter` : it moves those two algorithms to their folder

They all pass on my Mac.
Do we have this CMake flavor on CI already?

PR #1302, opened 2024-08-14
Update cmake scripts and add them to github actions

vissarion:
This PR: 

- creates github actions for cmake builds and tests (as discussed in https://github.com/boostorg/geometry/pull/1280#issuecomment-2104855981) 
- declares tests EXCLUDE_FROM_ALL (closes https://github.com/boostorg/geometry/issues/1282)
- tests compilation with C++17 (with cmake is much simpler to separate building from running tests, see https://github.com/boostorg/geometry/pull/1273)
- adds some missing tests in cmake (such as index)


barendgehrels:
Cool PR, I review later

test/CMakeLists.txt
```diff
@@ -8,33 +15,37 @@
...
+      "${PROJECT_SOURCE_DIR}/index/test"
       .)
 
   # To compile with C++14
   target_compile_features(${unit_test_name} PRIVATE cxx_std_14)
 
   # To be able to run ctest
-  add_test(NAME ${unit_test_name} COMMAND ${unit_test_name})
+  if (NOT ("boost_geometry_${prefix}_${item}" STREQUAL "boost_geometry_cs_undefined_setops1") AND
+      NOT ("boost_geometry_${prefix}_${item}" STREQUAL "boost_geometry_cs_undefined_setops2"))
```
barendgehrels, line 46: Isn't this too specific?

barendgehrels: And why is it excluded. I get a memory violation - so that might be the reason.
Can't we pass an optional boolean variable to this function, to add to the test suite, or not?
(and then for ones like this we should not, but with a comment why not).

And then, probably, it can still be added as a unit test - but not as dependency to the custom `tests` target

test/cs_undefined/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+    is
+    measure
+    other
+    relops1
+    relops2
+    setops1
+    setops2
+  )
+  boost_geometry_add_unit_test("cs_undefined" ${item})
+endforeach()
```
barendgehrels, line 21: So, in line with my comment above, you would get a second loop here

```
foreach(item IN ITEMS
    setops1
    setops2
  )
  boost_geometry_add_unit_test("cs_undefined" ${item} FALSE)
endforeach()
```

test/robustness/overlay/areal_areal/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+    recursive_polygons
+    star_comb
+    #ticket_9081
+  )
+  boost_geometry_add_unit_test("robustness" ${item})
+  target_include_directories(boost_geometry_robustness_${item}
+    PRIVATE
+      "${PROJECT_SOURCE_DIR}/test/robustness"
+      .)
+endforeach()
```
barendgehrels, line 24: These are not unit tests. These are programs.
I think we need to add a specific function for them.

But anyway, great they are at least compiled now.

barendgehrels (approved):
Thanks!

I made some comments, which can be applied either now, or in a subsequent PR (doesn't have to be you, I can also take care maybe) is also fine.

test/CMakeLists.txt
```diff
@@ -8,33 +15,37 @@
...
   # Include the main Geometry test folder and the current folder
   target_include_directories(${unit_test_name}
     PRIVATE
-      "${PROJECT_SOURCE_DIR}/test" 
+      "${PROJECT_SOURCE_DIR}/test"
+      "${PROJECT_SOURCE_DIR}/index/test"
       .)
 
   # To compile with C++14
   target_compile_features(${unit_test_name} PRIVATE cxx_std_14)
```
barendgehrels, line 42: The PR description mentions: `test compilation with C++17` 
But won't this setting override that?

test/robustness/overlay/areal_areal/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+    intersection_pies
+    intersection_stars
+    intersects
+    random_ellipses_stars
+    recursive_polygons
+    star_comb
+    #ticket_9081
+  )
+  boost_geometry_add_unit_test("robustness" ${item})
+  target_include_directories(boost_geometry_robustness_${item}
```
barendgehrels, line 20: You might use ${BOOST_GEOMETRY_UNIT_TEST_NAME} , it is set by previous function

test/cs_undefined/CMakeLists.txt
vissarion: Thanks, I added then one by one though.

test/CMakeLists.txt
vissarion: Done, thanks.

test/CMakeLists.txt
vissarion: It seems that cmake selects the highest version of the standard specified in both CMAKE_CXX_STANDARD and target_compile_features.
see also https://discourse.cmake.org/t/how-to-handle-different-cxx-standards-in-the-same-project/6510/6

However, it is not very neat to mix CMAKE_CXX_STANDARD and target_compile_features. As the code is right now I think it is simpler to just use CMAKE_CXX_STANDARD and avoid target_compile_features, as the former only needed to be set once while the latter need to be set for each individual test. But maybe for this PR it is OK as is. What do you think?

vissarion:
@barendgehrels thansk for the review and the comments. I addressed them all. Only the discussion on the c++ standard needs some check.

test/CMakeLists.txt
barendgehrels: I agree, maybe we should just use `CMAKE_CXX_STANDARD`
And for this PR, it is OK as it is.

Agreed.

test/CMakeLists.txt
```diff
@@ -8,55 +15,62 @@
...
 
   # To be able to run ctest
-  add_test(NAME ${unit_test_name} COMMAND ${unit_test_name})
+  set(extra_macro_args ${ARGN})
+  if (NOT extra_macro_args STREQUAL "not_run")
```
barendgehrels, line 46: 👍

barendgehrels:
Thanks for handling everything, completely fine!

Osyotr:
Why does `boost::geometry` depend on `Boost::program_options` and `Boost::crc` now?

PR #1303, opened 2024-08-17
[extensions] Fix nsphere access

awulkiew:


vissarion (approved):
Thanks! It seems to be useful to also test extensions on CI.

PR #1307, opened 2024-09-08
[examples] Update qt example

barendgehrels:
Recently I updated wxwidget example.
Similarly, this PR updates the qt example.
It's now simple to install on Mac, using brew

The first example `x01` was nearly the same as the other `x06` so the first is omitted.

Example of running it.
```
./qt_world_mapper.app/Contents/MacOS/qt_world_mapper ../../../../test/algorithms/buffer/data/gr.wkt
```

image


example/with_external_libs/wxwidgets/wxwidgets_world_mapper.cpp
```diff
@@ -34,6 +33,8 @@
 #include "wx/dcgraph.h"
 #endif
 
+#include "common/read_countries.hpp"
```
barendgehrels, line 36: They now share code

vissarion (approved):
Great, thanks!

I managed to build and run in Ubuntu 22 (although not that straightforward ;) 

First, Vulkan SDK was needed (see [here](https://vulkan.lunarg.com/doc/sdk/1.3.239.0/linux/getting_started_ubuntu.html)). 

Then I had several issues with qt local setup which I will not post here because it was probably due to my setup and not applicable to other users. 

Something that should be mentioned in this PR is the dependency upgrade from qt4 to qt6. In particular, QT6.3 which introduces `qt_standard_project_setup()` command. For example, ubuntu 22 has QT6.2.4 thus I replaced  `qt_standard_project_setup()` with 

```
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_AUTOMOC ON)
set(CMAKE_AUTOUIC ON)
set(CMAKE_AUTORCC ON)
```

barendgehrels:
Thanks for testing! Will update the README and consult you

barendgehrels:
BTW I could build it without vulcan. But also with `brew install vulkan-tools`. Will add that also

barendgehrels:
@vissarion I updated the readme according to your findings. Is this OK?

vissarion:
> BTW I could build it without vulcan. But also with `brew install vulkan-tools`. Will add that also

That could be also possible in ubuntu, I didn't chech, I just show the warning and fix it by installing vulcan.

> @vissarion I updated the readme according to your findings. Is this OK?

Thanks, I am OK!

barendgehrels:
> That could be also possible in ubuntu, I didn't chech, I just show the warning and fix it by installing vulcan.

OK, I rephrase it a bit, adding "might", and merge, thanks!

PR #1310, opened 2024-09-22
feat: enhance and fix documentation

barendgehrels:
The registration macros were not clear enough stating that `longitude` should come first, and then `latitude`. 
Adding that, I encountered that the documentation procedure was broken on my Mac. This was caused in the end by this 
commit: #116c7b0 so I wonder why it can run in the main branches.

Anyway, after fixing that locally I can generate the same documentation again. There were quite some warnings (by doxygen, or doxygen to qbk) which are fixed. Also some other unclarities are enhanced.

Some pages were not used for already 10 years and are removed. This was in doxygen - but we use quickbook now.

Documentation looks now like (some shots):

image

image


barendgehrels:
BTW, it's quite confusing that `b2` calls `make_qbk.py` and, vice versa, `make_qbk.py` can optionally call `b2` (in release mode).
Which one is effectively used in release documentation?

barendgehrels-tomtom:
> This was caused in the end by this commit: #116c7b0 so I wonder why it can run in the main branches.

This is https://github.com/boostorg/boostbook/commit/116c7b0f23b578ece6f5c530584a987f4db6947b

doc/debug_make_doc.sh
```diff
@@ -0,0 +1,147 @@
...
+
+echo "=== From DocBook to html"
+xsltproc ${params} --path ${folder_generated} -o "html/" ${xsl_html} ${file_docbook}
+
+echo "=== Resulting html, generated with:"
+cat html/index.html| grep -w -e generator
+
+echo "=== Resulting contents of generated 'adapted.html', there should be sections"
+cat html/geometry/reference/adapted.html | grep section
+
```
barendgehrels, line 147: @Lastique also for your information ('ll refer to it from elsewhere)

Issue #1311, opened 2024-09-24
`doxygen_xml2qbk` not found during documentation build


Lastique:
This is on Kubuntu 24.04, when building Boost.Geometry docs for the first time:

```
sh: 1: doxygen_xml2qbk: not found
Boost.Geometry is making .qbk files in /home/lastique/src/boost/libs/geometry/doc
Boost.Geometry is cleaning Doxygen files in /home/lastique/src/boost/libs/geometry/doc/doxy/doxygen_output/xml/
Boost.Geometry is cleaning Doxygen files in /home/lastique/src/boost/libs/geometry/doc/index/xml/
Traceback (most recent call last):
  File "/home/lastique/src/boost/libs/geometry/doc/make_qbk.py", line 209, in <module>
    exec(compile(open("make_qbk.py", "rb").read(), "make_qbk.py", 'exec'))
  File "make_qbk.py", line 38, in <module>
    run_command(cmd % ("classboost_1_1geometry_1_1index_1_1rtree", "rtree"))
  File "make_qbk.py", line 27, in run_command
    raise Exception("Error running %s" % command)
Exception: Error running doxygen_xml2qbk --xml xml/classboost_1_1geometry_1_1index_1_1rtree.xml --start_include boost/ --output_style alt --alt_max_synopsis_length 59 > generated/rtree.qbk

EXIT STATUS: 1
====== END OUTPUT ======

    
    PYTHONPATH="/home/lastique/src/boost/libs/geometry/doc"
export PYTHONPATH
 "/home/lastique/src/boost/libs/geometry/doc/make_qbk.py" --release-build --doxygen-xml2qbk "../../../bin.v2/libs/geometry/doc/src/docutils/tools/doxygen_xml2qbk/gcc-13/release/x86_64/link-static/threading-multi/visibility-hidden/doxygen_xml2qbk" > "../../../bin.v2/libs/geometry/doc/make_qbk.test/gcc-13/release/x86_64/link-static/threading-multi/visibility-hidden/make_qbk.output" 2>&1 < /dev/null
    status=$?
    echo >> "../../../bin.v2/libs/geometry/doc/make_qbk.test/gcc-13/release/x86_64/link-static/threading-multi/visibility-hidden/make_qbk.output"
    echo EXIT STATUS: $status >> "../../../bin.v2/libs/geometry/doc/make_qbk.test/gcc-13/release/x86_64/link-static/threading-multi/visibility-hidden/make_qbk.output"
[...]

PR #1312, opened 2024-09-24
Fix `doxygen_xml2qbk` execution from `make_qbk.py`

Lastique:
When `b2` invokes `doc/make_qbk.py`, it passes a relative path to the compiled `doxygen_xml2qbk` executable. This path becomes invalid if used from a directory different from doc, which may happen as `make_qbk.py` changes the current directory during its execution. Additionally, `doc/index/make_qbk.py` invokes `doxygen_xml2qbk` unqualified, which requires the executable to be in `PATH` and is never the case unless the user has pre-compiled and placed it accordingly.

To fix this, first resolve the paths to `doxygen` and `doxygen_xml2qbk` to absolute paths before changing the current directory. Additionally, pass the resolved path to `doxygen_xml2qbk` to `doc/index/make_qbk.py` via the `DOXYGEN_XML2QBK` environment variable.

Fixes https://github.com/boostorg/geometry/issues/1311.

### DIFF
--- a/doc/index/make_qbk.py
+++ b/doc/index/make_qbk.py
@@ -15,7 +15,24 @@
 
 import os, sys, shutil
 
-cmd = "doxygen_xml2qbk"
+# Resolves the path to an executable and returns an absolute path to it
+def resolve_executable(orig_path):
+    resolved_path = shutil.which(orig_path)
+    if resolved_path is None:
+        raise Exception("%s is not found or not executable" % orig_path)
+    return os.path.abspath(resolved_path)
+
+if 'DOXYGEN_XML2QBK' in os.environ:
+    doxygen_xml2qbk_cmd = os.environ['DOXYGEN_XML2QBK']
+elif '--doxygen-xml2qbk' in sys.argv:
+    doxygen_xml2qbk_cmd = sys.argv[sys.argv.index('--doxygen-xml2qbk')+1]
+else:
+    doxygen_xml2qbk_cmd = 'doxygen_xml2qbk'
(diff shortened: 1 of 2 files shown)


barendgehrels (approved):
Thanks a lot!
So this fixes a structural issue and makes making documentation easier.

For others, @Lastique also solved the "1.79.2" problem, as discussed here
https://github.com/boostorg/boostbook/pull/13

For that research, this PR was essential.

vissarion (approved):
Thanks a lot, looks good to me, I approve. 

Just a note on documentation, there is a [CI job](https://github.com/boostorg/geometry/actions/workflows/documentation.yml) that builds documentation and seems to work as expected. Of course it is on Ubuntu so it cannot catch issues on Mac/win etc.

Lastique:
The `documentation` CI workflow works because it explicitly builds `doxygen_xml2qbk` and copies it to `/usr/local/bin` prior to building the documentation. This is unconventional and should not be required. This PR fixes this, and building docs should now work by simply invoking `b2` in `libs/geometry/doc`. It may be a good idea to update the CI workflow to verify that.

vissarion:
> The `documentation` CI workflow works because it explicitly builds `doxygen_xml2qbk` and copies it to `/usr/local/bin` prior to building the documentation. This is unconventional and should not be required. This PR fixes this, and building docs should now work by simply invoking `b2` in `libs/geometry/doc`. It may be a good idea to update the CI workflow to verify that.

That is a good suggestion, I opened a [PR](https://github.com/boostorg/geometry/pull/1314). Please feel free to review it.

PR #1313, opened 2024-09-26
feat: add cmake support to examples

barendgehrels:
- replace BOOST_FOREACH

- replace Boost.Assign

- replace typedef with using

- other minor changes


doc/src/examples/algorithms/clear.cpp
```diff
@@ -56,7 +49,7 @@ int main()
 /*`
 Output:
 [pre
-(((0, 0), (0, 10), (11, 11), (0, 0)), ((0, 0), (0, 10), (11, 11), (0, 0)))
+(((0, 0), (0, 10), (10, 10), (10, 0), (0, 0)), ((1, 2), (8, 2), (4, 6), (1, 2)))
```
barendgehrels, line 52: This example was incorrect, the output did not correspond with the code

doc/src/examples/algorithms/for_each_segment_const.cpp
```diff
@@ -11,72 +11,37 @@
...
+    // Iterate over its segments to find the minimum and maximum length
+    double min_length = std::numeric_limits<double>::max();
+    double max_length = -std::numeric_limits<double>::max();
+    
+    boost::geometry::for_each_segment(polyline, 
+        [&](auto const& s)
+        {
+            const auto length = boost::geometry::length(s);
+            min_length = std::min(min_length, length);
+            max_length = std::max(max_length, length);
```
barendgehrels, line 38: Replaced with lambda

doc/src/examples/algorithms/make_2d_point.cpp
```diff
@@ -39,10 +38,9 @@ void construct_and_display()
 
 int main()
 {
-    construct_and_display<boost::geometry::model::d2::point_xy<double> >();
-    construct_and_display<boost::geometry::model::d2::point_xy<int> >();
-    construct_and_display<boost::tuple<double, double> >();
-    construct_and_display<boost::polygon::point_data<int> >();
+    construct_and_display<boost::geometry::model::d2::point_xy<double>>();
+    construct_and_display<boost::geometry::model::d2::point_xy<int>>();
+    construct_and_display<boost::tuple<double, double>>();
```
barendgehrels, line 43: `Boost.Polygon` is removed. The dependency is removed. This is probably not a use case for anyone, anymore.

doc/src/examples/io/svg.cpp
```diff
@@ -1,14 +1,17 @@
...
 // Use, modification and distribution is subject to the Boost Software License,
 // Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
 // http://www.boost.org/LICENSE_1_0.txt)
 
 //[svg_mapper
-//` Shows the usage of svg_mapper
+//` Shows the usage of svg_mapper.
+//` Make sure to first call only the method [^add] and then call only the methods [^map] or [^text]. 
+//` Because [^add] calculates the bounding box internally and that should be finished
+//` before calling mapping functions. This might require two loops.
```
barendgehrels, line 14: Oops, this is also part of the other PR (I continued with that one). Anyway, no problem, will be harmonized

doc/src/examples/geometries/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+    multi_point
+    multi_polygon
+    ring
+    segment
+  )
+
+  boost_geometry_add_example("geometries" ${item})
+  
+endforeach()
+
```
barendgehrels, line 24: Still needs to add adapted examples - will be added in a separate PR

doc/src/examples/algorithms/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+    envelope
+    equals
+    expand
+    for_each_point
+    for_each_point_const
+    for_each_segment_const
+    intersection_ls_ls_point
+    intersection_segment
+    intersects_linestring
+    is_simple
```
vissarion, line 42: Is this the complete list? For example, it seems that `is_empty` is missing. I am counting 82 items in the examples directory but here there are only 60 in `foreach` loop.

vissarion (commented):
Great addition, and modernization! Thanks!

I have only a question. Also it seems that something is missing and some examples cannot be compiled in CI.

doc/src/examples/algorithms/CMakeLists.txt
barendgehrels: Oh, I took over what was in the Jamfile.
So there are 22 more. I will add them.

barendgehrels:
> Great addition, and modernization! Thanks!
> 
> I have only a question. Also it seems that something is missing and some examples cannot be compiled in CI.

Indeed, will handle. Thanks.

doc/src/examples/algorithms/CMakeLists.txt
barendgehrels: ✅ added most of them. Some of them are not even used (added that as a comment).
An ancient one which was not used is deleted.

barendgehrels:
> > I have only a question. Also it seems that something is missing and some examples cannot be compiled in CI.
> 
> Indeed, will handle. Thanks.

✅

doc/src/examples/geometries/CMakeLists.txt
barendgehrels: ✅ added in this PR

doc/src/examples/algorithms/simplify_insert_with_strategy.cpp
barendgehrels, line 1: It didn't compile, is not used, and was so old that it was not worth to fix it here

doc/src/examples/algorithms/simplify_insert.cpp
```diff
@@ -18,18 +18,12 @@
...
-    line.push_back(P(3.1, 3.1));
-    line.push_back(P(4.9, 1.1));
-    line.push_back(P(3.1, 1.9));
+    using L = boost::geometry::model::linestring<boost::geometry::model::d2::point_xy<double>>;
 
+    L const line{{1.1, 1.1}, {2.5, 2.1}, {3.1, 3.1}, {4.9, 1.1}, {3.1, 1.9}};
     L simplified;
-    boost::geometry::simplify_inserter(line, std::back_inserter(simplified), 0.5);
+    
+    boost::geometry::detail::simplify::simplify_insert(line, std::back_inserter(simplified), 0.5);
```
barendgehrels, line 26: We probably should just delete all examples using `detail`, OR promote it to the general API (such as `assign_box_corners`). But not this one probably.

doc/src/examples/io/svg.cpp
barendgehrels: ✅

vissarion (approved):
Thanks, now all examples (except from problematic ones where comments are inserted) are included and the test are passing.

PR #1314, opened 2024-09-27
[ci] Simplify the documentation CI

vissarion:
Following the suggestion [here](https://github.com/boostorg/geometry/pull/1312#issuecomment-2373918114) this is a simplification for the documentation CI script.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -4,6 +4,10 @@
 # Copyright (c) 2020 Mateusz Loskot <mateusz@loskot.net>
 # Copyright (c) 2020 Adam Wulkiewicz, Lodz, Poland
 #
+# This file was modified by Oracle on 2024.
+# Modifications copyright (c) 2024 Oracle and/or its affiliates.
+# Contributed and/or modified by Vissarion Fysikopoulos, on behalf of Oracle
+#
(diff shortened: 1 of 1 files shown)


.github/workflows/documentation.yml
```diff
@@ -84,22 +88,13 @@ jobs:
...
         cd $BOOST_ROOT
         $BOOST_ROOT/b2 libs/geometry/doc/src/examples
 
-    - name: Run libs/geometry/doc/make_qbk.py
+    - name: Build documentation
       run: |
         cd $BOOST_ROOT
         cd libs/geometry/doc
-        python ./make_qbk.py
+        $BOOST_ROOT/b2
```
barendgehrels, line 100: This is perfect! So we call `b2` now instead of `make_qbk.py`, cool.

Can we then remove the fact that `make_qbk.py` calls `b2`? 
Is everyone following this workflow?
Or do we, if we do that, disturb the making of the official release documentation?

barendgehrels (approved):
👍

.github/workflows/documentation.yml
vissarion: I do not know what scripts/workflow is used to build the official release documentation. @barendgehrels do you know?

barendgehrels: No, I don't know that exactly...

PR #1320, opened 2024-10-02
[test] Add test case for goegrphic intersection for polygons touching North Pole

vissarion:
Adding test case from issue https://github.com/boostorg/geometry/issues/480


test/algorithms/intersects/intersects_sph_geo.cpp
```diff
@@ -79,3 +79,29 @@ int test_main( int , char* [] )
...
+  spherical::polygon bigTile,bigTileOn90,usa;
+  boost::geometry::read_wkt(bigTileWKt, bigTile);
+  boost::geometry::read_wkt(bigTileWKtOn90, bigTileOn90);
+  boost::geometry::read_wkt(usaWkt, usa);
+
+  spherical::multipolygon out1, out2;
+  boost::geometry::intersection(bigTileOn90,usa,out1);
+  boost::geometry::intersection(bigTile,usa,out2);
+  BOOST_CHECK(out1.size() == 1);
+  BOOST_CHECK(boost::geometry::area(out1) == boost::geometry::area(out2));
```
barendgehrels, line 106: Shouldn't we use a floating-point check with epsilon?

test/algorithms/intersects/intersects_sph_geo.cpp
```diff
@@ -79,3 +79,29 @@ int test_main( int , char* [] )
...
+  using point = boost::geometry::model::d2::point_xy<double, boost::geometry::cs::spherical_equatorial<boost::geometry::degree> >;
+  using linestring = boost::geometry::model::linestring<point>;
+  using polygon = boost::geometry::model::polygon<point>;
+  using multipolygon = boost::geometry::model::multi_polygon<polygon>;
+  using mpoint = boost::geometry::model::multi_point<point>;
+}
+
+BOOST_AUTO_TEST_CASE(usaIntersectBigTile90) {
+  auto bigTileWKt     = "POLYGON((-182.25 -2.25,-182.25 85,-90 85,4.5 85,4.5 -2.25,-90 -2.25,-182.25 -2.25))";
+  auto bigTileWKtOn90 = "POLYGON((-182.25 -2.25,-182.25 90,-90 90,4.5 90,4.5 -2.25,-90 -2.25,-182.25 -2.25))";
```
barendgehrels, line 94: I think we never use this camelCase convention...
Also (minor), can be const

barendgehrels (approved):
Minor comments but looks good to me

test/algorithms/intersects/intersects_sph_geo.cpp
vissarion: That should be OK without epsilon because the two computed polygons are exactly the same and those the areas.

test/algorithms/intersects/intersects_sph_geo.cpp
vissarion: Thanks you are right, I was just copy-pasting here from the original issue. Fixed now.

PR #1331, opened 2024-10-13
chore: change point_type to point_type_t and use 'using' at those places

barendgehrels:
This PR
* changes `typename point_type<X>::type` to `point_type_t<X>`
* at those places, use `using` i/o `typedef`
* in lines immediately before / after that, also use `using`
* sometimes also change to `coordinate_type_t` there (but that is not consistently)

I also verified `clang-tidy` with its usage `modernize-use-using` but that is not giving the desired results at this moment.

Clang tidy can, with its option `modernize-use-using`, replace `typedef` with `using. It sometimes works well, but:

* it can replace `typedef differential_quantities<CT, EnableReducedLength, EnableGeodesicScale, 1> quantities` with `using quantities = int` which is plainly wrong
* it can replace `typedef typename boost::range_value<Range const>::type point_t` with `using point_t = typename boost::range_value<const Range>::type` which moves the const to left, and that is not what is in our conventions
* it removes any newline and changes neatly formatted long typedefs into a very long one-lined using clause.

Therefore I decided to **not** use it. So this commit changes to `point_type_t`, plus `using`. A next might change to `coordinate_type_t` plus `using` and then we can change the rest manually, or semi-automatically (still using clang-tidy but judge per change)

### DIFF
--- a/include/boost/geometry/algorithms/discrete_hausdorff_distance.hpp
+++ b/include/boost/geometry/algorithms/discrete_hausdorff_distance.hpp
@@ -59,29 +59,29 @@ namespace detail { namespace discrete_hausdorff_distance
 struct point_range
 {
     template <typename Point, typename Range, typename Strategies>
-    static inline auto apply(Point const& pnt, Range const& rng,
+    static inline auto apply(Point const& pnt, Range const& range,
                              Strategies const& strategies)
     {
         typedef typename distance_result
             <
                 Point,
-                typename point_type<Range>::type,
+                point_type_t<Range>,
                 Strategies
             >::type result_type;
 
         typedef typename boost::range_size<Range>::type size_type;
 
-        boost::geometry::detail::throw_on_empty_input(rng);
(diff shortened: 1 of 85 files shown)


include/boost/geometry/algorithms/discrete_hausdorff_distance.hpp
```diff
@@ -141,27 +141,27 @@ struct range_range
 
 struct range_multi_range
 {
-    template <typename Range, typename Multi_range, typename Strategies>
-    static inline auto apply(Range const& rng, Multi_range const& mrng,
+    template <typename Range, typename MultiRange, typename Strategies>
+    static inline auto apply(Range const& range, MultiRange const& multi_range,
```
barendgehrels, line 145: In this file some additional style improvements are done.

awulkiew:
It looks good. Thanks.

awulkiew (approved):

vissarion (approved):
LGTM, thanks!

PR #1335, opened 2024-10-31
chore: replace typename coordinate_system with coordinate_system_t

barendgehrels:
and add detail meta function coordinate_system_unit

Next PR in series. This one a bit shorter. A few more to come.

### DIFF
--- a/include/boost/geometry/core/coordinate_system.hpp
+++ b/include/boost/geometry/core/coordinate_system.hpp
@@ -96,10 +96,25 @@ struct coordinate_system
         >::type;
 };
 
-
 template <typename Geometry>
 using coordinate_system_t = typename coordinate_system<Geometry>::type;
 
+#ifndef DOXYGEN_NO_DETAIL
+namespace detail {
+
+// Short cut for coordinate system units
+template <typename Geometry>
+struct coordinate_system_units
+{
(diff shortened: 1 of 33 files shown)


include/boost/geometry/core/coordinate_system.hpp
```diff
@@ -96,10 +96,25 @@ struct coordinate_system
...
 
-
 template <typename Geometry>
 using coordinate_system_t = typename coordinate_system<Geometry>::type;
 
+#ifndef DOXYGEN_NO_DETAIL
+namespace detail {
+
+// Short cut for coordinate system units
+template <typename Geometry>
```
vissarion, line 106: I probably missing something here but why not simply 
```
template <typename Geometry>
using coordinate_system_units_t = typename coordinate_system_units<Geometry>::type::units;
```

vissarion (approved):
LGTM, thanks!

include/boost/geometry/core/coordinate_system.hpp
barendgehrels: That was my first version. But `XXX_t` is in general always an alias for the meta function `typename XXX::type` and I did not want to deviate from that convention. So let's keep it like that.

PR #1337, opened 2024-11-05
chore: Fix typo in type alias

vissarion:


barendgehrels (approved):
Oops, sorry. Thanks for the fix!
How come it was not triggering a build failure...

vissarion:
> Oops, sorry. Thanks for the fix! How come it was not triggering a build failure...

In PRs the CI runs only minimal tests https://github.com/boostorg/geometry/blob/develop/.github/workflows/minimal.yml and the whole suite is executed after merging into develop branch.

barendgehrels:
> > Oops, sorry. Thanks for the fix! How come it was not triggering a build failure...
> 
> In PRs the CI runs only minimal tests https://github.com/boostorg/geometry/blob/develop/.github/workflows/minimal.yml and the whole suite is executed after merging into develop branch.

 But I compile all also locally with CMake... however, I see it now. Exactly this one was excluded...
```
if (NOT APPLE)
    # The results of these tests vary considerably on Apple/Darwin/arm64 using clang
    foreach(item IN ITEMS
        envelope_on_spheroid```
...

PR #1338, opened 2024-11-05
chore: replace typedef with using in overlay

barendgehrels:


include/boost/geometry/algorithms/detail/overlay/get_turns.hpp
```diff
@@ -1014,8 +1014,8 @@ inline void get_turns(Geometry1 const& geometry1,
 {
     concepts::check_concepts_and_equal_dimensions<Geometry1 const, Geometry2 const>();
 
-    typedef detail::overlay::get_turn_info<AssignPolicy> TurnPolicy;
-    //typedef detail::get_turns::get_turn_info_type<Geometry1, Geometry2, AssignPolicy> TurnPolicy;
+    using turn_policy_t = detail::overlay::get_turn_info<AssignPolicy>;
+    //using turn_policy_t = detail::get_turns::get_turn_info_type<Geometry1, Geometry2, AssignPolicy>;
```
vissarion, line 1018: Is this alternative policy still needed, if so could you please write a short comment?

barendgehrels: Indeed, not needed, I think. I'll check later, and remove (or comment)

barendgehrels: I investigated - the second (commented) line is the more common and generic way - specializing to the other way anyway for areal/areal.
I will keep the second way.

PR #1346, opened 2024-11-20
fix: add condition to handle_imperfect_touch

### LINKED ISSUE #1288: Difference or intersection fail depending on BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE
Depending on if we set BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE, either difference or intersection fail for some of our test scenarios.

The test case for difference:

MULTIPOLYGON(((-2.0 -1.5, 2.0 -1.5, 2.0 1.5, -2.0 1.5)))
POLYGON((-0.5 -1.49999999, -2.0 -0.1, -1.99999999 -1.5))

difference(MULTIPOLYGON, POLYGON, result)

result should be like MULTIPOLYGON(((-2.0 -0.1, -0.5 -1.49999999, 2.0 -1.5, 2.0 1.5, -2.0 1.5)))

If we do not define BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE, result is empty.
If we define BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE, we get the expected result.

The test case for intersection:

MULTIPOLYGON(((1.0 2.0, 3.0 1.0, 4.0 3.0, 2.0 4.0)))
POLYGON((2.0 1.5, 3.0 1.5, 3.0 2.5, 2.0 2.5))

intersection(MULTIPOLYGON, POLYGON, result)

result should be like MULTIPOLYGON(((2.0 1.5, 3.0 1.5, 3.0 2.5, 2.0 2.5))), which is similar to the input POLYGON.

If we do not define BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE, we get the expected result.
If we define BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE, result contains the bigger input MULTIPOLYGON.

Attached BoostGeometryTests.cpp contains these 2 test cases and some more, which can be enabled by defining ALL_TEST_CASES 1 (in the second line).
[BoostGeometryTests.cpp.txt](https://github.com/user-attachments/files/15955912/BoostGeometryTests.cpp.txt)

With Boost 1.71.0 all test cases succeed (in fact all but Intersection_04c1, where Intersection_04c2 is the same test with slightly adapted expectation - 5 points instead of 4).
With Boost 1.80.0 to 1.85.0 multiple tests fail (different, depending on the BOOST_GEOMETRY_ROBUSTNESS_ALTERNATIVE flag).

It is not important, that we get exactly the declared expected results, but reasonable ones.

### LINKED ISSUE #1345: Error in difference (extracted from #1241)
(extracted from #1241)

> Hi All,
> 
> 
> 
> I believe I hit this same issue and thought I may help by creating a few test cases showing to help:
> 
> 
> 
> ```
> 
> BOOST_AUTO_TEST_CASE(Boost_Diffence_ClearAll)
> 
> {
> 
> 	using PointXY = boost::geometry::model::d2::point_xy<double>;
> 
> 	using Polygon = boost::geometry::model::polygon<PointXY>;
> 
> 	using MultiPolygon = boost::geometry::model::multi_polygon<Polygon>;
> 
> 
> 
> 	MultiPolygon polyA;
> 
> 	read_wkt("MULTIPOLYGON((("
> 
> 	    "-2.0785311235613415 -0.6304193410175202,"
> 
>       "-2.0534946127981359 -0.6304193410175202,"
> 
>       "-2.0534946127981359 -0.8867932112327471,"
> 
>       "-2.3098684830133629 -0.8867932112327471,"
> 
>       "-2.3098684830133629 -0.6554558517807265,"
> 
>       "-2.2848319722501573 -0.6554558517807265,"
> 
>       "-2.0785311235613415 -0.6554558517807265,"
> 
>       "-2.0785311235613415 -0.6304193410175202)))", polyA);
> 
> 
> 
> 	MultiPolygon polyB;
> 
> 	read_wkt("MULTIPOLYGON((("
> 
> 			"-2.0785311235613420 -0.6304193410175202,"
> 
> 			"-2.0534946127981359 -0.6304193410175202,"
> 
> 			"-2.0534946127981359 -0.6554558517807265,"
> 
> 			"-2.0785311235613420 -0.6554558517807265,"
> 
> 			"-2.0785311235613420 -0.6304193410175202)))", polyB);
> 
> 
> 
> 	MultiPolygon polyAB;
> 
> 	//Clip off the small polyB appendix top/left in part polyA
> 
> 	difference(polyA, polyB, polyAB);
> 
> 
> 
> 	//Without an error we should get:
> 
> 	//BOOST_CHECK_CLOSE(area(polyAB), area(polyA) - area(polyB), 0.0001);
> 
> 	//BOOST_CHECK_EQUAL(polyAB.size(), 1u);
> 
> 	//BOOST_CHECK(polyAB[0].inners().empty());
> 
> 
> 
> 	//But we get this:
> 
> 	BOOST_CHECK(polyAB.empty());
> 
> }
> 
> 
> 
> BOOST_AUTO_TEST_CASE(Boost_Diffence_ClearsTooMuch)
> 
> {
> 
> 	using PointXY = boost::geometry::model::d2::point_xy<double>;
> 
> 	using Polygon = boost::geometry::model::polygon<PointXY>;
> 
> 	using MultiPolygon = boost::geometry::model::multi_polygon<Polygon>;
> 
> 
> 
> 	MultiPolygon polyA;
> 
> 	read_wkt("MULTIPOLYGON((("
> 
> 		"-6.1723606999999996 3.2834021000000000,"
> 
> 		"-6.1723606999999996 2.8006724999999992,"
> 
> 		"-5.7133718999999994 2.8006724999999992,"
> 
> 		"-5.7133718999999994 3.2834021000000000,"
> 
> 		"-5.6896310999999997 3.2834021000000000,"
> 
> 		"-5.6896310999999997 2.7769316999999996,"
> 
> 		"-6.1961014999999993 2.7769316999999996,"
> 
> 		"-6.1961014999999993 3.2834021000000000,"
> 
> 		"-6.1723606999999996 3.2834021000000000)))", polyA);
> 
> 
> 
> 	MultiPolygon polyB;
> 
> 	read_wkt("MULTIPOLYGON((("
> 
> 		"-6.1723606999999996 2.8006724999999997,"
> 
> 		"-5.7133718999999994 2.8006724999999997,"
> 
> 		"-5.7133718999999994 2.7769316999999996,"
> 
> 		"-6.1723606999999996 2.7769316999999996,"
> 
> 		"-6.1723606999999996 2.8006724999999997)))", polyB);
> 
> 
> 
> 	MultiPolygon polyAB;
> 
> 	//Clip off the small polyB at the bottom center part of polyA
> 
> 	difference(polyA, polyB, polyAB);
> 
> 
> 
> 	//Without an error we should get:
> 
> 	//BOOST_CHECK_CLOSE(area(polyAB), area(polyA) - area(polyB), 0.0001);
> 
> 	//BOOST_CHECK_EQUAL(polyAB.size(), 2u);
> 
> 	//BOOST_CHECK(polyAB[0].inners().empty());
> 
> 	//BOOST_CHECK(polyAB[1].inners().empty());
> 
> 
> 
> 	//But we get this:
> 
> 	BOOST_CHECK_CLOSE(area(polyAB), 0.5*(area(polyA) - area(polyB)), 0.0001);
> 
> 	BOOST_CHECK(polyAB.size() == 1u);
> 
> 	BOOST_CHECK(polyAB[0].inners().empty());
> 
> }
> 
> ``` 

 _Originally posted by @JohanDore in [#1241](https://github.com/boostorg/geometry/issues/1241#issuecomment-2183964934)_

barendgehrels:
Fixes #1345
**Edit** also fixes #1288 which is added because it was similar.

**About #1345**
Difference was wrong for both cases, caused by nearly opposite segments. The condition causing this had to be fine-tuned with an extra check.

Fixed behaviour:
*a* (`p`=big polygon containing all, `q`=small rectangle at top which subtly overlaps `p`)

Difference a-b (in image, this should be a rectangle, correct)
image

Difference b-a -> 0 (correct)

*b* (`p` = u-shape, `q` = the lower bar of it, subtly overlapping `p`)

Difference `a-b` (in image, this should be the two vertical bars of the u, correct)
Difference `b-a` -> 0 (correct)

image

**About #1288**
It handled a situation as a "touch" but both the middle points were not collinear. In those cases it should not be handled as a touch and use the "original" code path.

The "mystery" why it happened only for polygon and not for multi-polygon, and this affects also the validity check, is not yet solved - but the case itself is solved.
The multi/poly combination is not yet unit tested, I have another test for that in preparation but it's better to not have large chuncks in the beta-time for the release.


include/boost/geometry/algorithms/detail/overlay/get_turn_info_helpers.hpp
```diff
@@ -66,12 +73,6 @@ struct side_calculator
     inline int pk_wrt_q2() const { return m_side_strategy.apply(get_qj(), get_qk(), get_pk()); }
     inline int qk_wrt_p2() const { return m_side_strategy.apply(get_pj(), get_pk(), get_qk()); }
 
-    // Necessary when rescaling turns off:
```
barendgehrels, line 69: Moved above to order `i, j, k`

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -674,6 +682,7 @@ struct touch : public base_turn_handler
             {
                 if (side_qk_p1 == 0 && side_pk_q1 == 0
                     && has_pk && has_qk
+                    && opposite(side.pi_wrt_q1(), side.qk_wrt_p2())
```
barendgehrels, line 685: This is the fix

test/algorithms/overlay/overlay_cases.hpp
```diff
@@ -1231,6 +1231,59 @@ static std::string issue_1342_b[2] =
     "POLYGON((2.000000861538488 2.944711629068049e-16,2.000000861538488 2.0000008,23.52786548447633 2.0000008,23.52786548447633 2.944711629068049e-16,2.000000861538488 2.944711629068049e-16))"
 };
 
+static std::string issue_1345_a[2] =
+{
+    // Needs check for opposite in handle_imperfect_touch
+    R""""(
```
barendgehrels, line 1237: @vissarion this is (AFAIK) the first case written in this new multi-line format.
It doesn't have all those quotes and reflects the reported case.
Like it?

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -354,6 +356,17 @@ struct touch_interior : public base_turn_handler
...
         else // else prevents unreachable code warning
         {
+            if (same(side.pj_wrt_q1(), side.qj_wrt_p2())
+                || same(side.pj_wrt_q2(), side.qj_wrt_p1()))
+            {
+                // At a touch, the touching points (pj and qj) should be collinear
+                // with both other segments.
+                // If that is not the case (both left or both right), it should not be handled as a touch,
+                // (though the intersection point might be close to the end),
+                // because segments might cross each other or touch the other in the middle.
```
barendgehrels, line 366: This is the fix for #1288

test/algorithms/overlay/overlay_cases.hpp
vissarion: Looks good, thanks.

vissarion (approved):
Thanks!

Issue #1299, opened 2024-07-29
Difference does not work reliable for double precision


MauriceHubain:
This is a follow up ticket for  https://github.com/boostorg/geometry/issues/1288.

Adapting your test code from that ticket with a new data set, we get wrong results for double and long double precision:

```
#include <boost/geometry.hpp>
#include <boost/geometry/geometries/geometries.hpp>
#include <iostream>
#include <iomanip>

namespace bg = boost::geometry;

template <typename T>
void test_Difference_B83673_1_MP()
{
  using point_t = bg::model::point<T, 2, bg::cs::cartesian>;
  using polygon_t = bg::model::polygon<point_t, false, false>;
  using mp_t = bg::model::multi_polygon<polygon_t>;

  mp_t poly1;
  bg::read_wkt("MULTIPOLYGON(((1.2549999979079400 0.85000000411847698, -1.2550000020920500 0.84999999897038103, -1.2549999999999999 -0.85000000102961903, 1.2549999999999999 -0.84999999999999998)))", poly1);
  bg::correct(poly1);

  mp_t poly2;
  bg::read_wkt("MULTIPOLYGON(((-0.87500000000000000 -0.84999999999999998, -0.87500000000000000 -0.070000000000000201, -1.2549999999999999 -0.070000000000000201, -1.2549999999999999 -0.84999999999999998)))", poly2);
  bg::correct(poly2);

  mp_t diff, inters;
  bg::difference(poly1, poly2, diff);
  bg::intersection(poly1, poly2, inters);

  std::cout << std::setprecision(12) << std::boolalpha;
  std::cout
    << "Type: " << typeid(T).name()
    << " a: " << bg::area(poly1)
    << " b: " << bg::area(poly2)
    << " diff: " << bg::area(diff)
    << " ints: " << bg::area(inters)
    << " a-b: " << bg::area(poly1) - bg::area(poly2)
    << " d+i " << bg::area(diff) + bg::area(inters)
    << " validity diff: " << bg::is_valid(diff)
    << " ints: " << bg::is_valid(inters)
    << std::endl;
}

template <typename T>
void test_Difference_B83673_1_P()
{
  using point_t = bg::model::point<T, 2, bg::cs::cartesian>;
  using polygon_t = bg::model::polygon<point_t, false, false>;
  using mp_t = bg::model::multi_polygon<polygon_t>;

  mp_t poly1;
[...]

PR #1347, opened 2024-11-24
Fix/issue 893 and 1299

### LINKED ISSUE #893: Multipolygon difference computes wrong result
I compute intersection and difference of multipolygons p and q. The intersection is not empty but the difference is empty. This seems to be wrong.

Here is the code (also on godbolt https://godbolt.org/z/11fK95Wxj)

```cpp
#define BOOST_GEOMETRY_NO_ROBUSTNESS

#include <boost/geometry.hpp>
#include <iostream>
#include <string>

using Point = boost::geometry::model::d2::point_xy<double>;
using Polygon = boost::geometry::model::polygon<Point>;
using MultiPolygon = boost::geometry::model::multi_polygon<Polygon>;

struct Info {
  double area;
  bool isValid;
  std::string failureMsg;
  std::string name;
};

Info computeInfo(const MultiPolygon& multipolygon, const std::string& name) {
  std::string failureMsg;
  const auto isValid = bg::is_valid(multipolygon, failureMsg);
  const auto area = bg::area(multipolygon);

  return Info{area, isValid, failureMsg, name};
}

std::ostream& operator<<(std::ostream& os, const Info& info) {
  os << info.name << '\n';
  os << "area: " << info.area << '\n';
  os << "isValid: " << std::boolalpha << info.isValid << '\n';
  os << "failureMsg: " << info.failureMsg << '\n';

  return os;
}

int main() {
  MultiPolygon p{{{
      {-9133.3885344331684, 3976.3162451998137},
      {-6748.2449169873034, -5735.0734557728138},
      {12359.886942916415, -1042.0645095456412},
      {5126.3084924076147, 2226.9708710750697},
      {-1604.5619839035633, 573.85084904357439},
      {-9133.3885344331684, 3976.3162451998137},
  }}};

  MultiPolygon q{{{
      {-3228.4265340177531, 1307.7159344890201},
      {-4500.2645033380131, 1882.4913860267370},
      {-4294.7752070657516, 1045.8178117890784},
      {-3228.4265340177531, 1307.7159344890201},
  }}};

  bg::correct(p);
  std::cout << computeInfo(p, "p") << '\n';

  bg::correct(q);
  std::cout << computeInfo(q, "q") << '\n';

  std::cout << "p overlaps q: " << std::boolalpha << bg::overlaps(p, q) << '\n';
  std::cout << "q whithin p: " << std::boolalpha << bg::within(q, p) << '\n';
  std::cout << "p touches q: " << std::boolalpha << bg::touches(p, q) << '\n';
  std::cout << "q touches p: " << std::boolalpha << bg::touches(q, p) << '\n';
  // std::cout << "relation: " << bg::relation(q, p).str() << '\n';
  std::cout << '\n';

  MultiPolygon intersection;

  bg::intersection(p, q, intersection);
  std::cout << computeInfo(intersection, "intersection") << '\n';

  MultiPolygon difference;

  bg::difference(p, q, difference);
  std::cout << computeInfo(difference, "difference") << '\n';
}
```

Output:
```
p
area: 9.76869e+07
isValid: true
failureMsg: Geometry is valid

q
area: 473002
isValid: true
failureMsg: Geometry is valid

p overlaps q: true
q whithin p: false
p touches q: false
q touches p: false

intersection
area: 473002
isValid: true
failureMsg: Geometry is valid

difference
area: 0
isValid: true
failureMsg: Geometry is valid
```

multipolgons-p-q-intersection-difference

On the image it seems p is overlapping q and that they are touching. Maybe this is problematic. How should one handle those cases?

From the image it seems to be clear that the difference should be not empty.

barendgehrels:
Fixes #893 and #1299 which had the same cause.

This had to be tweaked for non-rescaling, there were some conditions not handled.


include/boost/geometry/algorithms/detail/overlay/overlay.hpp
```diff
@@ -102,38 +102,42 @@ inline void get_ring_turn_info(TurnInfoMap& turn_info_map, Turns const& turns, C
...
+            auto const& op = turn.operations[i];
+            auto const& other_op = turn.operations[1 - i];
             ring_identifier const ring_id = ring_id_by_seg_id(op.seg_id);
 
-            if (! is_self_turn<OverlayType>(turn)
-                && (
-                    (BOOST_GEOMETRY_CONDITION(target_operation == operation_union)
-                      && op.enriched.count_left > 0)
-                  || (BOOST_GEOMETRY_CONDITION(target_operation == operation_intersection)
-                      && op.enriched.count_right <= 2)))
```
barendgehrels, line 126: I'm glad this condition could be removed. These counts can have issues with floating point precision, because they count open and close polygons on either sides, and open/close can be ambiguous for cases where in/out segments are very close.

include/boost/geometry/algorithms/detail/overlay/overlay.hpp
```diff
@@ -102,38 +102,42 @@ inline void get_ring_turn_info(TurnInfoMap& turn_info_map, Turns const& turns, C
...
+            // If the turn (one of its operations) is used during traversal,
+            // and it is an intersection or difference, it cannot be set to blocked.
+            // This is a rare case, related to floating point precision,
+            // and can happen if there is, for example, only one start turn which is
+            // used to traverse through one of the rings (the other should be marked
+            // as not traversed, but neither blocked).
+            bool const can_block
+                = is_union
```
barendgehrels, line 131: I can refactor it more - because it basically means that all union (original) rings are blocked, so not included in the result - which makes sense for union (completely separate rings are handled differently).
Also the colocation check can be enhanced. I have a draft of these changes locally.
But for this PR (maybe for 1.87) I wanted to keep the changes as small as possible.

test/algorithms/set_operations/difference/difference_multi.cpp
```diff
@@ -152,19 +152,7 @@ void test_areal()
...
-        // 3: with rescaling A is considered as invalid (robustness problem)
-        ut_settings settings;
-        settings.validity_of_sym = true;
-        settings.validity_false_negative_sym = true;
-        TEST_DIFFERENCE_WITH(0, 1, bug_21155501,
-                             (count_set(1, 4)), expectation_limits(3.75893, 3.75894),
-                             (count_set(1, 4)), (expectation_limits(1.776357e-15, 7.661281e-15)),
-                             (count_set(2, 5)));
-    }
+    TEST_DIFFERENCE(bug_21155501, 1, 3.758937, 1, 1.78e-15, 1);
```
barendgehrels, line 155: This test was the only existing test affected by one of the conditions. Behaviour is better now, AFAICS

test/algorithms/overlay/multi_overlay_cases.hpp
```diff
@@ -1586,6 +1586,12 @@ static std::string bug_21155501[2] =
         "MULTIPOLYGON(((4.9658203125 18.729501999072138,-3.4868710311820115 24.246968623627644,8.3589904332912 33.833614418115445,8.3056640625 32.99023555965106,9.8876953125 31.728167146023935,9.7119140625 25.48295117535531,11.8212890625 23.563987128451217,4.9658203125 18.729501999072138)),((-3.88714525609152 24.508246314579743,-8.3935546875 27.449790329784214,-8.9208984375 29.458731185355344,-0.5712890625 32.02670629333614,-1.8896484375 35.60371874069731,8.5693359375 37.16031654673677,8.362166569827938 33.883846345901595,-3.88714525609152 24.508246314579743)))",
     };
 
+static std::string issue_1299[2] =
+{
+    "MULTIPOLYGON(((1.2549999979079400 0.85000000411847698, -1.2550000020920500 0.84999999897038103, -1.2549999999999999 -0.85000000102961903, 1.2549999999999999 -0.84999999999999998)))",
+    "MULTIPOLYGON(((-0.87500000000000000 -0.84999999999999998, -0.87500000000000000 -0.070000000000000201, -1.2549999999999999 -0.070000000000000201, -1.2549999999999999 -0.84999999999999998)))"
+};
```
barendgehrels, line 1593: This is on purpose put below, to avoid merge conflicts with #1346 
Same in the unit tests processing it.

Later (`1.48`) I make a PR changing this test approach and the `issue_` cases will be together again.

vissarion (approved):
LGTM, thanks!

Issue #1349, opened 2024-11-27
Difference of gpolygons giving wrong result


gastald88:
I have two polygons, p1 (orange) and p2 (blue) as in figure.
The difference of p1 and p2 returns a (wrong) empty result.
![Image](https://github.com/user-attachments/assets/d3a29178-a800-4c90-83df-b632a53cbba1)

This can be replicated in the following godbolt link:
https://godbolt.org/z/GT7aWaWzW

barendgehrels:
Detailed note (mainly for myself).

It's caused by the fix of #869 , a condition for difference added by me in 2021.

I have to revise that part of the code. Skipping the `reverseMeaningInTurn`  is fine for all test cases, except `issue_869` which has correct areas, but is considered as invalid.

In the `traversal_switch_detector.hpp` it should not be necessary change behavior for difference.

barendgehrels:
PR fixing this issue is published

PR #1350, opened 2024-12-04
fix: revise traversal_switch_detector for isolated areas

barendgehrels:
**Not ready for review**

~Fixes: issue 1349~

This is quite a revisal. Issue #1349 was caused by a logic error, and not by a precision error. It was easily fixed by removing the (in hindsight weird) condition (added by me) to reverse the meaning of isolated.

But that caused the original issue again. I made an inverse case of that one as well, and that failed anyway (also with its earlier "fix") for difference.

~I decided to go for another way to detect isolation, which is now clearer - but even then not that easy. Below some cases for reference.~

Below some pictures of intersections (blue, green: input, turquoise: both (= intersection), dotted red/green: resulting intersection.

Case 869 (only ii turns)
image

Case 128 (with a separate polygon connected as "multiple" with only ii turns (turn 6,9,18), it should not be isolated)

image

Case 141 (idem but it should be isolated and the "multiple" region inside (turn 13,14) is a separate polygon)

image

 Case case_recursive_boxes_81 which looks the same, but here it are two outer polygons with a part (at turn 26, 27, 42)

image


tinko92:
I think, this creates a regression. Consider

```cpp
#include <iostream>
#include <boost/geometry.hpp>

using point = bg::model::point<int, 2, bg::cs::cartesian>;
using multipoly = bg::model::multi_polygon<polygon>;

int main() {
    auto p1 = bg::from_wkt<multipoly>("MULTIPOLYGON(((2 10,2 8,0 8,0 10,2 10)),((10 8,10 2,8 2,8 0,0 0,0 4,2 4,2 8,4 8,4 10,6 10,6 8,6 6,8 6,8 8,10 8),(8 2,8 4,4 4,4 2,8 2)))");
    auto p2 = bg::from_wkt<multipoly>("MULTIPOLYGON(((2 6,2 4,2 2,2 0,0 0,0 6,2 6)),((2 10,2 8,0 8,0 10,2 10)),((6 8,6 6,2 6,2 8,6 8)),((8 4,8 2,6 2,6 4,8 4)),((10 8,10 6,8 6,8 8,10 8)),((8 10,8 8,6 8,6 10,8 10)))");
    multipoly p3;
    bg::sym_difference(p1, p2, p3);
    std::cout << bg::wkt(p3) << "\n";
    return 0;
}
```

Output on develop: MULTIPOLYGON(((10 6,10 2,8 2,8 0,2 0,2 4,0 4,0 6,10 6),(6 2,6 4,4 4,4 2,6 2)),((4 8,4 10,6 10,8 10,8 8,4 8)))
Output on this branch: MULTIPOLYGON(((10 6,10 2,8 2,8 0,2 0,2 4,0 4,0 6,10 6)),((4 8,4 10,6 10,8 10,8 8,4 8)))

barendgehrels:
> > I think, this creates a regression. Consider
> 
> 👍 I'll check

Indeed it fails now. I have to revise it.

barendgehrels:
The region information for #869 is incorrect - there is a fix that fixes this all but leads to yet another situation which goes wrong.
I have to revise the code, probably getting rid of the isolation concept.

barendgehrels:
@tinko92 we (@vissarion and I) added you as a collaborator, now I can add you as a reviewer, great.

For the rest this PR is still in draft and might be closed by a replacement later this month.

PR #1353, opened 2024-12-07
test: verify failures and remove some rescaling remainings

barendgehrels:
This PR:

* removes `BoostGeometryWriteExpectedFailures` (it was to compare rescaling with / without)
* verifies failing tests
* where not failing anymore, removes the defines or conditions

There were several tests which now succeeds. I didn't bisect them, but it is good to have them enabled now for future PR's.

### DIFF
--- a/test/algorithms/set_operations/difference/difference.cpp
+++ b/test/algorithms/set_operations/difference/difference.cpp
@@ -34,15 +34,6 @@
 namespace
 {
 
-// Change compiler defines to constexpr bools
-// to make conditions more readable
-// and to always compile all code.
-#if defined(BOOST_GEOMETRY_TEST_FAILURES)
-constexpr bool test_failures = true;
-#else
-constexpr bool test_failures = false;
-#endif
-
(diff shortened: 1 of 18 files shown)


test/algorithms/set_operations/difference/difference.cpp
```diff
@@ -34,15 +34,6 @@
 namespace
 {
 
-// Change compiler defines to constexpr bools
-// to make conditions more readable
-// and to always compile all code.
```
barendgehrels, line 39: This was only done, once now, in this source, so I harmonized it back with the others.

test/algorithms/set_operations/difference/difference.cpp
```diff
@@ -330,23 +298,25 @@ void test_all()
             count_set(1, 2), settings);
     }
 
-    /*** TODO: self-tangencies for difference
```
barendgehrels, line 333: This TODO was really old and the tests succeed now.
The 0,1 output is invalid because the input is invalid, and therefore it can be ignored.

vissarion (approved):
Thanks!

PR #1356, opened 2024-12-15
Fix/issue1349 second

barendgehrels:
This is an updated version of #1350 

See also the pictures there.

It should be build on top of #1355 In the past that was easy to configure, now I don't see it anymore.

Summary of fixes:
* #1349 
* #1221 
* #1350 (the comment made there) is kept valid (unlike using that PR)

~There might be a follow-up, but that is far from ready so I think we should go first for this one.~
I'm busy with the follow-up which is now in a promising state. Will close this one.

PR #1361, opened 2025-01-06
Test set operations on random grids with integer coordinates

tinko92:
As discussed in #1356 , this adds an overlay test from https://gist.github.com/tinko92/fad450e1058ce8ceebde66a6b86d45c1 that generates grids based on bit patterns and verifies the correctness of set operation by testing against the corresponding logical operation on the bit sets.

The test is cleaned up and generalized to take width and height options from the command line. It is not intended to be run by default as part of the test suite (I understand, this is not the case for the robustness tests in general?) because it is only useful with a large iteration count and thereby long-running. It is also meant to be built with optimizations to mitigate the high time cost and prints failures to stdout in WKT with the corresponding bitset rather than triggering assertions.

The design idea was to build a test that, in principle, could test every possible code path in overlay (for sufficiently large width and height). I lack the insight in the overlay details to judge whether it meets that goal. It is also meant to test only for non-numerical issues, so it intentionally only supports integer coordinates. The limitation to a rectangular grid is just for simplicity of implementation, but in principle every indexable, disjoint set of rings with corresponding test points should work.


barendgehrels:
Cool! I'll review it Wednesday

test/robustness/overlay/areal_areal/Jamfile
```diff
@@ -25,4 +25,4 @@ exe recursive_polygons : recursive_polygons.cpp ;
 exe star_comb : star_comb.cpp ;
 
 exe ticket_9081 : ticket_9081.cpp ;
-
+exe random_integer_grids : random_integer_grids.cpp ;
```
barendgehrels, line 28: Can you added it to the `CMakeLists.txt` too?

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+struct grid_settings
+{
+    int width              = 5;
+    int height             = 5;
+    int count              = 1;
+    prng::result_type seed = prng::default_seed;
+};
+
+constexpr int cell_width = 2;
```
barendgehrels, line 50: This is used for both width and height, which is confusing.
Can you either make two separate constants, or call it `cell_dimension` for example?

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,264 @@
...
+    for (size_t i = 0; i < grid.size(); ++i)
+    {
+        if( b[i / chunk_size][i % chunk_size] )
+        {
+            mp temp;
+            bg::union_(out, grid[i], temp);
+            out = temp;
+        }
+    }
+    const auto b_gs = [&settings](bits const& b) { return std::make_pair(b, settings); };
```
barendgehrels, line 117: what is `b_gs`?

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,264 @@
...
+template <typename BitOp>
+bits apply_for_each(bits a, bits const& b, BitOp bit_op)
+{
+    for (size_t i = 0; i < a.size(); ++i) a[i] = bit_op(a[i], b[i]);
+    return a;
+}
+
+template<typename BitOp, typename GeoOp>
+void test_op(bits const& bits1, bits const& bits2, mp const& geo1, mp const& geo2,
+             std::string op_label, BitOp bit_op, GeoOp geo_op, std::vector<point> const& points,
```
barendgehrels, line 157: op_label should be `const&` probably

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,267 @@
...
+                [](mp const& g1, mp const& g2) { mp g; bg::sym_difference(g1, g2, g); return g; },
+                points, grid, settings, all_success);
+        test_op(bits1, bits2, geo1, geo2, "difference g1 \\ g2", 
+                [](std::bitset<chunk_size> b1, std::bitset<chunk_size> b2) { return b1 & (~b2); },
+                [](mp const& g1, mp const& g2) { mp g; bg::difference(g1, g2, g); return g; },
+                points, grid, settings, all_success);
+        test_op(bits1, bits2, geo1, geo2, "difference g2 \\ g1", 
+                [](std::bitset<chunk_size> b1, std::bitset<chunk_size> b2) { return b2 & (~b1); },
+                [](mp const& g1, mp const& g2) { mp g; bg::difference(g2, g1, g); return g; },
+                points, grid, settings, all_success);
```
barendgehrels, line 214: This (the lines above) looks very good, great idea to test it like this.

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,267 @@
...
+#include <geometry_test_common.hpp>
+
+constexpr int chunk_size = 64;
+using bits = std::vector<std::bitset<chunk_size>>;
+using prng = std::mt19937_64;
+static_assert(sizeof(prng::result_type) >= chunk_size / 8, "PRNG output too narrow.");
+
+namespace bg = boost::geometry;
+
+using point = bg::model::d2::point_xy<int>;
```
barendgehrels, line 39: You probably mean here: `<BOOST_GEOMETRY_DEFAULT_TEST_TYPE>`, though it's a bit long.
It's actually a pity that it can only test integers (though you gave the rationale). I would like to have a version for `double` as well. But we can also change it later.

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,267 @@
...
+        {
+            int index = y * b_gs.second.width + x;
+            os << b_gs.first[index / chunk_size][index % chunk_size];
+        }
+        os << '\n';
+    }
+    return os;
+}
+
+bits geometry_to_bits(mp const& g, std::vector<point> const& tp)
```
barendgehrels, line 96: The names are short and sometimes cryptic.
I don't say I never use abbreviations myself (I do). But these ones I don't fully get all.
`tp`, what is it?
`b` is probably fine in this context.
`g` is short (but at least I understand it), I would use `geometry` here or make the `mp` a `multi_polygon_t` or `mp_t` and call it `mp` then here
`bm` (below)
`prng` and `gen`, I would use longer names

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,267 @@
...
+template<typename BitOp, typename GeoOp>
+void test_op(bits const& bits1, bits const& bits2, mp const& geo1, mp const& geo2,
+             std::string const& op_label, BitOp const& bit_op, GeoOp const& geo_op,
+             std::vector<point> const& points, std::vector<box> const& grid,
+             grid_settings const& settings, bool& success)
+{
+    auto test_geo = geo_op(geo1, geo2);
+    // Convenience lambda to pair bits with settings to use width/height in operator<<(os, ...)
+    const auto b_gs = [&settings](bits const& b) { return std::make_pair(b, settings); };
+    if ( ! bg::is_valid(test_geo) )
```
barendgehrels, line 166: You can get the reason (for invalidity) and output it, it is useful in these kind of programs

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,267 @@
...
+    {
+        std::cout << op_label << "(\n\t " << bg::wkt(geo1) << ",\n\t " << bg::wkt(geo2) << "\n), "
+                  << "generated from" << b_gs(bits1) << "and" << b_gs(bits2) << "is invalid.\n\n";
```
barendgehrels, line 169: Looks like some spaces are missing, do you have a sample output?

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,267 @@
...
+        success = false;
+    }
+    bits expected = apply_for_each(bits1, bits2, bit_op);
+    bits obtained = geometry_to_bits(test_geo, points);
```
barendgehrels, line 173: These can be const (not really important here, but anyway)

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,267 @@
...
+using point = bg::model::d2::point_xy<int>;
+using box   = bg::model::box<point>;
+using mp    = bg::model::multi_polygon<bg::model::polygon<point>>;
+
+struct grid_settings
+{
+    int width              = 5;
+    int height             = 5;
+    int count              = 1;
+    prng::result_type seed = prng::default_seed;
```
barendgehrels, line 48: Can we generate it somehow from current time?

test/robustness/overlay/areal_areal/random_integer_grids.cpp
barendgehrels: I get them now, it's fine because the `b_gs` lambda generates new lines

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,267 @@
...
+             po::value<decltype(settings.seed)>(&settings.seed)->default_value(settings.seed),
+             "Initialization seed for random generator")
+            ("count",
+             po::value<decltype(settings.count)>(&settings.count)->default_value(settings.count),
+             "Number of tests (-1 for infinite loop)")
+            ("width",
+             po::value<decltype(settings.width)>(&settings.width)->default_value(settings.width),
+             "Width of grid")
+            ("height",
+             po::value<decltype(settings.height)>(&settings.height)->default_value(settings.height),
```
barendgehrels, line 241: The decltype is elegant. But because it's an `int`, if I specify `-10` for height, I get a segmentation fault...

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,267 @@
...
+        po::variables_map varmap;
+        po::store(po::parse_command_line(argc, argv, description), varmap);
+        po::notify(varmap);
+
+        if (varmap.count("help"))
+        {
+            std::cout << description << std::endl;
+            return 1;
+        }
+        if ( ! test_all(settings) )
```
barendgehrels, line 254: About style.
We put conditions normally in curly braces.
It's allowed without (IIRC) if you make it one line, so `if (! test_all(settings)) return 1;`

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,267 @@
...
+    int height             = 5;
+    int count              = 1;
+    prng::result_type seed = prng::default_seed;
+};
+
+constexpr int cell_dimension = 2;
+
+std::vector<box> grid_cells(grid_settings const& settings)
+{
+    std::vector<box> out;
```
barendgehrels, line 55: You could reserve width*height, here and below - it's probably a bit faster when testing a lot. Though the bg operations might overshadow the gain, of course.

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,267 @@
...
+            return 1;
+    }
+    catch(std::exception const& e)
+    {
+        std::cout << "Exception " << e.what() << '\n';
+    }
+    catch(...)
+    {
+        std::cout << "Other exception" << '\n';
+    }
```
barendgehrels, line 264: I would like to see a summary of the number of tests, the number of failures, and the time spent.

barendgehrels:
> It is not intended to be run by default as part of the test suite (I understand, this is not the case for the robustness tests in general?) because it is only useful with a large iteration count and thereby long-running. 

This is completely fine, indeed the robustness tests are to be executed manually. I do it regularly (not all of them, but some).

I'll continue tonight or tomorrow - I had some remarks but basically it looks good, I'm glad with the addition.

tinko92:
The second commit was intentionally not squished to delineate the new changes from the already reviewed ones but I can rebase and force push if a single commit PR is prefered.

barendgehrels:
> The second commit was intentionally not squished to delineate the new changes from the already reviewed ones but I can rebase and force push if a single commit PR is prefered.

I usually look at `compare` so for me it's fine to squash it. But this is also fine.

test/robustness/overlay/areal_areal/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ foreach(item IN ITEMS
     recursive_polygons
     star_comb
     #ticket_9081
+    #random_integer_grid
```
barendgehrels, line 18: It should be `random_integer_grids` (with `s`) and not commented.

BTW, about the name, it's now really targeting integer. Suggestion: `random_bitset_grids` instead.

test/robustness/overlay/areal_areal/random_integer_grids.cpp
barendgehrels: Finally got it, `bit_gridsettings` probably.

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,381 @@
...
+                                : to_bits(settings.bits2);
+        test_bits(bits1, bits2, grid, points, settings, failures);
+    }
+    auto const t = std::chrono::high_resolution_clock::now();
+    auto const elapsed_ms = std::chrono::duration_cast<std::chrono::milliseconds>(t - t0).count();
+    int failure_count = std::accumulate(failures.begin(), failures.end(), 0,
+                                        [](int acc, auto const& kv) { return acc + kv.second; });
+    std::cout << "\niterations: " << settings.count
+              << " errors: " << failure_count
+              << " time: " << elapsed_ms/1000 << '\n';
```
barendgehrels, line 282: suggestion: `elapsed_ms / 1000.0` to avoid `0`

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,381 @@
...
+    if (failure_count != 0)
+    {
+        std::cout << "Failure counts by failure mode:\n";
+        for(auto const& fm : failures) std::cout << '\t' << fm.first << ": " << fm.second << '\n';
```
barendgehrels, line 286: Cool addition, thanks!

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,381 @@
...
+        {
+            std::cout << "Both bit patterns fixed, count is changed to 1.\n";
+            settings.count = 1;
+        }
+        if(settings.height < 1 || settings.width < 1)
+        {
+            std::cout << "Invalid dimensions, height and width need to be positive.\n";
+            return 1;
+        }
+        if (varmap.count("help"))
```
barendgehrels, line 364: Nit, the layout of the `if` statements has three variants.

I would prefer this one, so: `if` `space` `(` condition `)`, which I think is the most common in our repo (though it is not always consistent)

test/robustness/overlay/areal_areal/random_integer_grids.cpp
```diff
@@ -0,0 +1,381 @@
...
+            return 1;
+        }
+        if ( settings.bits1.size() != 0 && settings.bits2.size() != 0 && settings.count != 1 )
+        {
+            std::cout << "Both bit patterns fixed, count is changed to 1.\n";
+            settings.count = 1;
+        }
+        if(settings.height < 1 || settings.width < 1)
+        {
+            std::cout << "Invalid dimensions, height and width need to be positive.\n";
```
barendgehrels, line 361: Thanks for handling my comments! I'm OK with this PR and will approve,
but I prefer is @vissarion has a look too.

vissarion (approved):
Thanks Tinko, LGTM.

test/robustness/overlay/areal_areal/random_bitset_grids.cpp
```diff
@@ -0,0 +1,381 @@
...
+            return 1;
+        }
+        if (varmap.count("help"))
+        {
+            std::cout << description << std::endl;
+            return 1;
+        }
+        if (! test_all(settings)) return 1;
+    }
+    catch(std::exception const& e)
```
vissarion, line 371: please also add a space after `catch` here

test/robustness/overlay/areal_areal/random_bitset_grids.cpp
```diff
@@ -0,0 +1,381 @@
...
+    {
+        std::cout << "Exception " << e.what() << '\n';
+    }
+    catch(...)
```
vissarion, line 375: ..and here

vissarion (approved):

tinko92:
Thanks for the comments, @vissarion (I think some of the early ones maybe overlapped with my previous push) and @barendgehrels !

barendgehrels (approved):
👍

barendgehrels:
Can we merge this?

tinko92:
> Can we merge this?

I checked all conversations and resolved every point as addressed. Should I merge it or is that for someone with a boostorg member role to do?

PR #1364, opened 2025-01-10
Fix several strategies (add getters, fix compilation error and warning)

awulkiew:
The getters are missing for `centroid` and `closest_points` in `index` and global umbrella strategies preventing them to be passed to these algorithms and rtree.

I divided centroid into detail part and official part derived from base. It works for now because centroid doesn't have proper strategies for spherical and geographic. After they are added and e.g. geographic one uses `Formula` and `Spheroid` we'll probably have to implement this differently.

### DIFF
--- a/include/boost/geometry/strategies/geographic/distance_cross_track.hpp
+++ b/include/boost/geometry/strategies/geographic/distance_cross_track.hpp
@@ -1,6 +1,6 @@
 // Boost.Geometry (aka GGL, Generic Geometry Library)
 
-// Copyright (c) 2023 Adam Wulkiewicz, Lodz, Poland.
+// Copyright (c) 2023-2025 Adam Wulkiewicz, Lodz, Poland.
 
 // Copyright (c) 2016-2022, Oracle and/or its affiliates.
 // Contributed and/or modified by Vissarion Fysikopoulos, on behalf of Oracle
@@ -697,10 +697,13 @@ protected:
         point p2 = point(lon2, lat2);
         point p3 = point(lon3, lat3);
 
-        geometry::strategy::distance::cross_track<CT> cross_track(earth_radius);
+        using haversine_t = geometry::strategy::distance::haversine<CT>;
+        using cross_track_t = geometry::strategy::distance::cross_track<void, haversine_t>;
+
+        cross_track_t cross_track(earth_radius);
         CT s34_sph = cross_track.apply(p3, p1, p2);
 
-        geometry::strategy::distance::haversine<CT> str(earth_radius);
+        haversine_t str(earth_radius);
         CT s13_sph = str.apply(p1, p3);
 
         //CT s14 = acos( cos(s13/earth_radius) / cos(s34/earth_radius) ) * earth_radius;
--- a/include/boost/geometry/strategies/geographic/distance_cross_track.hpp
+++ b/include/boost/geometry/strategies/geographic/distance_cross_track.hpp
@@ -273,7 +273,7 @@ private:
                 pl_lat = res14.lat2;
             }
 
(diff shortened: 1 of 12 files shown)


include/boost/geometry/strategies/spherical/distance_cross_track_box_box.hpp
```diff
@@ -45,11 +47,11 @@ class cross_track_box_box_generic
...
-    template <typename Point, typename PPStrategy, typename PSStrategy>
-    ReturnType static inline diagonal_case(Point topA,
-                                           Point topB,
-                                           Point bottomA,
-                                           Point bottomB,
+    template <typename Point1, typename Point2, typename PPStrategy, typename PSStrategy>
+    ReturnType static inline diagonal_case(Point1 topA,
+                                           Point2 topB,
+                                           Point1 bottomA,
+                                           Point2 bottomB,
```
barendgehrels, line 54: Thanks, good to see a PR from your side!

So the point types can be different, I noticed. That's fine.
But they can be made `Point1 const&` (it was already before this PR - but now it's touched...), and the strategies too.
Also unusual is that `ReturnType` is specified before `static inline` but that was also already the case.

Anyway, I'me fine with the changes.

barendgehrels (approved):

include/boost/geometry/strategies/geographic/distance_cross_track.hpp
```diff
@@ -697,10 +697,13 @@ class geographic_cross_track
...
 
-        geometry::strategy::distance::cross_track<CT> cross_track(earth_radius);
+        using haversine_t = geometry::strategy::distance::haversine<CT>;
+        using cross_track_t = geometry::strategy::distance::cross_track<void, haversine_t>;
+
+        cross_track_t cross_track(earth_radius);
         CT s34_sph = cross_track.apply(p3, p1, p2);
 
-        geometry::strategy::distance::haversine<CT> str(earth_radius);
+        haversine_t str(earth_radius);
```
barendgehrels, line 706: They probably can be made `const`. But also that was already before.

awulkiew: Right, I automatically changed the types. I added const now, also in other parts of the file.

include/boost/geometry/strategies/spherical/distance_cross_track_box_box.hpp
awulkiew: Done.

awulkiew:
I added one more fix in geo distance_cross_track_box_box strategy.

include/boost/geometry/strategies/spherical/distance_cross_track_box_box.hpp
```diff
@@ -45,15 +47,15 @@ class cross_track_box_box_generic
...
-    template <typename Point, typename PPStrategy, typename PSStrategy>
-    ReturnType static inline diagonal_case(Point topA,
-                                           Point topB,
-                                           Point bottomA,
-                                           Point bottomB,
+    template <typename Point1, typename Point2, typename PPStrategy, typename PSStrategy>
+    static inline ReturnType diagonal_case(Point1 const& topA,
+                                           Point2 const& topB,
+                                           Point1 const& bottomA,
+                                           Point2 const& bottomB,
```
barendgehrels, line 54: Thanks @awulkiew !

awulkiew:
I missed that and used the old pattern. I altered the commit messages.

It would probably be useful to have it written down in the gudelines. I can see there are some differences between your messages and the article as well, e.g. doc vs docs, ext vs extension, starting from small vs big letter, etc. Also kinds/categories of changes are mixed with library parts like fix/feat vs extension. Fix/feat should probably be used for extensions as well.

PR #1365, opened 2025-01-11
[formulas] Fix warning (variable hiding member) in sjoberg_intersection

awulkiew:


include/boost/geometry/formulas/sjoberg_intersection.hpp
```diff
@@ -566,7 +566,6 @@ class sjoberg_geodesic
     {
         CT const c0 = 0;
         CT const dLj = d_lambda(c0);
-        CT const asin_tj_t0j = asin(Cj * tan_betaj / sqrt_1_Cj_sqr);
```
barendgehrels, line 569: Thanks. So it's a member var and indeed calculated in the constructor.
👍

barendgehrels (approved):

include/boost/geometry/formulas/sjoberg_intersection.hpp
tinko92: This looks correct assuming, Cj, tan_betaj and sqrt_1_Cj_sqr are all constant between construction and every call of this function (they are public, mutable members in a non-detail-namespaced class but since the class is undocumented, I assume it is not meant to be a public interface).

tinko92 (approved):

vissarion (approved):
Thanks for this fix Adam. I agree with merging.

PR #1367, opened 2025-01-15
feat: include changes and tests changes

barendgehrels:
Various changes done for traversal change, but put into a separate PR


include/boost/geometry/algorithms/detail/overlay/traversal.hpp
```diff
@@ -23,6 +23,7 @@
 #include <boost/range/value_type.hpp>
 
 #include <boost/geometry/algorithms/detail/overlay/cluster_info.hpp>
+#include <boost/geometry/algorithms/detail/overlay/debug_traverse.hpp>
```
barendgehrels, line 26: This file will be removed. But the debug info is used elsewhere, so it's moved out

include/boost/geometry/strategies/relate/cartesian.hpp
```diff
@@ -22,6 +22,8 @@
 #include <boost/geometry/strategies/cartesian/point_in_poly_winding.hpp>
 #include <boost/geometry/strategies/cartesian/disjoint_box_box.hpp>
 
+#include <boost/geometry/strategies/distance/detail.hpp>
+#include <boost/geometry/strategies/distance/services.hpp>
```
barendgehrels, line 26: Traveral will use the `distance` strategy from the umbrella strategy

test/algorithms/buffer/CMakeLists.txt
```diff
@@ -38,6 +37,17 @@ foreach(item IN ITEMS
   target_compile_definitions(${BOOST_GEOMETRY_UNIT_TEST_NAME} PRIVATE BOOST_GEOMETRY_TEST_ONLY_ONE_TYPE)
 endforeach()
 
+# Tests which take a long time and are not run by default.
+# Besides that, the buffer_countries test requires the WKT files.
+foreach(item IN ITEMS
+    buffer_countries
+    buffer_linestring
+    buffer_linestring_geo
```
barendgehrels, line 45: The `buffer_linestring_geo` was not yet in cmake lists.
It's added now but very slow. Likewise `buffer_linestring` is slow.
They are not run by default now.

test/robustness/CMakeLists.txt
```diff
@@ -5,6 +5,37 @@
 # Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
 # http://www.boost.org/LICENSE_1_0.txt)
 
+
+function(boost_geometry_add_robustness_test item)
```
barendgehrels, line 9: The robustness tests were compiled and run with ctest.
For the robustness tests that makes no sense and costs (depending on the options) a lot of time.
Therefore it's now a separate target, and not added to the test framework.

test/algorithms/buffer/buffer_countries.cpp
```diff
@@ -119,6 +119,12 @@ void test_all()
 
     const std::string base_folder = "data/";
     const std::string gr = read_from_wkt_file<mpt>(base_folder + "gr.wkt");
+    if (gr.empty())
+    {
+        // If this file is not found, the others won't be find either.
+        return;
```
tinko92, line 125: Maybe it is dangerous to return silently without warning that the test could be conducted or without returning a failure. (I do not know the history/purpose of that test though)

test/algorithms/set_operations/union/union_other_types.cpp
```diff
@@ -139,24 +139,14 @@ int test_main(int, char* [])
...
-    // (the rectangular case is correct; other input might give wrong results)
-    // The int16 version throws a <zero denominator> exception
-#if ! defined(BOOST_GEOMETRY_TEST_ONLY_ONE_TYPE)
-    test_areal<point_xy<boost::rational<std::int16_t>>>({exclude::all});
-    test_areal<point_xy<boost::rational<std::int32_t>>>({exclude::fp});
-#endif
-    test_areal<point_xy<boost::rational<std::int64_t>>>({exclude::fp});
+    // Boost multi precision ratio is currently NOT supported
+    // The sqrt is not supported for it, which would limit it support
+    // drastically to a very limited set of operations.
```
tinko92, line 148: This is not obvious for set operations to me. Will the upcoming major PR use distance in set operations in a context where comparable distance could not be used?

test/algorithms/buffer/read_from_wkt_file.hpp
```diff
@@ -38,8 +38,6 @@ inline std::string read_from_wkt_file(std::string const& filename)
         out << std::fixed << std::setprecision(20) << bg::wkt(mp);
     }
 
-    BOOST_CHECK(! out.str().empty());
```
tinko92, line 41: Same comment as below, this change to silently ignore a missing input seems dangerous in a test context.

include/boost/geometry/strategies/relate/cartesian.hpp
barendgehrels: Putting everything in one umbrella sounds reasonable to me. 

I still find it hard to not be able to use distance because of the mutual dependency. I've had it several times.
Resorting then to unwished things like https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp#L188
But now I cannot escape from it.

We could also look at why distance needs relate. But that's maybe a passed station.

test/algorithms/buffer/buffer_countries.cpp
barendgehrels: It should not be used as a standard unit test.
But yes, indeed - I can add a message. Will do.
Good point.

test/algorithms/set_operations/union/union_other_types.cpp
barendgehrels: Indeed. I cannot use comparable distance because I'm summing distances.
Essentially, for a cc turn (both continue), we need to go if we in the end go left or right, and when.
As long as the segments continue, I need to sum their lengths.

barendgehrels: While writing this, and re-reading, I realize: if they all continue, it's a straight line and could be computed with one call for comparable distance for the turn point and the end point.
I'll try to change it. Thanks for the comment.

test/algorithms/buffer/buffer_countries.cpp
barendgehrels: ✅ done

test/algorithms/set_operations/union/union_other_types.cpp
barendgehrels: ✅ It was possible and gave identical results.
So it is changed, and these changes in `union_other_types.cpp` could be reverted now.
Not that I think it will be very useful to use `boost::rational` - but at least it is not a change anymore.
Thanks again.

include/boost/geometry/strategies/relate/cartesian.hpp
awulkiew: Distance needs relate because it checks for intersection of geometries and returns 0 if that's the case. Consider e.g. distance(point, polygon). Intersects or covered_by has to be checked before or in addition to calculating the distance between point and boundary of polygon.

For sure only a subset of strategy getters is needed. Not all relational operations has to be available in distance. The question is whether or not there are any benefits of dividing the umbrella strategies into parts and maintaining it over putting everything in one strategy. I'm ok with both approaches.

But if we agreed that we should put everything needed by distance, relate (and therefore also setops) and index in one umbrella strategy (this also means area, centroid, envelope and expand strategies, maybe also less/compare, I can't remember) then there is not much left. So maybe we should consider having one umbrella strategy per CS defining every strategy needed by any algorithm and deprecate everything else?

barendgehrels: Thanks for the backgrounds, let's come back to that elsewhere.

Yes, if most umbrella strategies contain nearly everything - let's make one?
But is that really true? Some simpler algorithms, simplify, for example, needs just two strategies (AFAIR). Convex hull will not use much either. Maybe we can move this to the discussion section.

include/boost/geometry/strategies/relate/geographic.hpp
```diff
@@ -78,6 +80,16 @@ class geographic
             >(base_t::m_spheroid);
     }
 
+    template <typename Geometry1, typename Geometry2>
+    auto comparable_distance(Geometry1 const&, Geometry2 const&,
+                             distance::detail::enable_if_pp_t<Geometry1, Geometry2> * = nullptr) const
```
barendgehrels, line 85: I didn't see it used earlier - but I think we should name it like this.

vissarion (approved):
LGTM thanks!

include/boost/geometry/algorithms/detail/overlay/debug_traverse.hpp
```diff
@@ -0,0 +1,69 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
+
+// Copyright (c) 2007-2024 Barend Gehrels, Amsterdam, the Netherlands.
```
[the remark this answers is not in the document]
barendgehrels: ✅

PR #1368, opened 2025-01-16
fix: pass umbrella strategies in rtree spatial predicates

awulkiew:


index/test/rtree/rtree_with_strategies.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+BOOST_GEOMETRY_REGISTER_POINT_2D(point, double, cs::undefined, x, y)
+BOOST_GEOMETRY_REGISTER_BOX(box, point, mi, ma)
+BOOST_GEOMETRY_REGISTER_SEGMENT(segment, point, first, second)
+
+template <typename QueryValue, typename Rtree>
+constexpr bool is_bp_or_bb_v = bg::util::is_box<QueryValue>::value
+                               && bg::util::is_box<typename Rtree::value_type>::value
+                            || bg::util::is_box<QueryValue>::value
+                               && bg::util::is_point<typename Rtree::value_type>::value;
```
barendgehrels, line 32: Everywhere I work it is required to put brackets, there were `&&` and `||` are combined. Also compilers normally warn for it. Can we do it?
Though the layout is beautiful here, so it's not unclear to me.
And it's test code.

Nit, bp or bb (but the order is: bb or bp)

index/test/rtree/rtree_with_strategies.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+// Copyright (c) 2025 Adam Wulkiewicz, Lodz, Poland.
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#include <vector>
+
+#define BOOST_GEOMETRY_INDEX_DETAIL_EXPERIMENTAL_PREDICATES
```
barendgehrels, line 12: Is it still experimental? When can we promote it to "general"?

barendgehrels (approved):
👍

index/test/rtree/rtree_with_strategies.cpp
```diff
@@ -0,0 +1,156 @@
...
+    test_strategies<Value, Params, bg::strategies::index::geographic<>>();
+    test_strategies<Value, Params, bg::strategies::geographic<>>();
+}
+
+template <typename Value>
+void test_value()
+{
+    test_params<Value, bgi::linear<4>>();
+    test_params<Value, bgi::quadratic<4>>();
+    test_params<Value, bgi::rstar<4>>();
```
barendgehrels, line 146: Did we document somewhere what is the adviced option? `rstar` ?

index/test/rtree/rtree_with_strategies.cpp
```diff
@@ -0,0 +1,156 @@
...
+    test_queries<point>(rtree);
+    test_queries<box>(rtree);
+    test_queries<segment>(rtree);
+}
+
+template <typename Value, typename Params>
+void test_params()
+{
+    test_strategies<Value, Params, bg::strategies::index::cartesian<>>();
+    test_strategies<Value, Params, bg::strategies::cartesian<>>();
```
barendgehrels, line 134: What is the difference actually? Is it another umbrella?

index/test/rtree/rtree_with_strategies.cpp
```diff
@@ -0,0 +1,156 @@
...
+    std::enable_if_t<(!is_pb_or_bb_v<QueryValue, Rtree>), int> = 0
+>
+void test_queries_pb_bb(Rtree const& ) {}
+
+template
+<
+    typename QueryValue, typename Rtree,
+    std::enable_if_t<is_pp_or_bb_v<QueryValue, Rtree>, int> = 0
+>
+void test_overlaps_pp_bb(Rtree const& rtree)
```
vissarion, line 92: maybe rename to `test_overlaps_touches_pp_bb` ?

vissarion (approved):

index/test/rtree/rtree_with_strategies.cpp
awulkiew: Here it's only for `touches` and only to test if it compiles. I can't remember exactly but I think it's experimental because it makes little sense to calculate whether or not bounding boxes touch. It could be removed I guess.

There is also experimental path predicate, aka distance along linestring/segment, aka ray query which could be useful but it probably requires more thinking.

index/test/rtree/rtree_with_strategies.cpp
awulkiew: It depends on the data and algorithm using the rtree. It's described here: https://www.boost.org/doc/libs/1_87_0/libs/geometry/doc/html/geometry/spatial_indexes/introduction.html#geometry.spatial_indexes.introduction.r_tree

With that said I'd probably use quadratic balancing algorithm by default and packing if possible.

index/test/rtree/rtree_with_strategies.cpp
awulkiew: Yes, just like various algorithms has their own strategies index has it as well. And in addition to that it should work with the highest level one which can be passed anywhere.

index/test/rtree/rtree_with_strategies.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+template
+<
+    typename QueryValue, typename Rtree,
+    std::enable_if_t<is_pb_or_bb_v<QueryValue, Rtree>, int> = 0
+>
+void test_queries_pb_bb(Rtree const& rtree)
+{
+    // These predicates use algorithms that are not implemented for
+    // some geometry combinations
```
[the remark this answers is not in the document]
awulkiew: Yes, if we want to implement missing combinations. But I'm not sure if this is the correct place for this.

The combinations that are not implemented are related mostly to `within`/`covered_by` when:
- higher topo-dim geometry is checked to be inside lower topo-dim geometry (e.g. `box` in `point`)
- segment is any of the arguments (it seems that the compiling combination is `point` in `segment`)

`overlaps` is defined only for combinations of the same topo-dim and is not implemented for segments

`touches` is not implemented for segments, point/box, surprisingly it compiles for point/point which should always return false

So technically more combinations could be checked than are enabled now but I think it's enough for a test of rtree umbrella strategies.

Btw, in index/detail/predicates.hpp there is a table describing which algorithms are called for values stored in leafs and bounds stored in internal nodes:
```
// value_tag        bounds_tag
// ---------------------------
// contains(I,G)    covers(I,G)
// covered_by(I,G)  intersects(I,G)
// covers(I,G)      covers(I,G)
// disjoint(I,G)    !covered_by(I,G)
// intersects(I,G)  intersects(I,G)
// overlaps(I,G)    intersects(I,G)
// touches(I,G)     intersects(I,G)
// within(I,G)      intersects(I,G)
```

index/test/rtree/rtree_with_strategies.cpp
awulkiew: Right, I forgot to rename it to `test_queries_pp_bb`.

index/test/rtree/rtree_with_strategies.cpp
awulkiew: Sure, done.

index/test/rtree/rtree_with_strategies.cpp
barendgehrels: OK - I understand - but you'll probably have the best understanding of this code.

If it makes no sense - we should remove it?

awulkiew: Now that I think about it. The R-tree also works for segments so maybe `touches(segment, segment)` could be useful? But this should be done in a different PR anyway.

index/test/rtree/rtree_with_strategies.cpp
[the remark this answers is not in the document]
vissarion: Thanks for the details!

PR #1370, opened 2025-01-30
Relax half period check in spheroidal normalization

vissarion:
Currently when points (180,0) and (-180,0) are tested for equality (in spherical or geographic systems) we are getting the expected result which is true. 

However, if the points are in radians instead of degrees the accuracy of the representation of pi/2 or the conversion from 180 degrees to radians could result in wrong results of equals (see added tests in this PR for details). 

This PR fixes this issue by relaxing conditions in normalization of spheroidal coordinates.

### DIFF
--- a/include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
+++ b/include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
@@ -2,8 +2,9 @@
 
 // Copyright (c) 2017 Adam Wulkiewicz, Lodz, Poland.
 
-// Copyright (c) 2015-2022, Oracle and/or its affiliates.
+// Copyright (c) 2015-2025, Oracle and/or its affiliates.
 
+// Contributed and/or modified by Vissarion Fysikopoulos, on behalf of Oracle
 // Contributed and/or modified by Menelaos Karavelas, on behalf of Oracle
 // Contributed and/or modified by Adam Wulkiewicz, on behalf of Oracle
 // Contributed and/or modified by Adeel Ahmad, as part of Google Summer of Code 2018 program
@@ -236,17 +237,22 @@ protected:
     }
 
 public:
-    static inline void apply(CoordinateType& longitude)
+    static inline void apply(CoordinateType& longitude, bool exact = true)
     {
         // normalize longitude
-        if (math::equals(math::abs(longitude), constants::half_period()))
+        CoordinateType const epsilon = std::numeric_limits<float>::epsilon();
+        static constexpr bool is_integer = std::numeric_limits<CoordinateType>::is_integer;
+
+        if (exact || is_integer ? math::equals(math::abs(longitude), constants::half_period())
+            : math::abs(math::abs(longitude) - constants::half_period()) <= epsilon)
         {
             longitude = constants::half_period();
         }
         else if (longitude > constants::half_period())
         {
             longitude = normalize_up(longitude);
-            if (math::equals(longitude, -constants::half_period()))
+            if (exact || is_integer ? math::equals(longitude, -constants::half_period())
+                : math::abs(longitude + constants::half_period()) <= epsilon)
             {
                 longitude = constants::half_period();
             }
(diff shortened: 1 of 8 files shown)


include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
```diff
@@ -236,17 +237,22 @@ class normalize_spheroidal_coordinates
     }
 
 public:
-    static inline void apply(CoordinateType& longitude)
+    static inline void apply(CoordinateType& longitude, bool exact = true)
     {
         // normalize longitude
-        if (math::equals(math::abs(longitude), constants::half_period()))
+        CoordinateType const epsilon = CoordinateType(1e-07);
+        bool const is_integer = std::numeric_limits<CoordinateType>::is_integer;
```
barendgehrels, line 244: Usually we use `is_integral`

include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
```diff
@@ -236,17 +237,22 @@ class normalize_spheroidal_coordinates
...
+
+        if (exact || is_integer ? math::equals(math::abs(longitude), constants::half_period())
```
barendgehrels, line 246: `math::equals` is still not exact for floating point.
I'm not sure of the context, but the `1.e-7` looks a bit rough an arbitrary.

barendgehrels (approved):
I'm ok with some questions.
Thanks!

include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
tinko92: Interesting. According to cppreference 
> If the program adds specializations for std::is_integral or std::is_integral_v, the behavior is undefined.
https://en.cppreference.com/w/cpp/types/is_integral

but 

> Implementations may provide specializations of std::numeric_limits for implementation-specific types: e.g. GCC provides std::numeric_limits<__int128>. Non-standard libraries may [add specializations](https://en.cppreference.com/w/cpp/language/extending_std) for library-provided types, e.g. [OpenEXR](http://openexr.com/) provides std::numeric_limits<half> for a 16-bit floating-point type.
https://en.cppreference.com/w/cpp/types/numeric_limits

Given the design objective of genericity and the use of e.g. boost multiprecision in tests and it showing up in issues, maybe numeric_limits is the more appropriate choice.

tinko92 (approved):

include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
vissarion: Yes, it is indeed, I am not happy with this either. 

Selecting the right epsilon value depends on the threshold we want to set when declaring two points with longitude 180 to be equal. `1.e-7` is the smallest value so that the new tests are passing, actually set by [this test](https://github.com/boostorg/geometry/pull/1370/files#diff-6898dddc1f86c9a558e1a3871bf25eebfa1b492dcb7a0e7c7de1c86a78e400b5R200); if we only want to support [this test case](https://github.com/boostorg/geometry/pull/1370/files#diff-6898dddc1f86c9a558e1a3871bf25eebfa1b492dcb7a0e7c7de1c86a78e400b5R204)  a smaller value suffices. 
The latter case seems important since it could be the case the 180 degrees is converted to radian using the constants of the spheroid (that is the case of mysql).

vissarion: In any case, if we keep such an arbitrary value the selection should be documented in a comment.

include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
vissarion: Here, we need `is_integer` to be true even for not native types that have the "behavior" of integers. We may have to revisit other places where `is_integral` is used in the library.

tinko92:
I am okay, with setting the epsilon to `std::numeric_limits<float>::epsilon();` to make the intent clear to return half_period for longitudes that are given as single precision constants, as in the test that requires the large threshold. If that is not the intent, I would change the threshold to the longitude error that is desired, at least 1e-14 to support the two remaining test cases (which are not at but close to the limit of double precision). I included more detailed discussion inline. Also added a minor note to the serialisation with precision function.

PR #1373, opened 2025-02-07
Replace boost type_traits with stl type_traits.

gogagum:
With C++14, is possible to replace **Boost.TypeTraits** library usages with standard `type_traits`. This PR removes direct **Boost.TypeTraits** dependency in **Geometry** module.

### DIFF
--- a/include/boost/geometry/index/detail/serialization.hpp
+++ b/include/boost/geometry/index/detail/serialization.hpp
@@ -14,8 +14,7 @@
 #ifndef BOOST_GEOMETRY_INDEX_DETAIL_SERIALIZATION_HPP
 #define BOOST_GEOMETRY_INDEX_DETAIL_SERIALIZATION_HPP
 
-#include <boost/type_traits/alignment_of.hpp>
-#include <boost/type_traits/aligned_storage.hpp>
+#include <type_traits>
 
 #include <boost/serialization/serialization.hpp>
 #include <boost/serialization/split_member.hpp>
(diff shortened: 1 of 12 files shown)


include/boost/geometry/index/detail/serialization.hpp
```diff
@@ -65,10 +64,10 @@ class serialization_storage
     }
     T * address()
     {
-        return static_cast<T*>(m_storage.address());
+        return static_cast<void*>(&m_storage);
```
barendgehrels, line 67: @awulkiew is this fine?

barendgehrels (approved):
Thanks! Looks good to me, one question to Adam

vissarion (approved):
Thanks, this look OK to me.

include/boost/geometry/index/detail/serialization.hpp
awulkiew: `std::aligned_storage` is deprecated in C++23. AFAIR boost version is superior to the std one but I can't remember why besides the fact that it doesn't force us to cast address of the storage. A proposed replacement is something like this:
```
alignas(T) std::byte storage[sizeof(T)];
```
but it requires C++17 and would require careful testing. So I'd prefer to keep `boost::aligned storage` for now.

awulkiew: Btw, here https://en.cppreference.com/w/cpp/types/aligned_storage the required use of `std::launder` is mentioned. This change can introduce some unwanted effects. I guess I strongly suggest to keep the old version. :)

awulkiew (changes requested):

barendgehrels:
@gogagum , will you address the one review action item indicated by @awulkiew ?

barendgehrels:
@gogagum will you update it or can we close it? (we still can remove the dependency later)

gogagum:
@barendgehrels I am going to finish this work.

barendgehrels:
Replaced by #1459

Issue #629, opened 2019-10-24
MSVC Warning: C4244 'initializing': conversion from 'CoordinateType' to 'const PromotedType', possible loss of data


AlastairHolmes:
Minimal Code to produce related warning (boost 1.70.0 MSVC 2017 15.8.0 /std:c++latest /W4):

```
boost::geometry::model::polygon<
    boost::geometry::model::point<std::int64_t, 3, boost::geometry::cs::cartesian>
> test;
boost::geometry::remove_spikes(test);
```

If the `std::int64_t` in the above is changed to `int` there is no warning.

On line 245 (side_by_triangle.hpp) boost decides the type of PromotedType = `select_most_precise<double,std::int64_t>` => `double`. And then initializes PromotedType values with `std::int64_t` values.

This pattern of select_most_precise<double,...> is pretty common, and I have found the same warning being produced in other parts of boost that do the same thing when using with `std::int64_t`.

vschoech:
This is still a problem in boost 1.72.0.
`\boost_1_72_0\boost\geometry\strategies\cartesian\side_by_triangle.hpp(123): error C4244: 'initializing': conversion from 'CoordinateType' to 'PromotedType', possible loss of data`

vschoech:
This is still a problem in boost 1.73.0.
`\boost_1_73_0\boost\geometry\strategies\cartesian\side_by_triangle.hpp(123): error C4244: 'initializing': conversion from 'CoordinateType' to 'PromotedType', possible loss of data`

mloskot:
I think it should be safe to `... = static_cast<PromotedType>(...)` wherever the left-hand side is an explicitly promoted type.

@awulkiew What do you think?

Ne94fets:
Hello, I have a similar problem on Windows using msvc14.1. On linux there is no warning by the way.

I created a minimal example:
```
#include <boost/geometry.hpp>

namespace bg = boost::geometry;
typedef bg::model::polygon<bg::model::d2::point_xy<double>> polygon_t;
typedef bg::model::multi_polygon<polygon_t> mpolygon_t;

int main() {
    polygon_t p0{ {{-5,-5},{-5,5},{5,5},{5,-5},{-5,-5}} };
    polygon_t p1{ {{-4,-4},{-4,4},{4,4},{4,-4},{-4,-4}} };
    mpolygon_t out;

    bg::difference(p0, p1, out);
}
```

[...]

PR #1379, opened 2025-03-15
fix: avoid warnings for coordinate conversions and unused parameters

barendgehrels:
Fixes: #629

This was a long standing warning where we were pinged about several times.

I think this is the best solution at the moment. However, it solves one case. There is the message:

>This pattern of select_most_precise<double,...> is pretty common, and I have found the same warning being produced in other parts of boost that do the same thing when using with std::int64_t.

I cannot fix all of that now, it first have to wait at least for the big improvement in intersections.

Then we probably should fix it in `select_most_precise` itself.

Please give me your opinions.

If possible it should be included in `1.88`

### DIFF
--- a/include/boost/geometry/strategy/cartesian/side_by_triangle.hpp
+++ b/include/boost/geometry/strategy/cartesian/side_by_triangle.hpp
@@ -33,6 +33,7 @@
 #include <boost/geometry/strategies/compare.hpp>
 #include <boost/geometry/strategies/side.hpp>
 
+#include <boost/geometry/util/promote_integral.hpp>
 #include <boost/geometry/util/select_calculation_type.hpp>
 #include <boost/geometry/util/select_most_precise.hpp>
 
@@ -205,23 +206,29 @@ public :
     template <typename P1, typename P2, typename P>
     static inline int apply(P1 const& p1, P2 const& p2, P const& p)
     {
-        using coor_t = typename select_calculation_type_alt<CalculationType, P1, P2, P>::type;
-
-        // Promote float->double, small int->int
-        using promoted_t = typename select_most_precise<coor_t, double>::type;
-
-        bool const are_all_integral_coordinates =
+        constexpr bool are_all_integral_coordinates =
             std::is_integral<coordinate_type_t<P1>>::value
             && std::is_integral<coordinate_type_t<P2>>::value
             && std::is_integral<coordinate_type_t<P>>::value;
 
+        // Promote float to double
+        // For integer: short -> int -> long
+        // For larger integers: long, long long, std::int64_t all stay as they are (on a Mac)
+        using coor_t = typename select_calculation_type_alt<CalculationType, P1, P2, P>::type;
+        using promoted_t = std::conditional_t
+            <
+                are_all_integral_coordinates,
+                typename promote_integral<coor_t>::type,
+                typename select_most_precise<coor_t, double>::type
+            >;
+
         eps_policy< math::detail::equals_factor_policy<promoted_t> > epsp;
-        promoted_t s = compute_side_value
+        promoted_t const s = compute_side_value
             <
                 coor_t, promoted_t, are_all_integral_coordinates
             >::apply(p1, p2, p, epsp);
 
-        promoted_t const zero = promoted_t();
+        static promoted_t const zero = promoted_t();
         return math::detail::equals_by_policy(s, zero, epsp.policy) ? 0
             : s > zero ? 1
             : -1;
--- a/include/boost/geometry/algorithms/detail/overlay/check_enrich.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/check_enrich.hpp
@@ -21,6 +21,7 @@
 #include <cstddef>
 #include <vector>
 
+#include <boost/core/ignore_unused.hpp>
 #include <boost/range/begin.hpp>
 #include <boost/range/end.hpp>
 #include <boost/range/value_type.hpp>
@@ -66,6 +67,8 @@ inline void display(MetaTurn const& meta_turn, const char* reason = "")
         //<< " -> " << op_index
         << " " << reason
         << std::endl;
+#else
+boost::ignore_unused(meta_turn, reason);
 #endif
 }
 
--- a/include/boost/geometry/strategies/geographic/distance_cross_track_box_box.hpp
+++ b/include/boost/geometry/strategies/geographic/distance_cross_track_box_box.hpp
@@ -189,8 +189,7 @@ private:
 
 public:
     template <typename T>
-    static inline return_type apply(this_strategy const& strategy,
-                                    T const& distance)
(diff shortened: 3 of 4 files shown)


include/boost/geometry/strategy/cartesian/side_by_triangle.hpp
```diff
@@ -205,23 +206,27 @@ public :
...
-        // Promote float->double, small int->int
-        using promoted_t = typename select_most_precise<coor_t, double>::type;
-
-        bool const are_all_integral_coordinates =
+        constexpr bool are_all_integral_coordinates =
             std::is_integral<coordinate_type_t<P1>>::value
             && std::is_integral<coordinate_type_t<P2>>::value
             && std::is_integral<coordinate_type_t<P>>::value;
 
+        // Promote float->double, integers to std::int64_t
```
tinko92, line 214: Is the latter part of this comment true/intended to remain true, given 
```
// For now deactivate the use of multiprecision integers
// TODO: activate it later
#define BOOST_GEOMETRY_NO_MULTIPRECISION_INTEGER
```
<https://github.com/boostorg/geometry/blob/0900b5d61f6c0b304e78475d8e56a34cf168170d/include/boost/geometry/util/promote_integral.hpp#L14>
 for the case in #629 where std::int64_t is used ? Or is that TODO obsolete?

include/boost/geometry/strategy/cartesian/side_by_triangle.hpp
```diff
@@ -205,23 +206,27 @@ public :
...
+
         eps_policy< math::detail::equals_factor_policy<promoted_t> > epsp;
-        promoted_t s = compute_side_value
+        promoted_t const s = compute_side_value
             <
                 coor_t, promoted_t, are_all_integral_coordinates
             >::apply(p1, p2, p, epsp);
 
-        promoted_t const zero = promoted_t();
+        static promoted_t const zero = promoted_t();
```
tinko92, line 229: Is this a microoptimisation with a meaningful benefit or is there some other intention? For primitive types (almost every case, I'd expect), I wouldn't expect it. I suppose this is for custom number types were the initialisation of 0 may be costly.

I find this makes the code less clear, it made me wonder if that value was intended to be referenced/pointed to somewhere outside the scope or something like this, maybe it is just unfamiliarity on my part, though.

tinko92 (approved):
Looks good to me, I commented two things which are both minor. Given that you want it in 1.88 and it is a long-standing bug, I'm in favour of merging regardless of the resolution of my comments, so I selected "Approve".

include/boost/geometry/strategy/cartesian/side_by_triangle.hpp
barendgehrels: Thanks for your sharp observation.
I will change `int64_t` in the comments (this was my first approach) because integer will go to 'long'.
It is actually not related to `BOOST_GEOMETRY_NO_MULTIPRECISION_INTEGER` because that is really about the Boost.Multi Precision library and their integers. That is turned off indeed (I don't remember the `TODO` thanks - I think we should not turn it on automatically, it should be a conscious choice for users if they want it, or not).

So what happens now (at least on my mac) is:
`short` will to to `int`
`int` will go to `long` (instead of `double` before - even though the comment said it would not)
`int64` will stay `int64_t` (instead of `double` before)

include/boost/geometry/strategy/cartesian/side_by_triangle.hpp
barendgehrels: It's more harmonizing with other code than a micro optimization.
For example, in `segment_ratio.hpp`: `static thistype result(0, 1);`
In `get_turn_info.hpp`: `static decltype(dm.measure) const zero = 0;`
In `infinite_lien_functions.hpp`: `static Type const zero = 0;`

But there are also cases where it is not static.
I intended to make it `constexpr` instead but that is not supported (IIRC) by Boost.MultiPrecision.

include/boost/geometry/strategy/cartesian/side_by_triangle.hpp
barendgehrels: The full list (I added code to display it in a comment in #629)
```
t: s -> promoted_t: i
t: i -> promoted_t: l
t: l -> promoted_t: l
t: x -> promoted_t: x
t: x -> promoted_t: x
```

tinko92: I see, thanks. My concern was, if it was a TODO to remove that macro, if it were removed then int64_t would actually be promoted to a multiprecision guide because it is in the promotion list if that macro is not defined https://github.com/boostorg/geometry/blob/0900b5d61f6c0b304e78475d8e56a34cf168170d/include/boost/geometry/util/promote_integral.hpp#L237 . I agree that it should not be turned on by default.

barendgehrels: Agreed, I changed the todo now in this PR. I didn't want to make the real change now but created another ticket #1380 to not forget it.

tinko92:
One more thought on this: I can see this change affects the coordinate type change in the context of side-by-triangle, which evaluates a second-degree polynomial. The integer promotion promotes to an integral type with twice the digits, which will make this evaluate without integer overflow if the coordinate_type were e.g. an int32_t and the input geometry fully uses the range, the predicate would be evaluated in int64_t without overflow. For int64_t without int128 or a multiprecision type, this is not possible.

I think the choice of giving float a priority above int in select_most_precise and this use in side_by_triangle has some merit. 

Converting coordinates that use the full range of int64_t to double will lose 10 bits of precision, which is fair to warn about from the compiler, and may produce an incorrect result if 1) they use close to that full range of the type and 2) they were nearly collinear up to the last 10 (binary) digits of their original 64 digits.

Evaluating a second degree polynomial in side by triangle in int64_t if the coordinates use close to the full range of the type will likely overflow and may also produce the wrong result but now silently without a warning and even if the points are not near-collinear, with undefined behaviour (signed integer overflow) and with AFAIK no standard way to detect that it happened.

vissarion (approved):
I also agree with merging.

barendgehrels:
> One more thought on this
 
> I think the choice of giving floating-poitn a priority above integer in `select_most_precise` and this use in `side_by_triangle` has some merit.

Yes, going to double had some merit indeed. 

But I still think we should not have done it. Originally we didn’t. I don’t remember when it was changed - but the “integer” world is to be fully safe, without having FP issues in the calculations. Therefore the side should be stable, and operator on input=integer, output=integer. Same for other calculations. The intersections can fall off the grid, of course, but they are rounded.

So integer should stay integer, as much as possible.

And then you are right, whenever we multiply integers, we need to have the space for it. That was also discussed, long ago, with the “customer” (in this case: @vschoech). They (from Thinkcell) are aware to not use the full range of int64_t. But we should indeed document it properly. That either only part of a range can be used, or promotion should be possible and successful (for example, we could promote to Boost.Multiprecision using that define).

PR #1392, opened 2025-04-19
chore: update retired ubuntu 20.04

barendgehrels:
Trying to fix message

image

https://github.com/boostorg/geometry/actions/runs/14548002056/job/40816695847?pr=1369

# Note

This PR is probably not good. It gives many warnings and stops with `E: Package 'clang-13' has no installation candidate`

I'm not yet proficient with workflows, @vissarion can you help with this PR?


vissarion:
I opened a PR with a proposed fix https://github.com/boostorg/geometry/pull/1396 we can discuss more there.

PR #1396, opened 2025-04-24
Retire ubuntu 20.04

vissarion:
Retire ubuntu 20.04 from github actions. This force us to also remove clang versions > 6 and gcc < 7, since it is not straightforward to install them in ubuntu 22.04. In particular, if one tries to import the ubuntu 18 (bionic) archives then `apt install` reaches an impossible situation when trying to install older versions (e.g. clang 5.0) due to unmet dependencies.


barendgehrels (approved):
super, thanks

PR #789, opened 2020-12-28
Defined a geometry for polyhedron

Siddharth-coder13:
As suggested by @barendgehrels and @vissarion in comments #683, I defined geometry for polyhedral surfaces.

PR #1402, opened 2025-04-30
Add geometry polyhedral surface

vissarion:
This PR is based on https://github.com/boostorg/geometry/pull/789 and adds a 3D geometry type for polyhedral surfaces according to OGC standard. 

The main class, concepts, i/o and basic traits are added.

Comparing to https://github.com/boostorg/geometry/pull/789 it refines the code (since that was an old PR and the library evolved since then) and also defines the polyhedral surface as a set of polygons (instead of rings) as described in the standard. 

Finally, there are docs and unit tests.


example/08_polyhedralsurface_example.cpp
```diff
@@ -0,0 +1,65 @@
...
+int main()
+{
+    using point_t = bg::model::point<double, 3, bg::cs::cartesian>;
+    using polygon_t = bg::model::polygon<point_t>;
+    using polyhedral_t = bg::model::polyhedral_surface<polygon_t>;
+
+    // intializing an empty polyhedral surface (default constructor)
+    polyhedral_t polyhedral0;
+
+    // creating a polyhderal surface (cube) using standard initiallized list
```
barendgehrels, line 26: Typos ->  `polyhedral` and `initialized` (and `Create` and `using a`)

example/08_polyhedralsurface_example.cpp
```diff
@@ -0,0 +1,65 @@
...
+    polyhedral_t polyhedral1 = {{{{0, 0, 0}, {0, 1, 0}, {1, 1, 0}, {1, 0, 0}, {0, 0, 0}}},\
+                                {{{0, 0, 0}, {0, 1, 0}, {0, 1, 1}, {0, 0, 1}, {0, 0, 0}}},\
+                                {{{0, 0, 0}, {1, 0, 0}, {1, 0, 1}, {0, 0, 1}, {0, 0, 0}}},\
+                                {{{1, 1, 1}, {1, 0, 1}, {0, 0, 1}, {0, 1, 1}, {1, 1, 1}}},\
+                                {{{1, 1, 1}, {1, 0, 1}, {1, 0, 0}, {1, 1, 0}, {1, 1, 1}}},\
+                                {{{1, 1, 1}, {1, 1, 0}, {0, 1, 0}, {0, 1, 1}, {1, 1, 1}}}};
+
+    // creating a polyhderal surface (triangular pyramid) using append
```
barendgehrels, line 34: The typo is repeated several times, could you search/replace?

example/08_polyhedralsurface_example.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    // intializing an empty polyhedral surface (default constructor)
+    polyhedral_t polyhedral0;
+
+    // creating a polyhderal surface (cube) using standard initiallized list
+    polyhedral_t polyhedral1 = {{{{0, 0, 0}, {0, 1, 0}, {1, 1, 0}, {1, 0, 0}, {0, 0, 0}}},\
+                                {{{0, 0, 0}, {0, 1, 0}, {0, 1, 1}, {0, 0, 1}, {0, 0, 0}}},\
+                                {{{0, 0, 0}, {1, 0, 0}, {1, 0, 1}, {0, 0, 1}, {0, 0, 0}}},\
+                                {{{1, 1, 1}, {1, 0, 1}, {0, 0, 1}, {0, 1, 1}, {1, 1, 1}}},\
+                                {{{1, 1, 1}, {1, 0, 1}, {1, 0, 0}, {1, 1, 0}, {1, 1, 1}}},\
```
barendgehrels, line 31: The backslashes are probably from the WKT variant.
Does it even compile here?

example/08_polyhedralsurface_example.cpp
```diff
@@ -0,0 +1,65 @@
...
+                                    ((0 0 0, 0 1 0, 0 1 1, 0 0 1, 0 0 0)),\
+                                    ((0 0 0, 1 0 0, 1 0 1, 0 0 1, 0 0 0)),\
+                                    ((1 1 1, 1 0 1, 0 0 1, 0 1 1, 1 1 1)),\
+                                    ((1 1 1, 1 0 1, 1 0 0, 1 1 0, 1 1 1)),\
+                                    ((1 1 1, 1 1 0, 0 1 0, 0 1 1, 1 1 1)))", polyhedral2);
+
+    // Write polyhedral surface wkt
+    std::cout << bg::wkt(polyhedral0) << std::endl;
+    std::cout << bg::wkt(polyhedral1) << std::endl;
+    std::cout << bg::wkt(polyhedral2) << std::endl;
```
barendgehrels, line 62: Great you picked it up again!
* there should be credit to Sidharta in the copyright section
* how sure are we that this all works and is according to the standards? Are there rules for ordering faces? 
* should we check at least one property (for example area or volume) and compare it with another library
* or are there ways (other OS tools or so) to visualize it to see we didn't miss anything?

include/boost/geometry/geometries/concepts/polyhedral_surface_concept.hpp
```diff
@@ -0,0 +1,77 @@
...
+namespace boost { namespace geometry { namespace concepts
+{
+
+template <typename Geometry>
+class PolyhedralSurface
+{
+#ifndef DOXYGEN_NO_CONCEPT_MEMBERS
+    using polygon_type = typename boost::range_value<Geometry>::type;
+
+	BOOST_CONCEPT_ASSERT( (concepts::Polygon<polygon_type>) );
```
barendgehrels, line 29: Indent

include/boost/geometry/geometries/polyhedral_surface.hpp
```diff
@@ -0,0 +1,99 @@
...
+#include <boost/geometry/core/tags.hpp>
+#include <boost/config.hpp>
+
+namespace boost { namespace geometry
+{
+namespace model
+{
+
+/*!
+\brief A Polyhedral Surface is a contiguous collection of polygons,
```
barendgehrels, line 29: of 3 dimensional polygons?

include/boost/geometry/geometries/polyhedral_surface.hpp
```diff
@@ -0,0 +1,99 @@
...
+    which share common boundary segments.
```
barendgehrels, line 30: we don't check it. Suggestion: 
"The concepts and algorithms don't check if the boundary segments are common"

include/boost/geometry/io/wkt/read.hpp
```diff
@@ -458,6 +458,35 @@ struct polygon_parser
     }
 };
 
+template<typename PolyhedralSurface>
+struct polyhderal_surface_parser
```
barendgehrels, line 462: Typo also here

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -228,6 +228,30 @@ struct wkt_poly
 
 };
 
+template <typename Polyhedral_surface, typename PrefixPolicy>
+struct wkt_polyhedral
```
barendgehrels, line 232: we might add `_surface` and the template parameter `PolyhedralSurface`

test/geometries/polyhedral_surface.cpp
```diff
@@ -0,0 +1,86 @@
...
+    polyhedral_t polyhedral0;
+    BOOST_CHECK_EQUAL(boost::size(polyhedral0), 0);
+
+    // creating a polyhderal surface using standard initiallized list
+    polyhedral_t polyhedral1 = {{{{0, 0, 0}, {0, 1, 0}, {1, 1, 0}, {1, 0, 0}, {0, 0, 0}}},\
+                                {{{0, 0, 0}, {0, 1, 0}, {0, 1, 1}, {0, 0, 1}, {0, 0, 0}}},\
+                                {{{0, 0, 0}, {1, 0, 0}, {1, 0, 1}, {0, 0, 1}, {0, 0, 0}}},\
+                                {{{1, 1, 1}, {1, 0, 1}, {0, 0, 1}, {0, 1, 1}, {1, 1, 1}}},\
+                                {{{1, 1, 1}, {1, 0, 1}, {1, 0, 0}, {1, 1, 0}, {1, 1, 1}}},\
+                                {{{1, 1, 1}, {1, 1, 0}, {0, 1, 0}, {0, 1, 1}, {1, 1, 1}}}};
```
barendgehrels, line 41: Same comments as for the example.
I think this unit test is fine, testing 3 use cases.
But the example does not need to be an exact repetition and one example is enough.

include/boost/geometry/geometries/polyhedral_surface.hpp
vissarion: I also added that common boundary segments should be checked by is_valid()

test/geometries/polyhedral_surface.cpp
vissarion: Thanks, I removed the `example/08_polyhedralsurface_example.cpp` and kept the one in the docs.

example/08_polyhedralsurface_example.cpp
vissarion: > Great you picked it up again!
> 
>     * there should be credit to Sidharta in the copyright section

I have already added as a commit coauthor, he hasn't added copyright notes in his original PR, but I agree that should be mentioned in copyrights too, so I added. @Siddharth-coder13 please let me know if you agree with those additions. 

>     * how sure are we that this all works and is according to the standards? Are there rules for ordering faces?

We are not sure at this point, that check (consistent orientation of faces) could be part of validity checks. Another check should be related to dimension (assert that is 3). Does it make sense to add this in is_valid too? AFAIU this is similar to the overlapping polygon check for multi_polyogons (that check also applies for polyhedral surfaces).

The standard also defines the following methods (that will be implemented in a subsequent PR):
--- `NumPatches () : Integer` ⎯ Returns the number of including polygons
--- `PatchN (N: Integer): Polygon` ⎯ Returns a polygon in this surface, the order is arbitrary.
--- `BoundingPolygons (p: Polygon): MultiPolygon` ⎯ Returns the collection of polygons in this surface that
bounds the given polygon “p” for any polygon “p” in the surface.
--- `IsClosed (): Integer` - Returns 1 (True) if the polygon closes on itself, and thus has no boundary and
encloses a solid

>     * should we check at least one property (for example area or volume) and compare it with another library

Sure, but I think that should be after implementing `is_valid` and the basic methods above. 

>     * or are there ways (other OS tools or so) to visualize it to see we didn't miss anything?

there are a few visualization software (e.g. geomview) but I do not know of any that supports WKT for polyhedral surfaces.

vissarion:
Thanks @barendgehrels for the quick review! I've addressed all your comments and replied to questions.

example/08_polyhedralsurface_example.cpp
vissarion: EDIT: `NumPatches ()` and `PatchN (N: Integer)` are already implemented by calling `size()` and `[]`.

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -396,6 +420,15 @@ struct wkt<Polygon, polygon_tag>
         >
 {};
 
+template <typename Polyhedral_surface>
```
tinko92, line 423: Like previous comment, should this be PolyhedralSurface?

doc/concept/polyhedral_surface.qbk
```diff
@@ -0,0 +1,45 @@
...
+
+[heading Description]
+[concept PolyhedralSurface..polyhedral surface]
+
+[heading Concept Definition]
+
+The PolyhedralSurface Concept is defined as following:
+
+* There must be a specialization of the metafunction `traits::tag`, defining `polyhedral_surface_tag` as type
+* It must behave like a Boost.Range Random Access Range
```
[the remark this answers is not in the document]
barendgehrels: I'm not sure... we cannot yet make it dependent on C++20 yet...
Also it would be weird to let some concepts depend on the std-ranges/concepts, and others not...

doc/src/examples/geometries/polyhedral_surface.cpp
```diff
@@ -0,0 +1,79 @@
+// Boost.Geometry
+// QuickBook Example
+
+// Copyright (c) 2025 Siddharth Kumar, Roorkee, India.
```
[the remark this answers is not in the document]
Siddharth-coder13: Thanks @vissarion for mentioning me in the copyrights.

doc/src/examples/geometries/polyhedral_surface.cpp
```diff
@@ -0,0 +1,79 @@
...
+    using polyhedral_t = bg::model::polyhedral_surface<polygon_t>;
+
+    polyhedral_t polyhedral0; /*< Initializing an empty polyhedral surface (default constructor) >*/
+
+    polyhedral_t polyhedral1 = {{{{0, 0, 0}, {0, 1, 0}, {1, 1, 0}, {1, 0, 0}, {0, 0, 0}}},
+                                {{{0, 0, 0}, {0, 1, 0}, {0, 1, 1}, {0, 0, 1}, {0, 0, 0}}},
+                                {{{0, 0, 0}, {1, 0, 0}, {1, 0, 1}, {0, 0, 1}, {0, 0, 0}}},
+                                {{{1, 1, 1}, {1, 0, 1}, {0, 0, 1}, {0, 1, 1}, {1, 1, 1}}},
+                                {{{1, 1, 1}, {1, 0, 1}, {1, 0, 0}, {1, 1, 0}, {1, 1, 1}}},
+                                {{{1, 1, 1}, {1, 1, 0}, {0, 1, 0}, {0, 1, 1}, {1, 1, 1}}}}; /*< Creating a polyhedral surface (cube) using standard initialized list >*/
```
barendgehrels, line 35: I'm not experienced in this subject. I had to parse it mentally and it is a cube. Then I noticed it is written there - can that comment go first?

Same below - it's nice to have a pyramid, which is kind of realistic. But I think the example is not correct. I would expect at least a point in the center (2.5, 2.5) which is not there. Or it is not a pyramid. Didn't parse the rest...

Maybe a (squarish) pyramid would be a more realistic example. It should have 5 faces, the floor face having 4 points, the other faces having 3 points, if I'm correct. Or the closure points should be added in all faces. The 4 other faces should have one common point in the center.

And then there should a picture of it (somehow, for example in geomview, constructed in another way then).

Writing this gives me also the questions:
* should the surfaces be closed? Will it follow the polygon rules?
* should they be clockwise? Will it follow the polygon rules?

doc/concept/polyhedral_surface.qbk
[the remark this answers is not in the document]
tinko92: > we cannot yet make it dependent on C++20 yet...
Yes. To be clear, my thought was to document it based on the STL concept, not actually bumping the version requirement of the library. The behaviour could be used without depending on the actual C++ version in which it became part of the STL. The rationale is avoiding a future breaking change in this BG concept, if BG moves to a newer standard eventually.

But I understand the point about consistency with existing BG concepts.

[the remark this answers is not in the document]
vissarion: It is a good point @tinko92 but I think we should do it once for all concepts.

doc/src/examples/geometries/polyhedral_surface.cpp
vissarion: [Here](https://trinket.io/python3/cc089613aaa6) there is a python script that can be used to visualize POLYHEDRALSURFACE WKTs.

First, this is the triangular pyramid from the example
POLYHEDRALSURFACE(((0 0 0,5 0 0,0 5 0,0 0 0)),((0 0 0,0 5 0,0 0 5,0 0 0)),((0 0 0,5 0 0,0 0 5,0 0 0)),((5 0 0,0 5 0,0 0 5,5 0 0)))
Screenshot from 2025-05-05 11-44-12
(I added that picture in the docs)

Polyhedral surfaces is not necessarily closed as described in the standard. 

> If each such LineString is the boundary of exactly 2 Polygon patches, then the PolyhedralSurface is a simple,
closed polyhedron and is topologically isomorphic to the surface of a sphere.

Thus the cube and the tringular pyrimid examples are polyhedra but the following polyhedral surface is not. 

POLYHEDRALSURFACE(((0 0 0,0 1 0,1 1 0,1 0 0,0 0 0)),((0 0 0,0 1 0,0 1 1,0 0 1,0 0 0)),((0 0 0,1 0 0,1 0 1,0 0 1,0 0 0)))
open_polyhedral_surface2

Regarding the question on ordering (clockwise or counterclockwise), the polygons that describe the faces of the polyhedral surface will have some orientation (we mention that there should be a consistent orientation among polygons of the surface). The surface itself if closed then (coping from the standard) has some orientation:

>In this case, the “top” of the surface will either point inward or outward of the enclosed finite solid.
If outward, the surface is the exterior boundary of the enclosed surface. If inward, the surface is the interior of the
infinite complement of the enclosed solid. A Ball with some number of voids (holes) inside can thus be presented
as one exterior boundary shell, and some number in interior boundary shells.

tinko92 (approved):

doc/src/examples/geometries/polyhedral_surface.cpp
barendgehrels: 👍 thanks a lot for adding the pictures, and to the doc, and for your explanations!

barendgehrels (approved):
Cool.

Small thing, can we squash and rebase (no merge commit)? Especially commits like `fix typo` we should avoid in the branch, I think.

PR #1404, opened 2025-05-04
Fix/buffer use cluster for within piece

barendgehrels:
This is a follow-up of #1369 
I wrote it in February, after first PR - and just rebased it and retested it.
It is much smaller.

What can happen that one point is considered as within another piece (that's good) but another clustered piece is not considered as within, or it is considered as within the first piece. This is all caused by floating point behaviour.

It happens with both side strategies (robust and `by_triangle`)

This fix is simple, it considers clusters and the pieces involved, and does not even need to perform a within-check in those cases.

Behaviour, before (on current `develop`, this is the same as reported in #1369 but there it was on that branch):
```
geometries: 603809 errors: 506 type: d time: 67.478
geometries: 610519 errors: 417 type: d time: 63.547 (robust side)
```

Behaviour, after the fix
```
geometries: 608834 errors: 430 type: d time: 68.572
geometries: 615232 errors: 333 type: d time: 64.119 (robust side)
```

So roughly 25% improvement. A few of these are added to the unit test suite.

In terms of pictures, it's a big difference often - it cannot make the output if some turns are classified as "within". The orange/red dotted line is the resulting buffer.

Before (the turn labelled as `7` is classified wrongly)
Screenshot 2025-05-04 at 12 52 15

After:
Screenshot 2025-05-04 at 12 52 45

### DIFF
--- a/include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
@@ -38,6 +38,7 @@ template
 <
     typename CsTag,
     typename Turns,
+    typename Clusters,
     typename Pieces,
     typename DistanceStrategy,
     typename UmbrellaStrategy
@@ -46,6 +47,7 @@ template
 class turn_in_piece_visitor
 {
     Turns& m_turns; // because partition is currently operating on const input only
+    Clusters const& m_clusters;
     Pieces const& m_pieces; // to check for piece-type
     DistanceStrategy const& m_distance_strategy; // to check if point is on original or one_sided
     UmbrellaStrategy const& m_umbrella_strategy;
@@ -77,6 +79,39 @@ class turn_in_piece_visitor
         return false;
     }
 
+    // Returns true if the turn is part of a cluster, and one of the other turns in the same
+    // cluster is involved in the same piece as the operations in the turn.
+    template <typename Turn, typename Piece>
+    inline bool skip_by_same_cluster(Turn const& turn, Piece const& piece) const
+    {
+        auto it = m_clusters.find(turn.cluster_id);
+        if (it == m_clusters.end())
+        {
+            return false;
+        }
+
+        for (auto const& index : it->second.turn_indices)
+        {
+            if (index == static_cast<signed_size_type>(turn.turn_index))
+            {
+                continue;
+            }
+            auto const& other_turn = m_turns[index];
+            auto const& seg_id0 = other_turn.operations[0].seg_id;
+            auto const& seg_id1 = other_turn.operations[1].seg_id;
+
+            if (seg_id0.piece_index == piece.index
+                || seg_id1.piece_index == piece.index)
+            {
+                // One of the other turns in the same cluster is an intersection
+                // with the same piece.
+                // Therefore, the turn under inspection cannot be within that piece.
+                return true;
+            }
+        }
+        return false;
+    }
+
     template <typename NumericType>
     inline bool is_one_sided(NumericType const& left, NumericType const& right) const
     {
@@ -96,10 +131,11 @@ class turn_in_piece_visitor
 
 public:
 
-    inline turn_in_piece_visitor(Turns& turns, Pieces const& pieces,
+    inline turn_in_piece_visitor(Turns& turns, Clusters const& clusters, Pieces const& pieces,
                                  DistanceStrategy const& distance_strategy,
                                  UmbrellaStrategy const& umbrella_strategy)
         : m_turns(turns)
+        , m_clusters(clusters)
         , m_pieces(pieces)
         , m_distance_strategy(distance_strategy)
         , m_umbrella_strategy(umbrella_strategy)
@@ -140,7 +176,12 @@ public:
 
         if (piece.type == geometry::strategy::buffer::buffered_empty_side)
         {
-            return false;
+            return true;
+        }
+
+        if (skip_by_same_cluster(turn, piece))
+        {
+            return true;
         }
 
         if (piece.type == geometry::strategy::buffer::buffered_point)
--- a/test/algorithms/buffer/buffer_linestring_geo.cpp
+++ b/test/algorithms/buffer/buffer_linestring_geo.cpp
@@ -116,9 +116,11 @@ void test_linestring_aimes()
     std::set<int> const skip_cases_miter_flat{};
 #else
     // Cases where the algorithm is still failing.
-    std::set<int> const skip_cases_round_round{17, 22, 38, 181, 196};
-    std::set<int> const skip_cases_round_flat{17, 22, 38, 103, 196};
-    std::set<int> const skip_cases_miter_flat{17, 18, 22, 38, 103, 196};
+    // It might differ in debug/release mode.
+    // It might also depend on the compiler and on the operating system.
+    std::set<int> const skip_cases_round_round{17, 22, 38, 67, 75, 109, 163, 181, 143, 196};
+    std::set<int> const skip_cases_round_flat{17, 22, 38, 67, 75, 103, 109, 196};
+    std::set<int> const skip_cases_miter_flat{17, 18, 22, 38, 67, 75, 103, 109, 196};
 #endif
 
     for (auto const& enumerated : bg::util::enumerate(testcases_aimes))
--- a/test/algorithms/buffer/buffer_multi_polygon.cpp
+++ b/test/algorithms/buffer/buffer_multi_polygon.cpp
@@ -625,10 +625,7 @@ void test_all()
     TEST_BUFFER(rt_w15, join_miter, end_flat, 80.1348, 1.0);
     TEST_BUFFER(rt_w16, join_miter, end_flat, 31.6495, 1.0);
     TEST_BUFFER(rt_w17, join_miter, end_flat, 33.74264, 1.0);
-
-#if defined(BOOST_GEOMETRY_TEST_FAILURES)
-    TEST_BUFFER(rt_w18, join_miter, end_flat, 82.4779, 1.0);
-#endif
+    TEST_BUFFER(rt_w18, join_miter, end_flat, 83.4779, 1.0);
 
 #if defined(BOOST_GEOMETRY_TEST_FAILURES) || defined(BOOST_GEOMETRY_CONCEPT_FIX_ARRIVAL)
     // See comments at issue issue_1262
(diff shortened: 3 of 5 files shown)


include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
```diff
@@ -140,7 +181,12 @@ class turn_in_piece_visitor
 
         if (piece.type == geometry::strategy::buffer::buffered_empty_side)
         {
-            return false;
+            return true;
```
barendgehrels, line 184: This does not make any difference - but it should have been `true`

test/algorithms/buffer/buffer_multi_polygon.cpp
```diff
@@ -672,6 +666,11 @@ void test_all()
     TEST_BUFFER_VALIDITY_FALSE_NEGATIVE(rt_w35, join_round32, end_flat, 51.63174, 1.0);
 
     TEST_BUFFER(rt_w35, join_miter, end_flat, 57.6569, 1.0);
+    TEST_BUFFER(rt_w36, join_miter, end_flat, 60.1274, 1.0);
+    TEST_BUFFER(rt_w37, join_miter, end_flat, 30.6569, 1.0);
+    TEST_BUFFER(rt_w38, join_miter, end_flat, 68.2279, 1.0);
+    TEST_BUFFER(rt_w39, join_miter, end_flat, 46.2279, 1.0);
+    TEST_BUFFER(rt_w40, join_miter, end_flat, 49.0490, 1.0);
```
barendgehrels, line 673: These new 5 all fail in `develop`

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
```diff
@@ -77,6 +79,44 @@ class turn_in_piece_visitor
...
+
+        auto it = m_clusters.find(turn.cluster_id);
+        if (it == m_clusters.end())
+        {
+            return false;
+        }
+
+        for (auto const& index : it->second.turn_indices)
+        {
+            if (index == turn.turn_index)
```
tinko92, line 100: This produces a comparison of integers of different signedness (long int and size_t) warning in algorithms_buffer_linestring_aimes.test and other tests.

tinko92 (commented):
The code looks good to me and catching something through the algorithm logic before predicate calls seems good.

When running the full test suite on my ARM64 machine, I get 10 new test case failures in algorithms_buffer_linestring_geo that I do not get on develop:

```
[...]/test_buffer_geo.hpp(83): [...] aimes_67_rr has too many points too close 1 0.25
[...]/test_buffer_geo.hpp(107): [...] aimes_67_rr the area is too small, expected at least 1.0038978240812866 15735.3694290676200406610455466882215
[...]/test_buffer_geo.hpp(83): [...] aimes_67_rf has too many points too close 1 0.25
[...]/test_buffer_geo.hpp(107): [...] aimes_67_rf the area is too small, expected at least 1.0038978240812866 13870.0487909986804213579532542697212
[...]/test_buffer_geo.hpp(107): [...] aimes_67_mf the area is too small, expected at least 1.0038978240812866 13870.0487909986804213579532542697212
[...]/test_buffer_geo.hpp(83): [...] aimes_163_rr has too many points too close 0 nan
[...]/test_buffer_geo.hpp(107): [...] aimes_163_rr the area is too small, expected at least 0 6078.31722878943595932505319884877138
[...]/test_buffer_geo.hpp(107): [...] aimes_181_rr the area is too small, expected at least 198.67884304160947 13456.4538307680239549585864667961117
[...]/test_buffer_geo.hpp(83): [...] aimes_163_rr has too many points too close 0 nan
[...]/test_buffer_geo.hpp(107): [...] aimes_163_rr the area is too small, expected at least 0 6078.31722878943595932505319884877138

*** 10 failures are detected in the test module "Test Program"
```

If they don't occur for you maybe this is related to long double being float128 on my machine and can be ignored and addressed later (since that also causes other errors on develop already on my machine).

barendgehrels:
> When running the full test suite on my ARM64 machine, I get 10 new test case failures

Thanks for testing! Indeed I don't get this. I've arm64 but not float128. 
`Darwin Kernel Version 23.6.0: Thu Mar  6 21:58:56 PST 2025; root:xnu-10063.141.1.704.6~1/RELEASE_ARM64_T6020 arm64`

> if they don't occur for you maybe this is related to long double being float128 on my machine and can be ignored and addressed later

Good you mention this. I tested the buffer algorithms but not yet the geo-ones. I will recheck.

The aimes tests are fragile (some of them) - indeed we might ignore them.

barendgehrels:
> Good you mention this. I tested the buffer algorithms but not yet the geo-ones. I will recheck.

All fine. The only check failing is `boost_geometry_index_rtree_exceptions_rst` - which also fails for me on `develop`. It is not related - I would be happy if it could be fixed or skipped

barendgehrels:
> If they don't occur for you maybe this is related to long double being float128 on my machine

The `long double` is not tested in the default test suite configuration...

#if ! defined(BOOST_GEOMETRY_TEST_ONLY_ONE_TYPE)
    test_linestring<bg::strategy::andoyer, true, bg::model::point<long double, 2, bg::cs::geographic<bg::degree> > >();
#endif

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
vissarion: Does this mean that this case was not covered by a unit test? Could we add one?

test/algorithms/buffer/buffer_multi_polygon.cpp
```diff
@@ -637,10 +634,7 @@ void test_all()
 
     TEST_BUFFER(rt_w20, join_miter, end_flat, 63.0269, 1.0);
     TEST_BUFFER(rt_w21, join_miter, end_flat, 26.3137, 1.0);
-
-#if defined(BOOST_GEOMETRY_TEST_FAILURES)
-    TEST_BUFFER(rt_w22, join_miter, end_flat, 86.0416, 1.0);
-#endif    
+    TEST_BUFFER(rt_w22, join_miter, end_flat, 86.1274, 1.0);
```
vissarion, line 637: There is a change in area here. Did the expected result changed?

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
```diff
@@ -77,6 +79,44 @@ class turn_in_piece_visitor
         return false;
     }
 
+    // Returns true if the turn is part of a cluster, and one of the other turns in the same
+    // cluster is involved in the same piece as the operations in the turn.
+    template <typename Turn, typename Piece>
+    inline bool skip_by_same_cluster(Turn const& turn, Piece const& piece) const
+    {
+        if (turn.cluster_id <= 0)
```
vissarion, line 87: what does a negative cluster_id means and why we are not searching for it? It is the case that m_clusters does not contain negative values so this condition speeds up the search?

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
barendgehrels: Good point. Apparently not. Those empty sides are rare. 

I'm not sure how to test it.

What happens with `false` is that partition bails out and does not verify further cases.
```
            if (! visitor.apply(**it1, **it2))
            {
                return false; // Bail out if visitor returns false
            }
```

It might be that this change fixes some issues for linestrings with flat-ends, in rare cases.

Currently I'm not available to dive deeper into it - but I can make a follow-up ticket.

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
barendgehrels: ✅

test/algorithms/buffer/buffer_multi_polygon.cpp
barendgehrels: 👍  Yes, the earlier area was wrong.

This is the new result:

image


While earlier (however, not in `develop` but probably in one of my branches) it did not include:

image

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
barendgehrels: 👍 I removed it, the condition was indeed redundant and wrong. Thanks.

include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
barendgehrels: Added #1408 for this situation

barendgehrels:
> When running the full test suite on my ARM64 machine (`Linux magicbook 5.15.167.4-microsoft-standard-WSL2 #1 SMP Tue Nov 5 00:19:58 UTC 2024 aarch64 GNU/Linux`), I get 10 new test case failures in algorithms_buffer_linestring_geo that I do not get on develop:
> 

They didn't run earlier with cmake. They didn't fail for me, but I added them to the failing cases set. It was probably for `double` as well.

Thanks for the report.

test/algorithms/buffer/buffer_linestring_geo.cpp
```diff
@@ -116,9 +116,11 @@ void test_linestring_aimes()
...
 #else
     // Cases where the algorithm is still failing.
-    std::set<int> const skip_cases_round_round{17, 22, 38, 181, 196};
-    std::set<int> const skip_cases_round_flat{17, 22, 38, 103, 196};
-    std::set<int> const skip_cases_miter_flat{17, 18, 22, 38, 103, 196};
+    // It might differ in debug/release mode.
+    // It might also depend on the compiler and on the operating system.
+    std::set<int> const skip_cases_round_round{17, 22, 38, 67, 75, 109, 163, 181, 143, 196};
+    std::set<int> const skip_cases_round_flat{17, 22, 38, 67, 75, 103, 109, 196};
+    std::set<int> const skip_cases_miter_flat{17, 18, 22, 38, 67, 75, 103, 109, 196};
```
barendgehrels, line 123: A few cases were added - not (necessarily) by this PR but by the fact that earlier I tested in release mode, and now in debug mode. I also added `67` and `163 reported by @tinko92

Issue #1406, opened 2025-05-09
Wrong intersection of polygon and linestring


vissarion:
The following code

```cpp
#include <iostream>
#include <vector>
#include <boost/geometry.hpp>

namespace bg = boost::geometry;

int main() {
    // Define the polygon
    bg::model::polygon<bg::model::d2::point_xy<double>> polygon;
    bg::read_wkt("POLYGON((1 1.5,3.5 4.5,4 6,2.5 4.5,1 1.5))", polygon);

    // Define the linestring
    bg::model::linestring<bg::model::d2::point_xy<double>> linestring;
    bg::read_wkt("LINESTRING(1.5 1.5,0.5 1.5)", linestring);

    // Compute the intersection
    std::tuple<
        bg::model::multi_point<bg::model::d2::point_xy<double>>,
        bg::model::multi_linestring<bg::model::linestring<bg::model::d2::point_xy<double>>>
    > output;
    bg::intersection(polygon, linestring, output);

    // Print the results
    auto linestrings = std::get<1>(output);
    auto points = std::get<0>(output);
    std::cout << "Intersection linestrings: " << bg::wkt(linestrings) << std::endl;
    std::cout << "Intersection points: " << bg::wkt(points) << std::endl;

    return 0;
}
```

Gives the following incorrect output:
```
Intersection linestrings: MULTILINESTRING((1 1.5,0.5 1.5))
Intersection points: MULTIPOINT()
```

on current develop with clang-15 on ubuntu 22.

![Image](https://github.com/user-attachments/assets/13c74026-7013-4a8e-8fbf-91e10dab3419)

vissarion:
Similar issue, intersection of `POLYGON((1 1,3 5,4 6,3 4,1 1))` with `LINESTRING(0 0,1 1)`. Now the geometries intersect in a common vertex.

PR #1409, opened 2025-05-21
feat: Implement is_valid algorithm for polyhedral surfaces

vissarion:
This PR introduces support for the `is_valid` algorithm for polyhedral surface geometries. A polyhedral_surface is a 3D composite geometry formed by a collection of polygonal faces that meet at shared edges. The implementation follows the OGC Simple Features Specification and introduces structural and topological checks for polyhedral surfaces.

The following properties should hold (following OGC standard):

- Contiguity: All polygon patches must form a connected surface via shared boundary segments.
- Shared Edges: Shared boundaries must be finite LineStrings used in at most two polygons.
- Orientation: Shared boundaries must be traversed in opposite directions by adjacent polygons; global orientation must be consistent.

Six new validity failure types has been introduced that reflect the description of the standard. 

Below there is a matrix with valid and invalid cases with visualizations. Those cases (and a few more) are included in the unit tests of this PR. 

| Name                                               | Geometry                                                                                                                                                                                                                                     | Figure |
|----------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------|
| valid_surface                                      | POLYHEDRALSURFACE(((0 0 0,0 1 0,1 1 0,1 0 0,0 0 0)),((0 0 0,0 0 1,0 1 1,0 1 0,0 0 0)),((0 0 0,1 0 0,1 0 1,0 0 1,0 0 0)),((1 1 1,0 1 1,0 0 1,1 0 1,1 1 1)),((1 1 1,1 0 1,1 0 0,1 1 0,1 1 1)),((1 1 1,1 1 0,0 1 0,0 1 1,1 1 1))) | ![Screenshot from 2025-05-21 16-52-54](https://github.com/user-attachments/assets/0ddb2629-bf4b-49ab-9992-ddfba7b1ee18) |
| valid_surface_non_convex                          | POLYHEDRALSURFACE(((0 0 0,0 1 0,1 1 0,1 0 0,0 0 0)),((0 0 0,0 0 1,0 1 1,0 1 0,0 0 0)),((0 0 0,1 0 0,1 0 1,0 0 1,0 0 0)),((1 1 1,0 1 1,0 0 1,1 0 1,1 1 1)),((1 1 1,1 0 1,1 0 0,1 1 0,1 .5 .5,1 1 1)))               | ![Screenshot from 2025-05-21 16-53-30](https://github.com/user-attachments/assets/954adfe2-efb4-413e-9758-fd55c65c3d9e) |
| valid_surface_holes                                | POLYHEDRALSURFACE(((0 0 0,0 1 0,1 1 0,1 0 0,0 0 0),(0.2 0.2 0,0.2 0.5 0,0.5 0.5 0,0.5 0.2 0,0.2 0.2 0)),((0 0 0,0 0 1,0 1 1,0 1 0,0 0 0)),((0 0 0,1 0 0,1 0 1,0 0 1,0 0 0)),((1 1 1,0 1 1,0 0 1,1 0 1,1 1 1)),((1 1 1,1 0 1,1 0 0,1 1 0,1 1 1)),((1 1 1,1 1 0,0 1 0,0 1 1,1 1 1))) | ![Screenshot from 2025-05-21 16-55-21](https://github.com/user-attachments/assets/72a00aa7-cf2f-4249-9805-21cbb63c0f47)  |
| invalid_surface_non_planar_face                    | POLYHEDRALSURFACE(((0 0 0,1 0 0,0 1 0,0 0 1,0 0 0)),((1 0 0,1 1 0,0 1 0,1 0 0)))                                                                                                                        | ![Screenshot from 2025-05-21 17-20-01](https://github.com/user-attachments/assets/f0ac921c-6bbb-4d00-b3fd-36267e1d702d)   |
| invalid_surface_collinear_points                   | POLYHEDRALSURFACE(((1 0 0,1 1 0,0 1 0,1 0 0)),((0 0 0,1 0 0,2 0 0,0 0 0)))                                                                                                                              |  ![Screenshot from 2025-05-21 17-10-21](https://github.com/user-attachments/assets/0cbc3240-3088-44ba-b90d-bdfd87ce296c)     |
| invalid_surface_few_points                         | POLYHEDRALSURFACE(((1 0 0,1 1 0,0 1 0,1 0 0)),((0 0 0,1 0 0,0 0 0)))                                                                                                                                    |  ![Screenshot from 2025-05-21 17-11-01](https://github.com/user-attachments/assets/160455b0-b0a6-4393-9ff8-1e9caf368e29)     |
| invalid_surface_invalid_intersection_boundary_interior | POLYHEDRALSURFACE(((0 0 0,1 0 0,0 0 1,0 0 0)),((0 1 0,0.5 -0.5 0.5,1 1 0,0 1 0)))                                                                                                                   | ![Screenshot from 2025-05-21 17-11-20](https://github.com/user-attachments/assets/7805e68c-d576-414d-90d7-999d1ff319f1)      |
| invalid_surface_invalid_intersection_boundary_boundary | POLYHEDRALSURFACE(((0 0 0,1 0 0,1 0 1,0 0 1,0 0 0)),((0 1 0,0 -1 1,1 -1 1,1 1 0,0 1 0)))                                                                                                               | ![Screenshot from 2025-05-21 17-11-36](https://github.com/user-attachments/assets/abe12aea-b14c-480d-8a46-86f47abc980a)      |
| invalid_surface_invalid_intersection_in_common_vertex | POLYHEDRALSURFACE(((0 0 0,1 0 0,1 0 1,0 0 1,0 0 0)),((0 1 0,1 0 1,1 1 0,0 1 0)))                                                                                                                      |  ![Screenshot from 2025-05-21 17-11-57](https://github.com/user-attachments/assets/d702ef12-04b6-4d83-8528-172449472010)     |
| invalid_surface_invalid_intersection_partial_edge   | POLYHEDRALSURFACE(((0 0 0,1 0 0,1 0 1,0 0 1,0 0 0)),((0 1 0,0.5 0 1,1 0 1,1 1 0,0 1 0)))                                                                                                                | ![Screenshot from 2025-05-21 17-12-11](https://github.com/user-attachments/assets/a962160f-af17-40b1-85a2-357d514a97b1)      |
| invalid_surface_disconnected_polygons              | POLYHEDRALSURFACE(((0 0 0,1 0 0,0 1 0,0 0 0)),((1 1 0,2 1 0,2 2 0,1 2 0,1 1 0)))                                                                                                                      |  ![Screenshot from 2025-05-21 17-19-31](https://github.com/user-attachments/assets/909293eb-e24a-4593-8998-8eb6cbf34365)     |
| invalid_surface_invalid_intersection_vertex_edge    | POLYHEDRALSURFACE(((0 0 0,1 0 0,0 1 0,0 0 0)),((0.5 0.5 0,2 1 0,2 2 0,1 2 0,0.5 0.5 0)))                                                                                                                | ![Screenshot from 2025-05-21 17-12-52](https://github.com/user-attachments/assets/67035568-5a18-433d-b450-6906c43465a6)      |
| invalid_surface_invalid_intersection_vertex_vertex  | POLYHEDRALSURFACE(((0 0 0,1 0 0,0 1 0,0 0 0)),((1 0 0,2 1 0,2 2 0,1 2 0,1 0 0)))                                                                                                                        | ![Screenshot from 2025-05-21 17-13-07](https://github.com/user-attachments/assets/ea1c1288-3239-429b-acaa-744377128b16)       |
| invalid_surface_invalid_intersection_parallel_edges | POLYHEDRALSURFACE(((0 0 0,1 0 0,0 1 0,0 0 0)),((0 1 0,0.5 0.5 0,1 0 0,2 1 0,2 2 0,1 2 0,0 1 0)))                                                                                                        | ![Screenshot from 2025-05-21 17-13-24](https://github.com/user-attachments/assets/c8d158f6-ff5f-4ee9-920c-b8055502eeea)       |
| invalid_surface_invalid_intersection_operlapping_faces | POLYHEDRALSURFACE(((0 0 0,1 0.5 0,0.5 1 0,0 0 0)),((0 1 0,1 0 0,2 1 0,2 2 0,1 2 0,0 1 0)))                                                                                                           | ![Screenshot from 2025-05-21 17-13-45](https://github.com/user-attachments/assets/5f488f7a-464e-4782-8767-4ac2718782bd)      |
| invalid_surface_invalid_intersection_operlapping_faces2 | POLYHEDRALSURFACE(((0 0 0,1 0 0,0.5 1 0,0 0 0)),((0 1 0,1 0 0,2 1 0,2 2 0,1 2 0,0 1 0)))                                                                                                             | ![Screenshot from 2025-05-21 17-14-04](https://github.com/user-attachments/assets/ba5d7bad-04a1-4fe7-b1b5-b0a819b09b08)      |
| invalid_surface_invalid_face                        | POLYHEDRALSURFACE(((0 0 0,1 0 0,0 1 0,0 0 0)),((0 1 0,1 0 0,2 2 0,2 1 0,1 2 0,0 1 0)))                                                                                                                  | ![Screenshot from 2025-05-21 17-14-25](https://github.com/user-attachments/assets/61c2c2e4-0960-4fde-9de5-e05b467a66ab)      |
| invalid_surface_non_manifold_edge                  | POLYHEDRALSURFACE(((0 0 0,1 0 0,0 1 0,0 0 0)),((0 1 0,1 0 0,2 1 0,2 2 0,1 2 0,0 1 0)),((0 1 0,1 0 0,1 1 1,0 1 0)))                                                                                       | ![Screenshot from 2025-05-21 17-14-43](https://github.com/user-attachments/assets/47c1fbeb-f99c-4a4e-a099-14d07ca69af3)      |

Two unit tests are failing and are commented out. Those are dependent on https://github.com/boostorg/geometry/issues/1406

### DIFF
--- a/include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
+++ b/include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
@@ -1,0 +1,651 @@
+// Boost.Geometry
+
+// Copyright (c) 2025 Oracle and/or its affiliates.
+// Contributed and/or modified by Vissarion Fysikopoulos, on behalf of Oracle
+
+// Use, modification and distribution is subject to the Boost Software License,
+// Version 1.0. (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_GEOMETRY_ALGORITHMS_DETAIL_IS_VALID_POLYHEDRAL_SURFACE_HPP
+#define BOOST_GEOMETRY_ALGORITHMS_DETAIL_IS_VALID_POLYHEDRAL_SURFACE_HPP
+
+#include <boost/geometry/algorithms/buffer.hpp>
+#include <boost/geometry/algorithms/correct.hpp>
+#include <boost/geometry/algorithms/detail/is_valid/interface.hpp>
+#include <boost/geometry/algorithms/detail/is_valid/polygon.hpp>
+#include <boost/geometry/algorithms/detail/calculate_point_order.hpp>
+#include <boost/geometry/algorithms/distance.hpp> //needed for rtree
+#include <boost/geometry/index/predicates.hpp>
+#include <boost/geometry/index/rtree.hpp>
+#include <boost/geometry/io/wkt/wkt.hpp>
+#include <boost/geometry/iterators/point_reverse_iterator.hpp>
+
+#include <vector>
+#include <stack>
+
+namespace boost { namespace geometry
+{
+
+#ifndef DOXYGEN_NO_DETAIL
+namespace detail { namespace is_valid
+{
+
+struct polygon_frontier_type
+{
+    polygon_frontier_type(std::size_t size)
+        : m_stack(), m_visited(size, false)
+    {}
+
+    void push(std::size_t index)
+    {
+        if (!m_visited[index])
+        {
+            m_stack.push(index);
+            m_visited[index] = true;
+        }
+    }
+    bool empty()
+    {
+        return m_stack.empty();
+    }
+    std::size_t top()
+    {
+        return m_stack.top();
+    }
+    void pop()
+    {
+        m_stack.pop();
+    }
+private :
+    std::stack<std::size_t> m_stack;
+    std::vector<bool> m_visited;
+};
+
+// Implementation of the polygon segment intersection algorithm
+// This function checks if a segment (s1, s2) intersects with a 2D polygon in 3D space.
+// proj1 and proj2 are the indices of the coordinate axes used to determine the sign of three points
+// of the polygon after projection.
+// It returns:
+// - 0 if the segment does not intersect the polygon
+// - 1 if the segment intersects the polygon in the interior
+// - 2 if the segment intersects the polygon at a vertex
+// - 3 if the segment intersects the polygon at an edge
+// Reference: R.J.Segura, F.R.Feito - An algorithm for determining intersection segment-polygon in
+// 3D, Computers & Graphics 1998
+template <typename Point, typename Polygon, typename Strategy>
+int polygon_segment_intersection(Point const& s1,
+                                 Point const& s2,
+                                 Polygon const& polygon,
+                                 int proj1,
+                                 int proj2,
+                                 Strategy const& strategy)
+{
+    BOOST_GEOMETRY_ASSERT(geometry::dimension<Point>::value == 3);
+    BOOST_GEOMETRY_ASSERT(geometry::dimension<Polygon>::value == 3);
+
+    auto const pit_a = points_begin(polygon);
+    auto const& p_a = *pit_a;
+
+    int cut = 0;
+
+    for (auto pit = boost::next(pit_a); pit != boost::prior(points_end(polygon)); ++pit)
+    {
+        auto const& p_b = *pit;
+        auto const& p_c = *boost::next(pit);
+
+        int const sign = strategy.side().apply(p_a, p_b, p_c, proj1, proj2);
+
+        int const sign1 = strategy.side3().apply(s1, p_a, s2, p_b);
+        int const sign2 = strategy.side3().apply(s1, p_c, p_b, s2);
+        int const sign3 = strategy.side3().apply(s1, p_a, p_c, s2);
+
+        // Insects at p_b or p_a or p_c -> vertex
+        if ((sign1 == 0 && sign2 == 0) || (sign1 == 0 && sign3 == 0) || (sign2 == 0 && sign3 == 0))
+        {
+            return 2;
+        }
+
+        // Intersects at edge p_a, p_c -> internal edge
+        if (sign1 == 0 && sign2 == sign3) cut += sign;
+
+        // Intersects at edge p_b, p_c -> external edge
+        if (sign2 == 0 && sign1 == sign3) return 3;
+
+        // Intersects at edge p_a, p_b -> internal edge
+        if (sign3 == 0 && sign1 == sign2) cut += sign;
+
+        // Intersects inside p_a, p_b, p_c -> inside
+        if (sign1 == sign2 && sign2 == sign3) cut += 2 * sign;
+    }
+
+    return (cut == 2) ? 1 : 0;
+}
+
+template <typename PolyhedralSurface>
+struct valid_intersection
+{
+    valid_intersection(std::size_t num_of_polygons)
+        : m_polygon_projections(num_of_polygons)
+    {}
+
+    using coordinate_type = typename coordinate_type<PolyhedralSurface>::type;
+    using polygon_3d = typename boost::range_value<PolyhedralSurface>::type;
+    using ring_3d = typename boost::geometry::ring_type<polygon_3d>::type;
+    using point_3d = typename boost::geometry::point_type<ring_3d>::type;
+    using segment_3d = typename boost::geometry::model::segment<point_3d>;
+    using box_3d = typename boost::geometry::model::box<point_3d>;
+
+    using point_2d = boost::geometry::model::point
+        <coordinate_type, 2, boost::geometry::cs::cartesian>;
+    using linestring_2d = typename boost::geometry::model::linestring<point_2d>;
+    using polygon_2d = typename std::conditional
+        <
+            boost::geometry::point_order<polygon_3d>::value == boost::geometry::counterclockwise,
+            typename boost::geometry::model::polygon<point_2d, true>,
+            typename boost::geometry::model::polygon<point_2d, false>
+        >::type;
+    using ring_2d = typename boost::geometry::ring_type<polygon_2d>::type;
+
+private:
+
+    enum class intersection_type
+    {
+        none,
+        valid,
+        invalid
+    };
+
+    struct polygon_projection
+    {
+        polygon_2d projected_polygon;
+        int p1_idx;
+        int p2_idx;
+        int p3_idx;
+        int proj1;
+        int proj2;
+        int dir;
+    };
+
+    template <typename Strategy>
+    static std::tuple<int, int, int> affinely_independent(point_3d const& p1,
+                                                          point_3d const& p2,
+                                                          point_3d const& p3,
+                                                          Strategy const& strategy)
+    {
+        int side = strategy.side().apply(p1, p2, p3, 0, 1);
+        if (side != 0) return {0, 1, side};
+
+        side = strategy.side().apply(p1, p2, p3, 0, 2);
+        if (side != 0) return {0, 2, side};
+
+        side = strategy.side().apply(p1, p2, p3, 1, 2);
+        if (side != 0) return {1, 2, side};
+
+        return {-1, -1, 0}; // All points are collinear
+    }
+
+    template <typename Strategy>
+    static std::tuple<int, int, int> find_non_collinear_points(polygon_3d const& polygon,
+                                                               int& p1_index,
+                                                               int& p2_index,
+                                                               int& p3_index,
+                                                               Strategy const& strategy)
+    {
+        for (auto pit1 = points_begin(polygon); pit1 != points_end(polygon); ++pit1)
+        {
+            for (auto pit2 = boost::next(pit1); pit2 != points_end(polygon); ++pit2)
+            {
+                for (auto pit3 = boost::next(pit2); pit3 != points_end(polygon); ++pit3)
+                {
+                    // A set of 3 independent points is sufficient to construct a plane.
+                    auto res = affinely_independent(*pit1, *pit2, *pit3, strategy);
+                    if (std::get<0>(res) != -1)
+                    {
+                        p1_index = std::distance(points_begin(polygon), pit1);
+                        p2_index = std::distance(points_begin(polygon), pit2);
+                        p3_index = std::distance(points_begin(polygon), pit3);
+                        return res;
+                    }
+                }
+            }
+        }
+        return {-1, -1, 0}; // All points are collinear
+    }
+
+    template <typename VisitPolicy>
+    static bool has_correct_intersection_orientation(polygon_2d const& polygon,
+                                                     point_2d const& segment_point1,
+                                                     point_2d const& segment_point2,
+                                                     VisitPolicy& visitor)
+    {
+        if (boost::geometry::equals(segment_point1, segment_point2))
+        {
+            visitor.template apply<failure_duplicate_points>();
+            return false;
+        }
+
+        boost::geometry::closing_iterator<ring_2d> it(boost::geometry::exterior_ring(polygon));
+        auto begin = it;
+        boost::geometry::closing_iterator<ring_2d> end(boost::geometry::exterior_ring(polygon), true);
+
+        for (; it != end - 1; ++it)
+        {
+            if (boost::geometry::equals(*it, segment_point1))
+            {
+                auto prev = (it == begin) ? end - 1 : boost::prior(it);
+
+                if (boost::geometry::equals(segment_point2, *prev))
+                {
+                    // valid intersection and consistent orientation
+                    return true;
+                }
+
+                auto next = (it == end) ? begin : boost::next(it);
+
+                if (boost::geometry::equals(segment_point2, *next))
+                {
+                    // valid intersection but inconsistent orientation
+                    visitor.template apply<failure_inconsistent_orientation>();
+                    return false;
+                }
+            }
+        }
+
+        visitor.template apply<failure_invalid_intersection>();
+
+        return false;
+    }
+
+    point_2d project_point_to_coordinate_plane(point_3d const& point, int proj1, int proj2)
+    {
+        if (proj1 == 1 && proj2 == 2) return point_2d{geometry::get<1>(point), geometry::get<2>(point)};
+        if (proj1 == 0 && proj2 == 2) return point_2d{geometry::get<0>(point), geometry::get<2>(point)};
+        if (proj1 == 0 && proj2 == 1) return point_2d{geometry::get<0>(point), geometry::get<1>(point)};
+        return point_2d{};
+    }
+
+    // Check if a segment intersects a polygon in 3D space
+    template <typename VisitPolicy, typename Strategy>
+    intersection_type check_polygon_segment_intersection(point_3d const& segment_point1,
+                                                         point_3d const& segment_point2,
+                                                         polygon_3d const& polygon,
+                                                         std::size_t const polygon_index,
+                                                         VisitPolicy& visitor,
+                                                         Strategy const& strategy)
+    {
+        auto const& projection = m_polygon_projections[polygon_index];
+        auto const& projected_polygon = projection.projected_polygon;
+
+        auto const& p1 = polygon.outer()[projection.p1_idx];
+        auto const& p2 = polygon.outer()[projection.p2_idx];
+        auto const& p3 = polygon.outer()[projection.p3_idx];
+
+        int proj1 = projection.proj1;
+        int proj2 = projection.proj2;
+        int dir = projection.dir;
+
+        // Check if the segment intersects the plane of the polygon
+        auto sign1 = strategy.side3().apply(p1, p2, p3, segment_point1);
+        auto sign2 = strategy.side3().apply(p1, p2, p3, segment_point2);
+
+        if ((sign1 > 0 && sign2 > 0) || (sign1 < 0 && sign2 < 0))
+        {
+            return intersection_type::none; // No intersection, this is valid
+        }
+
+        // Check if the segment is on the same plane as the polygon
+        if (sign1 == 0 && sign2 == 0)
+        {
+            auto const& projected_p1 = project_point_to_coordinate_plane(segment_point1, proj1, proj2);
+            auto const& projected_p2 = project_point_to_coordinate_plane(segment_point2, proj1, proj2);
+
+            linestring_2d projected_segment = {projected_p1, projected_p2};
+
+            // Check if the projected segment intersects the polygon
+            std::tuple
+            <
+                boost::geometry::model::multi_point<point_2d>,
+                boost::geometry::model::multi_linestring<linestring_2d>> intersections;
+            boost::geometry::intersection(projected_segment, projected_polygon, intersections);
+
+            auto const& points = std::get<0>(intersections);
+            auto const& linestrings = std::get<1>(intersections);
+
+            if (boost::geometry::is_empty(points) && boost::geometry::is_empty(linestrings))
+            {
+                return intersection_type::none; // No intersection, this is valid
+            }
+
+            if (!points.empty())
+            {
+                if (points.size() == 1 &&
+                    (boost::geometry::equals(points.front(), projected_p1) ||
+                    boost::geometry::equals(points.front(), projected_p2)))
+                {
+                    return intersection_type::none; // Valid empty intersection
+                }
+                visitor.template apply<failure_invalid_intersection>();
+                return intersection_type::invalid; // Invalid intersection
+            }
+
+            if (linestrings.size() == 1)
+            {
+                if (boost::geometry::equals(linestrings.front(), projected_segment))
+                {
+                    auto const& pp1 = dir < 0 ? projected_p2 : projected_p1;
+                    auto const& pp2 = dir < 0 ? projected_p1 : projected_p2;
+                    if (!has_correct_intersection_orientation(projected_polygon, pp1, pp2, visitor))
+                    {
+                        return intersection_type::invalid;
+                    }
+                    return intersection_type::valid; // The only case of valid non empty intersection
+                }
+            }
+            visitor.template apply<failure_invalid_intersection>();
+            return intersection_type::invalid;
+        }
+
+        int inter = polygon_segment_intersection(segment_point1, segment_point2, polygon, proj1,
+            proj2, strategy);
+
+        if ((inter == 1 || inter == 3) || (inter == 2 && sign1 !=0 && sign2 != 0))
+        {
+            visitor.template apply<failure_invalid_intersection>();
+            return intersection_type::invalid;
+        }
+
+        return intersection_type::none;
+    }
+
+    // Helper function to check if any segment of a ring intersects a polygon
+    template <typename VisitPolicy, typename Strategy>
+    intersection_type check_polygon_polygon_intersection(polygon_3d const& polygon1,
+                                                         polygon_3d const& polygon2,
+                                                         std::size_t const& index2,
+                                                         VisitPolicy& visitor,
+                                                         Strategy const& strategy)
+    {
+        bool valid_intersection = false;
+        auto const& ring1 = boost::geometry::exterior_ring(polygon1);
+
+        for (auto it = boost::begin(ring1); it + 1 != boost::end(ring1); ++it)
+        {
+            auto result = check_polygon_segment_intersection(*it, *(it + 1), polygon2, index2,
+                visitor, strategy);
+
+            if (result == intersection_type::invalid)
+            {
+                return intersection_type::invalid; // Found invalid intersection
+            }
+
+            if (result == intersection_type::valid)
+            {
+                valid_intersection = true; // Found a valid intersection
+            }
+        }
+        // Either no intersection or valid intersection
+        return valid_intersection ? intersection_type::valid : intersection_type::none;
+    }
+
+ public:
+
+    template <typename VisitPolicy, typename Strategy>
+    intersection_type apply(polygon_3d const& polygon1,
+                            polygon_3d const& polygon2,
+                            std::size_t const& index1,
+                            std::size_t const& index2,
+                            VisitPolicy& visitor,
+                            Strategy const& strategy)
+    {
+        auto intersection1 = check_polygon_polygon_intersection(polygon1, polygon2, index2,
+            visitor, strategy);
+
+        if (intersection1 == intersection_type::invalid)
+        {
+            return intersection_type::invalid; // Found invalid intersection
+        }
+
+        auto intersection2 = check_polygon_polygon_intersection(polygon2, polygon1, index1,
+            visitor, strategy);
+
+        if (intersection2 == intersection_type::invalid)
+        {
+            return intersection_type::invalid; // Found invalid intersection
+        }
+
+        if (intersection1 == intersection_type::none && intersection2 == intersection_type::none)
+        {
+            return intersection_type::none; // No intersection found
+        }
+
+        return intersection_type::valid;
+    }
+
+    template <typename VisitPolicy, typename Strategy>
+    intersection_type initialize(polygon_3d const& polygon,
+                                 std::size_t const polygon_index,
+                                 VisitPolicy& visitor,
+                                 Strategy const& strategy)
+    {
+        // TODO: Support interior rings
+        ring_3d const& ring = boost::geometry::exterior_ring(polygon);
+
+        if (boost::size(ring) < detail::minimum_ring_size<ring_3d>::value)
+        {
+            visitor.template apply<failure_few_points_on_face>();
+            return intersection_type::invalid;
+        }
+
+        // Get three non-collinear points from the polygon to define the plane
+        int p1_idx = 0;
+        int p2_idx = 0;
+        int p3_idx = 0;
+
+        auto res = find_non_collinear_points(polygon, p1_idx, p2_idx, p3_idx, strategy);
+
+        if (std::get<0>(res) == -1)
+        {
+            visitor.template apply<failure_collinear_points_on_face>();
+            return intersection_type::invalid;
+        }
+
+        int proj1 = std::get<0>(res);
+        int proj2 = std::get<1>(res);
+        int dir = std::get<2>(res);
+
+        auto const& p1 = polygon.outer()[p1_idx];
+        auto const& p2 = polygon.outer()[p2_idx];
+        auto const& p3 = polygon.outer()[p3_idx];
+
+        // Check for non-coplanar points
+        for (auto pit = points_begin(polygon); pit != points_end(polygon); ++pit)
+        {
+            if (strategy.side3().apply(p1, p2, p3, *pit) != 0)
+            {
+                visitor.template apply<failure_non_coplanar_points_on_face>();
+                return intersection_type::invalid;
+            }
+        }
+
+        // Project the polygon
+        polygon_2d projected_polygon;
+
+        auto add_projected_points = [&](auto begin_it, auto end_it)
+        {
+            for (auto it = begin_it; it != end_it; ++it)
+            {
+                projected_polygon.outer().emplace_back(
+                    project_point_to_coordinate_plane(*it, proj1, proj2));
+            }
+        };
+
+        if (dir > 0) {
+            add_projected_points(points_begin(ring), points_end(ring));
+        } else {
+            add_projected_points(points_rbegin(ring), points_rend(ring));
+        }
+
+        // Check if the projected polygon is valid
+        if (!resolve_dynamic::is_valid<polygon_2d>::apply(projected_polygon, visitor, strategy))
+        {
+            return intersection_type::invalid;
+        }
+
+        // Cache the projection for future use
+        m_polygon_projections[polygon_index] = polygon_projection{std::move(projected_polygon),
+            p1_idx, p2_idx, p3_idx, proj1, proj2, dir};
+
+        return intersection_type::valid;
+    }
+
+private:
+    std::vector<polygon_projection> m_polygon_projections;
+};
+
+template <typename PolyhedralSurface>
+class is_valid_polyhedral_surface
+{
+public:
+    template <typename VisitPolicy, typename Strategy>
+    static inline bool apply(PolyhedralSurface const& surface,
+                             VisitPolicy& visitor,
+                             Strategy const& strategy)
+    {
+        boost::ignore_unused(strategy);
+
+        valid_intersection<PolyhedralSurface> valid_intersection(surface.size());
+
+        // Check for colinear/coplanar points and compute projections
+        for (std::size_t i = 0; i < surface.size(); ++i)
+        {
+            auto intersection = valid_intersection.initialize(surface[i], i, visitor, strategy);
+
+            using intersection_type = decltype(intersection);
+
+            if (intersection == intersection_type::invalid)
+            {
+                return false;
+            }
+        }
+
+        // Create an rtree to use it for polygon pair intersections
+        using box_3d = typename boost::geometry::detail::is_valid::valid_intersection
+            <PolyhedralSurface>::box_3d;
+        using value = std::pair<box_3d, std::size_t>;
+
+        boost::geometry::index::rtree<value, boost::geometry::index::rstar<4>> rtree;
+        std::vector<std::unique_ptr<box_3d>> boxes(surface.size());
+
+        for (std::size_t i = 0; i < surface.size(); ++i)
+        {
+            auto b = boost::geometry::return_envelope<box_3d>(surface[i], strategy);
+            boxes[i] = std::make_unique<box_3d>(b);
+            rtree.insert(std::make_pair(b, i)); // use balancing algorithm
+        }
+
+        // We check pairs of polygons for having a valid intersection.
+        // We iterate in a BFS manner to check connectivity of the surface.
+
+        auto it1 = boost::begin(surface);
+
+        polygon_frontier_type polygon_frontier(surface.size());
+        polygon_frontier.push(0);
+
+        std::vector<std::size_t> remaining_polygons(surface.size());
+        std::vector<std::size_t> positions(surface.size());
+
+        for (std::size_t i = 0; i < surface.size(); ++i)
+        {
+            remaining_polygons[i] = i;
+            positions[i] = i;
+        }
+
+        int remaining_count = surface.size();
+
+        while (!polygon_frontier.empty())
+        {
+            auto index1 = polygon_frontier.top();
+            polygon_frontier.pop();
+
+            it1 = boost::begin(surface) + index1;
+
+            // Remove index from remaining polygons and update positions
+            auto moving_index = remaining_polygons[remaining_count - 1];
+            remaining_polygons[positions[index1]] = moving_index;
+            positions[moving_index] = positions[index1];
+
+            // Query all polygons that intersect with the current polygon.
+            // We use a buffer to avoid precision issues with the rtree query.
+            std::vector<value> intersecting_boxes;
+            auto const& b_query = *boxes[index1];
+            auto const& b_query_buf = boost::geometry::return_buffer<box_3d>(b_query, 0.0001);
+
+            rtree.query(boost::geometry::index::intersects(b_query_buf),
+                        std::back_inserter(intersecting_boxes));
+
+            for (int i = 0; i < remaining_count - 1; ++i)
+            {
+                auto index2 = remaining_polygons[i];
+
+                // Skip if index2 is not in the intersecting boxes
+                if (std::find_if(intersecting_boxes.begin(), intersecting_boxes.end(),
+                    [index2](const value& v) { return v.second == index2; }
+                ) == intersecting_boxes.end())
+                {
+                    continue;
+                }
+
+                auto intersection = valid_intersection.apply(*it1, surface[index2], index1, index2,
+                    visitor, strategy);
+
+                using intersection_type = decltype(intersection);
+
+                if (intersection == intersection_type::invalid)
+                {
+                    return false; //  Invalid intersection found (including inconsistent orientation)
+                }
+                else if (intersection == intersection_type::valid)
+                {
+                    polygon_frontier.push(index2);
+                }
+            }
+            remaining_count--;
+        }
+
+        if (remaining_count > 0)
+        {
+            visitor.template apply<failure_disconnected_surface>();
+            return false; // Not all polygons were visited, hence disconnected
+        }
+
+        return true;
+    }
+};
+
+}} // namespace detail::is_valid
+#endif // DOXYGEN_NO_DETAIL
+
+#ifndef DOXYGEN_NO_DISPATCH
+namespace dispatch
+{
+
+
+// Reference (for validity of Polyhedral Surfaces): OGC
+// OpenGIS® Implementation Standard for Geographic
+// information - Simple feature access - Part 1: Common
+// architecture: 6.1.12
+template <typename PolyhedralSurface, bool AllowEmptyMultiGeometries>
+struct is_valid
+    <
+        PolyhedralSurface, polyhedral_surface_tag, AllowEmptyMultiGeometries
+    > : detail::is_valid::is_valid_polyhedral_surface<PolyhedralSurface>
+{};
+
+
+} // namespace dispatch
+#endif // DOXYGEN_NO_DISPATCH
+
+}} // namespace boost::geometry
+
+#endif // BOOST_GEOMETRY_ALGORITHMS_DETAIL_IS_VALID_POLYHEDRAL_SURFACE_HPP
--- a/include/boost/geometry/algorithms/detail/overlay/graph/assign_counts.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/graph/assign_counts.hpp
@@ -80,22 +80,22 @@ void assign_counts(Turn& turn)
         }
     };
 
-    auto assign_left_for = [&turn, &assign_for](counts_per_op_t const& op1, counts_per_op_t op2)
+    auto assign_left_for = [&assign_for](counts_per_op_t const& op1, counts_per_op_t op2)
     {
         assign_for(op1, op2, [](auto& enriched, auto count) { enriched.count_left = count; });
     };
 
-    auto assign_right_for = [&turn, &assign_for](counts_per_op_t const& op1, counts_per_op_t op2)
+    auto assign_right_for = [&assign_for](counts_per_op_t const& op1, counts_per_op_t op2)
     {
         assign_for(op1, op2, [](auto& enriched, auto count) { enriched.count_right = count; });
     };
 
-    auto assign_left_incoming_for = [&turn, &assign_for](counts_per_op_t const& op1, counts_per_op_t op2)
+    auto assign_left_incoming_for = [&assign_for](counts_per_op_t const& op1, counts_per_op_t op2)
     {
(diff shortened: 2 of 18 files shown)


include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,462 @@
...
+            {
+                if (boost::geometry::equals(point1, point2))
+                {
+                    visitor.template apply<failure_duplicate_points>();
+                    return false;
+                }
+                auto prev = (it == boost::begin(ring)) ? boost::end(ring) - 1 : boost::prior(it);
+                if (boost::geometry::equals(point2, *prev))
+                {
+                    // valid intersection and inconsistent orientation
```
tinko92, line 133: Should this be ~~in~~consistent?

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,462 @@
...
+                    set_vec_from_point::apply(*pit3, v2);
+                    if (!are_linearly_dependent(v1 - base, v2 - base))
+                    {
+                        return true;
+                    }
+                }
+            }
+        }
+        return false; // All points are collinear
+    }
```
tinko92, line 102: I think by the library design philosophy and also practically, if one wants to be able to plug in a robust computation later, this is a cartesian strategy.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,462 @@
...
+        vec3 base;
+        vec3 v1;
+        vec3 v2;
+        vec3 normal;
+    };
+
+    // Check if two vectors are linearly dependent (collinear)
+    static bool are_linearly_dependent(vec3 const& a, vec3 const& b)
+    {
+        return qvm::mag_sqr(qvm::cross(a, b)) < tolerance * tolerance;
```
tinko92, line 79: We may want some factor on this bound, I will recheck.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,462 @@
...
+        vec3 base;
+        vec3 v1;
+        vec3 v2;
+        vec3 normal;
+        polygon_2d projected_polygon;
+
+        // Check if the polygon has been projected before
+        std::ostringstream oss;
+        oss << boost::geometry::wkt(polygon);
+        if (m_polygon_projections.find(oss.str()) != m_polygon_projections.end())
```
tinko92, line 178: This seems rather expensive aside from pulling the wkt dependency into this. Can we use the indices of the polygons in their original datastracture as integer keys instead? (I understand that all boost geometry concepts require random access ranges).

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,462 @@
...
+    {
+        boost::ignore_unused(strategy);
+        bool disconnected = true;
+        // TODO: improve perofmance by using a spatial index
+        for (auto it1 = boost::begin(surface); it1 != boost::end(surface); ++it1)
+        {
+            for (auto it2 = boost::begin(surface); it2 != boost::end(surface); ++it2)
+            {
+                if (std::distance(boost::begin(surface), it1) !=
+                    std::distance(boost::begin(surface), it2))
```
tinko92, line 407: Is this different from it1 != it2?

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,462 @@
...
+{
+public:
+    template <typename VisitPolicy, typename Strategy>
+    static inline bool apply(PolyhedralSurface const& surface,
+                             VisitPolicy& visitor,
+                             Strategy const& strategy)
+    {
+        boost::ignore_unused(strategy);
+        bool disconnected = true;
+        // TODO: improve perofmance by using a spatial index
```
tinko92, line 401: Typo: performance

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,462 @@
...
+    using vec3 = boost::qvm::vec<CT, 3>;
+    using polygon_3d = typename boost::range_value<PolyhedralSurface>::type;
+    using ring_3d = typename boost::geometry::ring_type<polygon_3d>::type;
+    using point_3d = typename boost::range_value<ring_3d>::type;
+    using segment_3d = typename boost::geometry::model::segment<point_3d>;
+    using set_vec_from_point = typename strategy::transform::detail::matrix_transformer::
+        set_vec_from_point<point_3d, 0, 3>;
+
+    using point_2d = boost::geometry::model::point
+        <CT, 2, boost::geometry::cs::cartesian>;
```
tinko92, line 44: Are polyhedral surfaces always cartesian (I didn't see a corresponding constraint in the concept check https://github.com/boostorg/geometry/blob/develop/include/boost/geometry/geometries/concepts/polyhedral_surface_concept.hpp but I may have missed it) and if not should this be generic or is_valid constrained for now?

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,462 @@
...
+        vec3 v2;
+        vec3 normal;
+        polygon_2d projected_polygon;
+
+        // Check if the polygon has been projected before
+        std::ostringstream oss;
+        oss << boost::geometry::wkt(polygon);
+        if (m_polygon_projections.find(oss.str()) != m_polygon_projections.end())
+        {
+            auto& projection = m_polygon_projections[oss.str()];
```
tinko92, line 180: Minor: This seems to be only copied from, so maybe it can be declared const.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,462 @@
...
+                set_vec_from_point::apply(*pit, vi);
+                auto vol = qvm::dot(normal, vi)- dot(normal, base);
+                if (std::abs(vol) > tolerance)
+                {
+                    visitor.template apply<failure_non_coplanar_points_on_face>();
+                    return intersection_type::invalid;
+                }
+            }
+            // Project the polygon onto the 2D plane defined by v1 and v2
+            for (auto& point : ring)
```
tinko92, line 224: This point seems to only be read from.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,462 @@
...
+                }
+            }
+            visitor.template apply<failure_invalid_intersection>();
+            return intersection_type::invalid;
+        }
+
+        // Compute the intersection point
+        vec3 seg_dir = p2 - p1;
+        CT denom = dot(normal, seg_dir);
+        assert(denom != 0); // Ensure the segment is not parallel to the plane
```
tinko92, line 313: Is this necessarily a bug where the program should be halted in a build with asserts? Seems like something that could also happen with e.g. invalid coordinate inputs (e.g. denormal small coords so that product is zero, since normal is not normalised afaics) that should not be a halt but the check returning an is_valid failure.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,462 @@
...
+            v2 -= base;
+            normal = cross(v1, v2);
+
+            // Check for non-coplanar points
+            for (auto pit = points_begin(polygon); pit != points_end(polygon); ++pit)
+            {
+                vec3 vi;
+                set_vec_from_point::apply(*pit, vi);
+                auto vol = qvm::dot(normal, vi)- dot(normal, base);
+                if (std::abs(vol) > tolerance)
```
tinko92, line 217: This may be to narrow for e.g. float coordinates in the range of thousand, I think, the bound should probably be scaled to the input magnitude, I will revisit this later when I have more time.

tinko92 (commented):
I noted some points inline, mostly minor but importantly (imho) the question of separating some computations into strategy.

A minor point, this PR should probably include the removal of  https://github.com/boostorg/geometry/blob/d2e5137f4070d8bbcd8ab150ad4fe42713c8ceb3/include/boost/geometry/geometries/polyhedral_surface.hpp#L47

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
vissarion: Indeed, a lot of computations here are coordinate system spesific and should be moved in strategies. On one hand, I think that polyhedral surface can only be cartesian. But on the other hand, predicates should be moved to stategies anyways for robustness reasons (as you said).

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,462 @@
...
+                point_2d p2 = projected_polygon.outer()[2];
+
+                auto side = strategy.side().apply(p0, p1, p2);
+
+                auto const& orientation = boost::geometry::point_order<polygon_2d>::value;
+                if ((side > 0 && orientation == boost::geometry::clockwise) ||
+                    (side < 0 && orientation == boost::geometry::counterclockwise))
+                {
+                    boost::geometry::reverse(projected_polygon);
+                }
```
tinko92, line 245: Is this implictly assuming that 
1) The polygon is convex? If not, three points may be insufficient to determine its orientation. It is not obvious to me from what I can find on the simple feature spec that it has to be.
2) Its first three points are not collinear? They may be collinear, even for a valid, convex polygon.

vissarion: Note that the correct orientation is tested by is_valid call to projected polygon below.

barendgehrels:
Looks great! I will review it next Wednesday or Thursday.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+{
+
+template <typename PolyhedralSurface, bool AllowInconsistentOrientation = false>
+struct valid_intersection
+{
+    valid_intersection(size_t num_of_polygons)
+        : m_polygon_projections(num_of_polygons)
+    {}
+
+    using CT = typename coordinate_type<PolyhedralSurface>::type;
```
barendgehrels, line 38: Suggestion, `coor_t`? `CT` looks like a template parameter

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+    using vec3 = boost::qvm::vec<CT, 3>;
+    using polygon_3d = typename boost::range_value<PolyhedralSurface>::type;
+    using ring_3d = typename boost::geometry::ring_type<polygon_3d>::type;
+    using point_3d = typename boost::range_value<ring_3d>::type;
```
barendgehrels, line 42: Why not `point_type` ? It delivers the same but we usually use that. You don't need the ring then (but maybe it's needed anyway)

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
barendgehrels: I think it's a valid constraint at least for now

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+                             Strategy const& strategy)
+    {
+        boost::ignore_unused(strategy);
+        bool disconnected = true;
+        // TODO: improve perfomance by using a spatial index
+        for (auto it1 = boost::begin(surface); it1 != boost::end(surface); ++it1)
+        {
+            for (auto it2 = boost::begin(surface); it2 != boost::end(surface); ++it2)
+            {
+                if (it1 != it2)
```
barendgehrels, line 395: Suggestion, `if (it1 == it2) continue`

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
vissarion: I added that as a concept constraint together with the constraint on dimension (==3).

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+class is_valid_polyhedral_surface
+{
+public:
+    template <typename VisitPolicy, typename Strategy>
+    static inline bool apply(PolyhedralSurface const& surface,
+                             VisitPolicy& visitor,
+                             Strategy const& strategy)
+    {
+        boost::ignore_unused(strategy);
+        bool disconnected = true;
```
barendgehrels, line 389: Suggestion, call it `connected = false` and set it later to `true`. For me that would read a bit better

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+            boost::geometry::within(projected_intersection, projected_polygon, strategy);
+        if (intersects)
+        {
+            visitor.template apply<failure_invalid_intersection>();
+            return intersection_type::invalid;
+        }
+        return intersection_type::none;
+    }
+
+    // Helper function to check if any segment of a ring intersects a polygon
```
barendgehrels, line 324: Do we really support polygons? Or rings? It's not completely clear to me.
This is checking the outer ring only?

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+    template <typename VisitPolicy, typename Strategy>
+    bool check_polygon_polygon_intersection(polygon_3d const& polygon1,
+                                            polygon_3d const& polygon2,
+                                            size_t const& index2,
+                                            VisitPolicy& visitor,
+                                            bool& disconnected,
+                                            Strategy const& strategy)
+    {
+        auto const& ring1 = boost::geometry::exterior_ring(polygon1);
+        for (auto it = boost::begin(ring1); it != boost::end(ring1) - 1; ++it)
```
barendgehrels, line 334: Suggestion: `it + 1 != boost::end(ring1)` to avoid empty ring problems

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+            auto const& projection = m_polygon_projections[polygon_index];
+            base = projection.base;
+            v1 = projection.v1;
+            v2 = projection.v2;
+            normal = projection.normal;
+            projected_polygon = projection.projected_polygon;
+        }
+        else
+        {
+            auto const& ring = boost::geometry::exterior_ring(polygon);
```
barendgehrels, line 192: So this block does really the projection. Nice.
I think it's good to extract it from this class - and even to another source.
Such that it is useful on its own, and can be unit tested on its own.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+                vec3 vi;
+                set_vec_from_point::apply(*pit, vi);
+                auto vol = qvm::dot(normal, vi)- dot(normal, base);
+                if (std::abs(vol) > tolerance)
+                {
+                    visitor.template apply<failure_non_coplanar_points_on_face>();
+                    return intersection_type::invalid;
+                }
+            }
+            // Project the polygon onto the 2D plane defined by v1 and v2
```
barendgehrels, line 226: Looks interesting. I'm not experienced in this so cannot really comment.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+
+            // Store the projection for future use
+            m_polygon_projections[polygon_index] = polygon_projection{projected_polygon, base, v1,
+                    v2, normal};
+        }
+
+        // Check if the segment intersects the plane of the polygon
+        CT dot_product = dot(normal, base);
+        CT sign1 = dot(normal, p1) - dot_product;
+        CT sign2 = dot(normal, p2) - dot_product;
```
barendgehrels, line 249: This can all be const

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
tinko92: This is orient3d (see e.g. https://www.[cs.cmu.edu/afs/cs/project/quake/public/code/predicates.c](https://www.cs.cmu.edu/afs/cs/project/quake/public/code/predicates.c)), i.e. the 3D generalisation of side, a determinant involving `*pit`, `base`, `v1`, `v2`. Bound for a tolerant version can be derived the same way as for side_rounded_input, I'll try to do it by the end of the week. Calling orient3d for every point may look more expensive, but I think the compiler will know to pull most unchanging subexpressions out of the loop (same as dot(normal, base) could be pulled out of the loop).

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+            v1 = projection.v1;
+            v2 = projection.v2;
+            normal = projection.normal;
+            projected_polygon = projection.projected_polygon;
+        }
+        else
+        {
+            auto const& ring = boost::geometry::exterior_ring(polygon);
+            constexpr bool is_ring_open = boost::geometry::closure<ring_3d>::value
+                == boost::geometry::open;
```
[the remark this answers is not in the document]
tinko92: Since this is constexpr, this is only done once at compile time, there may be merit to scoping it for separation of logic,.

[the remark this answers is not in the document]
barendgehrels: Agreed - but the whole list of types is also at the top (not scoped). This one fits there.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+                        std::distance(boost::begin(surface), it2),
+                        visitor, strategy);
+                    using intersection_type = decltype(intersection);
+                    if (intersection == intersection_type::invalid)
+                    {
+                        return false; //  Invalid intersection found (including inconsistent orientation)
+                    }
+                    else if (intersection == intersection_type::valid)
+                    {
+                        disconnected = false;
```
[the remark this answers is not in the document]
vissarion: You are right. Currently, the check only checks that there is no disconnected polygon. More work is needed to check for connectedness of the surface.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
vissarion: I agree, I changed it to `coordinate_type` which seems more common as a use in the library.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+        <CT, 2, boost::geometry::cs::cartesian>;
+    using linestring_2d = typename boost::geometry::model::linestring<point_2d>;
+    using polygon_2d = typename std::conditional
+        <
+            boost::geometry::point_order<polygon_3d>::value == boost::geometry::counterclockwise,
+            typename boost::geometry::model::polygon<point_2d, true>,
+            typename boost::geometry::model::polygon<point_2d, false>
+        >::type;
+    using ring_2d = typename boost::geometry::ring_type<polygon_2d>::type;
+    using segment_2d = typename boost::geometry::model::segment<point_2d>;
```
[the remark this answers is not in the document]
vissarion: segment_2d is not used any more so I removed. Ring types we can get them from `std::dectype` if you prefer it, I am ok with those as is.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
vissarion: This is also removed. There is a similar structure/logic in `check_polygon_polygon_intersection` where I used your naming approach.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+            if (result == intersection_type::invalid)
+            {
+                return false; // Found invalid intersection
+            }
+            else if (result == intersection_type::valid)
+            {
+                disconnected = false; // Found valid intersection
+            }
+        }
+        return true; // Either no intersection or valid intersection
```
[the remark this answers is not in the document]
vissarion: Now all functions in `valid_intersection` return `intersection_type`, I think it is easier to follow the code.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+        set_vec_from_point::apply(point2, p2);
+
+        vec3 base;
+        vec3 v1;
+        vec3 v2;
+        vec3 normal;
+        polygon_2d projected_polygon;
+
+        // Check if the polygon has been projected before
+        if (m_polygon_projections[polygon_index].projected_polygon.outer().size() > 0)
```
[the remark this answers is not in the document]
vissarion: I used the `boost::geometry::is_empty` function.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,453 @@
...
+                    visitor.template apply<failure_non_coplanar_points_on_face>();
+                    return intersection_type::invalid;
+                }
+            }
+            // Project the polygon onto the 2D plane defined by v1 and v2
+            for (auto const& point : ring)
+            {
+                vec3 p;
+                set_vec_from_point::apply(point, p);
+                auto projected_p = project_point_to_plane_2d(v1, v2, p);
```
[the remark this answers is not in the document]
vissarion: I changes to `projected_polygon.outer().emplace_back(project_point_to_plane_2d(v1, v2, p));`
that should avoid copies

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
vissarion: That reduces the computation to orientation, great. I was thinking using the `project_point_to_plane_2d` to only do one orientation test but I think projecting to coordinate planes as you suggested is better since it does not involve any computations (substitutions, dot products etc). 
I change this part to use the side strategy provided by the umbrella strategy used in is_valid. This uses the side_by_triangle strategy as the default but the user can change it (as in many other cases in the library). At this point I think this is better then using a creating a new umbrella strategy that has side_rounded_input as a side strategy. We can think of substituting the side_by_triangle by side_rounded_input in general or in as many algorithms as possible.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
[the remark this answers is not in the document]
barendgehrels: It's fine. But did you push already? I don't see the updates...

vissarion:
I refactor the code addressing almost all of the comments. Some things are left as todo for next PR(s). What is missing from this PR is the predicates (places in the code that use tolerance) that should be written as orientation checks implemented in strategies. I will work on it in the next days.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
tinko92: I thought about it. I think, we want

```math
\begin{vmatrix}
    a_x-d_x & a_y -d_y & a_z - d_z \\
    b_x-d_x & b_y-d_y & b_z-d_z \\
    c_x-d_x & c_y-d_y & c_z-d_z
\end{vmatrix}
```

Since only one point changes (let's say $a=\text{vi}$), it makes sense to take the first row as cofactors and the minors will be the elements of $n$ (if all points, including $a$ were translated by base before). Then the determinant is 

$(a_x-d_x) \times ((b_y-d_y) \times (c_z - d_z) - (b_z - d_z) \times (c_y - d_y)) - (a_y - d_y) \times \ldots$

Then, I think, the tolerance (based on https://zenodo.org/records/7539355 with rounded input rules, rounding the second order part up because it doesn't matter for this use case):

```math
12\epsilon \times \left(\left(\left\lvert a_x\right\rvert  + \left\lvert d_x\right\rvert \right) \times \left(\left(\left\lvert b_y\right\rvert  + \left\lvert d_y\right\rvert \right) \times \left(\left\lvert c_z\right\rvert  + \left\lvert d_z\right\rvert \right) + \left(\left\lvert b_z\right\rvert  + \left\lvert d_z\right\rvert \right) \times \left(\left\lvert c_y\right\rvert  + \left\lvert d_y\right\rvert \right)\right) + \left(\left\lvert a_y\right\rvert  + \left\lvert d_y\right\rvert \right) \times \left(\left(\left\lvert b_x\right\rvert  + \left\lvert d_x\right\rvert \right) \times \left(\left\lvert c_z\right\rvert  + \left\lvert d_z\right\rvert \right) + \left(\left\lvert b_z\right\rvert  + \left\lvert d_z\right\rvert \right) \times \left(\left\lvert c_x\right\rvert  + \left\lvert d_x\right\rvert \right)\right) + \left(\left\lvert a_z\right\rvert  + \left\lvert d_z\right\rvert \right) \times \left(\left(\left\lvert b_x\right\rvert  + \left\lvert d_x\right\rvert \right) \times \left(\left\lvert c_y\right\rvert  + \left\lvert d_y\right\rvert \right) + \left(\left\lvert b_y\right\rvert  + \left\lvert d_y\right\rvert \right) \times \left(\left\lvert c_x\right\rvert  + \left\lvert d_x\right\rvert \right)\right)\right)
```

where $\epsilon$ is **half** of numerical_limits<CT>::epsilon(). This is the same structure as the expanded determinant. If I made no mistake, this will guarantee that the tolerance is sufficiently large such that for all input points that are correctly rounded approximations of actually collinear points, it will find them coplanar.

And this would be the equivalent of side_rounded for planes.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,501 @@
...
+    {
+        polygon_2d projected_polygon;
+        vec3 base;
+        vec3 v1;
+        vec3 v2;
+        vec3 normal;
+    };
+
+    template <int I, int J, typename Strategy>
+    static bool affinely_independent(point_3d const& p1, point_3d const& p2, point_3d const& p3,
```
barendgehrels, line 111: Suggestion for name: `is_affinely_independent`

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,501 @@
...
+                                     Strategy const& strategy)
+    {
+        point_2d const p1_2{geometry::get<I>(p1), geometry::get<J>(p1)};
+        point_2d const p2_2{geometry::get<I>(p2), geometry::get<J>(p2)};
+        point_2d const p3_2{geometry::get<I>(p3), geometry::get<J>(p3)};
+        return strategy.side().apply(p1_2, p2_2, p3_2) != 0;
+    }
+
+    // Try to find 3 non-collinear points
```
barendgehrels, line 120: in any plane?

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
barendgehrels: But exactly that (shared variables) will makes it harder to extract it later as well...

A generic polygon projection from 3d to 2d plane would be great. We could then immediately implement the surface-area algorithm on a 3d surface... 

It would be a generic algorithm, every patch (surface) would be a separate 2D polygon. Or is it different than what you do here (I couldn't look into all the details yet).

Anyway - I'm fine having it as a next PR - I'm only afraid that it would not really safe time.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,501 @@
...
+    {
+        for (auto pit1 = points_begin(polygon); pit1 != points_end(polygon); ++pit1)
+        {
+            for (auto pit2 = pit1; pit2 != points_end(polygon); ++pit2)
+            {
+                for (auto pit3 = pit2; pit3 != points_end(polygon); ++pit3)
+                {
+                    if (affinely_independent<0, 1>(*pit1, *pit2, *pit3, strategy)
+                        || affinely_independent<0, 2>(*pit1, *pit2, *pit3, strategy)
+                        || affinely_independent<1, 2>(*pit1, *pit2, *pit3, strategy))
```
barendgehrels, line 133: Suggestion, add comment `One three-point-set is already sufficient, with such three independent points a plane can be constructed` (or similar).

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,501 @@
...
+                        set_vec_from_point::apply(*pit3, v2);
+                        return true;
+                    }
+                }
+            }
+        }
+        return false; // All points are collinear
+    }
+
+    static point_2d project_point_to_plane_2d(vec3 const& v1, vec3 const& v2, vec3 const& p)
```
barendgehrels, line 146: Without comments, and drawings, and much experience, it's hard to follow this...

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,501 @@
...
+    template <typename VisitPolicy>
+    static bool has_correct_intersection_orientation(polygon_2d const& polygon,
+                                                     point_2d const& point1,
+                                                     point_2d const& point2,
+                                                     VisitPolicy& visitor)
+    {
+        auto ring = boost::geometry::exterior_ring(polygon);
+        if (boost::geometry::closure<ring_2d>::value == boost::geometry::closed)
+        {
+            ring.pop_back(); // Remove the closing point to make it open
```
barendgehrels, line 160: We have also an iterator doing this automatically. It closes it automatically (which is similar - but you can then always skip the last point).
That avoids the need for a copy of the ring.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,501 @@
...
+        }
+
+        for (auto it = boost::begin(ring); it != boost::end(ring); ++it)
+        {
+            if (boost::geometry::equals(*it, point1))
+            {
+                if (boost::geometry::equals(point1, point2))
+                {
+                    visitor.template apply<failure_duplicate_points>();
```
barendgehrels, line 169: Can't this check be made right at the beginning, before even iterating?
What are point1 and point2? Is it a segment?
Can we then rename them to `segment_point1` or just `s1` with a comment?

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,501 @@
...
+        }
+        return false; // All points are collinear
+    }
+
+    static point_2d project_point_to_plane_2d(vec3 const& v1, vec3 const& v2, vec3 const& p)
+    {
+        return point_2d{qvm::dot(v1, p), qvm::dot(v2, p)};
+    }
+
+    template <typename VisitPolicy>
```
barendgehrels, line 151: So what does it do exactly? It finds if the segment point1-point2 overlaps with any segment of the outer ring of the polygon?

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,501 @@
...
+            base = projection.base;
+            v1 = projection.v1;
+            v2 = projection.v2;
+            normal = projection.normal;
+            projected_polygon = projection.projected_polygon;
+        }
+        else
+        {
+            // TODO: Support interior rings
+            ring_3d const ring = boost::geometry::exterior_ring(polygon);
```
barendgehrels, line 224: It's a copy...

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,501 @@
...
+        : m_polygon_projections(num_of_polygons)
+    {}
+
+    using coordinate_type = typename coordinate_type<PolyhedralSurface>::type;
+    using vec3 = boost::qvm::vec<coordinate_type, 3>;
+    using polygon_3d = typename boost::range_value<PolyhedralSurface>::type;
+    using ring_3d = typename boost::geometry::ring_type<polygon_3d>::type;
+    using point_3d = typename boost::geometry::point_type<ring_3d>::type;
+    using segment_3d = typename boost::geometry::model::segment<point_3d>;
+    using set_vec_from_point = typename strategy::transform::detail::matrix_transformer::
```
barendgehrels, line 75: The name is not 100% clear to me.

But (I see now) we have that already for a long time.
It sets a vector from a point?
Anyway, the `using` clause is fine then...

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,501 @@
...
+                    visitor.template apply<failure_inconsistent_orientation>();
+                    return false;
+                }
+            }
+        }
+        visitor.template apply<failure_invalid_intersection>();
+        return false;
+    }
+
+    // Check if a segment intersects a polygon in 3D space
```
barendgehrels, line 191: So also here `point1` and `point2` form a segment, it would be good to make that clear.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,501 @@
...
+                                                         point_3d const& point2,
+                                                         polygon_3d const& polygon,
+                                                         size_t const& polygon_index,
+                                                         VisitPolicy& visitor,
+                                                         Strategy const& strategy)
+    {
+        vec3 p1;
+        vec3 p2;
+        set_vec_from_point::apply(point1, p1);
+        set_vec_from_point::apply(point2, p2);
```
barendgehrels, line 203: Why not `v1` and `v2` here, if these are vectors? 
And then below you have another set of `v1` and `v2`, it's not clear to me...

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,501 @@
...
+
+            // Cache the projection for future use
+            m_polygon_projections[polygon_index] = polygon_projection{projected_polygon, base, v1,
+                    v2, normal};
+        }
+
+        // Check if the segment intersects the plane of the polygon
+        coordinate_type const dot_product = dot(normal, base);
+        coordinate_type const sign1 = dot(normal, p1) - dot_product;
+        coordinate_type const sign2 = dot(normal, p2) - dot_product;
```
barendgehrels, line 276: The `p1` and `p2` are calculated long before...
Would it be possible to move that to here? (And rename them to something vector-like)?

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
vissarion: 3 non-colinear points will define a plane. Later in the code we check if the rest of the points of the polygon lay on this plane.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
vissarion: Thanks, I will use `closing_iterator`.

vissarion:
@barendgehrels thanks for you comments! I addressed most of them. I need to work more on this PR. Apart from the orient3d some other simplifications can be implemented (the unaddressed comments of Barend are related to those simplifications). I will work on it next week.

vissarion:
I have updated this PR. Now strategies are used (with 2d and 3d orientation predicates) and there are no constructions on new geometries apart from coordinate projections (that are not that vulnerable to numerical errors). Also a new algorithm for segment-polygon intersection in 3D only using orientation predicates is implemented.

include/boost/geometry/strategies/relate3/cartesian.hpp
```diff
@@ -0,0 +1,58 @@
+// Boost.Geometry
+
+// Copyright (c) 2025, Oracle and/or its affiliates.
+
+// Contributed and/or modified by Vissarion Fysikopoulos, on behalf of Oracle
+
+// Licensed under the Boost Software License version 1.0.
+// http://www.boost.org/users/license.html
```
barendgehrels, line 8: AFAIK we don't use this two-line version.
Also, the other file added contains:
```
```

I'm OK with the shorter version if:
* it is used in other libraries too (do you have a reference)
* it is consistent within this PR

and then we can modify it in all our sources (maybe in one go)

include/boost/geometry/strategies/cartesian/side_3d_rounded_input.hpp
```diff
@@ -0,0 +1,86 @@
+// Boost.Geometry (aka GGL, Generic Geometry Library)
```
barendgehrels, line 1: we don't need `aka GGL` anymore, in new sources I always omit it.

include/boost/geometry/strategies/cartesian/side_3d_rounded_input.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+// Copyright (c) 2025 Tinko Bartels, Berlin, Germany.
+// Copyright (c) 2025 Oracle and/or its affiliates.
+// Contributed and/or modified by Vissarion Fisikopoulos, on behalf of Oracle
+
+// This file was modified by Oracle on 2023.
+// Modifications copyright (c) 2023 Oracle and/or its affiliates.
```
barendgehrels, line 8: Should be `2025`

include/boost/geometry/strategies/cartesian/side_3d_rounded_input.hpp
```diff
@@ -0,0 +1,86 @@
...
+        coor_t const p3_y = geometry::get<1>(p3);
+        coor_t const p3_z = geometry::get<2>(p3);
+        coor_t const p_x = geometry::get<0>(p);
+        coor_t const p_y = geometry::get<1>(p);
+        coor_t const p_z = geometry::get<2>(p);
+
+        static coor_t const eps = std::numeric_limits<coor_t>::epsilon() / 2;
+        coor_t const det = (p_x - p1_x) * ((p2_y - p1_y) * (p3_z - p1_z) - (p3_y - p1_y) * (p2_z - p1_z))
+                         - (p_y - p1_y) * ((p2_x - p1_x) * (p3_z - p1_z) - (p3_x - p1_x) * (p2_z - p1_z))
+                         + (p_z - p1_z) * ((p2_x - p1_x) * (p3_y - p1_y) - (p3_x - p1_x) * (p2_y - p1_y));
```
barendgehrels, line 61: Nice... it could profit from some ASCII art

barendgehrels:
Cool changes, thanks for the splits - I'll try to finish my review this weekend

include/boost/geometry/strategies/relate3/cartesian.hpp
vissarion: Thanks for noticing. Actually, there are ~300 files in BG that use the shorter version, while ~1700 that use the longer one. Other boost projects use slightly different versions e.g.
```
Distributed under the Boost Software License, Version 1.0.
(See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
```
used by `hana` and a similar one by `beast`.

So for this PR I will use the longer one and we have to decide how to proceed with all the sources.

barendgehrels: Thanks. And wow... so we've 3 versions indeed, starting with either
```
// Distributed under the Boost Software License, Version 1.0.
// Licensed under the Boost Software License version 1.0.
```
and in the whole Boost there are 103 versions (some caused by different way of commenting, but still 50 then)

Actually - why did we start using the version starting with "Licensed"? Because we're the only ones... Was this prescribed by Oracle?

By far the most frequent I see (12411 times) the one you mention with `Distributed under the Boost Software License`

Let's go for that one. But we can have the following lines together (because we use the 100 col. ruler).

I actually also like the ones adding the repository (as done in `url`) and documentation (as done in `unordered`).

So I would propose:
```
// Distributed under the Boost Software License, Version 1.0.
// (See accompanying file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)

// Official repository: https://github.com/boostorg/geometry
// Documentation: http://www.boost.org/libs/geometry
```

tinko92 (approved):
Thanks for all the work on incorporating the requested changes! This looks good to me. I left some comments on formatting (with diffs, see attached txt files in comments) and compiler warnings, but nothing that would require another review as far as I can see, so I'm marking it as approve.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,648 @@
...
+#include <boost/geometry/iterators/point_reverse_iterator.hpp>
+
+namespace boost { namespace geometry
+{
+
+#ifndef DOXYGEN_NO_DETAIL
+namespace detail { namespace is_valid
+{
+
+struct polygon_frontier_type : public std::stack<size_t>
```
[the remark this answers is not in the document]
vissarion: It is a leftover, thanks.

vissarion:
Thanks for your reviews! I addressed all the comments. 
(However, I think @barendgehrels didn't finish his review yet)

include/boost/geometry/strategies/relate3/cartesian.hpp
vissarion: > Actually - why did we start using the version starting with "Licensed"? Because we're the only ones... Was this prescribed by Oracle?

I do not think so, probably it happened accidentally. 

>So I would propose:
```
// Distributed under the Boost Software License, Version 1.0.
// (See accompanying file LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)

// Official repository: https://github.com/boostorg/geometry
// Documentation: http://www.boost.org/libs/geometry
```
I agree.

include/boost/geometry/algorithms/detail/is_valid/polyhedral_surface.hpp
```diff
@@ -0,0 +1,650 @@
...
+// This function checks if a segment (s1, s2) intersects with a 2D polygon in 3D space.
+// proj1 and proj2 are the indices of the coordinate axes used to determine the sign of three points
+// of the polygon after projection.
+// It returns:
+// - 0 if the segment does not intersect the polygon
+// - 1 if the segment intersects the polygon in the interior
+// - 2 if the segment intersects the polygon at a vertex
+// - 3 if the segment intersects the polygon at an edge
+// Reference: R.J.Segura, F.R.Feito - An algorithm for determining intersection segment-polygon in
+// 3D, Computers & Graphics 1998
```
barendgehrels, line 75: 👍

include/boost/geometry/algorithms/detail/overlay/graph/assign_counts.hpp
```diff
@@ -80,22 +80,22 @@ void assign_counts(Turn& turn)
...
     };
 
-    auto assign_left_incoming_for = [&turn, &assign_for](counts_per_op_t const& op1, counts_per_op_t op2)
+    auto assign_left_incoming_for = [&assign_for](counts_per_op_t const& op1, counts_per_op_t op2)
     {
         assign_for(op1, op2, [](auto& enriched, auto count) { enriched.count_left_incoming = count; });
     };
 
-    auto assign_right_incoming_for = [&turn, &assign_for](counts_per_op_t const& op1, counts_per_op_t op2)
+    auto assign_right_incoming_for = [&assign_for](counts_per_op_t const& op1, counts_per_op_t op2)
```
barendgehrels, line 98: Thanks for the side activity ;-)

include/boost/geometry/strategies/cartesian/side_3d_rounded_input.hpp
```diff
@@ -0,0 +1,98 @@
...
+namespace strategy { namespace side
+{
+
+// This strategy is used for 3D side calculations.
+// It is used to determine the side of a point p with respect to a triangle/plane defined by three
+// points in 3D space namely p1, p2, p3.
+// The following determinant is used to calculate the side:
+//       | (p_x - p1_x)  (p_y - p1_y)  (p_z - p1_z)  |
+// det = | (p2_x - p1_x) (p2_y - p1_y) (p2_z - p1_z) |
+//       | (p3_x - p1_x) (p3_y - p1_y) (p3_z - p1_z) |
```
barendgehrels, line 41: 👍

barendgehrels (approved):
Big step! Thanks!

tinko92:
Thanks for the quick response. The latest force push broke `b2 test`, looks like a missing `git add test/geometries/polyhedral_surface_fail.cpp`.

PR #1413, opened 2025-07-12
feat: add supported combinations for convert

barendgehrels:
This PR adds combinations for `convert`. I need it for a next step (an alternative for `dissolve`).

It adds support for:
* areal features to multi-linestring
* linear and areal features to multi-point
* and a few combinations more

It also extends the test coverage to these combinations (and some combinations earlier missing).

The support matrix `.qbk` is updated. I use another mechanism for that now, which:
* is built inside the test code
* also verifies if the test coverage is complete

_The old mechanisms (there were two) both don't work currently. Also there is no cmake support. And the new mechanism is simpler, faster, and more powerful._

But that will be a next PR to keep things smaller and not mixing things.


test/algorithms/Jamfile
```diff
@@ -25,7 +25,6 @@ test-suite boost-geometry-algorithms
     [ run centroid_multi.cpp           : : : : algorithms_centroid_multi ]
     [ run comparable_distance.cpp      : : : : algorithms_comparable_distance ]
     [ run convert.cpp                  : : : : algorithms_convert ]
-    [ run convert_multi.cpp            : : : : algorithms_convert_multi ]
```
barendgehrels, line 28: Because it's not multi but also single to multi - it was more convenient to combine it into one source.

include/boost/geometry/algorithms/convert.hpp
```diff
@@ -277,8 +337,8 @@ namespace dispatch
 template
 <
     typename Geometry1, typename Geometry2,
-    typename Tag1 = tag_cast_t<tag_t<Geometry1>, multi_tag>,
-    typename Tag2 = tag_cast_t<tag_t<Geometry2>, multi_tag>,
+    typename Tag1 = tag_t<Geometry1>,
+    typename Tag2 = tag_t<Geometry2>,
```
barendgehrels, line 341: This was necessary to distinguish different multi to multi combinations

barendgehrels:
> Does this change the meaning of the algorithm and should the new meaning be explained? 

Maybe the documentation can be enhanced - but as far as I can tell, its meaning did not change.
It converts from one geometry to another... It's written there.

> Looking at the current develop support matrix, it so far only seems to allow conversions where the geometries can represent the same set

There were not so many examples but there was already box to point supported. So you will get four points of a box.


> Does this make the functionality overlap with the arguably more general, undocumented (but not in detail namespace and intended to be documented according to

Thanks for the pointer (pun intended), I forgot that functionality. I think it should not have been outside detail - apparently I was not alert enough in the review of that file....

There is already https://www.boost.org/doc/libs/latest/libs/geometry/doc/html/geometry/reference/algorithms/for_each/for_each_point.html which I thought about of using underneath. But the purpose of `for_each_point` is different - for what I want to achieve now.

I need a multi-linestring representation of all (rings, polygons, multi-polygons). We had nothing similar to that (`for_each_segment` is similar but inconvenient and I need to collect it into a `multi-linestring` anyway).

After considering the options, `convert` is the perfect candidate. Because it does already that, and similar things - but is incomplete. And indeed `areal` -> `linear` was not really there yet, or `areal` -> `point like` just partly. However it covers the name and the documentation page.

I don't really need multi-point representations of them, at this moment, but it made the change more complete AND it was useful for me to consider the design options. Because currently it was targeted to `single` -> `multi`

Hope this clarifies the intentions...

include/boost/geometry/algorithms/convert.hpp
```diff
@@ -148,17 +144,20 @@ struct range_to_range
...
     }
 
-    template <typename ConvertPointPolicy>
+    template <typename ConvertPointPolicy, typename Range1, typename Range2>
     static inline ConvertPointPolicy apply(Range1 const& source, Range2& destination,
                                            ConvertPointPolicy convert_point)
     {
         geometry::clear(destination);
 
+        constexpr bool Reverse = geometry::point_order<Range1>::value != geometry::point_order<Range2>::value;
```
vissarion, line 159: maybe lowercase "reverse" since this is not a template now

barendgehrels: ✅

PR #1401, opened 2025-04-28
Remove deprecated headers

vissarion:

PR #1415, opened 2025-07-16
Update release notes for 1.89

vissarion:


barendgehrels (approved):
Thanks!

vissarion:
>- Single typo in line 33 (improvements).

Thanks, fixed.

>- https://github.com/boostorg/geometry/pull/1401 is not mentioned but I think, it could break existing builds and make people look for the reason in the release notes. Maybe unnecessary to mention because it already had the deprecation warning in prior versions.

Good point, I think it doesn't hurt mentioning, so I added it (as a breaking change).

PR #1416, opened 2025-07-17
Fix windows runners in github actions

vissarion:
This PR removes the retired windows-2019 runner together with msvc-14.2 toolset. 

Also adds support for newer runner windows-2025 which gives access to  msvc-14.4 toolset but b2 returns an error so this case is commented out for now (needs more investigation). Related issue https://github.com/boostorg/boost/issues/914


barendgehrels (approved):
Thanks and looks also as a nice smplification

PR #1418, opened 2025-07-21
Refactor epsg, esri, iau2000 for faster compile times.

tinko92:
This is another proposal to reduce the compile times of epsg.hpp, esri.hpp, and iau2000.hpp. Since #1214  was rejected for added complexity, this does not include perfect hashing and leaves the tables in a more readable and more easily editable state.

It reduces the compile time under gcc for the following program from about 7 minutes with 1.5GB memory usage, which seems unreasonable, to under 4 seconds with half of that memory use.

```
#include <boost/geometry.hpp>
#include <boost/geometry/srs/epsg.hpp>
#include <boost/geometry/srs/projections/epsg_params.hpp>

int main() {
    auto epsg2000 = bg::projections::dynamic_parameters<bg::srs::epsg>::apply(bg::srs::epsg{2000});
    auto epsg2001 = bg::projections::dynamic_parameters<bg::srs::epsg>::apply(bg::srs::epsg{2001});
    return epsg2000.size();
}
```

With these changes, rather than generating thousands of assembly instructions to build the vectors at initialisation, the static data is put into the rodata section of the binary, which seems more appropriate for static data, and the vector is only build at retrieval for the code that is actually accessed.

This PR still results in a large binary, as does the current develop code. This could be solved with a changed, constexpr dpar::parameters, which would require two more helper classes, a constexpr static vector (could be replaced by std::inplace_vector when supports is bumped to C++26 in the far future) and a constexpr variant (could be replaced by std::variant when support is bumped to C++17 in the future). For compile-time constant codes, GCC and Clang could then recognise the opportunity for const propagation and not need store the entire parameter table in the binary at all (see e.g. https://godbolt.org/z/5reM6e3zf). But this idea is not included in the PR because it would change the interface and add two more classes (not too complicated ones but also not trivial).


barendgehrels:
Great, thanks. This keeps indeed everything readable and will be acceptable for me. The differences are not rendered in this PR but I downloaded the PR as a patch, compared, and it looks OK to me. I will approve it.

I'm just curious to try it. I never have these long compilation times. Maybe this part is not in CMake yet. Which source should I compile to notice the differences?

tinko92:
> I'm just curious to try it. I never have these long compilation times. Maybe this part is not in CMake yet. Which source should I compile to notice the differences?

I attached a minimal test case (file suffix .cpp.txt due to github limitations).

[epsg2000_size.cpp.txt](https://github.com/user-attachments/files/21398711/epsg2000_size.cpp.txt)

The compile times depend on compiler version and flags. I do not think they depend much on the arch and system, but for the record, I tested this on WSL Ubuntu 25.04 aarch64, Snapdragon X Elite CPU.

Compiling this with different compiler versions and flags and with `export BOOST_REPO=/path/to/my/boost/folder` in the following loop

```
for flags in "" "-O2"; do
    for compiler in g++-11 g++-12 g++-13 g++-14 g++-15 clang++-14 clang++-15 clang++-17 clang++-18 clang++-19 clang++-20; do
        echo $compiler $flags && {/usr/bin/time -f "%e" $compiler -I$BOOST_REPO epsg2000_size.cpp $flags -o $compiler-out-$flags -S;} 2> $compiler-$flags-time ;
    done;
done
```
So the compile command is g++-11 -I/path/to/my/boost/folder epsg2000_size.cpp and so on. 

I get the following results:
| Compiler | develop | develop ‑O2 | fix/srs‑compile‑times | fix/srs‑compile‑times ‑O2 |
|---|---|---|---|---|
| g++-11 | 7.77 | 120.16 | 4.06 | 3.79 |
| g++-12 | 540.69 | 1518.27 | 4.05 | 4.13 |
| g++-13 | 430.64 | 2624.35 | 3.71 | 3.58 |
| g++-14 | 951.42 | 4044.91 | 3.80 | 5.78 |
| g++-15 | 75.79 | 786.81 | 2.81 | 3.11 |
| clang++-14 | 6.00 | 8.87 Command terminated by signal 11 | 3.95 | 4.16 |
| clang++-15 | 5.09 | 7.85 Command terminated by signal 11 | 6.16 | 3.77 |
| clang++-17 | 5.59 | 63.41 | 3.74 | 3.73 |
| clang++-18 | 5.89 | 77.20 | 3.99 | 3.86 |
| clang++-19 | 6.24 | 68.63 | 3.96 | 4.07 |
| clang++-20 | 8.68 | 74.93 | 4.16 | 4.08 |

I did not systematically record RAM usage, but the highest I saw for current develop was 8.5GB (no swapping, though, so this wasn't the reason for the slowdown) with g++-14 -O2, and for fix/srs-compile-times I saw 1GB RAM usage for the same setting.

Develop and this PR lead to very different code generation. With g++-14 and -O2, for develop I get assembly that has tens of thousands of lines that look like this:

```
.LEHB4206:
        bl      _ZNSt6vectorIN5boost8geometry3srs4dpar9parameterIdEESaIS5_EE12emplace_backIJRNS3_6name_fEiEEERS5_DpOT_.isra.0
.LEHE4206:
        add     x1, sp, 16384
        mov     w0, 41248
        movk    w0, 0x7, lsl 16
        mov     x11, 28200
        mov     x12, 28204
        add     x2, sp, x11
        str     w0, [x1, 11816]
        mov     w0, 20
        str     w0, [x1, 11820]
        add     x0, sp, 172032
        add     x1, sp, x12
        add     x0, x0, 1080
.LEHB4207:
        bl      _ZNSt6vectorIN5boost8geometry3srs4dpar9parameterIdEESaIS5_EE12emplace_backIJRNS3_6name_fEiEEERS5_DpOT_.isra.0
.LEHE4207:
```

so it directly generates the assembly for all of those vector operations and the static data is intermingled with the code. The binary size for the appended test program with g++-15 -O2 is 2.8M on develop and it takes ~6ms to run (I suspect due to the thousands of calls to the allocator for each of the vectors in dpar::parameters at the initial static array initialisation).

With the branch fix/srs-compile-times, the content of arr itself is stored in a very long read-only data section in the binary
```
_ZZN5boost8geometry11projections6detail18epsg_to_parametersEiE3arr:
        .word   2000
        .zero   4
        .word   82
        .zero   4
        .word   124
        .zero   4
        .word   1
        .zero   4
        .word   25
        .zero   4
        .word   0
        .word   0
        .word   2
        .zero   4
        .word   31
        .zero   4
        .word   0
        .word   -1068564480
```
And the vectors for a specific dpar::parameters<> instance are only build when such an instance is returned from the method. The binary size for the appended test program with g++-15 -O2 is 1.5M on fix/srs-compile-times and it takes ~2ms to run.

For clang, though, on develop I get a 1.3M binary and a 1.5M binary on my branch. The fundamental difference in the kind of code generated (static data mixed with instructions on develop vs. long read only data section) is still the same but clang does the former more space efficiently than gcc apparently. The patched code still runs a little faster.

tinko92:
Thanks, Vissarion and Barend, for the quick feedback. Since this would not be reasonable to check for mistakes manually, the following test programs https://gist.github.com/tinko92/e75471ecf4f6eabc7117fd75308b8aaa produced no output for me on this test branch which includes old and new and (for testing this PR only) modifications to allow equality comparison of parameters: https://github.com/tinko92/geometry/tree/test-srs-projections-comparison .

include/boost/geometry/srs/projections/code.hpp
```diff
@@ -24,10 +25,83 @@ namespace boost { namespace geometry { namespace projections
...
         int code;
-        srs::dpar::parameters<> parameters;
+        std::array<code_parameter, 13> parameters;
+        std::array<double, 7> towgs84;
+        srs::dpar::parameters<double> to_parameters() const {
+            srs::dpar::parameters<double> out;
+            bool towgs84_to_be_inserted = towgs84[0] != INFINITY;
+            for (const auto& p : parameters)
+            {
+                if(towgs84_to_be_inserted && (p.id == srs::dpar::units || p.id == srs::dpar::no_defs))
```
vissarion, line 95: There is a space missing between `if` and `(`. Same to the if statement in line 101.

vissarion (approved):
I took another look an I have a minor comment before merging.

barendgehrels (approved):
Thanks for the revised PR and the explanations!

barendgehrels:
Minor comment: in two files:
```
        const code_element * first = arr.begin();
        const code_element * last = arr.end();
```
in `iau2000.hpp` you used `cbegin()` and `cend()`. You might harmonize it.

barendgehrels:
@tinko92 can we merge it?

tinko92:
> @tinko92 can we merge it?

@barendgehrels Yes. As a general rule shall I merge my PRs when they pass all reviews like this one or is that for the maintainers only?

barendgehrels:
> > @tinko92 can we merge it?
> 
> @barendgehrels Yes. As a general rule shall I merge my PRs when they pass all reviews like this one or is that for the maintainers only?

Good question. We've no strict rules. For a bit larger PR like this, I would say that you can merge it if there are two approvals and not remaining questions. For a smaller and simpler one like a one-liner (your previous), one approval is enough.

PR #1422, opened 2025-08-01
Define BOOST_GEOMETRY_DEBUG_SEGMENT_IDENTIFIER when BOOST_GEOMETRY_DEBUG_TRAVERSE_GRAPH is defined.

tinko92:
A trivial change because it is not possible to compile with BOOST_GEOMETRY_DEBUG_TRAVERSE_GRAPH, which requires the output of seg_ids, without BOOST_GEOMETRY_DEBUG_SEGMENT_IDENTIFIER.


barendgehrels (approved):
Thanks

PR #1423, opened 2025-08-01
fix: avoid stack overflow in traverse

barendgehrels:
See #1411 (last commit) where it was discovered and fixed.
This commit is meant to be merged to `1.89` beta

### DIFF
--- a/include/boost/geometry/algorithms/detail/overlay/graph/traverse_graph.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/graph/traverse_graph.hpp
@@ -178,73 +178,89 @@ struct traverse_graph
     bool continue_traverse(Ring& ring,
             signed_size_type component_id,
             signed_size_type start_node_id,
-            signed_size_type current_node_id)
+            signed_size_type target_node_id)
     {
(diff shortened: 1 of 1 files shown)


include/boost/geometry/algorithms/detail/overlay/graph/traverse_graph.hpp
```diff
@@ -178,73 +178,89 @@ struct traverse_graph
...
-        auto const next_target_nodes = get_target_nodes<target_operation>(m_turns, m_clusters,
-                current_turn_indices, component_id);
+        std::size_t iteration_count = 0;
 
-        if (next_target_nodes.empty())
+        // Keep traversing until it finds the start (successful finish), or it is stuck,
+        // or it find an already visited node during traversal.
+        // The iteration count is a defensive check to prevent endless loops and not iterate
+        // more than times there are turns (this should not happen).
+        while (iteration_count < m_turns.size())
```
barendgehrels, line 191: Best reviewing it with **Hide whitespace**.
Then you can see it just changes the recursive behaviour to a loop.

PR #1425, opened 2025-08-25
Replace several Boost headers with C++11/C++14 STL alternatives.

tinko92:
This PR removes a number of includes of Boost headers that are no longer required with C++14 because there are either language constructs (e.g. `static_assert`) or standard library headers that fullfil similar functions (e.g. `std::ignore` instead of `boost::ignore_unused`).

Where it wasn't needed code was removed entirely (e.g. `boost::ignore_unused` when the variable in question was passed to another function already). Where it seemed approriate, it was replaced with more explicit constructs (e.g. `static_asserts` on `std::is_convertible` in various concept headers rather than assignments to unused variables making it closer to the patterns established in C++ concepts).

Some modernisations/cleanups are included that I noticed along the way (e.g. the usual `typedef` -> `using`, `some_type* some_name = 0;` -> `some_type* some_name = nullptr`, removal of `distance_projected_point_ax.hpp`, which was deprecated earlier).

Because `std::ignore` is included it via `<tuple>` rather than `<utility>` so that it is compatible with older versions of the standard library (e.g. with GCC 11). This does not limit its intended usage with C++14 from today's perspective.

No non-trivial refactoring was done in vararray. This type should maybe be replaced with `static_vector` (`inplace_vector` in C++26), which seems to be based on it and duplicates its functionality in Boost.Container. This was not included in this PR because it led to a performance regression.

Some dependencies on Boost libraries now persist only for adapted geometry types. These were not touched because it would have changed documented behaviour. But IHMO it would be worth discussing whether these adapted geometries (Boost.Array and Boost.Tuple now, Boost.Range when the standard is raised to C++20 and Boost.Fusion when it is raised to C++26) could be deprecated for future removal, given that they are defined for types that are obsolete in C++14 which is the required minimum for Boost.Geometry.


include/boost/geometry/algorithms/detail/buffer/get_piece_turns.hpp
```diff
@@ -288,10 +287,8 @@ class piece_turn_visitor
 
     template <typename Section>
     inline bool apply(Section const& section1, Section const& section2,
-                    bool first = true)
+                    bool = true)
```
barendgehrels, line 290: So we can remove it. I'm fine with it like this in this PR, we can change it in a follow up

include/boost/geometry/algorithms/detail/intersection/areal_areal.hpp
```diff
@@ -80,14 +78,11 @@ struct intersection_areal_areal_<TupledOut, tupled_output_tag>
...
-                    <
-                        Areal1, Areal2, single_out,
-                        point_tag, linestring_tag, polygon_tag
-                    >
-            >();
+        (void)sizeof(geometry::detail::expect_output
+                <
+                    Areal1, Areal2, single_out,
+                    point_tag, linestring_tag, polygon_tag
+                >);
```
barendgehrels, line 85: This I don't like so much. `ignore_unused` was meant exactly to avoid this cryptic statement.
Is it solveable in another way?

(Why is the whole statement there anyway - but that is another question...)

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
```diff
@@ -61,20 +60,20 @@ namespace detail { namespace overlay
...
-    static bool const use_side_verification = true;
-    static bool const use_start_turn = true;
-    static bool const use_handle_as_touch = true;
-    static bool const use_handle_as_equal = true;
-    static bool const use_handle_imperfect_touch = true;
+    static bool constexpr use_side_verification = true;
+    static bool constexpr use_start_turn = true;
+    static bool constexpr use_handle_as_touch = true;
+    static bool constexpr use_handle_as_equal = true;
+    static bool constexpr use_handle_imperfect_touch = true;
```
barendgehrels, line 76: Is `static constexpr` valid? I didn't check but I remembered there was something with it.
Why is it actually changed? (A big PR can better have exactly one subject)

include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
```diff
@@ -197,7 +197,7 @@ class linear_intersections
     template <std::size_t I>
     ip_info const& get() const
     {
-        BOOST_STATIC_ASSERT(I < 2);
+        static_assert(I < 2, "I must not be greater than 1");
```
barendgehrels, line 200: This sounds like a funny sentence ;-)

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
barendgehrels: But anyway, if it is valid, it is fine.
Thanks for the PR, it seems a huge effort and a huge style improvement.

barendgehrels (approved):
I'm basically fine. Thanks a lot.

Some suggestions:
* split changes in `extensions` and other code, such that cherry-picking for `master` is more convenient. @vissarion to comment if this is still convenient, or it is done in another way
* make it smaller. It was called `boost alternatives` but there are also other changes such as changing `typedef` to `using`. That is indeed tempting - but it makes the scope larger
* I didn't review every line but more than half of it and it all makes sense to me (one question there)

include/boost/geometry/algorithms/detail/overlay/get_turn_info.hpp
tinko92: I think, it changes the meaning in so far as that these values will not have actual storage, which I think is appropriate in this context, since I understood that these are never used in ways that requires taking their addresses but rather for decisions that can all be resolved at compile time. But I can just take it out of this PR tomorrow to have a more clear scope.

include/boost/geometry/algorithms/detail/intersection/areal_areal.hpp
tinko92: I will test taking a pointer of this type and assign it to ignore.

tinko92:
Thanks for the very quick review! 🙏 I will update the PR to address the points you mentioned.

include/boost/geometry/algorithms/detail/overlay/get_turn_info_for_endpoint.hpp
barendgehrels: It is no problem, it just sounds like `I (me) should be too small` ;-) 
You can keep it.

barendgehrels:
> Thanks for the very quick review! 🙏 I will update the PR to address the points you mentioned.

I didn't mean you should revert the `typedef` -> `using`. That is more for next time. It can now stay in this PR.

But splitting out extensions would still be convenient, I think.

vissarion (approved):
Looks good to me! Nice modernization, Tinko, thanks! I am OK if the code splits in two commits (to separate extensions code).

PR #1429, opened 2025-09-10
Fix mistakes

ivanpanch:
I have corrected some minor mistakes.


tinko92 (approved):
No objections to the changes. Should the commits be squashed into one, though?

ivanpanch:
I have made some further changes, can you have a look? One last thing: There is `Line height in SVG pixels, in case the text contains \n`, but `\n` is not rendered verbatim here: https://www.boost.org/doc/libs/latest/libs/geometry/doc/html/geometry/reference/io/svg/svg_mapper.html

PR #1431, opened 2025-09-15
Use the dynamic type while calling epsilon().

### LINKED ISSUE #1430: Compiler error when using absl::uint28 as content_type for RTrees
Compile fails with the error:
```
...
../boost/geometry/index/detail/rtree/rstar/choose_next_node.hpp:135:33: error: conversion from 'type' (aka 'double') to 'uint128' is ambiguous
  135 |         if ( min_content_diff < -std::numeric_limits<double>::epsilon() || std::numeric_limits<double>::epsilon() < min_content_diff )
      |                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...
```

totakura:
Current call to epsilon is made with the assumption that the dynamic type `content_type` is compatible with double. This does not hold true for numeric types that are bigger than double, e.g: `absl::uint128`. This CL fixes it by binding the call to `content_type`'s type.

This fixes #1430.

### DIFF
--- a/include/boost/geometry/index/detail/rtree/rstar/choose_next_node.hpp
+++ b/include/boost/geometry/index/detail/rtree/rstar/choose_next_node.hpp
@@ -132,7 +132,7 @@ private:
 
         // is this assumption ok? if min_content_diff == 0 there is no overlap increase?
 
-        if ( min_content_diff < -std::numeric_limits<double>::epsilon() || std::numeric_limits<double>::epsilon() < min_content_diff )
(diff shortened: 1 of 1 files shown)


barendgehrels:
@totakura why is this PR closed?

totakura:
@barendgehrels I made a mistake in configuring my git repo and made a commit using my private email. Since I contributed to this work as part of my employment, I have to use my employer email address instead. So I closed this PR and created the other one (#1432) with my employer email address.

PR #1432, opened 2025-09-15
Use the dynamic type while calling epsilon().

totakura:
Current call to epsilon is made with the assumption that the dynamic type `content_type` is compatible with double. This does not hold true for numeric types that are bigger than double, e.g: `absl::uint128`. This CL fixes it by binding the call to `content_type`'s type.

This fixes https://github.com/boostorg/geometry/issues/1430.

### DIFF
--- a/include/boost/geometry/index/detail/rtree/rstar/choose_next_node.hpp
+++ b/include/boost/geometry/index/detail/rtree/rstar/choose_next_node.hpp
@@ -132,7 +132,7 @@ private:
 
         // is this assumption ok? if min_content_diff == 0 there is no overlap increase?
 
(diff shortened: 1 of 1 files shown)


totakura:
Made a mistake with the author email used to commit in the other PR. The current PR uses the correct email with my employer domain.

tinko92 (approved):
This looks good to me. epsilon for some custom integer type will be zero, so same behaviour as comparing to ~1e-16. It requires any content_type to have a numeric_limits specialisation but the header already requires that anyway.

This comparison to an unscaled epsilon looks dubious in general, but that's unrelated to the PR.

PR #1437, opened 2025-10-23
Move Boost::crc and Boost::program_options dependencies from library to unit and robustness tests

jeremy-murphy:
Not an actual dependency of the library. Discovered when I tried to make Boost.Geometry a dependency of Boost.Graph, the CMake breaks.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -34,7 +34,8 @@ function(boost_geometry_add_unit_test prefix item)
   # At least one unit test uses the crc library (but the Geometry library does not).
   target_link_libraries(${unit_test_name}
     PRIVATE
-      Boost::crc)
+      Boost::crc
+      Boost::program_options)
 
   # Include the main Geometry test folder and the current folder
   target_include_directories(${unit_test_name}
--- a/test/robustness/CMakeLists.txt
+++ b/test/robustness/CMakeLists.txt
@@ -15,6 +15,12 @@ function(boost_geometry_add_robustness_test item)
     PRIVATE
       Boost::geometry)
 
+  # Link to additional libraries required for testing.
+  target_link_libraries(${robustness_test_name}
+    PRIVATE
+      Boost::crc
+      Boost::program_options)
+
   # Include the main Geometry test folder and the current folder,
   # and the robustness test folder
   target_include_directories(${robustness_test_name}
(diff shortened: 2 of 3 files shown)


jeremy-murphy:
OK, reopened now that I've put the Boost::crc dependency where it needs to go (on the unit tests).

test/CMakeLists.txt
```diff
@@ -31,6 +31,11 @@ function(boost_geometry_add_unit_test prefix item)
     PRIVATE
       Boost::included_unit_test_framework)
 
+  # At least one unit test uses the crc library (but the Geometry library does not).
+  target_link_libraries(${unit_test_name}
+    PRIVATE
+      Boost::crc)
+
```
barendgehrels: 👍 thanks @jeremy-murphy !

jeremy-murphy:
The merge status is unknown and no amount of rechecking is fixing it, so I seem to be blocked from merging.

jeremy-murphy:
@barendgehrels are you able to merge this or does it require more approvals? I'm just getting an error from GitHub.

barendgehrels:
> @barendgehrels are you able to merge this or does it require more approvals? I'm just getting an error from GitHub.

I'll check.

barendgehrels:
@jeremy-murphy  
> Side note: There is one additional optimization possible, because we can do exactly the same for `program_options` But that requires two extra changes. I can do that in a follow-up if you want.

For convenience my diff (with branch `develop`):

[diff_crc.txt](https://github.com/user-attachments/files/23141532/diff_crc.txt)

But **ignore** the removal of `EXCLUDE_FROM_ALL`

jeremy-murphy:
`program_options` would cause the same problem, so I'll fix it here too.

test/robustness/CMakeLists.txt
```diff
@@ -15,6 +15,12 @@ function(boost_geometry_add_robustness_test item)
     PRIVATE
       Boost::geometry)
 
+  # Link to additional libraries required for testing.
+  target_link_libraries(${robustness_test_name}
+    PRIVATE
+      Boost::crc
+      Boost::program_options)
```
barendgehrels: This version works for me, thanks.

I will merge it today.

barendgehrels: Merged! Thanks again for your contribution!
It's the setup of our repo, not everyone can merge.

Issue #1439, opened 2025-12-02
Boost.Geometry intersection Bug: Returns Full Polygon Area for Two Non-Overlapping Polygons Summary


llpan91:
When computing the intersection of two non-overlapping simple polygons,
boost::geometry::intersection incorrectly returns a non-empty result, containing the full area of one polygon.

This is a critical topology error: a false-positive intersection.

Environment

Boost version: 1.79 / 1.81 / 1.82 (bug reproduced across versions)

Geometry models:

boost::geometry::model::d2::point_xy<double>

boost::geometry::model::polygon<Point>

Compilers: GCC / Clang (bug reproduced consistently)

Operating System: Linux / Ubuntu 22.04

\\ data

    BgPolygon p1 = createPolygon({
        BgBgPoint(-2.47089026, -86.03059246),

        BgBgPoint(-1.161944873, 146.3030596),
        BgBgPoint(3.40221214, 145.8628015),
        BgBgPoint(3.024695769, 141.949088),
        BgBgPoint(2.420586453, 111.9554564),

        BgBgPoint(1.208013978, -86.04069936),
        BgBgPoint(-2.47089026, -86.03059246)
    });

    BgPolygon p2 = createPolygon({
        BgPoint(-6.213278056, -88.01851748),
        BgPoint(-6.207382255, -86.51852906),
        BgPoint(-6.20539951, -86.01408604),
        BgPoint(-2.470873099, -86.02434575),
        BgPoint(-2.472221358, -86.53315586),
        BgPoint(-2.476196085, -88.03315059),
        BgPoint(-6.213278056, -88.01851748)
    });

barendgehrels:
Thanks for the report. Versions until `1.82` are quite old, per `1.88` is the intersection algorithm quite improved.

Could you try it with version `1.89` please? Thanks!

barendgehrels:
Sorry for the delay. I cannot reproduce it in `1.91`
Both polygons have a tiny intersection of `6.20003e-11`
So `Result: Intersection count: 1 (false-positive)` is not a `false positive` but the actual result.

I suggest to close the issue. I added the test (moved from `overlay.cpp` where it was added to `test/algorithms/set_operations/set_ops_areal_areal.cpp` which is the current location. Will create a PR of that.

barendgehrels:
For completeness, the intersection lives here (far zoomed in):

[...]

PR #1441, opened 2025-12-13
Add regression test for non-overlapping polygon intersection (issue #1439)

kashish2710:
This PR adds a minimal regression test reproducing issue #1439.

The test shows that `boost::geometry::intersection` currently produces
a non-empty result for two polygons that do not overlap.

The test fails on Boost 1.89 (tested on Linux / Google Colab),
confirming that the reported behavior is still present.

Once the underlying issue is fixed, this test will help ensure
that the behavior does not regress in future changes.

Screenshot 2025-12-13 at 4 32 31 PM

Still reproducible on Boost 1.89 (Linux).
See attached output showing non-empty intersection for non-overlapping polygons.

### DIFF
--- a/test/algorithms/intersection/issue_1439.cpp
+++ b/test/algorithms/intersection/issue_1439.cpp
@@ -1,0 +1,39 @@
+#include <boost/geometry.hpp>
+#include <boost/geometry/geometries/point_xy.hpp>
+#include <boost/geometry/geometries/polygon.hpp>
+
+#include <vector>
+
+namespace bg = boost::geometry;
+
+using point = bg::model::d2::point_xy<double>;
+using polygon = bg::model::polygon<point>;
+
+int main()
+{
+    polygon p1, p2;
+
+    bg::read_wkt(
+        "POLYGON((-2.47089026 -86.03059246,"
+        "-1.161944873 146.3030596,"
+        "3.40221214 145.8628015,"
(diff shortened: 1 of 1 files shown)


test/algorithms/intersection/issue_1439.cpp
```diff
@@ -0,0 +1,39 @@
...
+        "-2.472221358 -86.53315586,"
+        "-2.476196085 -88.03315059,"
+        "-6.213278056 -88.01851748))", p2);
+
+    std::vector<polygon> output;
+    bg::intersection(p1, p2, output);
+
+    // Expected: no intersection
+    return output.empty() ? 0 : 1;
+}
```
barendgehrels: Thanks a lot. But yes, please adapt it to our current framework.

The cases themselves can go to:
https://github.com/boostorg/geometry/blob/develop/test/algorithms/overlay/overlay_cases.hpp#L1273

The test can go to:
https://github.com/boostorg/geometry/blob/develop/test/algorithms/set_operations/intersection/intersection.cpp#L297

These tests test on area, so the expected area should be `0.0` 

The test will then fail. So you can put the test between
`#ifdef BOOST_GEOMETRY_TEST_ENABLE_FAILING`
and
`#endif`

barendgehrels (changes requested):
As you mentioned, and I added in the comment,
better to have it in the framework.

Issue #1375, opened 2025-02-17
Buffer (with strategies) Has a Unmentioned Hiden Requirement for Input Polygon


sssecs:
Dear Boost Geometry Developers,

First, I’d like to thank you for all the hard work you’ve put into Boost—it’s been invaluable to the community!

I’m new to Boost Geometry and encountered an issue while trying to buffer a polygon. I was following the documentation (https://beta.boost.org/doc/libs/1_82_0/libs/geometry/doc/html/geometry/reference/algorithms/buffer/buffer_7_with_strategies.html), and surprisingly, the example polygon worked as expected, but the polygon I created didn’t.

After some debugging and searching, I realized that the polygons in this related issue (https://github.com/boostorg/geometry/issues/1372) are defined as `using Polygon = bg::model::polygon<Point, false, false>`. However, the documentation defines the polygon type as `typedef boost::geometry::model::polygon<point> polygon`, which doesn’t specify the orientation of the vertices.

To resolve the issue, I changed the polygon definition to `using polygon = boost::geometry::model::polygon<point, false>`, ensuring the polygon is defined as counter-clockwise. Alternatively, adjusting the vertex order of my input polygon to counter-clockwise also solved the problem.

It would be very helpful if the buffer function could support both clockwise and counter-clockwise ordered polygons, or if this behavior could be clarified in the documentation. It would greatly assist newcomers like myself.

Here’s my (non-working under 1.87 and 1.82) version of the code for reference:

```
#include <boost/geometry.hpp>
#include <boost/geometry/algorithms/buffer.hpp>
#include <boost/geometry/geometries/point_xy.hpp>
#include <boost/geometry/geometries/geometries.hpp>
#include <iostream>

namespace bg = boost::geometry;

int main()
{
    using coordinate_type = double;
    using point = boost::geometry::model::d2::point_xy<coordinate_type>;
    using polygon = boost::geometry::model::polygon<point, false>;

    // Declare strategies
    const double buffer_distance = 1.0;
    const int points_per_circle = 16;
    boost::geometry::strategy::buffer::distance_symmetric<coordinate_type> distance_strategy(buffer_distance);
    boost::geometry::strategy::buffer::join_round join_strategy(points_per_circle);
    boost::geometry::strategy::buffer::end_round end_strategy(points_per_circle);
    boost::geometry::strategy::buffer::point_circle circle_strategy(points_per_circle);
    boost::geometry::strategy::buffer::side_straight side_strategy;

    // Declare output
[...]

PR #1442, opened 2025-12-13
Clarify polygon orientation requirement for buffer algorithm

kashish2710:
This PR adds a note in the buffer documentation (under [heading Strategies])
to clarify that the buffer algorithm expects polygons to be counter-clockwise.
If a polygon is clockwise, it should be reversed using:

    boost::geometry::reverse(poly);

A link to the Polygon Concept is also provided for more details.

This helps users avoid confusion when using buffer with custom polygons.
Related issue: #1375


barendgehrels (changes requested):
As mentioned in the comment

PR #1444, opened 2025-12-14
Add overlay regression case for issue 1439

kashish2710:
This PR adds a regression test case for Boost.Geometry issue #1439.

– Introduces case_1439 in overlay_cases.hpp
– The case represents two non-overlapping polygons
– This case is intended to be used by set operation tests

No algorithm behavior is changed; this only adds test data.

### DIFF
--- a/test/algorithms/overlay/overlay.cpp
+++ b/test/algorithms/overlay/overlay.cpp
@@ -369,12 +369,9 @@ void test_all()
     // TEST_DIFFERENCE_B(issue_893_multi, 97213916.0, 1, 1); // needs is_traverse
 
     
-    #ifdef BOOST_GEOMETRY_TEST_ENABLE_FAILING
-    // Issue 1439
-// Regression test: intersection of non-overlapping polygons should be empty.
-   TEST_INTERSECTION(case_1439, 0, 0, 0.0);
-   #endif
-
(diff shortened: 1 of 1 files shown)


kashish2710:
This PR adds the overlay regression case for issue #1439.

test/algorithms/overlay/overlay.cpp
```diff
@@ -368,6 +368,13 @@ void test_all()
 
     // TEST_DIFFERENCE_B(issue_893_multi, 97213916.0, 1, 1); // needs is_traverse
 
+    
+    #ifdef BOOST_GEOMETRY_TEST_ENABLE_FAILING
+    // Issue 1439
+// Regression test: intersection of non-overlapping polygons should be empty.
+   TEST_INTERSECTION(case_1439, 0, 0, 0.0);
```
vissarion: maybe issue_1439 is more descriptive

test/algorithms/overlay/overlay.cpp
```diff
@@ -368,6 +368,13 @@ void test_all()
 
     // TEST_DIFFERENCE_B(issue_893_multi, 97213916.0, 1, 1); // needs is_traverse
 
+    
+    #ifdef BOOST_GEOMETRY_TEST_ENABLE_FAILING
+    // Issue 1439
```
vissarion: this is not needed if the name of the test is changed

PR #1455, opened 2026-03-30
fix: clamp content differences to prevent FMA-induced numerical instability

### LINKED ISSUE #1452: Geometry Index GCC arch x86-64-v3

barendgehrels:
Might fix #1452

@tinko92 could you verify if this helps the problem?

The fix is by Claude Code (reviewed/directed by me), it argues:

```
 Analysis

  The core invariant being violated

  When box_exp is expanded from box_orig to include an indexable, we have box_exp ⊇ box_orig, so mathematically content(box_exp) >=
  content(box_orig) and content_diff >= 0. With FMA fusion, this invariant breaks: the multiply inside content() fuses with the
  subsequent subtraction differently for the two calls, yielding negative or spuriously non-zero results.

  The dangerous patterns

  Every call site follows one of two shapes:

  Shape 1 — "enlargement cost" (5 sites): content(expanded) - content(original)
  - insert.hpp:92
  - rstar/choose_next_node.hpp:120, 249
  - quadratic/redistribute_elements.hpp:294-295
  - linear/redistribute_elements.hpp:435-436

  Shape 2 — "free content" (1 site): content(union) - content(a) - content(b)
  - quadratic/redistribute_elements.hpp:77-79

  In all cases the result is mathematically guaranteed >= 0, but FMA can make it negative or spuriously non-zero.

  Why negative matters more than "slightly different positive"

  When content_diff becomes negative for a box that didn't actually expand, that box appears to shrink when receiving the new element —
  making it the "best" choice by a wide margin. This cascades: wrong node selections lead to a suboptimal tree structure (6685 nodes
  instead of 6763 in the reported case), which means far less effective pruning during queries.

  The fix: clamp to the known invariant

  Since we know these differences are >= 0, clamping them is mathematically sound and eliminates the FMA issue without suppressing any  compiler optimizations globally:

  content_type content_diff = (std::max)(content_type(0),
      content - index::detail::content(ch_i.first));

  This fixes both failure modes:
  - Negative values become exactly 0 (correct: the box didn't enlarge)
  - "Should be zero but is epsilon" — while (std::max) doesn't fix this, it doesn't need to. Two boxes that both don't enlarge will both   clamp to exactly 0, so equality comparisons work. Two boxes that both enlarge by a tiny positive amount are just heuristic  tie-breakers where sub-ULP differences are irrelevant.

  Interesting existing precedent

  The code at rstar/choose_next_node.hpp:135 already does epsilon-aware comparison:
  if ( min_content_diff < -std::numeric_limits<content_type>::epsilon()
    || std::numeric_limits<content_type>::epsilon() < min_content_diff )
  This shows the codebase already anticipated near-zero issues, just not systematically.

  What about the == comparisons?

  Lines like content_diff == smallest_content_diff and content1 == content2 are used as tie-breakers. After clamping, the most important
   case — two equal boxes both showing zero enlargement — will compare as exactly 0 == 0. For two genuinely different enlargements,
  sub-ULP differences in tie-breaking are harmless for the heuristic.

  Concrete changes needed:

  │ insert.hpp                          │ 92      │ Clamp content_diff                 │
  │ rstar/choose_next_node.hpp          │ 120     │ Clamp content_diff                 │
  │ rstar/choose_next_node.hpp          │ 249     │ Clamp content_diff                 │
  │ quadratic/redistribute_elements.hpp │ 77-79   │ Clamp free_content                 │
  │ quadratic/redistribute_elements.hpp │ 294-295 │ Clamp both content_increase values │
  │ linear/redistribute_elements.hpp    │ 435-436 │ Clamp both content_increase values │
 ```

I think this makes sense, looks good - and locally it did not break any unit tests (the RST always failed for me - on the fly that is fixed now as well, see other PR)

### DIFF
--- a/include/boost/geometry/index/detail/algorithms/content.hpp
+++ b/include/boost/geometry/index/detail/algorithms/content.hpp
@@ -112,6 +112,19 @@ typename default_content_result<Indexable>::type content(Indexable const& b)
             >::apply(b);
 }
 
+// Returns the content increase when expanding 'original' to 'expanded'.
+// Precondition: 'original' is covered by 'expanded'. Under this precondition
+// the result is mathematically non-negative, so clamping to 0 is valid and
+// guards against floating-point contraction artifacts (e.g. GCC FMA fusion
+// across function boundaries) that can make the difference negative or
+// spuriously non-zero for equal boxes. See GitHub issue #1452.
+template <typename Box>
+typename default_content_result<Box>::type content_diff(Box const& expanded, Box const& original)
+{
+    using content_type = typename default_content_result<Box>::type;
+    return (std::max)(content_type(0), content(expanded) - content(original));
+}
+
 #if defined(BOOST_GCC)
 #pragma GCC pop_options
 #endif
--- a/include/boost/geometry/index/detail/rtree/linear/redistribute_elements.hpp
+++ b/include/boost/geometry/index/detail/rtree/linear/redistribute_elements.hpp
@@ -16,6 +16,7 @@
 #ifndef BOOST_GEOMETRY_INDEX_DETAIL_RTREE_LINEAR_REDISTRIBUTE_ELEMENTS_HPP
 #define BOOST_GEOMETRY_INDEX_DETAIL_RTREE_LINEAR_REDISTRIBUTE_ELEMENTS_HPP
 
+#include <algorithm>
 #include <type_traits>
 
 #include <boost/core/ignore_unused.hpp>
@@ -432,8 +433,10 @@ struct redistribute_elements<MembersHolder, linear_tag>
                         content_type enlarged_content1 = index::detail::content(enlarged_box1);
                         content_type enlarged_content2 = index::detail::content(enlarged_box2);
 
-                        content_type content_increase1 = enlarged_content1 - content1;
-                        content_type content_increase2 = enlarged_content2 - content2;
+                        content_type const content_increase1
+                            = (std::max)(content_type(0), enlarged_content1 - content1);
+                        content_type const content_increase2
+                            = (std::max)(content_type(0), enlarged_content2 - content2);
 
                         // choose group which box content have to be enlarged least or has smaller content or has fewer elements
                         if ( content_increase1 < content_increase2 ||
--- a/include/boost/geometry/index/detail/rtree/quadratic/redistribute_elements.hpp
+++ b/include/boost/geometry/index/detail/rtree/quadratic/redistribute_elements.hpp
@@ -74,9 +74,10 @@ inline void pick_seeds(Elements const& elements,
 
             bounded_indexable_view bounded_ind1(ind1, strategy);
             bounded_indexable_view bounded_ind2(ind2, strategy);
-            content_type free_content = ( index::detail::content(enlarged_box)
-                                            - index::detail::content(bounded_ind1) )
-                                                - index::detail::content(bounded_ind2);
+            content_type free_content = (std::max)(content_type(0),
+                ( index::detail::content(enlarged_box)
+                    - index::detail::content(bounded_ind1) )
+                        - index::detail::content(bounded_ind2));
 
             if ( greatest_free_content < free_content )
             {
@@ -273,7 +274,7 @@ struct redistribute_elements<MembersHolder, quadratic_tag>
         typedef typename boost::iterator_value<It>::type element_type;
         typedef typename rtree::element_indexable_type<element_type, translator_type>::type indexable_type;
 
-        content_type greatest_content_incrase_diff = 0;
+        content_type greatest_content_increase_diff = 0;
         It out_it = first;
         out_content_increase1 = 0;
         out_content_increase2 = 0;
@@ -291,18 +292,20 @@ struct redistribute_elements<MembersHolder, quadratic_tag>
             content_type enlarged_content1 = index::detail::content(enlarged_box1);
             content_type enlarged_content2 = index::detail::content(enlarged_box2);
 
-            content_type content_incrase1 = (enlarged_content1 - content1);
-            content_type content_incrase2 = (enlarged_content2 - content2);
+            content_type const content_increase1
+                = (std::max)(content_type(0), enlarged_content1 - content1);
+            content_type const content_increase2
+                = (std::max)(content_type(0), enlarged_content2 - content2);
 
-            content_type content_incrase_diff = content_incrase1 < content_incrase2 ?
-                content_incrase2 - content_incrase1 : content_incrase1 - content_incrase2;
+            content_type content_increase_diff = content_increase1 < content_increase2 ?
+                content_increase2 - content_increase1 : content_increase1 - content_increase2;
 
-            if ( greatest_content_incrase_diff < content_incrase_diff )
+            if ( greatest_content_increase_diff < content_increase_diff )
             {
-                greatest_content_incrase_diff = content_incrase_diff;
+                greatest_content_increase_diff = content_increase_diff;
                 out_it = el_it;
-                out_content_increase1 = content_incrase1;
-                out_content_increase2 = content_incrase2;
+                out_content_increase1 = content_increase1;
+                out_content_increase2 = content_increase2;
             }
         }
 
--- a/include/boost/geometry/index/detail/rtree/rstar/choose_next_node.hpp
+++ b/include/boost/geometry/index/detail/rtree/rstar/choose_next_node.hpp
@@ -117,7 +117,7 @@ private:
 
             // areas difference
             content_type content = index::detail::content(box_exp);
-            content_type content_diff = content - index::detail::content(ch_i.first);
+            content_type content_diff = index::detail::content_diff(box_exp, ch_i.first);
 
             children_contents[i].set(i, content, content_diff);
 
@@ -246,7 +246,7 @@ private:
 
             // areas difference
             content_type content = index::detail::content(box_exp);
-            content_type content_diff = content - index::detail::content(ch_i.first);
+            content_type content_diff = index::detail::content_diff(box_exp, ch_i.first);
 
             // update the result
             if ( content_diff < smallest_content_diff ||
--- a/include/boost/geometry/index/detail/rtree/visitors/insert.hpp
+++ b/include/boost/geometry/index/detail/rtree/visitors/insert.hpp
@@ -89,7 +89,7 @@ public:
 
             // areas difference
             content_type content = index::detail::content(box_exp);
-            content_type content_diff = content - index::detail::content(ch_i.first);
+            content_type content_diff = index::detail::content_diff(box_exp, ch_i.first);
 
             // update the result
             if ( content_diff < smallest_content_diff ||

### CODE REVIEW

File: include/boost/geometry/index/detail/rtree/quadratic/redistribute_elements.hpp [hunk 1]
```diff
@@ -291,18 +292,18 @@ struct redistribute_elements<MembersHolder, quadratic_tag>
...
-            if ( greatest_content_incrase_diff < content_incrase_diff )
+            if ( greatest_content_increase_diff < content_increase_diff )
             {
-                greatest_content_incrase_diff = content_incrase_diff;
+                greatest_content_increase_diff = content_increase_diff;
                 out_it = el_it;
-                out_content_increase1 = content_incrase1;
-                out_content_increase2 = content_incrase2;
+                out_content_increase1 = content_increase1;
+                out_content_increase2 = content_increase2;
```
barendgehrels, line 308: I also fixed a typo, that's why the diff is a bit larger here

tinko92:
I tried something similar to what the language model outputs here before I wrote #1453 , but instead of clamping, I tested the values of the content calls directly for equality. I went into a different direction because

1) it added an extra branch and prevented contraction to FMA (because the compiler needed to compute both intermediate results for comparing anyway), so no advantage over turning of contraction for content directly. This clamping is better though because max is branchless and it still allows applying FMA, so 1. is resolved.

2) It did not fully resolve the issue.

Consider the following test using the instrumented [main.cpp](https://github.com/user-attachments/files/26345226/main.cpp.txt) and the [data.csv](https://github.com/user-attachments/files/25870653/data.csv) from the issue:
```bash
#!/usr/bin/env bash

(cd ../../boost/libs/geometry/ && git reset --hard -q && git checkout $1 -q && sed -i '504i\cross_track_calls++;' include/boost/geometry/strategies/geographic/distance_cross_track.hpp && git branch --show-current)
g++ -g -I../../boost/ -DNDEBUG -O3 -march=$2 main.cpp -std=c++20 -o a.out && ./a.out | grep -E "^2:|query all took|cross_track calls:"
```
(Edit: Line 504 in distance_cross_track.hpp is inside the private apply method.)

Then we get:
```sh
bartels@DESKTOP-BPNV3G3:~/dev/bg_1452/test$ sh test.sh develop x86-64-v2 # Baseline without issue (v2)
develop
2: 6763
query all took :   3460.036 ms
cross_track calls: 3303357

bartels@DESKTOP-BPNV3G3:~/dev/bg_1452/test$ sh test.sh develop x86-64-v3 # Issue (v3)
develop
2: 6685
query all took :   7346.178 ms
cross_track calls: 6117432

bartels@DESKTOP-BPNV3G3:~/dev/bg_1452/test$ sh test.sh fix/index-content-fp-contract-off x86-64-v3 #1453 
fix/index-content-fp-contract-off
2: 6763 # Probably the same tree structure
query all took :   3496.460 ms
cross_track calls: 3306476 # Minor difference in distance calls, might be due to difference in distance itself.

bartels@DESKTOP-BPNV3G3:~/dev/bg_1452/test$ sh test.sh fix/issue-1452-index-content x86-64-v3 # This PR
fix/issue-1452-index-content
2: 6697 # Different tree structure from baseline.
query all took :   4120.981 ms
cross_track calls: 4200675 # Still a severe regression in the number of distance calls.
```

So there must be call sites of content that were not covered fully by the LLM-generated patch. I still like the change in itself, though, because it seems like a sensibly defensive thing to do that causes no branching and prevents potentially nonsensical values. It may be beneficial to merge both PRs (the original PR also e.g. prevents issues with contraction of content at potential future call sites or existing sites unrelated to this bug if there are any).

include/boost/geometry/index/detail/rtree/linear/redistribute_elements.hpp
```diff
@@ -432,8 +433,8 @@ struct redistribute_elements<MembersHolder, linear_tag>
                         content_type enlarged_content1 = index::detail::content(enlarged_box1);
                         content_type enlarged_content2 = index::detail::content(enlarged_box2);
 
-                        content_type content_increase1 = enlarged_content1 - content1;
-                        content_type content_increase2 = enlarged_content2 - content2;
+                        content_type const content_increase1 = (std::max)(content_type(0), enlarged_content1 - content1);
+                        content_type const content_increase2 = (std::max)(content_type(0), enlarged_content2 - content2);
```
tinko92, line 437: Minor: These violate our max line length of 100.

tinko92 (approved):
I approve merging these changes, the one formatting comment is no blocker for me.

include/boost/geometry/index/detail/rtree/linear/redistribute_elements.hpp
barendgehrels: Fixed

tinko92:
On further inspection, #1452 is also affected by yet another source of rounding error, relating to antimeridian-crossing boxes in coordinate normalisation for boxes (which also affects x86-64-v2, although in a more minor way). E.g., if we have a box with lower lon 170°, upper long e.g. 190.1° it will first subtract -360°, then add 360°, and change it to something like 190.09999...°, and the content change from expansion becomes negative. I will add a separate PR with a test case to solve that (this is needed in addition to the fixes around content_diff).

~The content_diff-related issues can be fully resolved by making your content_diff robust with a filter, that not only clamps negative values but also rounds to 0 for very small positive values for which the actual content diff is 0. This approach makes both #1453 obsolete and is more general than the clamping proposed in this PR (#1455). I will try to write a new PR by end of next week, there are a few options, I want to compare first. Bascially we can have a static threshold which would be around 1.32e-10 for double (may be acceptable?), but for float it would be around 0.071 which may be too coarse. Or we can compute it per content diff call depending on the boxes which would be more accurate and preferable if the performance overhead turns out to be negligible in testing.~

EDIT: After a lot of testing, the filtering looked good on ARM but failed to resolve the issue with GCC on x86 (similar results to this PR #1455, roughly ~4M distance calls, rather than ~2.6M). As of now, #1453 is the only solution that works for me across all platforms.

include/boost/geometry/index/detail/rtree/linear/redistribute_elements.hpp
tinko92, line 437: Let's call content_diff here for consistency and in particular to retain consistency if that function is changed for robustness in the future. Recomputing content1 and content2 is not a performance concern here, all major compilers will pull that out of the loop on O2 ([tested for GCC, Clang, MSVC, ICX here](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZgQDbYB2AhgLbYgDkAjF%2BTXRMiAZVQtGIHgBYBQogFUAztgAKAD24AGfgCsp5eiyahUAUgBMAIUtXyKxqiIEh1ZpgDC6egFc2TEAtydwAZAiZsADk/ACNsUikATnIAB3QlYhcmL19/QNT0zKEwiOi2OISeZIdsJyyRIhZSIhy/AKCauqEGpqISqNj4pPtG5ta8jtG%2B8IHyoaqASnt0H1JUTi5LAGZw1F8cAGozLY9UNhYiJGPcMy0AQW3d/ewjk4A3WpJSa9uH%2B6UiKQfE4DjF0OojgB2Gz3A5wg6YFYxRgHej0dTkVH0ACemNWGIOq2xxxhD0hABESb9fkRsGwUkZaccPERsSlmOwXlZwT97oifMiXhhhMwiBBuRDhQDLAA2UELKGk%2BEHUjYIirJgHCAxAB0%2BIOAFpQTq0eoFQAqLW6omG41o7ELKn3MwUo7OvlIlFscIAfUINBo2vBBylRFloMxAMwIBAHycZGZYPU1xDQmlFjlMSUCpdSvh/MFoOwANe5IOAAkFABxXA%2BgBqdxCTruysEpAgLB8JFTTHTmZiBxAoOzv2VyoIgeFtOE2oVRqnou1MRzJyLAMdsLHW7iJeOZYXM%2BXtoPYpiy%2BbytV6tImp3Ya2pJd5K4S3o3AArPwAlwdOR0NwPFsWwDiUFY1hebY%2BHIIhtBfJYAGsQC2aQdQsCwqi2LYLGkGUZWkLQZUhQxuGkfg2ECSEdWkSE8J4d8tESaRMMSGUgm/X9/y4fglBALRoNgpY4FgFAMHpBh4koahRJScSEmAOigjoehaVIHjtVg8gYnCJpsW4KCtNYUhsQAeRiXRPj0/hRI4YRjKYHENJwGIfGADwJHoHjeH4HBzhMSQf28ghVTqD5PN/bB1FqLsNig8JaTfALDAIGJSB0rwcA0wECHIrzyA%2BUgwRUck6WMYB6HCUAAqWGgjGAJQ6wIbAAHdjPZb8oMEYQxAkTgZDkYRlDUTREv0LZDFKtAgOsJKs3gJZ0BSZw024A1jIsfh0Hy0gCBwHjICWToloCCB3HGAIeGCJhMH6MoKgMNIMiOs77sKI6bsGSp7GwRwjp6MZvDaAxDvqKZ3rmT6AV6Z6Lsh5owbungDrA9YpFfD8vw0ziDnUAAOGUDTwg5gFQVADjo1CtXwYgyCOCwtkR/gYKqpYkGwFgcASCA0a4UjyHIrZEh1HH3xx6REi0d9cK0LZIUhWR2I27huN4/iqvIITkDQdAxMYCgqAgaTZJAAhUHURliyIAQGBUtSYg0gydMszTtKM0zzKcJ3rNFOyHMSpyXLctFPKgnzSv839CGC5xQo0iKotpJ24u%2BjTypStKsA2X8spyqD8sK7Bit8sqKoEgRavqxqWrap3OtEcRJD62vBo0DT9FkIwTEm6xbBmvaub/Rask81b1r/LaduLOavp%2BrI3Cu6HLuumZbvmApHqyBeHqKJh4dX4HuimBf96YP7plKD6gcPgG8hh0Hl4vxHllWFHH4Sz9yAVv9uGxvGCekA4TYQgsFoCw74dRbB1DwSmhAvi03pgsRmAkWZsw5tQbmvN%2BYWB1LheijFmKsQzO%2BD%2BmMlb2BVkzHQCwEIgHotg3C0gxaJGQokHGEsCLES4FsDGiVOKILVhrARIltYyV1pJA2wijZ7FKmhHgfElI22oHbRKDsjJOxUSZMyFlcpe1svZXSftsDOVcu5YO3kSp%2BUzoFKOBAY6JTjqgaKicRQJV/KnVKRl0qWOgttHO/A87pALuY4unc1Y1RYHVBqzVWrMBrvIbqDdZBNxUC3EaBgO6mCmnYVOfd5qD2WlwEeG1x67SnsfOenhr7nUXrvT6W8nqVJeuvYo99wZA2%2Bp8EGUMGkw3aV0E%2Bd9z6tNvl03IVTYZn1mAjJGz9erc3fp/LGuN8aEykSYA4aFIE6i0NA6mpA4EM1VpQ6hYtwGYXfPRAi9EqjvguglDBNC%2BILNITxPiFCXzq0QBAIROsJL60NrrWMqAUgpB9G8KoPp6L1hEDwHGPoLA%2BiWVbZS8Rbb2xdvo/S6K3ZaKgjoogPt9ER0MQHExTtQ4WMckFDptjwqRQcQnXKScXH8DcenDKiVs5O38UVIJ5UQmULLuEiuUTq65VrvE3qiT5DN2Gr%2BfQQR0ldxsNNbJU8FpHR4gAejWmPeI20Sn7Wnh01wJ157dNkKEFpd1pBBDqRvc1a9t41JADao1fTT7PWqL036AzJlDFdeMz1IxejOptdM8CqMOHzJIVwH%2Bco2BKDeKTMFgt6IHDrNCnGFMIBU1gZBBBhy4LkFZuzIY/c7lkQecQnhzzyFIPIIhE5mEtjnK0JcrQ1zbncC4dWjipC3lUI4aPJ5XFC2DvyhkVw0ggA)).

include/boost/geometry/index/detail/rtree/quadratic/redistribute_elements.hpp
```diff
@@ -291,18 +292,20 @@ struct redistribute_elements<MembersHolder, quadratic_tag>
             content_type enlarged_content1 = index::detail::content(enlarged_box1);
             content_type enlarged_content2 = index::detail::content(enlarged_box2);
 
-            content_type content_incrase1 = (enlarged_content1 - content1);
-            content_type content_incrase2 = (enlarged_content2 - content2);
+            content_type const content_increase1
+                = (std::max)(content_type(0), enlarged_content1 - content1);
```
tinko92, line 296: I think L296 and L297 should call content_diff, as discussed in comment on redistribute_elements.hpp:437.

include/boost/geometry/index/detail/algorithms/content.hpp
```diff
@@ -98,6 +100,17 @@ typename default_content_result<Indexable>::type content(Indexable const& b)
...
 }
 
+// Returns the content increase when expanding 'original' to 'expanded'.
+// Clamped to 0 to guard against floating-point contraction artifacts (e.g.
+// GCC FMA fusion across function boundaries) that can make the difference
+// negative or spuriously non-zero for equal boxes. See GitHub issue #1452.
+template <typename Box>
+typename default_content_result<Box>::type content_diff(Box const& expanded, Box const& original)
+{
+    using content_type = typename default_content_result<Box>::type;
```
tinko92, line 110: I propose an asssertion on original being within expanded here, because that is a precondition for that general clamping being valid and should prevent accidental misuse of this function for general content differences where arg1 is not an expansion of arg2.

tinko92 (commented):
Because this was questioned https://github.com/boostorg/geometry/pull/1453#issuecomment-4232744623 , I am still in favour of merging this PR. The clamping is still reasonable and the refactoring is useful for potential future formula changes specific to content_diff to make this maybe more robust in the future.

include/boost/geometry/index/detail/algorithms/content.hpp
barendgehrels: I'm not sure about it. It is already documented, in namespace detail (not general usage), and an assertion on a Box would either mean dragging in geometry::covered_by, or making a custom similar function here (code duplication).

include/boost/geometry/index/detail/rtree/quadratic/redistribute_elements.hpp
barendgehrels: ✅

include/boost/geometry/index/detail/rtree/linear/redistribute_elements.hpp
barendgehrels: ✅

barendgehrels:
most points applied, thanks. Pushed.

include/boost/geometry/index/detail/algorithms/content.hpp
tinko92: Fair, then the existing comment is ok for me.

barendgehrels: 👍 thanks for your approvals! Merging this one too.

PR #1456, opened 2026-04-07
Skip unnecessary longitude normalization on antimeridian-crossing box

tinko92:
The program in issue #1452 triggers some cases in which a box shrinks when being expanded with a point that is already within it, due to rounding errors introduced unnecessarily in coordinate normalization. When a box crosses the antimeridian, the longitude of its max point is first normalized to be within [-180, 180], then increased by half_period() again, which can introduce rounding error, including shrinking the box. the shrinking then seems to produce unstable behaviour in the rtree when the computed area difference between a boxes expansion and itself is negative. This is only a small part of the tree instability in #1452 and does not resolve most of the issue but isolated enough from the other issues to be treated in a separated PR.

This PR contains the following changes:
1. Adds a flag that detects when a box crosses the antimeridian and has otherwise valid max longitude, and passes this flag down to coordinate normalization, so that this longitude is then not normalized down.
2. Add a test case with tolerance=0.0, in which the box would shrink without this change.
3. Update the box_equal code to respect the passed tolerance in longitude comparison rather than using the (more lenient) bg::math::equals unconditionally.

### DIFF
--- a/include/boost/geometry/util/normalize_spheroidal_box_coordinates.hpp
+++ b/include/boost/geometry/util/normalize_spheroidal_box_coordinates.hpp
@@ -48,8 +48,13 @@ public:
                              CoordinateType& latitude2,
                              bool band)
     {
+        bool const allow_antimeridian_crossing =
+               longitude1 >= constants::min_longitude()
+            && longitude1 <= constants::max_longitude()
+            && longitude2 > constants::max_longitude()
+            && longitude2 - constants::period() < longitude1;
         normalize::apply(longitude1, latitude1, false);
-        normalize::apply(longitude2, latitude2, false);
+        normalize::apply(longitude2, latitude2, false, true, allow_antimeridian_crossing);
 
         latitude_convert_if_polar<Units, IsEquatorial>::apply(latitude1);
         latitude_convert_if_polar<Units, IsEquatorial>::apply(latitude2);
--- a/test/algorithms/envelope_expand/test_envelope_expand_on_spheroid.hpp
+++ b/test/algorithms/envelope_expand/test_envelope_expand_on_spheroid.hpp
@@ -246,7 +246,7 @@ struct box_check_equals
             && (bg::get<0, 1>(box) < 0
                 ? equals(bg::get<0, 1>(box), lat_min)
                 : equals_with_eps(bg::get<0, 1>(box), lat_min))
-            && equals_with_eps(bg::get<1, 0>(box), lon_max)
+            && equals(bg::get<1, 0>(box), lon_max)
             && (bg::get<1, 1>(box) > 0
                 ? equals(bg::get<1, 1>(box), lat_max)
                 : equals_with_eps(bg::get<1, 1>(box), lat_max));
--- a/include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
+++ b/include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
@@ -223,7 +223,8 @@ protected:
     }
 
 public:
-    static inline void apply(CoordinateType& longitude, bool exact = true)
+    static inline void apply(CoordinateType& longitude, bool exact = true,
+                             bool allow_antimeridian_crossing = false)
     {
         // normalize longitude
         CoordinateType const epsilon = std::numeric_limits<float>::epsilon();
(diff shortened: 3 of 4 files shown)


test/algorithms/envelope_expand/test_envelope_expand_on_spheroid.hpp
```diff
@@ -246,7 +246,7 @@ struct box_check_equals
             && (bg::get<0, 1>(box) < 0
                 ? equals(bg::get<0, 1>(box), lat_min)
                 : equals_with_eps(bg::get<0, 1>(box), lat_min))
-            && equals_with_eps(bg::get<1, 0>(box), lon_max)
+            && equals(bg::get<1, 0>(box), lon_max)
             && (bg::get<1, 1>(box) > 0
```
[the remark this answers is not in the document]
tinko92: This change is intentionally minimal. I do not touch the tolerances on the other coordinate checks unrelated to this PR since I do not know for which failing tests these were made more lenient and my dev platform (ARM64) differs in some test behaviours from the most common platform (x86).

include/boost/geometry/util/normalize_spheroidal_box_coordinates.hpp
```diff
@@ -49,7 +49,11 @@ class normalize_spheroidal_box_coordinates
                              bool band)
     {
         normalize::apply(longitude1, latitude1, false);
-        normalize::apply(longitude2, latitude2, false);
+        const bool crosses_antimeridian =
+               longitude2 > longitude1
+            && longitude2 > constants::half_period()
+            && longitude2 <= constants::period();
+        normalize::apply(longitude2, latitude2, false, true, crosses_antimeridian);
```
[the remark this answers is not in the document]
tinko92: Addressed. The conditions are now very narrow (longitudes already have to be normalized). This is still useful, especially for circumstances in which this may be called on the same box multiple times (e.g. in tree insertion). Not resolving yet to wait for other opinions.

include/boost/geometry/util/normalize_spheroidal_box_coordinates.hpp
```diff
@@ -48,8 +48,13 @@ class normalize_spheroidal_box_coordinates
                              CoordinateType& latitude2,
                              bool band)
     {
+        const bool crosses_antimeridian =
```
barendgehrels: Officially we use: `bool const`
(I'm glad I can contribute something copilot didn't spot ;-) )

include/boost/geometry/util/normalize_spheroidal_box_coordinates.hpp
```diff
@@ -48,8 +48,13 @@ class normalize_spheroidal_box_coordinates
...
+               longitude1 >= -constants::half_period()
+            && longitude1 <= constants::half_period()
+            && longitude2 > constants::half_period()
+            && longitude2 <= constants::period();
         normalize::apply(longitude1, latitude1, false);
-        normalize::apply(longitude2, latitude2, false);
+        normalize::apply(longitude2, latitude2, false, true, crosses_antimeridian);
```
barendgehrels: elsewhere the same parameter is called "allow_antimeridian_crossing". 
Should we harmonize the names? It's a bit confusing.

include/boost/geometry/util/normalize_spheroidal_box_coordinates.hpp
tinko92: Thanks, I updated it in a new commit.

include/boost/geometry/util/normalize_spheroidal_box_coordinates.hpp
tinko92: My thoughts with the naming were that when looking at a box we can check whether it crosses the meridian and the point normalization needs to know whether exceeding the max longitude is allowed for this point (without knowing about the entire box) but I agree that consistent naming is more clear here, I changed it in a new commit.

include/boost/geometry/util/normalize_spheroidal_box_coordinates.hpp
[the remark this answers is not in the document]
tinko92: Changed the condition one more time. I hope this is semantically correct and robust. longitude1 needs to be normal (in [min_longitude, max_longitude]), longitude2 needs to be >max_longitude but <longitude1 + period (rewritten as longitude2 - period < longitude1, which should be exact due to Sterbenz lemma in the relevant range of longitude2 in [180°, 540°] and guaranteed false otherwise):

```cpp
        bool const allow_antimeridian_crossing =
               longitude1 >= constants::min_longitude()
            && longitude1 <= constants::max_longitude()
            && longitude2 > constants::max_longitude()
            && longitude2 - constants::period() < longitude1;
```

barendgehrels (approved):
Thanks for the fix! I'm OK

vissarion (approved):
Look OK, thanks!

PR #1458, opened 2026-04-19
Drop varray implementation in favor of Boost.Container static_vector.

tinko92:
This PR proposes to drop the varray implementation in favour of Boost.Container static_vector, and Boost.Container vector in favour of std::vector. From my understanding static_vector and varray share a common ancestry. This would drop several tests from the regular runs and more than 4000 lines to maintain.

I noticed recent increased activity and optimizations in Boost.Container and retested varray vs static_vector with the builtin benchmark.cpp from the index/examples directory and Boost.Container checked out to the most recent develop.

Build for the release variation:
```sh
b2 cxxflags="-std=c++20 -O3 -DNDEBUG" variant=release
```

Difference in measurements (Apple M5, Clang21):
```sh
➜  ~ head -n200 benchmark_varray.txt | grep "insert" | gawk '{print $1}' | awk '{sum+=$1} END {print "Average = ", sum/NR}'       
Average =  0.182573
➜  ~ head -n200 benchmark_static_vector.txt | grep "insert" | gawk '{print $1}' | awk '{sum+=$1} END {print "Average = ", sum/NR}' 
Average =  0.182281
➜  ~ head -n200 benchmark_varray.txt | grep "query" | gawk '{print $1}' | awk '{sum+=$1} END {print "Average = ", sum/NR}' 
Average =  0.203972
➜  ~ head -n200 benchmark_static_vector.txt | grep "query" | gawk '{print $1}' | awk '{sum+=$1} END {print "Average = ", sum/NR}'
Average =  0.208022
➜  ~ head -n200 benchmark_varray.txt | grep "remove" | gawk '{print $1}' | awk '{sum+=$1} END {print "Average = ", sum/NR}'
Average =  0.082499
➜  ~ head -n200 benchmark_static_vector.txt | grep "remove" | gawk '{print $1}' | awk '{sum+=$1} END {print "Average = ", sum/NR}' 
Average =  0.086818
```

The difference seems none to negligible and Boost.Container is actively maintained. Switching to this could be a step in the upgrade path to `std::inplace_vector` once Boost.Geometry switches to C++26. Unfortunately I couldn't test inplace_vector yet, since it's not yet implemented in the STL of clang 21 (with the beman implementation I get 0.175794, 0.193613, 0.0790506 respectively, though, so the spec may allow implementations that remain both drop-in-compatible for our use case and faster).


include/boost/geometry/index/detail/rtree/iterators.hpp
```diff
@@ -17,6 +17,7 @@
 
 #include <iterator>
 
+#include <boost/geometry/index/detail/assert.hpp>
```
tinko92, line 20: Not directly related to this change, but this is necessary to fix a CI failure with this change, probably due to relying on indirectly importing this for the `BOOST_GEOMETRY_INDEX_ASSERT` macro through some removed header.

index/test/Jamfile
```diff
@@ -27,8 +27,6 @@ project boost-geometry-index-test
 test-suite boost-geometry-index-detail
     :
     [ run minmax_heap.cpp ]
-    [ run varray_old.cpp ]
```
tinko92, line 30: The rationale for dropping these tests is that testing static_vector is done upstream in Boost.Container.

include/boost/geometry/index/detail/rtree/kmeans/split.hpp
```diff
@@ -79,11 +81,11 @@ class split<MembersHolder, split_kmeans_tag>
     typedef typename MembersHolder::leaf leaf;
 
 public:
-    typedef index::detail::varray
+    using nodes_container_type = boost::container::static_vector
```
barendgehrels, line 84: Pretty nice, removing the whole `varray`!

It is nowhere part of our public interface?

index/test/movable.hpp
barendgehrels, line 1: Only used in varray - so fine to drop indeed

barendgehrels (approved):
👍 thanks!

include/boost/geometry/index/detail/rtree/kmeans/split.hpp
tinko92: As far as I can see, it is not. 

1. The files and classes are in detail folders and namespaces respectively. 
2. The only mention I see in the documentation is the detail header listing and the credit to Andrew Hundt. 
3. The serialization test (patched because it doesn't currently build due to deprecation of the "boost/timer.h" header file) in index examples runs with and without the patch for me, and when patched to load only, it seems to be able to load from the archive created by the version from current develop, so it does not seem to break serialization (when done as in this example). Side note: Do we want a stable binary serialization format for rtree in a header only library? If so, we may want to move this from an experimental example into a test.

barendgehrels: OK Clear, thanks. Yes, I agree, it is most probably not.

We should anyway add a note in the release notes for this larger change (BTW, that applies for my both PR's too, though smaller).

I'm fine with my approval but curious to @vissarion 's insights.

About binary serialization - not sure - we could open a "discussion" about it.

barendgehrels:
@tinko92 can you merge it?

vissarion (approved):
Thanks Tinko! Reading your changes, the explanations, remarks and benchmarks, looks like a safe choice that will improve the code base.

PR #1459, opened 2026-04-25
Refactor/reduce misc boost dependencies

tinko92:
This PR removes a number of dependencies to other Boost libraries. It is more narrow in scope than the prior #1425 (closed it in favour of this) and cleanly (I hope) separates changes to extensions from changes that go into a release. I tried to limit it to headers and features that have one to one replacements to make it more reviewable. I will add comments where I deviate from that. 
Varray was intentionally untouched because I propose to remove it in #1458 and don't want to create merge conflicts. This PR is intended to come after that, if the remove does not happen, I can update this one.

PR #1463, opened 2026-05-05
Remove boost_static_assert dependency from b2 build

vissarion:


tinko92:
This change is also proposed in the not yet merged PR #1459 , which removes the code dependencies on boost static_assert which are still in current develop.

vissarion:
Oh sure, I missed that while trying to quickly fix the CI builds, I will review https://github.com/boostorg/geometry/pull/1459 instead.

PR #1474, opened 2026-06-15
Fix degenerate linestring relation on boundary of areal

### LINKED ISSUE #1473: Surprising relation results for degenerate linestring on boundary of areal
E.g. `covered_by` returns `false` for `LINESTRING(0 0,0 0)` and `POLYGON((0 0, 0 2, 2 2, 2 0, 0 0))`. I suppose the library does not guarantee a sensible result since `is_valid` is `false` for that linestring with `failure_wrong_topological_dimension`. If possible I think it would be great if the library could give less surprising results though, even for such invalid linestrings.

Seems like https://github.com/mjacobse/geometry/commit/0fe1eaca0c9efe3acee9106c5b03661b7a2ee2ca could indeed be a quick fix, the 
"on corner" and "on edge" test cases now pass where they wouldn't before. Though I do not quite understand how this fits into the "no turns" case and what should be done with the interrupt flags.

Feel free to close this as wontfix if such degenerate linestrings are just not intended to be supported. Otherwise I'd be happy to discuss the potential fix further.

mjacobse:
Basically makes it so `covered_by` returns `true` for `LINESTRING(0 0,0 0)` and `POLYGON((0 0, 0 2, 2 2, 2 0, 0 0))` instead of `false` as it did previously, which could be quite surprising.

Fixes #1473

The bit flags for interrupt aren't exactly pretty, but it follows the present implementation.

### DIFF
--- a/include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
+++ b/include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
@@ -65,26 +65,26 @@ public:
         , m_result(res)
         , m_strategy(strategy)
         , m_boundary_checker(boundary_checker)
-        , m_interrupt_flags(0)
+        , m_interrupt_flags(0x00)
     {
         if ( ! may_update<interior, interior, '1', TransposeResult>(m_result) )
         {
-            m_interrupt_flags |= 1;
+            m_interrupt_flags |= 0x01;
         }
 
         if ( ! may_update<interior, exterior, '1', TransposeResult>(m_result) )
         {
-            m_interrupt_flags |= 2;
+            m_interrupt_flags |= 0x02;
         }
 
         if ( ! may_update<boundary, interior, '0', TransposeResult>(m_result) )
         {
-            m_interrupt_flags |= 4;
+            m_interrupt_flags |= 0x04;
         }
 
         if ( ! may_update<boundary, exterior, '0', TransposeResult>(m_result) )
         {
-            m_interrupt_flags |= 8;
+            m_interrupt_flags |= 0x08;
         }
 
         if ( ! may_update<interior, boundary, '1', TransposeResult>(m_result) )
@@ -119,7 +119,7 @@ public:
         if ( pig > 0 )
         {
             update<interior, interior, '1', TransposeResult>(m_result);
-            m_interrupt_flags |= 1;
+            m_interrupt_flags |= 0x01;
         }
         else if ( pig == 0 )
         {
@@ -132,7 +132,7 @@ public:
         else
         {
             update<interior, exterior, '1', TransposeResult>(m_result);
-            m_interrupt_flags |= 2;
+            m_interrupt_flags |= 0x02;
         }
 
         // check if there is a boundary
@@ -143,12 +143,12 @@ public:
             if ( pig > 0 )
             {
                 update<boundary, interior, '0', TransposeResult>(m_result);
-                m_interrupt_flags |= 4;
+                m_interrupt_flags |= 0x04;
             }
             else
             {
                 update<boundary, exterior, '0', TransposeResult>(m_result);
-                m_interrupt_flags |= 8;
+                m_interrupt_flags |= 0x08;
             }
         }
 
--- a/include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
+++ b/include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
@@ -99,7 +99,7 @@ public:
         std::size_t const count = boost::size(linestring);
 
         // invalid input
-        if ( count < 2 )
+        if ( count < 1 )
         {
             // ignore
             // TODO: throw an exception?
--- a/include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
+++ b/include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
@@ -86,6 +86,11 @@ public:
         {
             m_interrupt_flags |= 8;
         }
+
+        if ( ! may_update<interior, boundary, '1', TransposeResult>(m_result) )
+        {
+            m_interrupt_flags |= 0x10;
+        }
     }
 
     template <typename Linestring>
@@ -102,7 +107,7 @@ public:
         }
 
         // if those flags are set nothing will change
-        if ( m_interrupt_flags == 0xF )
+        if ( m_interrupt_flags == 0x1F )
         {
             return false;
         }
@@ -110,12 +115,19 @@ public:
         int const pig = detail::within::point_in_geometry(range::front(linestring),
                                                           m_geometry2,
                                                           m_strategy);
-        //BOOST_GEOMETRY_ASSERT_MSG(pig != 0, "There should be no IPs");
 
         if ( pig > 0 )
         {
             update<interior, interior, '1', TransposeResult>(m_result);
             m_interrupt_flags |= 1;
+        }
+        else if ( pig == 0 )
+        {
+            // no turns but still point on boundary can actually happen when the
+            // linestring is degenerate. so handle this case explicitly here.
+            // for consistency let's still report dimension 1 instead of 0
+            update<interior, boundary, '1', TransposeResult>(m_result);
+            m_interrupt_flags |= 0x10;
         }
         else
         {
@@ -140,7 +152,7 @@ public:
             }
         }
 
-        return m_interrupt_flags != 0xF
+        return m_interrupt_flags != 0x1F
             && ! m_result.interrupt;
     }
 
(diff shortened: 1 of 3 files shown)


include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
barendgehrels, line 105: Now we handle degenerate as well, we could remove this comment.
It's a different situation (two points i/o) but further the same.
Maybe if count == 1, it now passes the same way as count == 1 with duplicate points

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
```diff
@@ -86,6 +86,11 @@ class no_turns_la_linestring_pred
         {
             m_interrupt_flags |= 8;
         }
+
+        if ( ! may_update<interior, boundary, '1', TransposeResult>(m_result) )
+        {
+            m_interrupt_flags |= 0x10;
```
barendgehrels, line 92: Thanks. Agreed the `m_interrupt_flags` are not pretty, I don't think the name is clear, and it mixes hexadecimal with decimal (already did). 

Maybe we can at least make them all hexadecimal (in this PR, I mean).

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
barendgehrels, line 148: Should we handle `pig == 0` here too?

barendgehrels:
Thanks for the PR, is it a bit larger than earlier? Anyway, that's not a problem, will continue the review later

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
mjacobse: A degenerate linestring does not have a boundary itself, so this case shouldn't be possible I believe.

mjacobse:
> Thanks for the PR, is it a bit larger than earlier? Anyway, that's not a problem, will continue the review later

Adding the handling of the interrupt flags for this case now made it a bit bigger yes

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
mjacobse: Oh yeah interesting, I hadn't considered that case, will have to try if it could work the same way now

mjacobse: Changing this to `count < 1` and adding copies of the tests with just one point in the linestring indeed seems to work. Would you like me to include that in this PR or keep this for later?

barendgehrels: Thanks!
Let's include that in this PR, it fits and will be less review/merge/administrative work

barendgehrels: All right

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
```diff
@@ -143,12 +143,12 @@ class no_turns_la_linestring_pred
...
             {
                 update<boundary, interior, '0', TransposeResult>(m_result);
-                m_interrupt_flags |= 4;
+                m_interrupt_flags |= 0x04;
             }
             else
             {
                 update<boundary, exterior, '0', TransposeResult>(m_result);
-                m_interrupt_flags |= 8;
+                m_interrupt_flags |= 0x08;
```
barendgehrels, line 151: Thanks!

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
```diff
@@ -65,26 +65,31 @@ class no_turns_la_linestring_pred
...
 
         if ( ! may_update<boundary, exterior, '0', TransposeResult>(m_result) )
         {
-            m_interrupt_flags |= 8;
+            m_interrupt_flags |= 0x08;
+        }
+
+        if ( ! may_update<interior, boundary, '1', TransposeResult>(m_result) )
+        {
+            m_interrupt_flags |= 0x10;
```
tinko92, line 92: This is not so easy to review. Something like 
```cpp
private:
constexpr static unsigned flag_interior_interior = 0x01;
// ...
constexpr static unsigned flag_interior_boundary = 0x10;
constexpr static unsigned flag_all = flag_interior_interior | /* ... */ | flag_interior_boundary;
// ...
m_interrupt_flags |= flag_interior_boundary;
```
may be more clear.

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
```diff
@@ -94,33 +99,40 @@ class no_turns_la_linestring_pred
...
                                                           m_strategy);
-        //BOOST_GEOMETRY_ASSERT_MSG(pig != 0, "There should be no IPs");
 
         if ( pig > 0 )
         {
             update<interior, interior, '1', TransposeResult>(m_result);
-            m_interrupt_flags |= 1;
+            m_interrupt_flags |= 0x01;
+        }
+        else if ( pig == 0 )
```
tinko92, line 124: Do we want some BOOST_GEOMETRY_... guard macro around this? Since this changes behavior we may want to include in the next changelog after this gets pulled as breaking, and if a user (possibly unknowingly since past issue reports suggest the geometric predicate names can be non-intuitive to some) relies on the old behaviour, they could define it.

tinko92 (commented):
Thank you for the PR. The changes look suitable to me for but the goal of this PR is but I have not yet run the full test suite locally.

I have two general questions on this approach:

1. Have you considered your approach against an implementation on the dispatch level, i.e. before the dispatch to the detail/relate/linear_areal code test the linestring for degeneracy and call the detail/relate/point_geometry code in case of degeneracy? This way may be easier and and more clear to wrap into a macro guard to allow restoring the old behaviour.
2. Related to this (and mentioned in another code comment) since we change behavior here, do we want a macro that can be defined to restore the old behavior?

barendgehrels:
> 2\. Related to this (and mentioned in another code comment) since we change behavior here, do we want a macro that can be defined to restore the old behavior?

I'm not sure if this is necessary for this fix.

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
barendgehrels: I would think that we don't need macro's for changed behaviour on degenerate linestrings.
But we should add it to the release notes.

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
mjacobse: I agree, I was just following what was there before, trying to keep the change set small. I am happy to follow the suggestion to make it more readable if desired

mjacobse:
> 1. Have you considered your approach against an implementation on the dispatch level, i.e. before the dispatch to the detail/relate/linear_areal code test the linestring for degeneracy and call the detail/relate/point_geometry code in case of degeneracy? This way may be easier and and more clear to wrap into a macro guard to allow restoring the old behaviour.

My concern with that was that checking the linestring for degeneracy didn't seem completely trivial. Checking `boost::geometry::is_valid` for `failure_wrong_topological_dimension` would have been my first approach which would be O(n) I guess? Probably negligible compared to what follows, but still some non-trivial extra code only to make some arguably out-of-contract behavior less surprising. On the other hand the present approach basically just handles a natural third case in an already present `if` statement. Seemed simpler, but I'd be happy to explore the dispatch level option if that's preferred.

barendgehrels:
> > 1. Have you considered your approach against an implementation on the dispatch level, i.e. before the dispatch to the detail/relate/linear_areal code test the linestring for degeneracy and call the detail/relate/point_geometry code in case of degeneracy? This way may be easier and and more clear to wrap into a macro guard to allow restoring the old behaviour.
> 
> My concern with that was that checking the linestring for degeneracy didn't seem completely trivial. Checking `boost::geometry::is_valid` for `failure_wrong_topological_dimension` would have been my first approach which would be O(n) I guess? Probably negligible compared to what follows, but still some non-trivial extra code only to make some arguably out-of-contract behavior less surprising. On the other hand the present approach basically just handles a natural third case in an already present `if` statement. Seemed simpler, but I'd be happy to explore the dispatch level option if that's preferred.

In general we dispatch by type, not by contents. So I think the choice made here is the correct one.

include/boost/geometry/algorithms/detail/relate/linear_areal.hpp
barendgehrels: If we change it, I would change the "interrupt flags" to an enum with matrix cells

```
enum class la_flag : std::size_t
{
    interior_interior,  // I/I, dim '1'
    interior_exterior,  // I/E, dim '1'
    boundary_interior,  // B/I, dim '0'
    boundary_exterior,  // B/E, dim '0'
    count
};
```

and then have them as a set `std::bitset<static_cast<std::size_t>(flag::count)> m_outcome;`

Possibly with a few helper functions or lambdas

This would (again) change the diff - so I would keep it separate and for now accept the constants.

I will approve current state - but if you think it should be made more readable in this PR, in either way, I'm fine with that as well.

barendgehrels (approved):
Thanks for the improvement!

tinko92 (approved):
> In general we dispatch by type, not by contents.

Sorry, I was imprecise, I was thinking of right after the type dispatch, to keep it more easily separable by a macro if details change later. But if we don't need a macro for to preserve the prior behaviour as an option, then that's moot. I approve the current state (with the release calendar in mind in case you want to merge it for 1.92).

barendgehrels:
@mjacobse is this PR as is OK?
@tinko92 @vissarion will we merge this as is?

I don't think @mjacobse can, so we'll have to take care

mjacobse:
OK for me, yes :+1:

vissarion (approved):
I am OK with merging, thanks! 
Should we add it in 1.92? If so @mjacobse feel free to add an item in release notes in this PR, otherwise I will open a new PR with this addition.

PR #1478, opened 2026-06-28
Narrow the condition for the parent assignment optimisation to union …

tinko92:
This PR fixes case_bitset_3 which fails for intersection due to the optimisation for one outer ring, which seems only safe for union (that seems to be consistent with the comment). The local is_difference variable was removed because it is now unused and no new variable was created because this  condition is checked only on one line and is readable. The test case was enabled.


barendgehrels (approved):
Thanks! This looks completely right

vissarion (approved):
Thanks!

Issue #1472, opened 2026-06-13
I cannot get Dissolve to compile in version 1.91


JohanDore:
Hi @barendgehrels,

I am sorry but it seems dissolve does not compile under the latest version 1.91.

Regarding the dissolve_using_correct., would it be possible for me to test? At least i cannot find it.

Thanks a lot

Johan

barendgehrels:
hi @JohanDore , sorry for the inconvenience, so it was solved last year but still not integrated in the `develop` branch.

I'll try to do that in a week from now, end of June.

barendgehrels:
hi @JohanDore , the PR of last year was undersnowed. It was actually unfinished, partially research, and contained three variants. 
You mention "dissolve_using_correct" - is that the one you were using in the end?
Earlier you used "dissolve"

JohanDore:
> hi [@JohanDore](https://github.com/JohanDore) , the PR of last year was undersnowed. It was actually unfinished, partially research, and contained three variants. You mention "dissolve_using_correct" - is that the one you were using in the end? Earlier you used "dissolve"

Hi @barendgehrels: We ended up using "dissolve" and I couldn't figure out how to try "dissolve_using_correct"

barendgehrels:
> > hi [@JohanDore](https://github.com/JohanDore) , the PR of last year was undersnowed. It was actually unfinished, partially research, and contained three variants. You mention "dissolve_using_correct" - is that the one you were using in the end? Earlier you used "dissolve"
> 
> Hi [@barendgehrels](https://github.com/barendgehrels): We ended up using "dissolve" and I couldn't figure out how to try "dissolve_using_correct"

It had a merge error indeed that is fixed now. I'm preparing a new PR. The algorithms are quite different, and I investigated some differences in them. I suggest you stick with "dissolve" . More details later.

barendgehrels:
hi @JohanDore the PR (still in #1411) is updated. Will this unblock you?

JohanDore:
Hi @barendgehrels: I can confirm that #1411 compiles with 1.91 and all initial tests are passing. I am now to run the full suite of tests and will let you know if something fails

JohanDore:
Hi @barendgehrels : Yes all tests passes at my end now. Huge thanks

barendgehrels:
hi @JohanDore : good to hear

[...]

PR #1479, opened 2026-06-29
fix: nearly identical input should give an intersection

### LINKED ISSUE #1471: Two almost identical rectangles have no intersection
Hi All,

I was looking into a bug reported for our software and found it boiling down to how Boost Geometry deals with almost identical rectangles.

Here is a test case showing the problem:

BOOST_AUTO_TEST_CASE(Boost_Intersection_Fails)
{
  using PointXY = boost::geometry::model::d2::point_xy<double>;
  using Polygon = boost::geometry::model::polygon<PointXY>;
  using MultiPolygon = boost::geometry::model::multi_polygon<Polygon>;

  MultiPolygon polyA;
  read_wkt("MULTIPOLYGON((("
    "20.0970444515398512 -0.3449179410282249,"
    "21.9812157674818778 -0.3449179410282249,"
    "21.9812157674818778 -2.2293927797379025,"
    "20.0970444515398512 -2.2293927797379025,"
    "20.0970444515398512 -0.3449179410282249)))", polyA);

  MultiPolygon polyB;
  read_wkt("MULTIPOLYGON((("
    "20.0970444515398476 -0.3449179410282246,"
    "21.9812157674818813 -0.3449179410282246,"
    "21.9812157674818813 -2.2293927797379034, "
    "20.0970444515398476 -2.2293927797379034, "
    "20.0970444515398476 -0.3449179410282246)))", polyB);

  MultiPolygon polyAB;
  intersection(polyA, polyB, polyAB);

  // The intersection of polyA & polyB should be the ~same as polyA & polyB:
  /*
    BOOST_CHECK_CLOSE(area(polyAB), area(polyA), 0.0001);
    BOOST_CHECK_CLOSE(area(polyAB), area(polyB), 0.0001);
    BOOST_CHECK_EQUAL(polyAB.size(), 1);
    BOOST_CHECK(polyAB[0].inners().empty());
  */
  // But boost returns an empty object
  BOOST_CHECK_EQUAL(polyAB.size(), 0);
}

It is also interesting that simply removing the last decimal for each x, y coordinate makes intersection return the right rectangle.

I tested this in both version 1.88 and 1.91.

Thanks a lot for your hard work on Boost Geometry

Johan

barendgehrels:
Fixes: #1471

✅ AI assisted

# Nearly identical polygons should intersect

Two almost-identical polygons (coordinates differing by ~1 epsilon) produced an empty intersection, a doubled union, and a wrong difference.

Root cause: such input yields zero turns, so select_rings decides containment from a single representative point — but that point is a ring vertex, which lands exactly on the other ring's boundary (range_in_geometry == 0), and an on-boundary point was treated as "not within".

Fix: in `select_rings`, when the overlay produced no turns, treat an entirely-on-boundary ring as "within" for the binary set operations (union/intersection/difference). 

Related to commit 2780c9dc2 (which changed this code, reversal would have fixed this as well - but would have given regression in other cases).

This PR also enhances the test (that initially passed for #1471): if two geometries intersect but don't touch, their intersection must be non-empty — independent of the overlay path, so it catches this class of bug that area-balance checks cannot.


vissarion (approved):
LGMT thanks!

barendgehrels:
> LGMT thanks!

@vissarion thanks for the quick review. It's a bugfix, 1.92 is still open, I think we can make that. Then I'll also include it in the release notes.
@tinko92 also OK with this?

vissarion:
> @vissarion thanks for the quick review. It's a bugfix, 1.92 is still open, I think we can make that. 

I agree, if Tinko is OK, I will merge it to main.

> Then I'll also include it in the release notes. @tinko92 also OK with this?

Thanks, I will then update the boost release notes.

include/boost/geometry/algorithms/detail/overlay/select_rings.hpp
```diff
@@ -261,23 +262,35 @@ inline void update_ring_selection(Geometry1 const& geometry1,
...
 
+        // Code 0: the point is located on the boundary. Normally such a
+        // situation is resolved by the turns (the rings are traversed), but if
+        // the overlay produced no turns at all the inputs are (nearly)
+        // coincident without any segment intersection being detected,
+        // such as issue 1471.
+        constexpr bool is_binary_operation
+            = OverlayType == overlay_union
+            || OverlayType == overlay_intersection
+            || OverlayType == overlay_difference;
```
tinko92: I have not reviewed dissolve sufficiently to comment on how it should be treated here. This is not a blocker for me for merging since it is in extensions.

test/algorithms/set_operations/set_ops_areal_areal.cpp
```diff
@@ -155,6 +155,17 @@ void test_detail(std::string const& name, std::string const& wkt1, std::string c
...
     BOOST_CHECK_MESSAGE(bgeo::math::abs(balance) < eps,
         "Case: " << name << " wrong union or intersection " << balance);
+
+    // The relate-based predicates do not depend on the overlay ring traversal
+    // If two geometries intersect but do not merely touch, their interiors overlap
+    // and the intersection cannot be empty. This catches cases such as issue 1471.
+    if (bgeo::intersects(geometry1, geometry2)
+            && ! bgeo::touches(geometry1, geometry2))
+    {
+        BOOST_CHECK_MESSAGE(! result_intersection.empty(),
```
tinko92: Minor: Probably more consistent to pull all conditions into the macro since we only `if` on settings here while many checks have compound conditions.

test/algorithms/overlay/select_rings.cpp
```diff
@@ -55,7 +55,7 @@ void test_geometry(std::string const& wkt1, std::string const& wkt2,
             Geometry1, Geometry2
         >::type strategy_type;
 
-    bg::detail::overlay::select_rings<OverlayType>(geometry1, geometry2, empty, selected, strategy_type());
+    bg::detail::overlay::select_rings<OverlayType>(geometry1, geometry2, empty, selected, strategy_type(), false);
```
tinko92: Why do we pass `false` here? I think the winded case has turns and the argument that would preserve the prior within_other = range_in_geometry  > 0 behavior would be `true` since it's negated.

include/boost/geometry/extensions/algorithms/dissolve.hpp
```diff
@@ -241,7 +241,7 @@ struct dissolve_ring
 
         std::map<ring_identifier, properties> selected;
 
-        detail::overlay::select_rings<overlay_dissolve>(input_ring, turn_info_per_ring, selected, strategy);
+        detail::overlay::select_rings<overlay_dissolve>(input_ring, turn_info_per_ring, selected, strategy, ! turns.empty());
```
tinko92: This value may be true or false but is never used if I understand correctly.

tinko92 (approved):
Sorry that I couldn't review this yesterday, there are some lines that were non-obvious to me (I left comments). I am OK with merging this before the release since it fixes cases and does not break anything, so my comments are not blockers.

Most comments are about passing something (constant false or a computed value) for overlay_has_turns being ambiguous. Maybe some argument name that indicates "decide within for the no turns special case" with a default value would be more clear here. But again, I am OK with merging this since it is well-limited to a special case, does not seem to break anything and fixes a real problem.

include/boost/geometry/algorithms/detail/overlay/select_rings.hpp
barendgehrels: The `dissolve` is excluded here because of commit 2780c9d

I didn't try it with the updated branch for #1472 - because it's in another branch and I don't have all the test cases Johan uses.

For buffer, it is probably not necessary. I'll try that too.

So I agree that it might be `true` always ==> just omit it.

include/boost/geometry/extensions/algorithms/dissolve.hpp
barendgehrels: Indeed. And I should have split the commit.

test/algorithms/overlay/select_rings.cpp
barendgehrels: There are no selections, it can be either way

test/algorithms/set_operations/set_ops_areal_areal.cpp
barendgehrels: Yes that makes sense

barendgehrels:
> overlay_has_turns

Maybe `include_on_border` is clearer (and then it should be reverted). I could give it a default value `false` (avoiding the change in dissolve - and then it will have `turns.empty()`).

barendgehrels:
I'll update the PR later today. Thanks for the approvals!

include/boost/geometry/algorithms/detail/overlay/select_rings.hpp
barendgehrels: ✅

include/boost/geometry/extensions/algorithms/dissolve.hpp
barendgehrels: ✅

barendgehrels:
> > overlay_has_turns
> 
> Maybe `include_on_border` is clearer (and then it should be reverted). I could give it a default value `false` (avoiding the change in dissolve - and then it will have `turns.empty()`).

✅  (`include_on_boundary` seems the best OGC fitting term)

include/boost/geometry/algorithms/detail/overlay/select_rings.hpp
barendgehrels: All your comments processed, thanks! It also makes the PR shorter and avoids the change in dissolve

test/algorithms/set_operations/set_ops_areal_areal.cpp
barendgehrels: ✅

tinko92:
Thanks for the quick changes and the fix! In my opinion this can be merged.

barendgehrels:
> Thanks, I will then update the boost release notes.

Merged! ✔️

PR #1481, opened 2026-07-03
test: include #1439 in test set and clean up of related issues

barendgehrels:


test/algorithms/set_operations/set_ops_areal_areal.cpp
```diff
@@ -306,6 +306,8 @@ int test_main(int, char* [])
     TEST_CASE(issue_1363);
 #endif
 
+    TEST_CASE_WITH(issue_1439, 0, 1, ut_settings().ignore_validity_diff_sym());
```
barendgehrels, line 309: Some (non commited) debug output gives:
```
issue_1439_0_1 875.977 7.49641 883.474 6.20003e-11 1 875.977 7.49641
```
which is area a, area b, the union, the intersection, "intersects" (`1`), diff a-b, diff b-a

The intersection is small but present. Its WKT is the same as the one of the reporter in #1439. So we can consider it as closed. The one imperfection is that the sym-diff is invalid, over such a small common area - it might be a false positive.

test/algorithms/relate/relate_linear_linear.cpp
```diff
@@ -380,7 +380,7 @@ void test_linestring_multi_linestring()
     test_geometry<ls, mls>("LINESTRING(0 0,5 5)",
                            "MULTILINESTRING((0 0,5 5),(5 5,6 6))",
                            "1FF00F102");
-#ifdef BOOST_GEOMETRY_TEST_ENABLE_FAILING
+#ifdef BOOST_GEOMETRY_TEST_FAILURES
```
barendgehrels, line 383: A few years back we harmonized these defines - apparently it diverged again after that, a bit, in this way. Better to have just one define for this purpose.

test/algorithms/overlay/overlay_cases.hpp
```diff
@@ -1297,10 +1297,10 @@ static std::string issue_1345_b[2] =
...
-    "MULTIPOLYGON(((2.0461606968426484521 0.90344999999999986429,2.001160696842647635 0.91095000000000003748,1.8511606968426477238 0.91094999999999992646,1.813660696842647635 0.91844999999999998863,1.813660696842647635 0.9409499999999999531,1.8211606968426479192 1.1059499999999999886,1.8286606968426479813 1.263449999999999962,1.9636606968426479902 1.263449999999999962,2.0461606968426484521 1.2559499999999998998,2.0536606968426478481 1.2409499999999999975,2.1286606968426480258 1.2409499999999999975,2.1286606968426480258 1.1059499999999999886,2.1211606968426477415 0.92594999999999982876,2.1136606968426479014 0.89594999999999980211,2.091160696842648381 0.88844999999999996199,2.0461606968426484521 0.90344999999999986429)))"
+    "POLYGON((-2.47089026 -86.03059246, -1.161944873 146.3030596, 3.40221214 145.8628015, 3.024695769 141.949088, 2.420586453 111.9554564, 1.208013978 -86.04069936, -2.47089026 -86.03059246))",
+    "POLYGON((-6.213278056 -88.01851748, -6.207382255 -86.51852906, -6.20539951 -86.01408604, -2.470873099 -86.02434575, -2.472221358 -86.53315586, -2.476196085 -88.03315059, -6.213278056 -88.01851748))"
```
barendgehrels, line 1303: This is the version from the sample of the report in #1439

test/algorithms/overlay/overlay.cpp
```diff
@@ -315,12 +315,6 @@ static std::string issue_1100_multi[2] =
...
-    "MULTIPOLYGON(((2.0461606968426484521 0.90344999999999986429,2.001160696842647635 0.91095000000000003748,1.8511606968426477238 0.91094999999999992646,1.813660696842647635 0.91844999999999998863,1.813660696842647635 0.9409499999999999531,1.8211606968426479192 1.1059499999999999886,1.8286606968426479813 1.263449999999999962,1.9636606968426479902 1.263449999999999962,2.0461606968426484521 1.2559499999999998998,2.0536606968426478481 1.2409499999999999975,2.1286606968426480258 1.2409499999999999975,2.1286606968426480258 1.1059499999999999886,2.1211606968426477415 0.92594999999999982876,2.1136606968426479014 0.89594999999999980211,2.091160696842648381 0.88844999999999996199,2.0461606968426484521 0.90344999999999986429)))"
-};
```
barendgehrels, line 322: Unrelated, it was duplicate

test/algorithms/overlay/overlay.cpp
```diff
@@ -364,15 +358,9 @@ void test_all()
...
     // TEST_INTERSECTION(case_bitset_3, 16.0, 1, 0);
     // TEST_INTERSECTION(case_149_multi, 48.0, 2, 2); // instruction
-// 
 
     // TEST_DIFFERENCE_B(issue_893_multi, 97213916.0, 1, 1); // needs is_traverse
 
-    
-#ifdef BOOST_GEOMETRY_TEST_ENABLE_FAILING
-    TEST_INTERSECTION(issue_1439, 0, 0, 0.0);
-#endif
```
barendgehrels, line 374: This `overlay.cpp` is more a debug test - for "real" tests we should prefer `test/algorithms/set_operations/set_ops_areal_areal.cpp`

tinko92 (approved):
Looks good to me based on reading the diff. Because I have many unrelated test failures on my current machine, I did not test locally and trust the CI on this (since it only touches  ./test/ anyway).

The comment on `overlay.cpp` is helpful, I was unsure about it's relationship to `set_ops_areal_areal.cpp` during earlier work. As an aside, is there any significance to some of its tests being commented out rather than behind `BOOST_GEOMETRY_TEST_FAILURES`?

barendgehrels:
> Looks good to me based on reading the diff. Because I have many unrelated test failures on my current machine, I did not test locally and trust the CI on this (since it only touches ./test/ anyway).
> 
> The comment on `overlay.cpp` is helpful, I was unsure about it's relationship to `set_ops_areal_areal.cpp` during earlier work. As an aside, is there any significance to some of its tests being commented out rather than behind `BOOST_GEOMETRY_TEST_FAILURES`?

Thanks for the approval and the comment!

Indeed, `overlay.cpp` is mostly a debugging tool and has the visuals I need. Mainly we should (ever) move it outside the unit tests... The cases tested are usually duplicate. They can just go. 

There is also a `buffer_overlay.cpp` with the same purpose.

And yes, tests are commented out - they could actually just be removed from these two sources.

barendgehrels:
Taking the freedom to merge also this, it's test only and has an approval

PR #1489, opened 2026-09-01
Fix Doxygen warnings

tinko92:
This PR fixes warnings currently produced by doxygen when building the BG doc. 

Some warnings were related to deprecated options. This is fixed with doxygen -u -s Doxyfile, which updates it to the newest version. I used Doxygen 1.14 for that, which seems to be used by the Boost release container. Unfortunately this is (imho) needlessly verbose and adds every available option with its default values. Where it added entire sections, I removed those. For the comments that looked manually written (e.g. explaining the role of HTML output for BG within Boost), I put them manually back in since doxygen removes all comments. Our CI uses Doxygen 1.9.8 from Ubuntu 24.04, so I removed the options added by 1.14 that produced warnings in doxygen 1.9.8.

One warning was about a syntax error in Alias for an unused alias. I removed all unused aliases.

Some warnings were related to missing groups (added) and missing parameter documentation (from an earlier PR of mine, also added).

PR is split into two commits because the first one should be neutral (not change the documentation output at all besides timestamps). To allow verifying that in review, I kept it in a separate commit.

Doxygen has an option to produce a non-zero exit code. My original intent was to fix the warnings, then enable that so the CI can warn about future drift. But since our documentation CI job does not necessarily use the same doxygen version as the Boost release image, I decided against it to avoid this causing problems in the future (e.g. Boost release updating to a future Doxygen version, then erroring on some obsolete option). We could consider adding some script to our documentation CI job that looks at the output and errors on warnings to avoid overlooking e.g. undocumented parameters like in my earlier PR.

### DIFF
--- a/doc/doxy/Doxyfile
+++ b/doc/doxy/Doxyfile
@@ -1,4 +1,4 @@
-# Doxyfile 1.5.5
+# Doxyfile 1.14.0
 
 #---------------------------------------------------------------------------
 # Project related configuration options
@@ -6,8 +6,12 @@
 DOXYFILE_ENCODING      = UTF-8
 PROJECT_NAME           = "Boost.Geometry (aka GGL)"
 PROJECT_NUMBER         =
+PROJECT_BRIEF          =
+PROJECT_LOGO           =
 OUTPUT_DIRECTORY       = ./doxygen_output
 CREATE_SUBDIRS         = NO
+CREATE_SUBDIRS_LEVEL   = 8
+ALLOW_UNICODE_NAMES    = NO
 OUTPUT_LANGUAGE        = English
 BRIEF_MEMBER_DESC      = YES
 REPEAT_BRIEF           = YES
@@ -15,13 +19,15 @@ ABBREVIATE_BRIEF       =
 ALWAYS_DETAILED_SEC    = YES
 INLINE_INHERITED_MEMB  = NO
 FULL_PATH_NAMES        = YES
-STRIP_FROM_PATH        = c:/_svn/boost/trunk/boost/geometry  \
+STRIP_FROM_PATH        = c:/_svn/boost/trunk/boost/geometry \
                          c:/_svn/boost/trunk/libs/geometry/doc/doxygen_input/pages
 STRIP_FROM_INC_PATH    =
 SHORT_NAMES            = NO
 JAVADOC_AUTOBRIEF      = NO
+JAVADOC_BANNER         = NO
 QT_AUTOBRIEF           = NO
 MULTILINE_CPP_IS_BRIEF = NO
+PYTHON_DOCSTRING       = YES
 INHERIT_DOCS           = YES
 SEPARATE_MEMBER_PAGES  = NO
 TAB_SIZE               = 8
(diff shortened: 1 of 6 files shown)


barendgehrels (approved):
👍 looks good to me! Thanks!

doc/doxy/Doxyfile
```diff
@@ -37,154 +43,184 @@ TAB_SIZE               = 8
...
+                         "tparam_first_box=first box type" \
+                         "tparam_second_point=second point type" \
+                         "tparam_second_box=second box type" \
+                         "tparam_segment_point=segment point type" \
+                         "tparam_strategy{1}=Any type fulfilling a \1 Strategy Concept" \
+                         "tparam_strategy_overlay=Compound strategy for segment intersection" \
+                         "param_box=A model of the specified Box Concept" \
+                         "param_box_or_segment=A box modelling the specified Box Concept or segment modelling the specified Segment Concept" \
+                         "param_geometry=A model of the specified concept" \
+                         "param_out{1}=The output iterator, to which \1 geometries are feeded" \
```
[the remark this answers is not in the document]
tinko92: Not really a change (just shows up in the diff due to Doxygen's syntax update) but fixed.

include/boost/geometry/util/normalize_spheroidal_coordinates.hpp
```diff
@@ -345,6 +345,7 @@ inline void normalize_angle_cond(CoordinateType& angle)
 \tparam CoordinateType The type of the coordinates
 \param longitude Longitude
 \param latitude Latitude
+\param exact Whether to use exact antimeridian checks; if false, a relaxed tolerance is used
```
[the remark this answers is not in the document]
tinko92: Not in the changes but since it fits the scope of the PR, addressed.

include/boost/geometry/io/wkt/write.hpp
```diff
@@ -546,10 +546,7 @@ struct wkt<Geometry, geometry_collection_tag>
 \details Stream manipulator, streams geometry classes as \wkt streams
 \par Example:
 Small example showing how to use the wkt class
-\dontinclude doxygen_1.cpp
-\skip example_as_wkt_point
-\line {
-\until }
+\snippet doxygen_1.cpp example_as_wkt_point
```
barendgehrels: Looks basically good to me but I don't remember the difference... (how it will look like in the doc)

tinko92: I'll build it to double check and share a screenshot when I get home.

tinko92: It actually doesn't show up in the documentation at all. It is changed to silence a warning, which read as 
```sh
/Users/tinko/dev/boost/boost/geometry/io/wkt/write.hpp:552: warning: referenced pattern '{' for command '\line' not found
```
and the change to the snippet command silences that warning. But the class template wkt\_manipulator does not show up in the documentation at least since 1.56 (https://www.boost.org/doc/libs/1_56_0/libs/geometry/doc/html/geometry/reference/io/wkt/wkt.html). I thought, it was because the template function of the same name shows up with a different snippet example, that is included differently via wkb. I mostly tested the commit by looking at the diff between the built documentation from develop and this branch to check that nothing changed for the worse. There is not git history with the original line command, it is in the first git commit, so I don't know when this snippet last produced visible output.

I can just delete the added snippet command from the commit, since it does not do anything. The snippet that it would show also doesn't relate to this particular class template, it's a more general example (that does not show in the documentation anywhere, so it may also be dead code): https://github.com/boostorg/geometry/blob/196d04c614c12a8d788212b63fa65d25c8b7ea86/doc/doxy/doxygen_input/sourcecode/doxygen_1.cpp#L46

barendgehrels: Thanks for testing! Yes, these things happen unfortunately...

In this case, yes, I agree, please delete that snippet. Better no snippet than an unshown (maybe) wrong snippet.

tinko92: Snippet removed altogether.

Issue #1421, opened 2025-07-30
Small integer multipolygon (sym_)difference failure case post traversal rewrite


tinko92:
Here is a very small failure case for both difference and sym_difference that first fails on commit 805ff654e869030c4ba70f37fad6c3825f59c077, i.e. the traversal rewrite.

```cpp
#include <boost/geometry.hpp>

namespace bg = boost::geometry;

using point = bg::model::d2::point_xy<int>;
using polygon = bg::model::polygon<point>;
using multi_polygon = bg::model::multi_polygon<polygon>;

int main() {
  const auto geo1 = bg::from_wkt<multi_polygon>("MULTIPOLYGON(((8 8,8 6,8 4,6 4,6 2,6 0,0 0,0 6,2 6,2 8,8 8),(2 4,2 2,4 2,4 4,2 4),(6 6,4 6,4 4,6 4,6 6)))");
  const auto geo2 = bg::from_wkt<multi_polygon>("MULTIPOLYGON(((6 6,6 4,4 4,4 6,6 6)),((6 6,6 8,8 8,8 6,6 6)),((4 6,2 6,2 8,4 8,4 6)),((4 4,4 2,2 2,2 4,4 4)),((4 2,6 2,6 0,4 0,4 2)),((2 6,2 4,0 4,0 6,2 6)),((2 2,2 0,0 0,0 2,2 2)),((6 2,6 4,8 4,8 2,6 2)))");
  multi_polygon diff, symdiff;
  bg::difference(geo1, geo2, diff);
  bg::sym_difference(geo1, geo2, symdiff);

  std::cout << "Diff:\t\t" << bg::to_wkt(diff) << '\n';
  std::cout << "Symdiff:\t" << bg::to_wkt(symdiff) << '\n';
  return 0;
}
```
Prints 
```sql
Diff:           MULTIPOLYGON(((8 6,8 4,6 4,6 6,8 6)),((2 2,4 2,4 0,2 0,2 2)),((2 4,2 2,0 2,0 4,2 4)),((6 6,4 6,4 8,6 8,6 6)))
Symdiff:        MULTIPOLYGON(((8 2,6 2,6 4,4 4,4 6,4 8,6 8,6 6,8 6,8 2)),((4 0,2 0,2 2,0 2,0 4,4 4,4 0)))
```
should print
```sql
[...]

PR #1491, opened 2026-09-10
Fix/generalized randomized grid test and fixes

### LINKED ISSUE #1238: within returns a wrong result with multi-linestrings
The following program returns true while this is clearly not the case.

```cpp
#include <boost/geometry.hpp>

int main() {
    using Point = boost::geometry::model::d2::point_xy<double>;
    using Linestring = boost::geometry::model::linestring<Point>;
    using MultiLinestring = boost::geometry::model::multi_linestring<Linestring>;

    MultiLinestring ml1, ml2;
    boost::geometry::read_wkt("MULTILINESTRING((10 10, 20 20 ), (15 15 , 30 15))", ml1);
    boost::geometry::read_wkt("MULTILINESTRING((10 10, 20 20), (60 -60, 30 15))", ml2);

    std::cout << boost::geometry::within(ml2, ml1) << std::endl;

    return 0;
}
```

### LINKED ISSUE #1406: Wrong intersection of polygon and linestring

### LINKED ISSUE #1420: Small integer multipolygon union failure case post traversal rewrite
Here is a very small failure case for union that first fails on commit 805ff654e869030c4ba70f37fad6c3825f59c077, i.e. the traversal rewrite.

```cpp
#include <boost/geometry.hpp>

using point = bg::model::d2::point_xy<int>;
using multi_polygon = bg::model::multi_polygon<polygon>;

int main() {
  const auto geo1 = bg::from_wkt<multi_polygon>("MULTIPOLYGON(((4 6,2 6,2 4,0 4,0 8,4 8,6 8,6 6,6 4,4 4,4 6)),((2 2,2 4,4 4,4 2,2 2)),((2 2,2 0,0 0,0 2,2 2)),((4 2,6 2,6 0,4 0,4 2)))");
  const auto geo2 = bg::from_wkt<multi_polygon>("MULTIPOLYGON(((4 2,6 2,6 0,4 0,4 2)),((4 2,2 2,2 4,4 4,4 2)))");
  multi_polygon res;
  bg::union_(geo1, geo2, res);
  std::cout << bg::to_wkt(res) << '\n';
  return 0;
}
```
Prints 
```wkt
MULTIPOLYGON(((2 4,0 4,0 8,4 8,6 8,6 6,6 4,4 4,4 2,2 2,2 4)),((4 2,6 2,6 0,4 0,4 2)),((2 2,2 0,0 0,0 2,2 2)))
```
should print
```wkt
MULTIPOLYGON(((6 2,6 0,4 0,4 2,6 2)),((2 4,0 4,0 8,4 8,6 8,6 6,6 4,4 4,4 6,2 6,2 4)),((2 4,4 4,4 2,2 2,2 4)),((2 2,2 0,0 0,0 2,2 2)))
```

### LINKED ISSUE #1490: difference() produces invalid (self-intersecting) result for valid polygon-with-hole input
### Version of Boost

`1.91.0`

### Actual and Expected Behavior

**Actual:** `boost::geometry::difference()` on two valid `multi_polygon` operands (a polygon with a hole, minus another polygon) returns a result that fails `boost::geometry::is_valid()` — it contains a self-intersection introduced at the point where the hole's ring touches the subtracted polygon's boundary.

I discovered this by seeing that a point, that is in `A \ B`, is reported as also being within `B`, which violates the basic definition of set difference.

**Expected:** `boost::geometry::difference()` should return a valid geometry (satisfying `is_valid()`) for valid input operands.

**Minimal complete program:** attached [testing.cpp](https://github.com/user-attachments/files/31964256/testing.cpp)
```
#include <boost/geometry.hpp>
#include <cassert>
#include <iostream>

using int_point_t = bg::model::point<int64_t, 2, bg::cs::cartesian>;
using int_poly_t = bg::model::polygon<int_point_t>;
using int_multi_poly_t = bg::model::multi_polygon<int_poly_t>;

int main()
{
   int_multi_poly_t poly_A;
   boost::geometry::read_wkt("MULTIPOLYGON(((0 0,20 0,20 20,0 20,0 0),(5 15,14 14,9 6,5 15)))", poly_A);

   int_multi_poly_t poly_B;
   boost::geometry::read_wkt("MULTIPOLYGON(((5 15,5 5,15 5,15 15,5 15)))", poly_B);

   bg::correct(poly_A);
   bg::correct(poly_B);

   assert(bg::is_valid(poly_A));
   assert(bg::is_valid(poly_B));

   int_multi_poly_t result;
   boost::geometry::difference(poly_A, poly_B, result);

   std::cout << boost::geometry::to_wkt(result) << std::endl;
   if (std::string message; !bg::is_valid(result, message))
      std::cerr << "Invalid geometry: " << message << std::endl;

}
```
Output when run:
> MULTIPOLYGON(((0 0,0 20,20 20,20 0,0 0),(5 15,9 6,14 14,5 15),(5 15,5 5,15 5,15 15,5 15)))
> Invalid geometry: Geometry has invalid self-intersections. A self-intersection point was found at (5, 15); method: t; operations: u/i; segment IDs {source, multi, ring, segment}: {0, 0, 0, 2}/{0, 0, 1, 3}

**Pictures of input:**
Image Image

**Picture of output:**
Image

### Steps necessary to reproduce the problem

1. Build the attached `testing.cpp` against Boost 1.91.0.
2. Run it.
3. Observe that `bg::is_valid(result, message)` reports `false` with a self-intersection at the point where polygon A's hole ring touches polygon B's boundary.

Both input polygons (`poly_A`, `poly_B`) are individually valid (confirmed via `is_valid()` before the `difference()` call in the program), so the invalidity is introduced by `difference()` itself.

### All relevant compiler information

- Compiler:  GCC 14 (`g++-14.3.0`)
- Standard: C++23 (`-std=gnu++23`)
- OS: Linux

tinko92:
This PR draft is vibe-coded and not as clean as I would like it to be but I share it early to avoid duplication of efforts. 

The main idea is to generalize random_bitset_grids (#1361) and use it as a test harness for LLMs to figure out the remaining algorithmic bugs in Boost.Geometry, as far as the generalisation of the mentioned older test permits.

<details>
<summary>reminder of what #1361 was</summary>
As a reminder, the original test considers an integer grid with horizontal and vertical uniformly spaced lines, associates each cell, associates it with a bitset, such that each cell corresponds to one bit’s index in the bitset, and then generates two bitsets, the corresponding multipolygons and checks if a set operation (union, intersection, sym_diff, …) produces a geometry that maps to the bitset produced by the corresponding bit operation (OR, AND, XOR, …). The idea is to autogenerate many problems automatically for which we can easily predict the correct result, do so in a way that can represent many topological situations in principle and then just let it run a long time. This also produces relatively simple test cases. Another important part of this test was, to separate this test entirely from the concern for floating-point issues, so all intermediate results must be integers.
</details>

We generalize this in the following ways:
1. We add two more line families 3x+4y=const and 3x-4y=const. This requires spacing the lines out more to keep all intersections at integer points. It increases the maximum degree of vertices in the grid, allowing more topological situations to be represented (e.g. 4 non-overlapping rings meeting in one point). For illustration, the following is a 2x2 grid with the added directions. It has 8 cells, 9 vertices, and 16 edges. v_0 is at (0 0), v_8 is at (8 6). All intersections are at numbered vertices at integer coordinates. By repeating this pattern, we get larger grids:

image

2. In addition to multipolygons (unions of some choice of the cells as in the picture above). We also consider multipoints (same for the vertices as in the picture above) and multilinestrings (unions of some choice of the edges as in the picture above) across the grid-lines. This is a rather natural generalisation.

3. We add testing of the relate methods. We can determine the expected DE9IM matrix from the bitsets because the adjacency and incidents relationships of the vertices, edges, and cells follow regular patterns.

4. We add testing for the buffer method with miter joins. For that it is important that all directions follow pythagorean triplets, so that there is a rational (not involving square roots) common distance such that for every line segment, the buffered line is again on the grid. Unfortunately buffer doesn’t fully do integer coordinates, but since our test points are comfortably within each cell, this shouldn’t affect results.

In principle this could be further generalised by adding more directions (there are infinite integer pythagorean triplets) but it gets very ugly very quickly and I think with 4 directions, allowing 8 nonoverlapping segments to meet in one point and having many such points, we can already represent a lot.

Running this at small sizes, generated a variety of failures for all three operations. which I used Codex GPT 5.6/6 to solve until running out of tokens. There are still some failures in buffer left. It also seems to fix some open issues:
59787a382 fixes #1238
a10fc80b4 fixes #1420, #1421
bbb1acdc9 fixes #1406
eb4427739 fixes #1490
The other commits fix bugs uncovered by the new tests that have no associated issues.

The individual fixes are intentionally kept in separate commits to allow for git bisection.

The changes create one failure in a buffer test (specifically here: https://github.com/tinko92/geometry/actions/runs/34663822728/job/103471604820#step:10:129 ), that I believe to be spurious. I think the test has numerical robustness issues that get unmasked by changed behaviour from unrelated buffer fixes.

I hope to understand every fix before marking it not a draft and possibly clean up the test code more. The test code already went through several iterations to be more reasonable than the original LLM output (except for the SVG drawing which I'm not so worried about) but miter buffer cells computation for example remains quite ugly. This may be acceptable for test-only code that is not intended to become part of the CI. I left the README.md that the agent tends to write in for now, for reviewability.

### DIFF
--- a/test/algorithms/crosses/crosses.cpp
+++ b/test/algorithms/crosses/crosses.cpp
@@ -58,8 +58,8 @@ void test_ll()
 
     test_geometry<mls, mls>("MULTILINESTRING((0 0,2 2,4 4),(3 0,3 4))", "MULTILINESTRING((0 1,4 1),(0 2,4 2))", true);
 
-    // spike - boundary and interior on the same point
-    test_geometry<ls, ls>("LINESTRING(3 7, 8 8, 2 6)", "LINESTRING(5 7, 10 7, 0 7)", true);
+    // The revisited endpoint is boundary, so neither contact is interior/interior.
+    test_geometry<ls, ls>("LINESTRING(3 7, 8 8, 2 6)", "LINESTRING(5 7, 10 7, 0 7)", false);
 }
 
 template <typename P>
--- a/include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/piece_border.hpp
@@ -231,7 +231,7 @@ struct piece_border
     template <typename TurnPoint, typename State>
     bool point_on_piece(TurnPoint const& point,
                         bool one_sided, bool is_linear_end_point,
-                        State& state) const
+                        State& state, bool is_exposed_start = false, bool is_exposed_end = false) const
     {
         if (ring_or_original_empty())
         {
@@ -253,12 +253,13 @@ struct piece_border
         geometry::strategy::buffer::place_on_ring_type const por_original
             = adapted_place_on_ring(geometry::strategy::buffer::place_on_ring_original,
                                     one_sided, is_linear_end_point);
+        // At flat ends and concave joins, helper segments also belong to the offsetted ring.
         geometry::strategy::buffer::place_on_ring_type const por_from_offsetted
             = adapted_place_on_ring(geometry::strategy::buffer::place_on_ring_from_offsetted,
-                                    one_sided, is_linear_end_point);
+                                    one_sided || is_exposed_end, is_linear_end_point);
         geometry::strategy::buffer::place_on_ring_type const por_to_offsetted
             = adapted_place_on_ring(geometry::strategy::buffer::place_on_ring_to_offsetted,
-                                    one_sided, is_linear_end_point);
+                                    one_sided || is_exposed_start, is_linear_end_point);
 
         bool continue_processing = true;
         if (m_original_size == 1)
--- a/include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
+++ b/include/boost/geometry/algorithms/detail/buffer/turn_in_piece_visitor.hpp
@@ -19,6 +19,7 @@
 
 #include <boost/geometry/algorithms/comparable_distance.hpp>
 #include <boost/geometry/algorithms/covered_by.hpp>
+#include <boost/geometry/algorithms/equals.hpp>
 #include <boost/geometry/algorithms/detail/disjoint/point_box.hpp>
 #include <boost/geometry/algorithms/detail/disjoint/box_box.hpp>
 #include <boost/geometry/algorithms/detail/dummy_geometries.hpp>
@@ -207,10 +208,21 @@ public:
         // Check if buffer is one-sided (at this point), because then a point
         // on the original border is not considered as within.
         bool const one_sided = has_zero_distance_at(turn.point);
+        // Another linestring's endpoint does not expose this piece's interior.
+        bool const is_linear_end_point = turn.is_linear_end_point
+            && border.m_original_size > 0
+            && ((piece.is_flat_start && geometry::equals(turn.point,
+                    border.m_originals[border.m_original_size - 1], m_umbrella_strategy))
+                || (piece.is_flat_end && geometry::equals(turn.point,
+                    border.m_originals[0], m_umbrella_strategy)));
 
         typename Border::state_type state;
         if (! border.point_on_piece(turn.point, one_sided,
-                                    turn.is_linear_end_point, state))
+                                    is_linear_end_point, state,
+                                    piece.is_flat_start
+                                        || m_pieces[piece.left_index].type == strategy::buffer::buffered_concave,
+                                    piece.is_flat_end
+                                        || m_pieces[piece.right_index].type == strategy::buffer::buffered_concave))
         {
             return true;
         }
--- a/include/boost/geometry/algorithms/detail/overlay/follow.hpp
+++ b/include/boost/geometry/algorithms/detail/overlay/follow.hpp
@@ -18,6 +18,7 @@
 
 #include <cstddef>
 #include <set>
+#include <memory>
 #include <type_traits>
 
 #include <boost/range/begin.hpp>
@@ -32,6 +33,7 @@
 #include <boost/geometry/algorithms/detail/overlay/debug_traverse.hpp>
 #include <boost/geometry/algorithms/detail/overlay/turn_info.hpp>
 #include <boost/geometry/algorithms/detail/point_on_border.hpp>
+#include <boost/geometry/algorithms/detail/relate/follow_helpers.hpp>
 #include <boost/geometry/algorithms/detail/relate/turns.hpp>
 #include <boost/geometry/algorithms/detail/tupled_output.hpp>
 #include <boost/geometry/core/static_assert.hpp>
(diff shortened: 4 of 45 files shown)


barendgehrels-tomtom:
> This PR draft is vibe-coded and not as clean as I would like it to be but I share it early to avoid duplication of efforts. 

...

Looks like a good solid extension of the test! Thanks! 

I will look in more detail in two weeks or so, currently it is not possible.

test/algorithms/crosses/crosses.cpp
```diff
@@ -58,8 +58,8 @@ void test_ll()
 
     test_geometry<mls, mls>("MULTILINESTRING((0 0,2 2,4 4),(3 0,3 4))", "MULTILINESTRING((0 1,4 1),(0 2,4 2))", true);
 
-    // spike - boundary and interior on the same point
-    test_geometry<ls, ls>("LINESTRING(3 7, 8 8, 2 6)", "LINESTRING(5 7, 10 7, 0 7)", true);
+    // The revisited endpoint is boundary, so neither contact is interior/interior.
+    test_geometry<ls, ls>("LINESTRING(3 7, 8 8, 2 6)", "LINESTRING(5 7, 10 7, 0 7)", false);
```
tinko92, line 62: This change is justified by the model quoting 

> The domain of geometric objects considered is those that are topologically closed. The interior of a geometric object consists of those Points that are left when the boundary Points are removed. The exterior of a geometric object consists of Points not in the interior or boundary

from https://docs.ogc.org/is/06-103r4/06-103r4.pdf . I.e. saying that POINT(5 7) being an end point and thereby boundary makes it removed from the interior. I'm not familiar enough with this to be sure it is the intent of the standard, but I can see where the interpretation is coming from.

